PR #2, opened 2016-03-29
Use redis for VID generator and move RID translation to syncd

kcudnik:


lguohan:
why closed?

2016-03-31 17:16 GMT-07:00 Kamil Cudnik <notifications@github.com>:

PR #4, opened 2016-04-12
syncd: adding MLNX SAI support

stcheng:
Adding MLNX SAI support in syncd.cpp

### DIFF
--- a/syncd/syncd.cpp
+++ b/syncd/syncd.cpp
@@ -4,6 +4,8 @@
 swss::Table *g_vidToRid = NULL;
 swss::Table *g_ridToVid = NULL;
 
+std::map<std::string, std::string> gProfileMap;
+
 void sai_diag_shell()
 {
     sai_status_t status;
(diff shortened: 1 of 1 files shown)


syncd/syncd.cpp
```diff
@@ -555,6 +560,11 @@ int main(int argc, char **argv)
     getRequest = new swss::ConsumerTable(db, "GETREQUEST");
     getResponse  = new swss::ProducerTable(db, "GETRESPONSE");
 
+#ifdef MLNXSAI
+    std::string mlnx_config_file = "/etc/ssw/ACS-MSN2700/sai_2700.xml";
```
kcudnik: currently this can be hardcoded but at later stage this could be passed from cmdline

kcudnik: Ack

stcheng: yes. that's what i'm thinking about.
later we either parsing a config file or getting the info from cmd. 
I'll also update the configure.ac file later.

PR #5, opened 2016-04-15
Reinit

kcudnik:


syncd/syncd.cpp
```diff
@@ -554,26 +859,34 @@ int main(int argc, char **argv)
...
-        SYNCD_LOG_ERR("fail to sai_initialize_switch: %lld", status);
+        SWSS_LOG_ERROR("fail to sai_initialize_switch: %d", status);
         exit(EXIT_FAILURE);
     }
 
-    std::thread bcm_diag_shell_thread = std::thread(sai_diag_shell);
-    bcm_diag_shell_thread.detach();
+    for (int i = 0; i < argc; i++)
+    {
+        if (strcmp(argv[i],"--diag") == 0)
```
lguohan: use getopt?

kcudnik: Yes, i wanted to join hrachyam since he should work on that so i wanted to switch to getopt when his implementation would be available

syncd/syncd_reinit.cpp
```diff
@@ -0,0 +1,558 @@
...
+            exit(EXIT_FAILURE);
+        }
+    }
+}
+
+sai_object_id_t redisGetDefaultVirtualRouterId()
+{
+    SWSS_LOG_ENTER();
+
+    auto redisVrId = g_redisClient->hget(HIDDEN, DEFAULT_VIRTUAL_ROUTER_ID);
```
lguohan: I am not sure I quite follow the need to have this HIDDEN table? why not put them into the ASIC view? Because, there is not set function?

kcudnik: yes, thats the point, that those objcts are created inside sai implementation and are "hidden"

lguohan:
ack

PR #6, opened 2016-04-15
make: adding --with-sai arg to select which SAI to compile with

stcheng:


kcudnik:
We had this param previously and we removed it why to add it again ?

kcudnik:
Ack

PR #8, opened 2016-04-19
pkg: using environmental variable to decide which SAI to compile

stcheng:


kcudnik:
Ack

debian/rules
```diff
@@ -16,6 +16,7 @@ include /usr/share/dpkg/default.mk
 # package maintainers to append LDFLAGS
 #export DEB_LDFLAGS_MAINT_APPEND = -Wl,--as-needed
 
+DEB_SAIREDIS_FLAGS ?= "--with-sai=bcm"
```
lguohan: name to generic? 

how about DEB_SAI_FLAVOR ?= bcm

debian/rules
```diff
@@ -28,7 +29,7 @@ include /usr/share/dpkg/default.mk
 #	-DCMAKE_LIBRARY_PATH=$(DEB_HOST_MULTIARCH)
 
 override_dh_auto_configure:
-	dh_auto_configure --
+	dh_auto_configure -- $(DEB_SAIREDIS_FLAGS)
```
lguohan: dh_auto_configure -- --with-sai=$(DEB_SAI_FLAVOR)

stcheng:
updated

PR #9, opened 2016-04-19
Add asic compile/switch skeleton in redis/syncd

kcudnik:


lib/src/sai_redis_switch.cpp
```diff
@@ -1,9 +1,17 @@
 #include "sai_redis.h"
 #include <thread>
 
+// if we will not get response in 60 seconds when
+// notify syncd to compile new state or to switch
+// to compiled state, then there is something wrong
+#define NOTIFY_SYNCD_TIMEOUT (60*1000)
+
+#define NOTIFY_COMPILE  "compile"
+#define NOTIFY_SWITCH   "switch"
```
lguohan: is this switch_sai_view, and compile for compile_sai_view. how about using the full name for clarity?

lguohan:
ack

lib/src/sai_redis_switch.cpp
kcudnik: yes, i can change that

PR #12, opened 2016-04-28
Add port counters collector thread

kcudnik:


syncd/syncd_counters.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+    swss::DBConnector db(COUNTERS_DB, "localhost", 6379, 0);
+    swss::Table countersTable(&db, "COUNTERS");
+
+    while(g_runCountersThread)
+    {
+        collectCounters(countersTable);
+
+        // collect counters every second
+        sleep(1);
```
lguohan: make this a tunable parameter?

lguohan: make the interval a tunable parameter?

syncd/syncd_counters.cpp
```diff
@@ -0,0 +1,299 @@
...
+std::string getCounterName(sai_port_stat_counter_t counter)
+{
+    if (counter < g_counterNameVector.size())
+        return g_counterNameVector[counter];
+
+    // TODO exit here ?
+    return std::to_string(counter);
+}
+
+sai_port_stat_counter_t collectCountersIds[] = {
```
lguohan: not sure if mlnx supports all these counters. I remember mlnx might not support some of those counters, for example IF_OUT_QLEN.

dplore: Is there a ‘get sai capabilities’ method or some way a SAI implementation can express which attributes it supports?

lguohan: not yet. one thing we can do is to do a query at the startup time, if the return value is not_implementated or not supported, then do not query it any more.

kcudnik: should that query be per port ? can different port support different counters? or can i query one port and assume all other ports supports the same counters ?

lguohan:
ack

PR #19, opened 2016-05-10
common: Removing deprecated SAI type VLAN_PORT_LIST

stcheng:


lguohan:
:shipit:

kcudnik:
Object id list will be used instead of that ?

stcheng:
this is replaced by vlan member i think

kcudnik:
Ack

PR #20, opened 2016-05-11
Add support for syncd first run detection

kcudnik:


stcheng:
I use "sai_init_view" and "sai_apply_view" to indicate: creating a new view, and actually applying the view to the ASIC. Which seems to be more readable from my point of view.

kcudnik:
Lets sync with Guohan offline on that

stcheng:
sai_init_view & sai_apply_view :) that's what i propose

kcudnik:
Added in another pull request

stcheng:
ack. regenerate this pull request against the newest master.

PR #29, opened 2016-07-07
remove MLNX/BRCM specific initialization code

lguohan:


syncd/syncd.cpp
```diff
@@ -1167,7 +1161,7 @@ int main(int argc, char **argv)
 
     initialize_common_api_pointers();
 
-    sai_status_t status = sai_switch_api->initialize_switch(0, "0xb850", "", &switch_notifications);
+    sai_status_t status = sai_switch_api->initialize_switch(0, "", "", &switch_notifications);
```
kcudnik: is this value not needed for bcm ?

kcudnik:
Where then will be specific code executed ?

stcheng:
I think an option -p will be added to provide the profile.

stcheng:
ACK

syncd/syncd.cpp
lguohan: Not needed.

PR #41, opened 2016-07-21
Introduce rpc server

pavel-shirshov:
Introducing support of saithrift rpc server for SONiCv2

### DIFF
--- a/syncd/Makefile.am
+++ b/syncd/Makefile.am
@@ -37,7 +37,7 @@ if SAITHRIFT
 SAI_RPC_INCLUDE_DIR = ../libsaiserver/include
 SAI_RPC_LIB_DIR = ../libsaiserver/lib
 syncd_CPPFLAGS += -I$(SAI_RPC_INCLUDE_DIR) -DSAITHRIFT=yes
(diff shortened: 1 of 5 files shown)


syncd/Makefile.am
```diff
@@ -33,6 +33,12 @@ syncd_CPPFLAGS = $(DBGFLAGS) $(AM_CPPFLAGS) $(CFLAGS_COMMON) \
 				 -I/usr/include/sai $(SAIFLAGS)
 syncd_LDADD = -lhiredis -lswsscommon -lsai -lpthread
 
+if SAITHRIFT
+SAI_RPC_INCLUDE_DIR = ../libsaiserver/include
+SAI_RPC_LIB_DIR = ../libsaiserver/lib
+syncd_CPPFLAGS += -I$(SAI_RPC_INCLUDE_DIR) -DSAITHRIFT=yes
+syncd_LDADD += -L$(SAI_RPC_LIB_DIR) -lrpcserver -lthrift
```
lguohan: >-lthrift [](start = 48, length = 8)

do we need thrift explicitly? I thought rpcserver already has a dependency on that.

PR #42, opened 2016-07-22
Introduce rpc server

pavel-shirshov:
Second attempt.

-lthrift is required in Makefile. rpcserver needs it too on link stage.


lguohan:
Ack

PR #46, opened 2016-07-25
bug fix: fix syncd.init undefined variables when stopping daemons

stcheng:


debian/syncd.init
```diff
@@ -16,36 +16,39 @@ DAEMON_SYNCD=/usr/bin/syncd
...
+start_mlnx()
+{
+    mkdir -p /dev/sxdevs
+    [ -e /dev/sxdevs/sxcdev ] || mknod /dev/sxdevs/sxcdev c 231 193
 }
 
 case "$1" in
 start)
-    . /etc/machine.conf
+    [ -x /etc/machine.conf ] && . /etc/machine.conf
```
lguohan: /etc/machine.conf does not need to be excutable

stcheng: updated.

lguohan:
Ack

PR #50, opened 2016-07-30
[sai-redis] Add policer and Trap Group serialization

hrachya-mellanox:
* Added policer and trap group serialization for COPP.
Signed-off-by: hrachya@mellanox.com


lguohan:
Ack

PR #52, opened 2016-08-06
Patch init script with saithrift start parameters.

pavel-shirshov:
Add portmap file for saithrift server


debian/rules
```diff
@@ -28,5 +28,23 @@ DEB_SAI_FLAVOR ?= "bcm"
...
+
+override_dh_install:
+  ifeq ($(saithrift), no)
+	rm -f etc/syncd.d/msn2700_port_map.ini
+  endif
+	dh_install
+
+override_dh_installinit:
+  ifeq ($(saithrift), yes)
+	sed -i 's/SAITHRIFT_ARGS_MLNX=""/SAITHRIFT_ARGS_MLNX="-r -m \/etc\/syncd.d\/msn2700_port_map.ini"/' debian/syncd.init
```
lguohan: it is probably more readable to setup ENABLE_SAITHRIFT=1 in syncd.init here, and in syncd.init 

if [ $(enable_saishfit) == 1]; then DAEMON_ARGS += "-r -m /etc/syncd.d/msn2700_port_map.ini"; fi.

and

debian/rules
```diff
@@ -28,5 +28,23 @@ DEB_SAI_FLAVOR ?= "bcm"
...
+ifeq ($(findstring --enable-rpcserver=yes,${CONFIGURE_ARGS}),--enable-rpcserver=yes)
+  saithrift = yes
+else
+  saithrift = no
+endif
+
 override_dh_auto_configure:
 	dh_auto_configure -- --with-sai=$(DEB_SAI_FLAVOR) ${CONFIGURE_ARGS}
+
+override_dh_install:
```
lguohan: that's fine. we can copy port_map to the deb package no matter what.

etc/syncd.d/msn2700_port_map.ini
```diff
@@ -0,0 +1,33 @@
+# alias          lanes
```
stcheng: do we want to put the file here or use ansible to deploy the file like this and pass the path to the daemon

pavel-shirshov: I'd say it's better to have everything in the package. Otherwise why do we need this package? We can deploy our application using ansible only without any packages.

lguohan:
Ack

PR #57, opened 2016-08-16
Change vid db format for counters

kcudnik:


stcheng:
I am okay with the current code change. For a long term consideration, I would propose a serialization/deserialization functionality put inside the common library so that all types of data can be stored as the same format in the database from all applications.

PR #61, opened 2016-08-18
Add basic param validation for several SAI objects

kcudnik:


lguohan:
are these validations also applied to virtual switch?

stcheng:
this validations happen on sairedis layer.

kcudnik:
Not direclty, but via syncd, but i also though how to make those apply on each sai, but i coudnt found right solution yet, expect a separate intermidiate library, lets meet today, i have few question about tunell

kcudnik:
Merged this change to metadata branch

PR #62, opened 2016-08-20
Add OA switch_shutdown_request notification on syncd exit

kcudnik:
When syncd will exit unexpectedly because of db failure or any SAI api failure it will send notification to OA so it can know and respond and take action. This notification is not guaranteed to be send on unexpected exit or crash.


stcheng:
ACK

PR #64, opened 2016-08-23
Introduced Cavium platform

garyachy:


garyachy:
@stcheng pls review

syncd/syncd.cpp
```diff
@@ -40,6 +40,7 @@ void exit_and_notify(int status)
     exit(status);
 }
 
+#ifdef BRCMSAI
```
stcheng: there's ifdef on line 1325, no need to add more here.

garyachy: This code breaks compilation as far as SAI_SWITCH_ATTR_CUSTOM_RANGE_BASE is not supported anymore.

kcudnik: SAI_SWITCH_ATTR_CUSTOM_RANGE_BASE  was changed to SAI_SWITCH_ATTR_CUSTOM_RANGE_START 

simple change could satisfy both: and later on when sai will be stable it can be changed back

int32_t id = -1;

#ifndef SAI_SWITCH_ATTR_CUSTOM_RANGE_BASE 
id = SAI_SWITCH_ATTR_CUSTOM_RANGE_BASE;
#elsif  SAI_SWITCH_ATTR_CUSTOM_RANGE_START
id = SAI_SWITCH_ATTR_CUSTOM_RANGE_START;  
#else
#error invalid atrib custom range not defined
#endif

garyachy: Please advise why other vendors need to compile in BRCM specific code?

Denys.

stcheng: I didn't realize that SAI_SWITCH_ATTR_CUSTOM_RANGE_BASE is not supported. Will merge it.

PR #70, opened 2016-09-19
Add static metadata

kcudnik:


stcheng:
Notice it is hard to review it.
Sorry, we could not display the entire diff because it was too big. - github

kcudnik:
Can you review it offline or should i split it file by file?

stcheng:
will submit review asap

meta/sai_meta.h
```diff
@@ -0,0 +1,791 @@
...
+    /*
+     * Attribute with this flag indicates that value of the attribute is
+     * dynamic and can change in time (like a attribute counter value, or port
+     * operational status). Change may happen independently or when other
+     * attribute was created or modified (creating vlan member will change vlan
+     * member list). Can be combined with READ_ONLY flag. In future it maybe
+     * possible to combine this attribute with other flags.
+     */
+
+    //SAI_ATTR_FLAGS_DYNAMIC             = (1 << 5),
```
stcheng: these flags are commented but are used below?

meta/sai_meta.cpp
```diff
@@ -0,0 +1,4884 @@
...
+
+    sai_object_type_t object_type = SAI_OBJECT_TYPE_NULL;
+
+    for (uint32_t i = 0; i < count; ++i)
+    {
+        sai_object_id_t oid = list[i];
+
+        if (oids.find(oid) != oids.end())
+        {
+            META_LOG_ERROR(md, "object on list [%u] oid 0x%llx is duplicated, but not allowed", i, oid);
```
erohsik, line 847: @kcudnik Can you please provide some context on this check?

kcudnik: this checks whether duplicated oids are on the list of the parameters, this is fdom 2016 code, why intrest on it now?

erohsik: I was trying to find some context on this check
https://github.com/sonic-net/sonic-sairedis/blob/master/meta/Meta.cpp#L6188
and followed the git history to get the original PR.
I was trying to get some reasoning behind this check. Is it invalid for an OID to be repeated in an OID list?

kcudnik: it make no sense to repeat oid on the list

PR #72, opened 2016-09-28
Handle default trap group and remove vlan members

kcudnik:


syncd/syncd.cpp
```diff
@@ -846,7 +846,7 @@ swss::Logger::Priority redisGetLogLevel()
     auto plevel = g_redisClient->get("LOGLEVEL");
 
     if (plevel == NULL)
-        return swss::Logger::SWSS_NOTICE;
+        return swss::Logger::SWSS_DEBUG;
```
stcheng: keep NOTICE

syncd/syncd.cpp
```diff
@@ -1250,7 +1250,7 @@ bool isVeryFirstRun()
 
 int main(int argc, char **argv)
 {
-    swss::Logger::getInstance().setMinPrio(swss::Logger::SWSS_NOTICE);
+    swss::Logger::getInstance().setMinPrio(swss::Logger::SWSS_DEBUG);
```
stcheng: keep NOTICE

stcheng:
let's do some basic test before merging

syncd/syncd_hard_reinit.cpp
```diff
@@ -157,10 +157,56 @@ void checkAllIds()
...
+    vlanMemberList.resize(portList.size());
+
+    sai_attribute_t attr;
+
+    attr.id = SAI_VLAN_ATTR_MEMBER_LIST;
+
+    attr.value.objlist.count = vlanMemberList.size();
+    attr.value.objlist.list = vlanMemberList.data();
+
+    sai_status_t status = sai_vlan_api->get_vlan_attribute(DEFAULT_VLAN_NUMBER, 1, &attr);
```
kcudnik: than it still should return zero members on list?

syncd/syncd.cpp
kcudnik: fixed

syncd/syncd.cpp
kcudnik: fixed

syncd/syncd_hard_reinit.cpp
lguohan: then you need resize the vlanMemberList otherwise, in line 191 it enumernate many members.

stcheng: +1 to resize again

kcudnik: ah yes

PR #74, opened 2016-09-30
Add api recording capability and fix getrequest

kcudnik:


lib/src/sai_redis_record.cpp
```diff
@@ -0,0 +1,105 @@
...
+std::string getTimestamp()
+{
+    SWSS_LOG_ENTER();
+
+    char buffer[64];
+    struct timeval tv;
+
+    gettimeofday(&tv, NULL);
+
+    size_t size = strftime(buffer, 64 ,"%Y-%m-%d.%T.", localtime(&tv.tv_sec));
```
qiluo-msft: 64 - 6, to make later sprintf very safe?

lib/src/sai_redis_record.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+std::ofstream recording;
+
+void recordLine(std::string s)
+{
+    SWSS_LOG_ENTER();
+
+    if (recording.is_open())
+    {
+        recording << getTimestamp() << "," << s << std::endl;
```
qiluo-msft: std::endl includes a flush()?

lib/src/sai_redis_record.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+        // performance hit
+        recording.flush();
+    }
+}
+
+std::string recfile = "dummy.rec";
```
qiluo-msft: User may expect to specif the file name and path. Currently it is in the working directory, so it may be possible to created in /.

lib/src/sai_redis_record.cpp
kcudnik: I dont get this comment, why -6 ?

lib/src/sai_redis_record.cpp
kcudnik: yes, forgot about that

lib/src/sai_redis_record.cpp
kcudnik: currently every start recording user gets new file name based on timestamp, later we could provide some interface to add abbility for user to specify desired filename, this is not priority right now

lib/src/sai_redis_record.cpp
qiluo-msft: because sprintf print a string of length 6, appending in the same buffer.

kcudnik:
Ah i see what you mean, agree

lib/src/sai_redis_record.cpp
qiluo-msft: cloesd.

common/saiserialize.cpp
```diff
@@ -119,7 +119,10 @@ sai_serialization_map_t sai_get_serialization_map()
     map[SAI_OBJECT_TYPE_SWITCH][SAI_SWITCH_ATTR_INGRESS_BUFFER_POOL_NUM] = SAI_SERIALIZATION_TYPE_UINT32;
     map[SAI_OBJECT_TYPE_SWITCH][SAI_SWITCH_ATTR_QOS_MAX_NUMBER_OF_CHILDS_PER_SCHEDULER_GROUP] = SAI_SERIALIZATION_TYPE_INT32;
     map[SAI_OBJECT_TYPE_SWITCH][SAI_SWITCH_ATTR_DEFAULT_TRAP_GROUP] = SAI_SERIALIZATION_TYPE_OBJECT_ID;
+
+#ifdef BRCMSAI
```
stcheng: no need to have BRCMSAI specific

qiluo-msft:
Please also check with other reviewers.

PR #76, opened 2016-10-01
Remove metadata generator

kcudnik:


stcheng:
this will be moved to SAI repo?

kcudnik:
No, this was generator used to generate metadata for sai headers, since we already have them its not needed

PR #79, opened 2016-10-06
Support default queues and priority groups

kcudnik:


syncd/syncd.cpp
```diff
@@ -1248,10 +1248,12 @@ bool isVeryFirstRun()
 
 int main(int argc, char **argv)
 {
-    swss::Logger::getInstance().setMinPrio(swss::Logger::SWSS_NOTICE);
+    swss::Logger::getInstance().setMinPrio(swss::Logger::SWSS_DEBUG);
```
lguohan: keep it as notice?

syncd/syncd.cpp
```diff
@@ -1248,10 +1248,12 @@ bool isVeryFirstRun()
...
 
     SWSS_LOG_ENTER();
 
+    swss::Logger::getInstance().setMinPrio(swss::Logger::SWSS_NOTICE);
+
```
lguohan: why do we need to setMinPrio here?

syncd/syncd_reinit.cpp
```diff
@@ -610,6 +617,174 @@ void helperCheckVlanId()
...
+
+std::vector<sai_object_id_t> saiGetPortQueues(sai_object_id_t portId)
+{
+    SWSS_LOG_ENTER();
+
+    uint32_t queueCount = saiGetPortNumberOfQueues(portId);
+
+    std::vector<sai_object_id_t> queueList;
+
+    queueList.resize(queueCount);
```
lguohan: if the queueCount is zero, we can just return?

syncd/syncd_reinit.cpp
```diff
@@ -610,6 +617,174 @@ void helperCheckVlanId()
...
+
+std::vector<sai_object_id_t> saiGetPortPriorityGroups(sai_object_id_t portId)
+{
+    SWSS_LOG_ENTER();
+
+    uint32_t pgCount =  saiGetPortNumberOfPriorityGroups(portId);
+
+    std::vector<sai_object_id_t> pgList;
+
+    pgList.resize(pgCount);
```
lguohan: if pgCount is zero, can we just return from here?

syncd/syncd.cpp
kcudnik: i use it becasue i want "main" entry to be logged as debug

syncd/syncd_reinit.cpp
kcudnik: not significant optimalization

kcudnik: and it will never be zero

syncd/syncd_reinit.cpp
kcudnik: and it will never be zero

syncd/syncd_hard_reinit.cpp
```diff
@@ -323,7 +327,56 @@ sai_object_id_t processSingleVid(sai_object_id_t vid)
...
+        {
+            // this is one of default priority group id
+            // we don't need to create it, just set attributes if any
+
+            rid = pgRid;
+
+            createObject = false;
+
+            SWSS_LOG_DEBUG("default priority group will not be created, processed VID %llx to RID %llx", vid, rid);
+        }
```
qiluo-msft: The 3 branches above are quite similar, is it possible to reuse most of the code?

kcudnik: not yet, after there will be need to add something similar

syncd/syncd_hard_reinit.cpp
```diff
@@ -140,9 +140,11 @@ void checkAllIds()
         g_vidToRidMap.erase(it);
     }
 
-    if (g_vidToRidMap.size() != 0)
+    size_t size = g_vidToRidMap.size();
+
+    if (size != 0)
     {
-        SWSS_LOG_ERROR("vid to rid map is not empty after translation");
+        SWSS_LOG_ERROR("vid to rid map is not empty (%lu) after translation", size);
```
kcudnik: fixed

syncd/syncd.cpp
kcudnik: it is notice after ENTER

syncd/syncd_hard_reinit.cpp
qiluo-msft: Closed.

syncd/syncd_reinit.cpp
lguohan: yes, on barefoot p4 switch, it is zero.

lguohan: I think it's better to skip if the count is zero. I am more worried about hitting bugs in the sai.

kcudnik: what ? is zero ? then why there is inconsistency whit that ? as you said before how we can't have queues ?

syncd/syncd_reinit.cpp
```diff
@@ -64,6 +65,7 @@ std::vector<sai_object_id_t> saiGetPortList()
     attr.value.objlist.count = portCount;
     attr.value.objlist.list = portList.data();
 
+    // we assube port list is always returned in the same order
```
stcheng: assume

kcudnik: yea, so far there are many assumptions :P

stcheng: :D

PR #81, opened 2016-10-06
change remove MANDATORY_ON_CREATE for SAI_ROUTE_ATTR_NEXT_HOP_ID

lguohan:


meta/tests.cpp
```diff
@@ -2000,11 +2000,6 @@ void test_route_entry_create()
 
     SWSS_LOG_NOTICE("create tests");
 
-    SWSS_LOG_NOTICE("zero attribute count (but there are mandatory attributes)");
-    attr.id = SAI_ROUTE_ATTR_NEXT_HOP_ID;
-    status = meta_sai_create_route_entry(&route_entry, 0, &attr, &dummy_success_sai_create_route_entry);
-    META_ASSERT_FAIL(status);
-
     SWSS_LOG_NOTICE("attr is null");
```
kcudnik: comment this out instead remove ?

lguohan: is there any you want to me to add, i can do that.

PR #82, opened 2016-10-10
add a7050 qx32 platform support

lguohan:


debian/syncd.init
```diff
@@ -48,6 +48,10 @@ start)
         start_bcm
         DAEMON_ARGS+=" -p /etc/syncd.d/dell_s6100.profile -N"
 
+    elif [ "$aboot_platform" == "x86_64-arista_7050_qx32" ]; then
+        start_bcm
+        DAEMON_ARGS+=" -p /etc/syncd.d/arista_a7050_qx32.profile"
```
qiluo-msft: Remove the leading blank? The same to above.

stcheng: the blank line makes it clear to read?

qiluo-msft: I mean the blank before "-p"

qiluo-msft: Then the last branch is not consistent.

stcheng: hmm yes. need to be consistent on adding space for this

PR #85, opened 2016-10-11
Fix serialization and refactor virtual switch init logic

kcudnik:


vslib/src/sai_vs_generic_create.cpp
```diff
@@ -86,6 +105,8 @@ sai_status_t vs_generic_create(
 {
     SWSS_LOG_ENTER();
 
+    std::string otstr;
```
qiluo-msft: Not used.

vslib/src/sai_vs_vlan.cpp
```diff
@@ -78,10 +161,23 @@ sai_status_t vs_remove_vlan_member(
...
+    sai_status_t status = meta_sai_remove_oid(
             SAI_OBJECT_TYPE_VLAN_MEMBER,
             vlan_member_id,
             &vs_generic_remove);
+
+    if (status == SAI_STATUS_SUCCESS)
+    {
+        auto it = vlan_members_map[vlan_id].find(vlan_member_id);
+
+        vlan_members_map[vlan_id].erase(it);
```
qiluo-msft: Will it be set::end? If never, add assertion?

kcudnik: no, as you can se its only executed after success, so vlan member id will exists

vslib/src/sai_vs_generic_create.cpp
qiluo-msft: Closed.

syncd/syncd.cpp
```diff
@@ -1180,21 +1223,21 @@ void handlePortMap(const std::string& portMapFile)
...
-        std::string fp_value = line.substr(0, pos);		
+
+        std::string fp_value = line.substr(0, pos);
         std::string lanes    = line.substr(pos + 1);
         lanes.erase(lanes.begin(), std::find_if(lanes.begin(), lanes.end(), std::not1(std::ptr_fun<int, int>(std::isspace))));
         std::istringstream iss(lanes);
         std::string lane_str;
         std::set<int> lane_set;
-		
+
```
stcheng: you could use tokenize() function in swss-common

vslib/src/sai_vs_init.cpp
```diff
@@ -130,135 +127,349 @@ sai_status_t initialize_default_objects()
...
+{
+    SWSS_LOG_ENTER();
+
+    SWSS_LOG_INFO("create vlan members for all ports");
+
+    vlan_member_list.clear();
+
+    vlan_members_map[DEFAULT_VLAN_NUMBER] = {};
+
+    for (auto &port_id : port_list)
```
stcheng: i doubt if we will add this to vs behavior. default vlan 1 needs to be there, but we could have 0 members in the default vlan. the behavior of adding all ports into default vlan by default is optional.

PR #86, opened 2016-10-11
Add metadata for SAI_INGRESS_PRIORITY_GROUP_ATTR_BUFFER_PROFILE

kcudnik:


meta/tests.cpp
```diff
@@ -3284,6 +3284,26 @@ void test_null_list()
     META_ASSERT_SUCCESS(status);
 }
 
+void test_priority_group()
+{
+    SWSS_LOG_ENTER();
+
+    meta_init_db();
+
+    sai_status_t    status;
```
stcheng: extra spaces

meta/tests.cpp
```diff
@@ -3284,6 +3284,26 @@ void test_null_list()
...
+
+    sai_attribute_t attr;
+
+    sai_object_id_t pg = insert_dummy_object(SAI_OBJECT_TYPE_PRIORITY_GROUP);
+
+    SWSS_LOG_NOTICE("set");
```
stcheng: add more meaningful logs

meta/tests.cpp
kcudnik: used to have same alignemnt with attr and pg, but fixed

PR #87, opened 2016-10-12
Add sai recorded log player

kcudnik:


player/player.cpp
```diff
@@ -0,0 +1,1144 @@
...
+{
+    on_switch_state_change,
+        on_fdb_event,
+        on_port_state_change,
+        on_port_event,
+        on_switch_shutdown_request,
+        on_packet_event
+};
+
+#define EXIT_ON_ERROR(x)\
```
qiluo-msft: Move this function to a header? I met it in previous code review.

player/player.cpp
```diff
@@ -0,0 +1,1144 @@
...
+{\
+    sai_status_t s = (x);\
+    if (s != SAI_STATUS_SUCCESS)\
+    {\
+        SWSS_LOG_ERROR("fail status %d", s);\
+        exit(EXIT_FAILURE);\
+    }\
+}
+
+std::vector<std::string> split(const std::string &s, char delim = ',')
```
qiluo-msft: Same as https://github.com/Azure/sonic-swss-common/blob/master/common/tokenize.cpp?

player/player.cpp
```diff
@@ -0,0 +1,1144 @@
...
+        sai_attr_serialization_type_t serialization_type;
+        sai_status_t status = sai_get_serialization_type(object_type, attr.id, serialization_type);
+
+        if (status != SAI_STATUS_SUCCESS)
+        {
+            SWSS_LOG_ERROR("unable to find serialization type for object type %x, attribute %x", object_type, attr.id);
+            exit(EXIT_FAILURE);
+        }
+
+#define CHECK_LIST(x)\
```
qiluo-msft: Define macro function outside any function?

player/player.cpp
```diff
@@ -0,0 +1,1144 @@
...
+    }
+
+    if (oid != redis_to_local[get_oid])
+    {
+        SWSS_LOG_ERROR("match failed, oid order is mismatch :( oid 0x%lx get_oid 0x%lx second 0x%lx",
+                oid,
+                get_oid,
+                redis_to_local[get_oid]);
+
+        exit(1);
```
qiluo-msft: magic number?

player/player.cpp
```diff
@@ -0,0 +1,1144 @@
...
+            handle_get_response(object_type, attr_count, attr_list, response);
+        }
+    }
+
+    infile.close();
+
+    return 0;
+}
+
+int main(int argc, char **argv)
```
qiluo-msft: suggest to add a 'help' parameter and output usage.

player/player.cpp
kcudnik: there was return_on_error if i remember which does different thing

player/player.cpp
kcudnik: fixed

player/player.cpp
kcudnik: fixed

player/player.cpp
kcudnik: added

player/player.cpp
qiluo-msft: Closed.

qiluo-msft:
:clock1:

player/player.cpp
qiluo-msft: So it makes sense to collect similar things together.

qiluo-msft: Ignore this comment if you cannot find it.

PR #98, opened 2016-10-24
Add more static metadata

kcudnik:


meta/saiserialize.h
```diff
@@ -77,7 +78,7 @@ typedef enum _sai_attr_serialization_type_t
     SAI_SERIALIZATION_TYPE_TUNNEL_MAP_LIST
 
 } sai_attr_serialization_type_t;
-
+*/
```
stcheng: since they are moved to sai_meta.h, why not remove it from here?

meta/sai_meta_hostintf.cpp
```diff
@@ -92,6 +92,117 @@ const char* metadata_sai_hostif_type_t_enum_values_short_names[] = {
...
+    "MLD_V2_REPORT",
+    "IP2ME",
+    "SSH",
+    "SNMP",
+    "L3_MTU_ERROR",
+    "TTL_ERROR",
+    "CUSTOM_EXCEPTION_RANGE_BASE",
+    NULL
+};
+const size_t metadata_sai_hostif_trap_type_t_enum_values_count = 33;
```
stcheng: use the size of the char pointer array instead of using number here?

meta/saiserialize.cpp
```diff
@@ -1770,3 +1770,28 @@ void sai_deserialize_route_entry(
 
     sai_deserialize_ip_prefix(s, index, re.destination);
 }
+
+// new methods
+
```
stcheng: why add these two functions?

meta/saiserialize.h
kcudnik: it will be removed in one of next pr

meta/sai_meta_hostintf.cpp
kcudnik: this is temprorary its copy from auto generated metadata and it needs to be this way since value needs to be known on compile time

meta/saiserialize.cpp
kcudnik: since they will be needed !

meta/sai_meta_sanity.cpp
```diff
@@ -837,6 +950,29 @@ std::string get_attr_info(const sai_attr_metadata_t& md)
...
+
+        auto it = attrset.find(attrid);
+
+        if (it != attrset.end())
+        {
+            return it->second;
+        }
+    }
+
+    SWSS_LOG_ERROR("unable to find metadata for object type %d attr id %d", objecttype, attrid);
```
lguohan: is it WARNING or ERROR?

kcudnik: error, since we always should find attribute id

PR #100, opened 2016-10-25
Convert meta to saimetadata library

kcudnik:


Makefile.am
```diff
@@ -1 +1 @@
-SUBDIRS = lib vslib syncd
+SUBDIRS = meta lib vslib syncd player
```
lguohan: which package will include this player binary? syncd.deb?

kcudnik: there will be separate package saimetadata.deb

lguohan: player seems to be counterpart of syncd. Maybe better to be syncd? saimetadata seems to be metadata checking library that can be used by others?

PR #104, opened 2016-10-26
Remove SAI_SWITCH_ATTR_CUSTOM_RANGE_START in player

stcheng:


kcudnik:
why you want to remoev it ?

stcheng:
1. this attribute is deprecated and no longer in use and not fully supported across different platforms
2. the default is false so no need to add it here.

kcudnik:
this attribute is internal usage only by sairedis

stcheng:
updated to use RANGE_BASE instead of using RANGE_START

PR #107, opened 2016-10-27
Update .gitignore file

stcheng:


.gitignore
```diff
@@ -13,6 +13,12 @@ debian/libsairedis.debhelper.log
 debian/libsairedis.postinst.debhelper
 debian/libsairedis.postrm.debhelper
 debian/libsairedis.substvars
+debian/libsaimetadata-dev.debhelper.log
```
qiluo-msft: Make it more general so we can reuse it?

PR #109, opened 2016-10-28
Add qos map serialization

kcudnik:


meta/saiserialize.cpp
```diff
@@ -827,6 +835,74 @@ std::string sai_serialize_number_list(
...
+    return j.dump();
+}
+
+std::string sai_serialize_tunnel_map_list(
+        _In_ const sai_qos_map_list_t& qosmap,
+        _In_ bool countOnly)
+{
+    SWSS_LOG_ENTER();
+
+    // TODO
```
qiluo-msft: What is to do?

kcudnik: TODO later on populate this function

PR #110, opened 2016-10-29
increase the max list count to 256

lguohan:


meta/sai_meta.h
```diff
@@ -15,7 +15,7 @@ extern "C" {
 
 #include "swss/logger.h"
 
-#define MAX_LIST_COUNT 128
+#define MAX_LIST_COUNT 256
```
kcudnik: why you need more objects on list ? and where ?

lguohan:
P4 switch, they have 256 ports

meta/sai_meta.h
kcudnik: ok

PR #112, opened 2016-11-02
Add serialization support for tunnel map

kcudnik:


syncd/syncd.cpp
```diff
@@ -1269,7 +1269,7 @@ bool isVeryFirstRun()
 
 int main(int argc, char **argv)
 {
-    swss::Logger::getInstance().setMinPrio(swss::Logger::SWSS_NOTICE);
+    swss::Logger::getInstance().setMinPrio(swss::Logger::SWSS_DEBUG);
```
qiluo-msft: Simplify this pattern?
   
    swss::Logger::getInstance().setMinPrio(swss::Logger::SWSS_DEBUG);
    SWSS_LOG_ENTER();
    swss::Logger::getInstance().setMinPrio(swss::Logger::SWSS_NOTICE);

kcudnik: what pattern ?

lguohan:
The build is failing. it seems like this PR is not rebased to master. can you rebase this PR so that it can pass the test?

kcudnik:
When i click on details i get page not found dont know what was wrong without logs

PR #114, opened 2016-11-07
debian: Add player into syncd package

stcheng:


debian/syncd.install
```diff
@@ -1,2 +1,3 @@
 usr/bin/syncd*
+usr/bin/player*
```
qiluo-msft: Just one concern, is this name too general?

stcheng: what about record-player?

lguohan: sai-player?

stcheng:
updated.

saiplayer/Makefile.am
```diff
@@ -0,0 +1,13 @@
+AM_CPPFLAGS = -I/usr/include/sai
```
qiluo-msft: You moved the file, and delete one -I, which break the build

PR #116, opened 2016-11-07
Add asicdump tool to view asic state

kcudnik:


asicdump/asicdump.cpp
```diff
@@ -0,0 +1,291 @@
...
+struct CmdOptions
+{
+    bool shortNames;
+    bool disableColors;
+    bool skipAttributes;
+    bool followReferences;
+};
+
+CmdOptions g_cmdOptions;
+std::map<sai_object_id_t, const TableMap*> g_oid_map;
```
qiluo-msft: >std:: [](start = 0, length = 5)

'std::' can be removed.

asicdump/asicdump.cpp
```diff
@@ -0,0 +1,291 @@
...
+    return options;
+}
+
+size_t get_max_attr_len(const TableMap& map)
+{
+    SWSS_LOG_ENTER();
+
+    size_t max = 0;
+
+    for (const auto&field: map)
```
qiluo-msft: add a blank?

asicdump/asicdump.cpp
kcudnik: i use std:: every time in cpp

asicdump/asicdump.cpp
kcudnik: ??

asicdump/asicdump.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+        ss << field.second;
+
+        std::cout << ss.str() << std::endl;
+    }
+}
+
+int main(int argc, char ** argv)
+{
+    swss::Logger::getInstance().setMinPrio(swss::Logger::SWSS_DEBUG);
```
kcudnik: i want log enter to be logged in main to know when new process started, i will add leter on other notice enter

Makefile.am
```diff
@@ -1 +1 @@
-SUBDIRS = meta lib vslib syncd player
+SUBDIRS = meta lib vslib syncd player asicdump
```
lguohan: mellanox has a sdkdump to dump all sdk state, to differentiate with sdkdump, can we call it saidump? for the player, can we call it saiplayer?

stcheng: i'll change player to saiplayer

asicdump/asicdump.cpp
qiluo-msft: auto &field

or 

    auto& field

kcudnik:
since this is strictly combined with redis maybe sairedisdump and sairedisplayer ?

PR #118, opened 2016-11-08
Update redis and syncd with notify logic

kcudnik:


lib/src/sai_redis_switch.cpp
```diff
@@ -1,19 +1,12 @@
 #include "sai_redis.h"
+#include "sairedis.h"
```
stcheng: two files with one named sai_redis.h one sairedis.h.
this might be confusing.

kcudnik: yes, i will change that later, new one will be distributes in the package, previous one is internal

player/player.cpp
```diff
@@ -1084,14 +1098,27 @@ int replay(int argc, char **argv)
...
 {
-    std::cout << "Usage: player [-h] recordfile" << std::endl;
+    std::cout << "Usage: player [-h] recordfile" << std::endl << std::endl;
+    std::cout << "    -C --skipNotifySyncd:" << std::endl;
+    std::cout << "        Will not send notify init/apply view to syncd" << std::endl << std::endl;
+    std::cout << "    -d --enableDebug:" << std::endl;
+    std::cout << "        Enable syslog debug messages" << std::endl << std::endl;
     std::cout << "    -h --help:" << std::endl;
-    std::cout << "        Print out this message" << std::endl;
+    std::cout << "        Print out this message" << std::endl << std::endl;
```
kcudnik: yes for better readability

lib/src/sai_redis_switch.cpp
stcheng: that's fine. i'll merge it. let's change that later soon.

PR #125, opened 2016-11-12
Support init and apply view in saiplayer and syncd

kcudnik:


saiplayer/saiplayer.cpp
```diff
@@ -944,8 +945,89 @@ void handle_get_response(
...
 
+void performSleep(const std::string& line)
+{
+    SWSS_LOG_ENTER();
+
+    // timestamp|action|sleeptime
+    auto v = swss::tokenize(line, '|');
+
+    uint32_t useconds;
+    sai_deserialize_number(v[2], useconds);
```
qiluo-msft: Possible index out of range? Would you like check before that?

saiplayer/saiplayer.cpp
```diff
@@ -994,6 +1067,23 @@ int replay(int argc, char **argv)
...
+            case 'a':
+                {
+                    std::string response;
+
+                    do
+                    {
+                        // this line may be notification, we need to skip
+                        std::getline(infile, response);
+                    }
+                    while (response[response.find_first_of("|")+1] == 'n');
```
qiluo-msft: Also check infile status?

PR #127, opened 2016-11-15
Add support for scheduler groups and trap group workaround

kcudnik:


syncd/syncd_hard_reinit.cpp
```diff
@@ -717,15 +817,25 @@ sai_unicast_route_entry_t getRouteEntryFromString(const std::string &strRouteEnt
...
     SWSS_LOG_ENTER();
 
+    SWSS_LOG_TIMER("apply routes");
+
     for (auto &kv: g_routes)
     {
         const std::string &strRouteEntry = kv.first;
         const std::string &asicKey = kv.second;
 
+        bool isDefault = strRouteEntry.find("\"0.0.0.0/0\"") != std::string::npos ||
```
qiluo-msft: >/0 [](start = 54, length = 2)

Just search "/0"?

syncd/syncd_hard_reinit.cpp
```diff
@@ -300,6 +301,82 @@ bool shouldSkipCreateion(
...
+    }
+    else
+    {
+        // default trap group
+        return;
+    }
+
+    sai_object_type_t objectType = SAI_OBJECT_TYPE_TRAP_GROUP;
+
+    SWSS_LOG_INFO("creating trap group and setting attributes 1 by 1 as workaroung");
```
lguohan: >workaroung [](start = 72, length = 10)

workaround

syncd/syncd_hard_reinit.cpp
```diff
@@ -717,15 +817,24 @@ sai_unicast_route_entry_t getRouteEntryFromString(const std::string &strRouteEnt
...
         const std::string &strRouteEntry = kv.first;
         const std::string &asicKey = kv.second;
 
+        bool isDefault = strRouteEntry.find("/0") != std::string::npos;
+
+        if (defaultOnly ^ isDefault)
+        {
+            continue;
+        }
+
```
lguohan: now we need to enumerate all routes in the table twice? but it looks like you have benchmarked this function, so how long does it take for this function when defaultOnly is true?

syncd/syncd_reinit.cpp
```diff
@@ -748,6 +748,182 @@ void helperCheckPriorityGroupsIds()
...
+                else
+                {
+                    SWSS_LOG_ERROR(
+                            "unexpected object type %s on scheduler group child list 0x%llx",
+                            sai_serialize_object_type(objectType).c_str(),
+                            schedGroupId);
+
+                    exit_and_notify(EXIT_FAILURE);
+                }
+            }
```
lguohan: this seems unnecessary, the childlist scheduler groups should be include in the get schedulergroup list

kcudnik: removed

PR #129, opened 2016-11-18
Add useTempView flag

kcudnik:


saiplayer/saiplayer.cpp
```diff
@@ -1241,6 +1246,10 @@ void handleCmdLine(int argc, char **argv)
                 swss::Logger::getInstance().setMinPrio(swss::Logger::SWSS_DEBUG);
                 break;
 
+            case 'u':
+                g_useTempView = false;
```
qiluo-msft: >false [](start = 32, length = 5)

true?

kcudnik: no, read description in sairedis.h

PR #132, opened 2016-11-23
Add scheduler groups support to virtual switch

kcudnik:


meta/saiserialize.cpp
```diff
@@ -84,6 +84,12 @@ sai_status_t transfer_list(
     {
         transfer_primitive(src_element.count, dst_element.count);
 
+        if (src_element.list == NULL && src_element.count > 0)
```
qiluo-msft: > if (src_element.list == NULL && src_element.count > 0) [](start = 7, length = 55)

replace src with dst, and move if-block before previous statement?

syncd/Makefile.am
```diff
@@ -20,6 +20,12 @@ if SAICAVM
 SAIFLAGS = -DCAVMSAI
 endif
 
+if SAIVS
+SAILIB=-L$(top_srcdir)/vslib/src/.libs -lsaivs
```
qiluo-msft: >.libs [](start = 33, length = 5)

.libs is not necessary. This is an example: [https://github.com/Azure/sonic-swss-common/blob/master/tests/Makefile.am](https://github.com/Azure/sonic-swss-common/blob/master/tests/Makefile.am)

vslib/src/sai_vs_init.cpp
```diff
@@ -439,6 +399,232 @@ sai_status_t create_hostif_traps()
...
+    // tree index
+    // 0 = 1 2
+    // 1 = 3 4 5 6 7 8 9 a
+    // 2 = b c (bug on brcm)
+
+    // 3..c - have both QUEUES, each one 2
+
+    // sg 0 (2 groups)
+    {
+        sai_object_id_t sg_0 = sgs.at(0);
```
qiluo-msft: >sgs [](start = 31, length = 3)

Do you have assumption on sgs.size()? Add assertion or runtime check?

vslib/src/sai_vs_init.cpp
```diff
@@ -439,6 +399,232 @@ sai_status_t create_hostif_traps()
...
+
+        CHECK_STATUS(vs_generic_set(SAI_OBJECT_TYPE_SCHEDULER_GROUP, sg_1, &attr));
+
+        // now assign queues to level 1 scheduler groups,
+
+        for (size_t i = 0; i < list.size(); ++i)
+        {
+            sai_object_id_t childs[2];
+
+            childs[0] = queues[queue_index];    // first half are in queues
```
qiluo-msft: >queue_index [](start = 31, length = 11)

i?

vslib/src/tests.cpp
```diff
@@ -44,13 +46,175 @@ service_method_table_t test_services = {
     profile_get_next_value
 };
 
+#define SUCCESS(x) \
+    if (x != SAI_STATUS_SUCCESS) \
```
qiluo-msft: >x [](start = 8, length = 1)

(x)

vslib/src/tests.cpp
```diff
@@ -44,13 +46,175 @@ service_method_table_t test_services = {
...
+    attr.id = SAI_SWITCH_ATTR_PORT_LIST;
+    attr.value.objlist.count = expected_ports;
+    attr.value.objlist.list = ports.data();
+
+    SUCCESS(sai_switch_api->get_switch_attribute(1, &attr));
+
+    ASSERT_TRUE(attr.value.objlist.count == expected_ports);
+}
+
 int main()
```
qiluo-msft: >main [](start = 4, length = 4)

Just wondering why not use gtest?

meta/saiserialize.cpp
kcudnik: no, src must be check since from it it will be copy

syncd/Makefile.am
kcudnik: its needed since i want to prefer local build lib agains the one installed in the system which is causing a lot of confusion and require to build twice

vslib/src/sai_vs_init.cpp
kcudnik: at() is runtime check

vslib/src/tests.cpp
kcudnik: yes

vslib/src/tests.cpp
kcudnik: you are wellcome to move all tests to gtest

meta/saiserialize.cpp
qiluo-msft: Sorry, I thought the wrong direction. So move if-block before previous statement? So that there is no corrupted dst_element when throw?

syncd/Makefile.am
qiluo-msft: Actually that is actual behavior. libtool does the magic.

vslib/src/sai_vs_init.cpp
kcudnik: Depends how to look at this, since this is one off, if written correctly it will never fail, and if some actions will lead to throw it will need to be corrected right away since this logic is executed in vs tests when doing make check, then is auto tested

syncd/Makefile.am
kcudnik: Not sure about this, i will need to investigate on that, for now i leave this as is since its working

meta/saiserialize.cpp
kcudnik: I don't quite get that, when you will get currupted dst_element ?

vslib/src/sai_vs_init.cpp
kcudnik: Actually all those CHECK_STATUS could throw anyway

meta/saiserialize.cpp
kcudnik: Ah in this case, sure i can move that, anyway this partial override doesen't matter since api will fail

syncd/Makefile.am
```diff
@@ -20,6 +20,12 @@ if SAICAVM
...
+else
+SAILIB=-lsai -lopennsl -ldl
```
lguohan: >-lopennsl [](start = 13, length = 9)

this should be brcm specific.

PR #133, opened 2016-12-03
Support temporary view mode in syncd

kcudnik:


syncd/syncd.cpp
```diff
@@ -781,45 +1000,206 @@ sai_status_t notifySyncd(const std::string& op)
...
+        sai_status_t status;
+
+        switch (object_type)
+        {
+            case SAI_OBJECT_TYPE_FDB:
+            case SAI_OBJECT_TYPE_NEIGHBOR:
+            case SAI_OBJECT_TYPE_ROUTE:
+            case SAI_OBJECT_TYPE_TRAP:
+
+                SWSS_LOG_ERROR("get is not supported on %s in init view mode", sai_serialize_object_type(object_type).c_str());
```
lguohan: >SWSS_LOG_ERROR("get is not supported on %s in init view mode", sai_serialize_object_type(object_type).c_str()); [](start = 16, length = 111)

why these objects are processed differently than the default?

syncd/syncd.cpp
```diff
@@ -864,12 +1243,20 @@ sai_status_t processEvent(swss::ConsumerTable &consumer)
...
     uint32_t attr_count = list.get_attr_count();
 
+    if (isInitViewMode())
+    {
+        return processEventInInitViewMode(object_type, str_object_id, api, attr_count, attr_list);
+    }
+
+    if (api != SAI_COMMON_API_GET)
+    {
+        // can't be called on init view, since some VID's may have no RID yet
```
lguohan: >// can't be called on init view, since some VID's may have no RID yet [](start = 8, length = 69)

we probably do not need this comment here as init view is handled completely differently.

syncd/syncd.cpp
```diff
@@ -781,45 +1000,206 @@ sai_status_t notifySyncd(const std::string& op)
...
+
+            default:
+
+                {
+                    sai_object_id_t object_id;
+                    sai_deserialize_object_id(str_object_id, object_id);
+
+                    SWSS_LOG_DEBUG("generic get (init view) for object type %s", sai_serialize_object_type(object_type).c_str());
+
+                    // object must exists, we can't call GET on created object in init view mode
```
lguohan: >// object must exists, we can't call GET on created object in init view mode [](start = 20, length = 76)

when object does not exists, can you return the one in the database?

I am not sure what exactly are the use case here since we do not call get on the objects we created usually, but it would be good to add more comments here so that it might be easier to revisit this in the future.

syncd/syncd.cpp
```diff
@@ -351,6 +529,13 @@ void internal_syncd_get_send(
                 attr_count,
                 attr_list,
                 false);
+
+        if (isInitViewMode())
+        {
+            // all oid values here are VID's
+
+            snoop_get_response(object_type, str_object_id, attr_count, attr_list);
+        }
```
lguohan: why is that the snoop op is only applied in the init view mode? The snoop op is trying to capture the asic state which are created by default in the sai library, so that we can use them for later comparison purpose.

for example, we first get 8 queues from asic, and remove 2 of them. In the asic view, we should have remaining 6. the snoop should be performed even during non init view phase.

syncd/syncd.cpp
kcudnik: Leftover

syncd/syncd.cpp
kcudnik: You can call get on ezisting object like default trap Group or hash etc to get some vendor values

PR #142, opened 2017-01-03
Add support for stp instance

kcudnik:


qiluo-msft:
>         syncdApplyView();

handle the return value? I guess it can be used in sendResponse()?

---
Refers to: syncd/syncd.cpp:1055 in 5746542. [](commit_id = 5746542de11bd52aa53455e118aadda93131dc14, deletion_comment = False)

syncd/syncd_hard_reinit.cpp
```diff
@@ -435,6 +435,13 @@ sai_object_id_t processSingleVid(sai_object_id_t vid)
             SWSS_LOG_INFO("default virtual router will not be created, processed VID 0x%lx to RID 0x%lx", vid, rid);
         }
     }
+    if (objectType == SAI_OBJECT_TYPE_STP_INSTANCE)
```
qiluo-msft: >if [](start = 4, length = 2)

else if? Or change to switch-case?

syncd/syncd_reinit.cpp
```diff
@@ -389,6 +425,15 @@ void redisSetDefaultTrapGroup(sai_object_id_t vr_id)
     g_redisClient->hset(HIDDEN, DEFAULT_TRAP_GROUP_ID, strVrId);
 }
 
+void redisSetDefaultStpInstance(sai_object_id_t stp_id)
```
qiluo-msft: >redisSetDefaultStpInstance [](start = 5, length = 26)

I may miss the context. What is "Stp" actually mean? How it is used?

kcudnik: spanning tree protocol

syncd/syncd_hard_reinit.cpp
kcudnik: yes, forgot to add that

PR #150, opened 2017-01-20
Support more objects for comparison

kcudnik:


syncd/syncd_applyview.cpp
```diff
@@ -672,10 +965,314 @@ std::shared_ptr<SaiObj> findCurrentBestMatchForRouterInterface(
...
+                    return c;
+                }
+            }
+        }
+    }
+
+    return nullptr;
+}
+
+std::shared_ptr<SaiObj> findCurrentBestMatchForNextHop(
```
qiluo-msft: >findCurrentBestMatchForNextHop [](start = 24, length = 30)

Why call it best? Is it just findNextHop?

syncd/syncd_applyview.cpp
```diff
@@ -1203,6 +1796,7 @@ sai_status_t applyViewTransition(
     // and then we will need second pass on all objects to see
     // if we have the same attributes
 
+    //return SAI_STATUS_SUCCESS;
     return SAI_STATUS_NOT_IMPLEMENTED;
```
qiluo-msft: >SAI_STATUS_NOT_IMPLEMENTED [](start = 11, length = 26)

Is it implemented?

syncd/syncd_applyview.cpp
kcudnik: because function searches for current existing object best match so it's more explaining itself
anyway this code will be removed in final version when i implement generic logic

syncd/syncd_applyview.cpp
kcudnik: when i develop i change it time to time since version with success produces more checks and produces a lot of noise, but time to time i want to see if it actually works to produce noise logs and not crashes, and forgot to remove, in final version in this way should return success

PR #151, opened 2017-01-20
[syncd]: Enable runtime loglevel setting via Redis DB

marian-pritsak:

### DIFF
--- a/syncd/syncd.cpp
+++ b/syncd/syncd.cpp
@@ -1,4 +1,5 @@
 #include <iostream>
+#include <map>
 #include "syncd.h"
 #include "sairedis.h"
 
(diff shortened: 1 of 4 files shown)


kcudnik:
Please keep loglevel functionality since its SWSS_LOG_ and its different from SAI API log levels
SAI LOG levels should be next to that

marian-pritsak:
@kcudnik 
SAI loglevels are a subset of what SWSS supports
https://github.com/Azure/sonic-sairedis/pull/151/files#diff-91b3e380b7f6f4a0a048ca8296a12e0dR1605
So you can set any SAI loglevel using SWSS

kcudnik:
SWSS log levels are defined here https://github.com/Azure/sonic-swss-common/blob/master/common/logger.h#L21 and even if they match sai log levels this is coincidenc rather than by design

marian-pritsak:
Changed logic so that syncd handles SAI loglevels explicitly without mapping from swss

PR #152, opened 2017-01-24
Fixed the same-MAC-address-bug.

antonpatenko:
Fixed the same-MAC-address-bug by reading it from eth0 and writing into .profile file.

### DIFF
--- a/debian/syncd.init
+++ b/debian/syncd.init
@@ -58,6 +58,20 @@ start)
 
     elif [ "$onie_platform" == "x86_64-mlnx_x86-r5.0.1400" ]; then
         start_mlnx
+
+        # Check if there's no MAC in .profile already.
+        if ! grep DEVICE_MAC_ADDRESS /etc/syncd.d/mlnx_2700.profile > /dev/null
+        then
+            # Read MAC address and align the last 6 bits.
+            MAC_ADDRESS=`ip link show eth0 | grep ether | awk '{print $2}'`
+            last_byte=`python -c "print '$MAC_ADDRESS'[-2:]"`
+            aligned_last_byte=`python -c "print format(int(int('$last_byte', 16) & 0b11000000), 'x')"`  # put mask and take away the 0x prefix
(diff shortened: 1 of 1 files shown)


debian/syncd.init
```diff
@@ -58,6 +58,20 @@ start)
...
+        if ! grep DEVICE_MAC_ADDRESS /etc/syncd.d/mlnx_2700.profile > /dev/null
+        then
+            # Read MAC address and align the last 6 bits.
+            MAC_ADDRESS=`ip link show eth0 | grep ether | awk '{print $2}'`
+            last_byte=`python -c "print '$MAC_ADDRESS'[-2:]"`
+            aligned_last_byte=`python -c "print format(int(int('$last_byte', 16) & 0b11000000), 'x')"`  # put mask and take away the 0x prefix
+            ALIGNED_MAC_ADDRESS=`python -c "print '$MAC_ADDRESS'[:-2] + '$aligned_last_byte'"`          # put aligned byte into the end of MAC
+
+            # Write MAC address into *.profile file.
+            echo "DEVICE_MAC_ADDRESS=$ALIGNED_MAC_ADDRESS" >> /etc/syncd.d/mlnx_2700.profile
```
kcudnik: on my box it produces: DEVICE_MAC_ADDRESS=08:00:27:51:2d:0
which is missing extra zero at the end, original mac is 08:00:27:51:2d:31

lguohan:
this zeros out the last 6 bits of mac, but the problem is that we may have two mlnx switches to be neighbor and their mgmt mac differnece is with the last 6 bits of the mac.  Then, the will have the same mac.

antonpatenko:
All our switches do have 128 MAC addresses reserved per switch.
It means that MAC from this interval might be used:

                XX:XX:XX:XX:XX:80 - XX:XX:XX:XX:XX:FF (last 7 bits)

What I do is write “zero” into the last 6 bits. So I leave 1 bit to make unique MAC address for each device.

PR #163, opened 2017-02-23
support centec platforms

sonicyao1985:


debian/syncd.init
```diff
@@ -36,6 +36,12 @@ start_mlnx()
     [ -e /dev/sxdevs/sxcdev ] || ( mkdir -p /dev/sxdevs && mknod /dev/sxdevs/sxcdev c 231 193 )
 }
 
+start_centec()
+{
+    [ -e /dev/linux_dal ] || ( insmod /usr/lib/dal.ko && mknod /dev/linux_dal c 198 0 )
```
lguohan: kernel modules are packaged into debian package and installed into base image, check here. https://github.com/Azure/sonic-buildimage/blob/master/platform/broadcom/one-image.mk#L6

for the tun drivers, if you need to use it, you can insmod in the package init script.

debian/syncd.init
```diff
@@ -85,6 +91,39 @@ start)
...
+        start_centec
+
+        if ! grep DEVICE_MAC_ADDRESS /etc/syncd.d/centec_e582_48x6q.profile > /dev/null
+        then
+            # Read MAC address and align the last 6 bits.
+            MAC_ADDRESS=`ip link show eth0 | grep ether | awk '{print $2}'`
+            last_byte=`python -c "print '$MAC_ADDRESS'[-2:]"`
+            aligned_last_byte=`python -c "print format(int(int('$last_byte', 16) & 0b11000000), '02x')"`  # put mask and take away the 0x prefix
+            ALIGNED_MAC_ADDRESS=`python -c "print '$MAC_ADDRESS'[:-2] + '$aligned_last_byte'"`          # put aligned byte into the end of MAC
+
```
lguohan: duplicated code, can you add a function to extract mac address.

PR #165, opened 2017-02-27
[everflow]: Remove SAI_MIRROR_SESSION_ATTR_VLAN_ID from VLAN validation.

oleksandrivantsiv:
0 is valid value for SAI_MIRROR_SESSION_ATTR_VLAN_ID attribute. Means that
outer packet should be untagged. On attempt to create session with VLAN 0
error was returned.

### DIFF
--- a/meta/sai_meta.h
+++ b/meta/sai_meta.h
@@ -623,11 +623,8 @@ typedef struct _sai_attr_metadata_t {
             return true;
         }
 
-        if (objecttype == SAI_OBJECT_TYPE_MIRROR &&
-                attrid == SAI_MIRROR_SESSION_ATTR_VLAN_ID)
-        {
-            return true;
-        }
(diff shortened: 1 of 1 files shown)


oleksandrivantsiv:
Fixes item 1 in https://github.com/Azure/sonic-swss/issues/154 issue.
Related to https://github.com/Azure/sonic-swss/pull/169 PR.

stcheng:
@kcudnik i think we still need this to return true but add 0 as a valid VLAN ID internally.

oleksandrivantsiv:
@stcheng I can add VLAN 0 to VLANs list internally as it is done for default VLAN. But this will affect other components (port and router interface assume that VLAN 0 is invalid value). @lguohan @kcudnik are you agree with suggestion to add VLAN 0 internally? It would nice to have a possibility to specify a range of allowed values for attributes but it seems that this is not supported.

PR #166, opened 2017-03-02
simply syncd start logic on different platforms (requires one image)

lguohan:


jleveque:
WIth this PR, your should also be able to delete the etc/syncd.d directory from this repo.

debian/syncd.init
```diff
@@ -34,68 +51,47 @@ start_bcm()
...
-    elif [ "$onie_platform" == "x86_64-m3000-r1" ]; then
+    if [ $sonic_asic_type == "broadcom" ]; then
         start_bcm
-        DAEMON_ARGS+=" -p /etc/syncd.d/m3000_r1.profile"
-
-    elif [ "$onie_platform" == "x86_64-mlnx_x86-r5.0.1400" ] || [ "$onie_platform" == "x86_64-mlnx_msn2700-r0" ]; then
+        DAEMON_ARGS+=" -p $HWSKU_DIR/sai.profile "
+    elif [ $sonic_asic_type == "mellanox" ]; then
         start_mlnx
+        DAEMON_ARGS+=" -p /tmp/sai.profile "
```
stcheng: question (may not be related): why mellanox sai.profile in /tmp/ folder instead of $HWSKU_DIR/ folder

lguohan: see line 62 and 63. HWSKU_DIR is mounted as RO, so I cannot modify the file.

stcheng: i see. thanks.

stcheng:
Addressing https://github.com/Azure/sonic-sairedis/issues/167

PR #179, opened 2017-05-01
Add Conflicts: entries into control, preventing syncd-rpc be installed with syncd

pavel-shirshov:


debian/control
```diff
@@ -53,6 +55,7 @@ Package: syncd-dbg
 Architecture: any
 Section: debug
 Priority: extra
+Conflicts: syncd-dbg-rpc
```
qiluo-msft: >syncd-dbg-rpc [](start = 11, length = 13)

syncd-rpc-dbg

PR #184, opened 2017-05-19
Refactor syncd init to share code between systemd and supervisord

jleveque:
- systemd init script and supervisor start script now share code via syncd_init_common.sh
- Use sonic-cfggen to obtain ASIC type (requires syncd dockers inherit from docker-config-engine. I will make this change once this PR is merged and I update the submodule in sonic-buildimage.
- Added Cavium-specific startup steps so that platform can share the same init files


debian/syncd.install
```diff
@@ -1,3 +1,4 @@
 usr/bin/saidump
 usr/bin/saiplayer
 usr/bin/syncd*
+syncd/usr/bin/* usr/bin
```
pavel-shirshov: Do you need both old content and new one? Why?

debian/rules
```diff
@@ -43,15 +43,15 @@ binary-syncd-rpc:
 #	-DCMAKE_LIBRARY_PATH=$(DEB_HOST_MULTIARCH)
 
 override_dh_auto_configure:
+    ifeq ($(shell cat /tmp/syncd-build), --enable-rpcserver=yes)
+	sed -i 's/ENABLE_SAITHRIFT=0/ENABLE_SAITHRIFT=1/' syncd/usr/bin/syncd_init_common.sh
+    else
+	sed -i 's/ENABLE_SAITHRIFT=1/ENABLE_SAITHRIFT=0/' syncd/usr/bin/syncd_init_common.sh
+    endif
```
pavel-shirshov: Interesting.
My path starts with debian, so current directory is root of the project.
Your path starts with syncd, I suspect it's inside of the debian directory.
Did you check this code with our build infrastructure?

debian/rules
```diff
@@ -43,15 +43,15 @@ binary-syncd-rpc:
 #	-DCMAKE_LIBRARY_PATH=$(DEB_HOST_MULTIARCH)
 
 override_dh_auto_configure:
+    ifeq ($(shell cat /tmp/syncd-build), --enable-rpcserver=yes)
```
pavel-shirshov: Is it possible to do this after dh clean, but before dh build?
I mean changing this flag?

syncd/usr/bin/syncd_init_common.sh
```diff
@@ -0,0 +1,116 @@
...
+        exit 1
+    fi
+
+    if [ ${ENABLE_SAITHRIFT} == 1 ]; then
+        CMD_ARGS+=" -r -m $HWSKU_DIR/port_config.ini"
+    fi
+
+    [ -r $PLATFORM_DIR/syncd.conf ] && . $PLATFORM_DIR/syncd.conf
+}
+
```
pavel-shirshov: I'd suggest to check this script with https://www.shellcheck.net/
Probably you'll find some.

debian/syncd.install
jleveque: Yes. `usr/bin/syncd*` installs the standard files that are copied to `debian/tmp/usr/bin/syncd*`. My new files are in syncd/usr/bin/ and do not get copied to `debian/tmp/usr/bin/`.

debian/rules
jleveque: My path is outside of `debian/`. My new files are in `syncd/usr/bin/`, and `syncd/` is on the same level as `debian/`. I have tested this with our build infrastructure.

debian/rules
jleveque: It shouldn't matter, as my scripts are not in a debian build directory, so they will not get cleaned up after the change. Do you have other concerns here I'm missing?

debian/rules
lguohan: I think it both starts from the working directory.

---
In reply to: [117566129](https://github.com/Azure/sonic-sairedis/pull/184#discussion_r117566129) [](ancestors = 117566129)

debian/rules
pavel-shirshov: Yesterday we discussed possibility of removing "/tmp/syncd-build" file and doing everything on upper layer. Have you tried that?

debian/rules
pavel-shirshov: ok

syncd/usr/bin/syncd_init_common.sh
lguohan: the website is nice. there are some good advices on the code. 

---
In reply to: [117566693](https://github.com/Azure/sonic-sairedis/pull/184#discussion_r117566693) [](ancestors = 117566693)

debian/syncd.install
pavel-shirshov: Now I understand. You put your files into syncd/usr/bin directory.
And make install will not put the files into right places.
I think you need to use bin_SCRIPTS option of automake.
I think it allows you to install your scripts into right places by make install.

P.S.: I think it's better to put your scripts into scripts directory in the root of the project.

syncd/usr/bin/syncd_init_common.sh
pavel-shirshov: I asked this because we're using variables like this
$VAR
"$VAR"
${VAR}
I think we should choose second or third way only and stick with it

jleveque: That's a great tool! Thanks for the link, Pavel! It gave a few warnings about code that was already in the file before I modified it, so I just cleaned it up and checked in an update.

debian/rules
jleveque: Yes. I tried that. Unfortunately, it appears my files get copied before the `binary` target is called, so my files get copied before the sed command executes. I moved the sed command to `override_dh_auto_configure` because it technically a configuration step.

debian/syncd.install
jleveque: I was thinking of putting them in a `scripts` directory in the root, but since this directory contains the source of multiple Debian packages, I felt they were better off placed in the project directory. I can move them if you still think they belong under the root.

debian/rules
jleveque: @lguohan: The behavior is as follows:

1. Try to find file in path relative to root (the level with the `debian/` directory.
2. If file is not found, try to find file relative to `debian/tmp/`

syncd/usr/bin/syncd_init_common.sh
jleveque: I agree we should choose a style and use it universally. Is there any actual difference in the expansion between `"$VAR"` and `${VAR}`?

pavel-shirshov: http://stackoverflow.com/questions/8748831/when-do-we-need-curly-braces-in-variables-using-bash

debian/syncd.install
pavel-shirshov: It's up to you. But I think making hierarchy syncd/usr/bin/ is not required. Probably it's better make it syncd/scripts? Make or not make bin_SCRIPTS is up to you.

debian/rules
pavel-shirshov: run sed command against copied version? If it has wrong value it would be overwritten. If it has right value it would be skipped.
What do you think?

syncd/usr/bin/syncd_init_common.sh
jleveque: Looks like there's not much difference between the two, but using curly braces is more versatile (e.g., for concatenating strings), so that's probably the convention we should use (it's normally what I use, anyway).

debian/syncd.install
jleveque: I changed the hierarchy to syncd/scripts.

debian/rules
jleveque: Done. Now only running sed against the version of the script that gets copied to the syncd-rpc staging area.

PR #186, opened 2017-05-26
Add recording log output dir attribute to sairedis

kcudnik:


lib/src/sai_redis_record.cpp
```diff
@@ -104,3 +107,43 @@ std::string joinFieldValues(
...
 }
+
+sai_status_t setRecordingLogOutputDir(
+        _In_ const sai_attribute_t &attr)
+{
+    SWSS_LOG_ENTER();
+
+    if (attr.value.s8list.count == 0)
+    {
+        logOutputDir = ".";
```
qiluo-msft: >logOutputDir [](start = 8, length = 12)

Suggest also check W_OK for "."

kcudnik: open() will fail then anyway

stcheng: I think this is checked later on line 138.

lib/inc/sai_redis.h
```diff
@@ -29,6 +29,8 @@ extern "C" {
 
 extern volatile bool                    g_record;
 extern void setRecording(bool record);
+extern sai_status_t setRecordingLogOutputDir(
```
qiluo-msft: >setRecordingLogOutputDir [](start = 20, length = 24)

Just feel log is strange. Call it setRecordingDir or  setRecordingOutputDir ?

kcudnik: ok

lib/inc/sairedis.h
```diff
@@ -68,6 +68,20 @@ typedef enum _sai_redis_switch_attr_t
      */
     SAI_REDIS_SWITCH_ATTR_FLUSH,
 
+    /**
+     * @brief Recording output directory.
+     *
+     * By defaut is current directory. Also setting empty will force default
```
stcheng: default

lib/src/sai_redis_record.cpp
```diff
@@ -104,3 +107,43 @@ std::string joinFieldValues(
...
+    {
+        SWSS_LOG_ERROR("list pointer is NULL");
+        return SAI_STATUS_FAILURE;
+    }
+
+    size_t len = strnlen((const char *)attr.value.s8list.list, attr.value.s8list.count);
+
+    if (len != (size_t)attr.value.s8list.count)
+    {
+        SWSS_LOG_ERROR("count (%u) is different then strnlen (%zu)", attr.value.s8list.count, len);
```
stcheng: than?

PR #188, opened 2017-05-30
[syncd]: Remove SWSS_LOG_TIMER and update log level

stcheng:
- Remove counter collection timer
- Change counter thread log to INFO

### DIFF
--- a/syncd/syncd_counters.cpp
+++ b/syncd/syncd_counters.cpp
@@ -12,8 +12,6 @@ void collectCounters(swss::Table &countersTable,
 
     SWSS_LOG_ENTER();
 
-    SWSS_LOG_TIMER("get counters");
-
     uint32_t countersSize = (uint32_t)supportedCounters.size();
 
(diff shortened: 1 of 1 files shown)


syncd/syncd_counters.cpp
```diff
@@ -136,10 +134,10 @@ void endCountersThread()
...
     if (g_countersThread != NULL)
     {
-        SWSS_LOG_NOTICE("counters thread join");
+        SWSS_LOG_INFO("counters thread join");
 
         g_countersThread->join();
     }
 
-    SWSS_LOG_NOTICE("counters thread ended");
+    SWSS_LOG_INFO("counters thread ended");
```
kcudnik: those are not needed i think to change to info, but they can stay

stcheng: I think they are more code related logs that end users are  not going to care about. So I move it to INFO.

kcudnik: ok

PR #189, opened 2017-05-31
[sairedis]: Increase the GET response timeout to 6 min

stcheng:


stcheng:
@kcudnik yes. I have sent out https://github.com/Azure/sonic-swss/pull/230 as well.

PR #190, opened 2017-06-14
Clear istringstream before reuse

andriymoroz-mlnx:


syncd/syncd.cpp
```diff
@@ -1715,6 +1715,7 @@ void handlePortMap(const std::string& portMapFile)
         std::string name, lanes, alias;
         iss >> name >> lanes >> alias;
 
+        iss.clear();
```
stcheng: shall this line after the `iss.str(lanes);`? btw, is it okay to use `tokenize()` function?

andriymoroz: I didn't try other options just used the shortest fix

stcheng: for the first question, i think this line shall be after the line 1719.

andriymoroz: many examples I found recommend 
- use
- clear
- reuse
and I tested this code it works fine
is there any special reason to change this? it will take several hours to rebuild and test a new code...

stcheng: i've found out both before/after the `str()` function are okay. thanks.

PR #191, opened 2017-06-16
Add sleep on syncd crash to keep data plane active

kcudnik:


syncd/syncd.cpp
```diff
@@ -67,7 +68,19 @@ void exit_and_notify(int status)
         SWSS_LOG_ERROR("Unknown runtime error");
     }
 
-    exit(status);
+    if (options.disableCrashSleep)
+    {
+        exit(status);
+    }
+
+    SWSS_LOG_WARN("syncd crashed, but will sleep forever to keep data plane active");
```
qiluo-msft: >crashed [](start = 25, length = 7)

technically syncd is not crashed, nearly crashed.

stcheng: 💀

stcheng:
build failed.

PR #192, opened 2017-06-16
Add support for logrotate in sairedis

kcudnik:


lib/src/sai_redis_record.cpp
```diff
@@ -22,10 +22,34 @@ std::string getTimestamp()
...
 // recording needs to be enabled explicitly
 volatile bool g_record = false;
+volatile bool g_logrotate = false;
 
 std::ofstream recording;
 std::mutex g_recordMutex;
 
+std::string recfile = "dummy.rec";
+
+void logRotate()
```
qiluo-msft: >logRotate [](start = 5, length = 9)

'Rotate' here is misleading. It just releases and fetches the file handle.

lib/src/sai_redis_record.cpp
```diff
@@ -22,10 +22,34 @@ std::string getTimestamp()
...
+
+    recording.close();
+
+    /*
+     * On log rotate we will use the same file name, we are assuming that
+     * logrotate deamon move filename to filename.1 and we will create new
+     * empty file here.
+     */
+
+    recording.open(recfile);
```
qiluo-msft: >open [](start = 14, length = 4)

For performance concern, refer [https://stackoverflow.com/a/164770/2514803](https://stackoverflow.com/a/164770/2514803). Just FYI.

lib/src/sai_redis_record.cpp
kcudnik: can you suggest other name ?

qiluo-msft: How about 'logShift‘？

kcudnik: sure

lib/src/sai_redis_record.cpp
lguohan: this is periodically, perhaps once per several hours. not sure about the performance concern here.

lib/src/sai_redis_record.cpp
lguohan: logfileRelease

lguohan: or logfileReopen

kcudnik: done

qiluo-msft:
Just throw the question without a quick solution: Is it possible not to add new sairedis interface and be transparent to application?

lguohan:
the interface is provided to send reopen the log file, need an api to do it.

lguohan:
:shipit:

kcudnik:
i think thats the point to have our specific apis that can influence sairedis over standard sai interface

PR #199, opened 2017-07-21
Added SAI dump support to syncd

andrii-savka:


kcudnik:
What is purpose of this change? there is already saidump tool which dumps redis database to file

andrii-savka:
To @kcudnik 
The purpose of this change is to dump the real SAI state including SDK log. 
The "saidump" tool dumps only the sairedis database - not a SAI layer and it could be different. 
It is new feature introduced for dumping the SAI and SDK log and add them to sysdump.

kcudnik:
This could be done easier if you would use connect to existign sai with INIT == false, similar like saidiscovery tool

PR #200, opened 2017-07-24
Add virtual switch mlnx/brcm SAI tests

kcudnik:


kcudnik:
build errors are caused by missing sai_switch_id_query in libsai, it needs to be updated (last valid build was a month ago)

tests/Makefile.am
```diff
@@ -0,0 +1,24 @@
...
+				../syncd/syncd_saiswitch.cpp \
+				../syncd/syncd_hard_reinit.cpp \
+				../syncd/syncd_notifications.cpp \
+				../syncd/syncd_counters.cpp \
+				../syncd/syncd_applyview.cpp
+
+vssyncd_CPPFLAGS = $(DBGFLAGS) $(AM_CPPFLAGS) $(CFLAGS_COMMON) $(SAIFLAGS)
+vssyncd_LDADD = -lhiredis -lswsscommon $(SAILIB) -lpthread -L$(top_srcdir)/meta/.libs -lsaimetadata -lsaimeta -ldl
+
+TESTS = mlnx.pl brcm.pl
```
qiluo-msft: >mlnx.pl brcm.pl [](start = 8, length = 15)

Should we create the same set of test cases for every SAI vendors? The vendor specific behavior may be captured in a parameter file or a config file.

qiluo-msft: I agree the recording files are per vendor. Still the test cases themselves should be the same set.

kcudnik: no, since if we decide to inspect output from the test (currently we are just looking for crash) then output result can be different per vendor and different steps could be taken

PR #201, opened 2017-07-25
[saimetadata]: Fixed "get_attr_name: object type 20 attr id 256 not found in map!"

Mrak-IW:
Attribute SAI_QUEUE_ATTR_INDEX added into attribute map for object type SAI_OBJECT_TYPE_QUEUE

Signed-off-by: Denis Maslov <Denis.Maslov@cavium.com>


lguohan:
@kcudnik, we also saw these in our logs. is this the right way to solve the problem?

kcudnik:
Yes, shoud work if queue index is defined, anyway this problem will go away on its own when switched to v1.0

PR #211, opened 2017-08-17
Fix thread deadlock issue

kcudnik:
Add separate lock for notification and reuse
global syncd lock to protect only db access

### DIFF
--- a/syncd/syncd.cpp
+++ b/syncd/syncd.cpp
@@ -5,7 +5,7 @@
 #include "swss/tokenize.h"
 #include <limits.h>
 
-std::mutex g_mutex;
+std::mutex g_db_mutex;
 
 swss::RedisClient *g_redisClient = NULL;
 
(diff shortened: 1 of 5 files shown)


syncd/syncd.cpp
```diff
@@ -148,6 +148,8 @@ void remove_rid_and_vid_from_local(
 sai_object_id_t translate_rid_to_vid(
         _In_ sai_object_id_t rid)
 {
+    std::lock_guard<std::mutex> lock(g_mutex);
```
yxieca: Does it make sense to rename g_mutex to g_db_mutex?

kcudnik: yes

syncd/syncd_counters.cpp
```diff
@@ -8,7 +8,11 @@ void collectCounters(swss::Table &countersTable,
     // collect counters should be under mutex
     // sice configuration can change and we
     // don't want that during counters collection
-    std::lock_guard<std::mutex> lock(g_mutex);
+    //
+    // XXX currently lock for counters is disabled
+    // since on new 0.9.4 can cause deadlock since
+    // interanl sai implementaion also uses locks
+    // std::lock_guard<std::mutex> lock(g_db_mutex);
```
lguohan: >g_db_mutex [](start = 40, length = 10)

can we move this lock to line 35 so that we avoid the lock for the sai api?

kcudnik: sure

syncd/syncd.cpp
yxieca: Thanks Buddy!

PR #213, opened 2017-08-22
[sycnd]: Support for PFC Watchdog

marian-pritsak:


syncd/syncd.cpp
```diff
@@ -2309,6 +2310,62 @@ sai_status_t processEvent(
...
+
+    sai_object_id_t queueVid = SAI_NULL_OBJECT_ID;
+    sai_deserialize_object_id(key, queueVid);
+    sai_object_id_t queueId = translate_vid_to_rid(queueVid);
+
+    const auto values = kfvFieldsValues(kco);
+    for (const auto& valuePair : values)
+    {
+        const auto field = fvField(valuePair);
+        const auto value = fvValue(valuePair);
```
kcudnik: what is format of field/value here /

syncd/syncd_pfc_watchdog.cpp
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    auto it = wd.m_counterIdsMap.find(queueVid);
+    if (it != wd.m_counterIdsMap.end())
+    {
+        (*it).second->portCounterIds = counterIds;
+        return;
+    }
+
+    auto pfcCounterIds = std::shared_ptr<PfcCounterIds>(new PfcCounterIds(queueId,
```
kcudnik: you can use std;;make_shared<> to not use new operator

syncd/syncd_pfc_watchdog.cpp
```diff
@@ -0,0 +1,260 @@
...
+}
+
+void PfcWatchdog::pfcWatchdogThread(void)
+{
+    SWSS_LOG_ENTER();
+
+    swss::DBConnector db(COUNTERS_DB, swss::DBConnector::DEFAULT_UNIXSOCKET, 0);
+    swss::Table countersTable(&db, COUNTERS_TABLE);
+
+    while(m_runPfcWatchdogThread)
```
kcudnik: space after while

syncd/syncd_pfc_watchdog.cpp
```diff
@@ -0,0 +1,260 @@
...
+    SWSS_LOG_ENTER();
+
+    if (m_runPfcWatchdogThread.load() == true)
+    {
+        return;
+    }
+
+    m_runPfcWatchdogThread = true;
+
+    m_pfcWatchdogThread = std::shared_ptr<std::thread>(new std::thread(&PfcWatchdog::pfcWatchdogThread, this));
```
kcudnik: make_shard

syncd/syncd_pfc_watchdog.h
```diff
@@ -0,0 +1,65 @@
...
+        void collectCounters(
+                _In_ swss::Table &countersTable);
+        void pfcWatchdogThread(void);
+        void startWatchdogThread(void);
+        void endWatchdogThread(void);
+
+        // Key is a queue Virtual ID
+        std::map<sai_object_id_t, std::shared_ptr<PfcCounterIds>> m_counterIdsMap;
+
+        std::atomic_bool m_runPfcWatchdogThread = { false };
```
kcudnik: use regular bool and make it volatile ?

syncd/syncd_pfc_watchdog.cpp
```diff
@@ -0,0 +1,260 @@
...
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to get port Id of queue 0x%lx: %d", queueId, status);
+        return SAI_NULL_OBJECT_ID;
+    }
+
+    return attr.value.oid;
+}
+
+void PfcWatchdog::collectCounters(
```
kcudnik: i would move this to SaiSwich class since threre is imilar method callectCounters which doues similar thing, only difference is that that you collect this counters every 50ms and interface counters are queried every 1second

syncd/syncd_pfc_watchdog.cpp
```diff
@@ -0,0 +1,260 @@
...
+        .value =
+        {
+            .oid = queueId,
+        }
+    };
+
+    sai_status_t status = sai_metadata_sai_queue_api->get_queue_attribute(queueId, 1, &attr);
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to get port Id of queue 0x%lx: %d", queueId, status);
```
kcudnik: user sai serialize oid/status to get strings representation which will be more meaningful

kcudnik:
Can you explain what is purpose of this PR ? for 32 ports how many PFC counters you need to query? in your pr you query them every 50ms, and how much time it takes to query all counters once ? this seems like not efficient method to me (pool and check in loop), and i would assume that hardware based event/notifications would be more efficient, i believe there is a threshold event already added in SAI spec for telemetry data (is that also including PFC?) this would be much efficient just to wait for a hardware event/notification.

did you made any test how much CPU this new thread is  taking ?

syncd/syncd_pfc_watchdog.cpp
marian-pritsak: Done

syncd/syncd_pfc_watchdog.cpp
marian-pritsak: Done

syncd/syncd_pfc_watchdog.h
marian-pritsak: Depending on template type, implementation will use a lock or declare it volatile, so for bool type, it's the same.

syncd/syncd.cpp
marian-pritsak: Field is either port counter ID list or queue counter ID list

syncd/syncd_pfc_watchdog.cpp
marian-pritsak: Done

syncd/syncd_pfc_watchdog.cpp
marian-pritsak: They are different in a way that regular counter thread is fetching all available counters, but PFC WD thread is only fetching those counters to which orchagent did the subscription. PFC WD might run only on a subset of ports, only on lossless queues, and there is a mechanism of subscription, which is not really related to switch class.

syncd/syncd_pfc_watchdog.cpp
marian-pritsak: syncd PFC WD will have a few more changes in the next iteration. I will make sure to refactor all error messages to pint serialized values in next PR.

marian-pritsak:
In general, the purpose of this PR is to add the functionality of polling counters for SW implementation of watchdog. It is designed for platforms, that do not have it in hardware. As for last discussions, polling period will be changed to 100 ms as soon as redis scripts will be moved from orchagent to syncd for better polling. Regarding SAI calls, PFC WD polls every lossless queue on every WD enabled port every 100 ms.

I did perform a test to see how much CPU is consumed, was within normal system never was overloaded.

syncd/syncd_pfc_watchdog.cpp
```diff
@@ -0,0 +1,260 @@
...
+    if (it != wd.m_counterIdsMap.end())
+    {
+        (*it).second->portCounterIds = counterIds;
+        return;
+    }
+
+    auto pfcCounterIds = std::make_shared<PfcCounterIds>(queueId,
+                                                         portId,
+                                                         counterIds,
+                                                         std::vector<sai_queue_stat_t>());
```
lguohan: >sai_queue_stat_t [](start = 69, length = 16)

should be sai_port_stat_t

syncd/syncd_pfc_watchdog.cpp
```diff
@@ -0,0 +1,260 @@
...
+        _In_ const std::vector<sai_port_stat_t> &portIds,
+        _In_ const std::vector<sai_queue_stat_t> &queueIds):
+    queueId(queue), portId(port), portCounterIds(portIds), queueCounterIds(queueIds)
+{
+}
+
+void PfcWatchdog::setPortCounterList(
+        _In_ sai_object_id_t queueVid,
+        _In_ sai_object_id_t queueId,
+        _In_ const std::vector<sai_port_stat_t> &counterIds)
```
lguohan: when you query the port counter list, why do you give queueId? For example, for PFC TX and RX, you should be able to query for both queue 3 and queue 4 for the same port. You do not need to give queue id here.

syncd/syncd_pfc_watchdog.cpp
```diff
@@ -0,0 +1,260 @@
...
+        for (size_t i = 0; i != queueCounterIds.size(); i++)
+        {
+            const std::string &counterName = sai_serialize_queue_stat(queueCounterIds[i]);
+            values.emplace_back(counterName, std::to_string(queueStats[i]));
+        }
+
+        // Write counters to DB
+        std::string queueVidStr = sai_serialize_object_id(queueVid);
+
+        countersTable.set(queueVidStr, values, "");
```
lguohan: >queueVidStr [](start = 26, length = 11)

your counter table is always index by queue vid, I think it is probably easier for upper layer management?

syncd/syncd_pfc_watchdog.h
```diff
@@ -0,0 +1,65 @@
...
+            PfcCounterIds(
+                    _In_ sai_object_id_t queue,
+                    _In_ sai_object_id_t port,
+                    _In_ const std::vector<sai_port_stat_t> &portIds,
+                    _In_ const std::vector<sai_queue_stat_t> &queueIds);
+
+            sai_object_id_t queueId;
+            sai_object_id_t portId;
+            std::vector<sai_port_stat_t> portCounterIds;
+            std::vector<sai_queue_stat_t> queueCounterIds;
```
lguohan: it is better to organize your counterIDmap based on the port id for the all port counters.

In this case, you can query all the port counters for one port together. later, you can write to the counter db indexed by queue id, since it will be easier for the lua script.

the pfccounterids has following structure

portqueuecounterids
{
portid;  // valid only for for portcounterid
std::vector<tuple<queueid, sai_port_stat_t>> portCounterIds; // valid only for portcounter id
queueid; // valid only for queuecounterid.
vector<sai_queue_stat_t> queueids // validonly for queuecounter id
}

syncd/syncd_pfc_watchdog.cpp
```diff
@@ -0,0 +1,260 @@
+#include "syncd_pfc_watchdog.h"
```
lguohan: this a generic port queue counter collection thread. we can rename it to syncd_flex_counters.cpp

syncd/syncd_pfc_watchdog.h
```diff
@@ -0,0 +1,65 @@
...
+extern "C" {
+#include "sai.h"
+}
+
+#include <atomic>
+#include <vector>
+#include <condition_variable>
+#include "swss/table.h"
+
+class PfcWatchdog
```
lguohan: the class can be rename to FlexCounterCollector

lguohan:
@kcudnik, the TAM proposal needs hardware support, not all asic supports that, so we choose to use software to do the polling. as noted by marian, the timer will be changed to 100ms in the next iteration. I think it will be fine.

syncd/syncd_pfc_watchdog.h
marian-pritsak: Agree. I will split subscription for queue counters and port counters so that port counters will be queried only once.

meta/saiserialize.h
```diff
@@ -196,4 +199,12 @@ void sai_deserialize_free_port_oper_status_ntf(
         _In_ uint32_t count,
         _In_ sai_port_oper_status_notification_t* portoperstatus);
 
+void sai_deserialize_port_stat(
```
kcudnik: look at line 192, its notification segment and you are adding enum segments here, can you move it up ?

syncd/syncd_pfc_watchdog.cpp
kcudnik: still not fixed

syncd/syncd_pfc_watchdog.cpp
kcudnik: this actually dont matter, saiswitch class can contain "subscribed"counters list and this will collect the counders, i wanted this to move to saiswitch class because its collecting counters in similar fashion as regular counters, and also you would not need to probide lock here but in the thread directly before calling this method

PR #214, opened 2017-08-24
VID RID mapping may not exist for partial upgrade

JipanYanga:


kcudnik:
When this can happen? All those ids after hard reinut should be consistent

JipanYanga:
Hitting the exception about two months ago,  without fully understanding the root cause I put the workaround there.  
I have forgot the exact scenario, maybe I was trying to run "/usr/bin/syncd --diag -p /usr/share/sonic/hwsku/sai.profile" directly to debug some issue.

Not seeing the problem any more,  close this PR to avoid confusion.

PR #219, opened 2017-09-06
Add lock for get response

kcudnik:


stcheng:
is this needed for v1.0.3m branch?

yxieca:
It only matters for TH2 platform. So I guess no. We need the fix in 2.3 and later.

On Sep 6, 2017 6:35 PM, Shuotian Cheng <notifications@github.com> wrote:

is this needed for v1.0.3m branch?

stcheng:
Then we need to be cautious when merging the master to v1.0.3 branch. Thanks for the info.

PR #220, opened 2017-09-10
Refactor mutexes

kcudnik:


syncd/syncd_counters.cpp
```diff
@@ -9,6 +9,8 @@ void collectCounters(swss::Table &countersTable,
     // sice configuration can change and we
```
qiluo-msft: >sice [](start = 7, length = 4)

since?

syncd/syncd_notifications.cpp
```diff
@@ -1,8 +1,13 @@
 #include "syncd.h"
 #include "sairedis.h"
 
-// mutex to protect notification send call
-std::mutex g_ntf_mutex;
+#include <queue>
+#include <memory>
+#include <condition_variable>
```
qiluo-msft: include those standard headers at very beginning

syncd/syncd_notifications.cpp
```diff
@@ -185,61 +178,293 @@ void on_port_event(
...
         _In_ const sai_attribute_t *attr_list)
 {
-    std::lock_guard<std::mutex> lock(g_ntf_mutex);
-
     SWSS_LOG_ENTER();
 
     SWSS_LOG_ERROR("not implemented");
+}
+
+// determine wheter notification thread is running
```
qiluo-msft: >wheter [](start = 13, length = 6)

typo?

syncd/syncd.cpp
```diff
@@ -5,7 +5,7 @@
 #include "swss/tokenize.h"
 #include <limits.h>
 
-std::mutex g_db_mutex;
+std::mutex g_mutex;
```
qiluo-msft: >g_mutex [](start = 11, length = 7)

more descriptive name?

syncd/syncd_notifications.cpp
```diff
@@ -185,61 +178,293 @@ void on_port_event(
...
+bool tryDequeue(
+        _Out_ swss::KeyOpFieldsValuesTuple &item)
+{
+    std::lock_guard<std::mutex> lock(queue_mutex);
 
-       sai_serialize_buffer(buffer, buffer_size, s);
+    SWSS_LOG_ENTER();
 
-       std::vector<swss::FieldValueTuple> entry;
+    if (ntf_queue.size() == 0)
```
yxieca: I saw Qi commenting on another review, asking to use <queue>.empty instead of testing size() ==0, is this an agreed or preferred method? If yes, please change.

Thanks,
Ying

syncd/syncd_notifications.cpp
kcudnik: common practice is to put std libs as last includes and not put everything into one header to reduce compilation time of source and remove headers that are not needed cor specific source file

syncd/syncd.cpp
kcudnik: done

syncd/syncd_counters.cpp
kcudnik: fixed

syncd/syncd_notifications.cpp
kcudnik: done

syncd/syncd_notifications.cpp
qiluo-msft: I cannot agree with "common practice is to put std libs as last includes"

---
In reply to: [138274252](https://github.com/Azure/sonic-sairedis/pull/220#discussion_r138274252) [](ancestors = 138274252)

syncd/syncd_notifications.cpp
kcudnik: fixed

PR #223, opened 2017-09-11
Add the Queue stat counters to Counter DB

andrii-savka:

PR #230, opened 2017-09-27
Queue stat counters

andrii-savka:
It's a updated pull request with fixed comments from https://github.com/Azure/sonic-sairedis/pull/223


qiluo-msft:
Could you continue the pull request discussion on original one? We lost all the context here, and open issues are hard to track.

I fast-forward v1.0.3m branch to latest master, so when the PR is done there, I can help merge the code for you.

BTW, there is a force-push happens on master branch, please tune you pull request if needed.

lguohan:
can you update description for this PR? What it does? I need to get a little bit context in order to review the code. I do not find the description in #223 either.

andrii-savka:
@qiluo-msft 
Unfortunately, the original changes for v1.0.3m branch were removed. I can't push the changes in the #223. 
I've rebased all PR to the latest version and I will push them tomorrow after verification.

andrii-savka:
Fixed review remarks from #223. 
The description of PR is added below. 

The changes have a following goals:
- Add and update queue's counters for each port to Counter DB
- Read and store the SAI oids for ports and queues once on init instead of polling the SAI each time when counters are updating

syncd/syncd_saiswitch.cpp
```diff
@@ -550,55 +539,102 @@ void SaiSwitch::collectCounters(
...
                     sai_serialize_status(status).c_str());
+
             continue;
         }
 
+<<<<<<< HEAD
         sai_object_id_t vid = translate_rid_to_vid(port_rid, m_switch_vid);
 
         std::string strPortId = sai_serialize_object_id(vid);
+=======
```
qiluo-msft: build break?

syncd/syncd_saiswitch.cpp
```diff
@@ -550,55 +539,102 @@ void SaiSwitch::collectCounters(
...
-            SWSS_LOG_ERROR("failed to collect counters for port RID %s: %s",
-                    sai_serialize_object_id(port_rid).c_str(),
+            SWSS_LOG_ERROR("failed to collect queue counters for queue RID %s: %s",
+                    sai_serialize_object_id(queue_rid).c_str(),
                     sai_serialize_status(status).c_str());
+
             continue;
         }
 
+<<<<<<< HEAD
```
lguohan: resolve conflict incorrectly?

syncd/syncd_saiswitch.cpp
```diff
@@ -550,55 +539,102 @@ void SaiSwitch::collectCounters(
...
+
+            sai_object_id_t vid = translate_rid_to_vid(port_rid, m_switch_vid);
+
+            std::vector<swss::FieldValueTuple> values;
+
+            for (size_t idx = 0; idx < counters.size(); idx++)
+            {
+                const std::string &field = sai_serialize_port_stat(m_supported_counters[idx]);
+                const std::string &value = std::to_string(counters[idx]);
+                values.push_back( swss::FieldValueTuple(field, value) );
```
lguohan: >  [](start = 33, length = 1)

no space here.

lguohan: >  [](start = 69, length = 1)

no space here.

syncd/syncd_saiswitch.cpp
```diff
@@ -1220,9 +1392,27 @@ SaiSwitch::SaiSwitch(
 
     helperInternalOids();
 
-    helperCheckLaneMap();
+    m_port_rid_list = saiGetPortList();
```
lguohan: this won't work with the port break out mode. This one is executed before you remove/add port, so the port list needs to be updated once the break out logic is executed.

syncd/syncd_saiswitch.cpp
```diff
@@ -1185,6 +1221,142 @@ sai_object_id_t SaiSwitch::getDefaultValueForOidAttr(
...
+    std::vector<uint64_t> counters;
+    counters.resize(countersSize);
+
+    swss::DBConnector db(COUNTERS_DB, swss::DBConnector::DEFAULT_UNIXSOCKET, 0);
+    swss::Table countersTable(&db, "COUNTERS");
+    sai_status_t status;
+
+    for (auto &port_queues: m_queue_rid_map)
+    {
+        for (auto queue_it = port_queues.second.begin(); queue_it != port_queues.second.end();)
```
lguohan: >; [](start = 93, length = 1)

you do not need semicolon here.

syncd/syncd_saiswitch.cpp
```diff
@@ -1185,6 +1221,142 @@ sai_object_id_t SaiSwitch::getDefaultValueForOidAttr(
...
+
+                    continue;
+                }
+
+                sai_object_id_t vid = translate_rid_to_vid(*queue_it, m_switch_vid);
+                std::string strQueueId = sai_serialize_object_id(vid);
+                std::vector<swss::FieldValueTuple> values;
+                const std::string &field = sai_serialize_enum(SAI_QUEUE_ATTR_TYPE, &sai_metadata_enum_sai_queue_attr_t);
+                const std::string &value = sai_serialize_enum(attr.value.s32, &sai_metadata_enum_sai_queue_type_t);
+                values.push_back( swss::FieldValueTuple(field, value) );
```
lguohan: >  [](start = 69, length = 1)

no spaces here.

syncd/syncd_saiswitch.cpp
andrii-savka: I'm sorry, I've not added the merge fixes to commit. It's already updated.

syncd/syncd_saiswitch.cpp
andrii-savka: I'm sorry, I've not added the merge fixes to commit. It's already updated.

syncd/syncd_saiswitch.cpp
andrii-savka: Fixed

andrii-savka: Fixed

syncd/syncd_saiswitch.cpp
andrii-savka: In this case I will revert the changes and syncd will read the port and queue list each time on the counter’s update.
The reason for that revert is we don’t have a notification in the syncd when portsyncd in swss removes/creates the ports.

syncd/syncd_saiswitch.cpp
andrii-savka: Fixed.

syncd/syncd_saiswitch.cpp
lguohan: can you hook with remove_port api to remove the port from the port_list?

syncd/syncd_saiswitch.cpp
```diff
@@ -550,55 +539,102 @@ void SaiSwitch::collectCounters(
...
+    for (auto &port_rid: m_port_rid_list)
     {
-        /*
-         * There are not supported counters :(
-         */
+        if (!m_supported_counters.empty())
+        {
+            uint32_t countersSize = (uint32_t)m_supported_counters.size();
+            std::vector<uint64_t> counters;
+            counters.resize(countersSize);
```
qiluo-msft: >resize [](start = 21, length = 6)

This line can be merged into ctor

syncd/syncd_saiswitch.cpp
```diff
@@ -550,55 +539,102 @@ void SaiSwitch::collectCounters(
 {
     SWSS_LOG_ENTER();
 
-    if (m_supported_counters.size() == 0)
+    for (auto &port_rid: m_port_rid_list)
```
qiluo-msft: >auto [](start = 9, length = 4)

auto const

syncd/syncd_saiswitch.cpp
```diff
@@ -550,55 +539,102 @@ void SaiSwitch::collectCounters(
...
+                continue;
+            }
+
+            sai_object_id_t vid = translate_rid_to_vid(port_rid, m_switch_vid);
+
+            std::vector<swss::FieldValueTuple> values;
+
+            for (size_t idx = 0; idx < counters.size(); idx++)
+            {
+                const std::string &field = sai_serialize_port_stat(m_supported_counters[idx]);
```
qiluo-msft: >const std::string & [](start = 16, length = 19)

string/auto is just ok

syncd/syncd_saiswitch.cpp
qiluo-msft: >values.push_back( swss::FieldValueTuple(field, value) ); [](start = 16, length = 56)

emplace_back(field, value)

syncd/syncd_saiswitch.cpp
```diff
@@ -550,55 +539,102 @@ void SaiSwitch::collectCounters(
...
+void SaiSwitch::collectQueueCounters(
+        _In_ sai_object_id_t port_rid,
+        _In_ swss::Table &countersTable) const
+{
+    SWSS_LOG_ENTER();
 
+    uint32_t countersSize = (uint32_t)m_supported_queue_counters.size();
     std::vector<uint64_t> counters;
-
     counters.resize(countersSize);
```
qiluo-msft: >resize [](start = 13, length = 6)

The same

syncd/syncd_saiswitch.cpp
```diff
@@ -550,55 +539,102 @@ void SaiSwitch::collectCounters(
...
+    if (m_queue_rid_map.find(port_rid) == m_queue_rid_map.end())
+    {
+        SWSS_LOG_ERROR("failed to collect queues counters for port RID %s",
+                sai_serialize_object_id(port_rid).c_str());
 
-    for (auto &port_rid: ports)
+        return;
+    }
+
+    const std::vector<sai_object_id_t>& queue_list = m_queue_rid_map.at(port_rid);
```
qiluo-msft: >at [](start = 69, length = 2)

Already 'find' by previous statement, no need to 'at'.
I did add this comment on previous PR: [https://github.com/Azure/sonic-sairedis/pull/223](https://github.com/Azure/sonic-sairedis/pull/223)

syncd/syncd_saiswitch.cpp
```diff
@@ -550,55 +539,102 @@ void SaiSwitch::collectCounters(
...
         for (size_t idx = 0; idx < counters.size(); idx++)
         {
-            const std::string &field = sai_serialize_port_stat(m_supported_counters[idx]);
+            const std::string &field = sai_serialize_enum(m_supported_queue_counters[idx], &sai_metadata_enum_sai_queue_stat_t);
             const std::string &value = std::to_string(counters[idx]);
-
-            swss::FieldValueTuple fvt(field, value);
-
-            values.push_back(fvt);
+            values.push_back(swss::FieldValueTuple(field, value));
```
qiluo-msft: >push_back [](start = 19, length = 9)

The same

syncd/syncd_saiswitch.cpp
```diff
@@ -1185,6 +1221,142 @@ sai_object_id_t SaiSwitch::getDefaultValueForOidAttr(
...
+        if (status != SAI_STATUS_SUCCESS)
+        {
+            SWSS_LOG_ERROR("failed to collect queues list for port RID %s: %s",
+                    sai_serialize_object_id(port_rid).c_str(),
+                    sai_serialize_status(status).c_str());
+
+            continue;
+        }
+
+        queue_map[port_rid] = queue_list;
```
qiluo-msft: >queue_list [](start = 30, length = 10)

Could you optimize the vector copy?

syncd/syncd_saiswitch.cpp
qiluo-msft: actually needed. I suggest rewrite as while-loop for readability.

---
In reply to: [143101748](https://github.com/Azure/sonic-sairedis/pull/230#discussion_r143101748) [](ancestors = 143101748)

syncd/syncd_saiswitch.cpp
```diff
@@ -1185,6 +1221,142 @@ sai_object_id_t SaiSwitch::getDefaultValueForOidAttr(
...
+
+                ++queue_it;
+            }
+            else
+            {
+                /* Remove queue which doesn't support the statistic counters */
+                queue_it = port_queues.second.erase(queue_it);
+
+                SWSS_LOG_ERROR("failed to get stat counters for queue RID %s: %s",
+                        sai_serialize_object_id(*queue_it).c_str(),
```
qiluo-msft: >queue_it [](start = 49, length = 8)

The failed one is already erased, right?

syncd/syncd_saiswitch.cpp
```diff
@@ -1185,6 +1221,142 @@ sai_object_id_t SaiSwitch::getDefaultValueForOidAttr(
...
+    }
+}
+
+std::vector<sai_queue_stat_t> SaiSwitch::saiGetSupportedQueueCounters(sai_object_id_t queue_rid) const
+{
+    SWSS_LOG_ENTER();
+
+    std::vector<sai_queue_stat_t> supported_queue_counters;
+
+    for (uint32_t idx = 0; idx < sai_metadata_enum_sai_queue_stat_t.valuescount; ++idx)
```
qiluo-msft: >sai_metadata_enum_sai_queue_stat_t [](start = 33, length = 34)

where is it defined? why a variable name ends with _t (type)

syncd/syncd_saiswitch.cpp
```diff
@@ -1185,6 +1221,142 @@ sai_object_id_t SaiSwitch::getDefaultValueForOidAttr(
...
+    {
+        sai_queue_stat_t counter = (sai_queue_stat_t)sai_metadata_enum_sai_queue_stat_t.values[idx];
+        uint64_t value;
+        sai_status_t status = sai_metadata_sai_queue_api->get_queue_stats(queue_rid, 1, &counter, &value);
+
+        if (status != SAI_STATUS_SUCCESS)
+        {
+            const std::string &name = sai_serialize_enum(counter, &sai_metadata_enum_sai_queue_stat_t);
```
qiluo-msft: >const std::string & [](start = 12, length = 19)

auto/string

syncd/syncd_saiswitch.cpp
lguohan: I think the best approach is to do something simliar to pfc watchdog, let swss to tell syncd the ports and counters to query periodically.

andrii-savka: First of all, thank you for your suggestions regarding to syncd port/queue list implementation. 

From my point of view, the best approach is to update the port/queue list on remote_port api notification. 
I’ve checked the port break out behavior. We get the event from database when port is created/removed and could add a new case:
If port or queue SAI object is created/removed, we should update the port and queue list.
Also I have an following open question to this suggestion:
Should we update the port/queue list on a set operation?

In other hand, we can update the port/queue counters in similar way as it’s done in pfc watchdog. 
We can add/remove the port’s oid  to DB when port add/remove api in swss is called and update it in syncd_flex_counter component. 
The main disadvantage of this approach is that pfc watchdog has a very small updating interval = 100 msec. 
In this case, we need to do one from below things:
•	duplicate the implementation of syncd_flex_counter component for port/queue updating 
•	refactor the syncd_flex_counter component and make the updating interval configurable for different oid


Which of above approaches should be implemented for optimizing the port and queue counters update?

andrii-savka:
Updated the implementation to use Flex counter for review. Under manual testing.

syncd/syncd.cpp
```diff
@@ -2365,8 +2365,14 @@ void processFlexCounterEvent(
     const auto &key = kfvKey(kco);
     const auto &op = kfvOp(kco);
 
+    std::size_t delimiter = key.find_last_of(":");
+    const auto intervalStr = key.substr(delimiter + 1);
```
qiluo-msft: >delimiter [](start = 40, length = 9)

check delimiter first?

syncd/syncd.cpp
```diff
@@ -2365,8 +2365,14 @@ void processFlexCounterEvent(
...
+    const auto vidStr = key.substr(0, delimiter);
+
+    int pollInterval = std::stoi(intervalStr);
```
qiluo-msft: >stoi [](start = 28, length = 4)

catch exception?

syncd/syncd_flex_counter.cpp
```diff
@@ -119,11 +167,12 @@ void FlexCounter::removePort(
...
 
 void FlexCounter::removeQueue(
-        _In_ sai_object_id_t queueVid)
+        _In_ sai_object_id_t queueVid,
+        _In_ uint32_t pollInterval)
 {
     SWSS_LOG_ENTER();
 
-    FlexCounter &fc = getInstance();
+    FlexCounter &fc = getInstance(pollInterval);
```
qiluo-msft: >getInstance [](start = 22, length = 11)

if FlexCounter(pollInterval) doesn't exist, you will create one, and find nothing. The memory is never collected timely.

syncd/syncd.cpp
qiluo-msft: >pollInterval [](start = 8, length = 12)

I am not aware of the design discussion. Is there any used cases for multiple intervals in one switch? If yes, what is the usage for both set and remove?

qiluo-msft:
Please try your best not to modify existing commits that already reviewed. Otherwise all the previous comments are lost with context. We have no way to check incrementally.

syncd/syncd.cpp
andrii-savka: Done

syncd/syncd.cpp
andrii-savka: Done

syncd/syncd_flex_counter.cpp
andrii-savka: Let me explain the design notes. 
Once orchagent has finished the port or queue initialization it sets the sai id with poll interval and list of stat counters  to Flex counter DB (for each port and queue). Syncd handles the events, and add the counter to appropriate Flex counter instance. We have a few FC instances for each poll interval because stat counters and PFC WD counters are polling with different intervals.  
So, FC instances are created for each poll inverval and consist own lists with ports, queues and counters to update.

syncd/syncd.cpp
andrii-savka: Port or queue could be removed as part of breakout process.

andrii-savka:
Please review only. Still under testing. 
Also please review swss part:
https://github.com/Azure/sonic-swss/pull/358

andrii-savka:
Testing has finished. Please review and merge.

syncd/syncd_flex_counter.cpp
qiluo-msft: Thanks for the information. However, nothing is related to my comment above.

andrii-savka: @qiluo-msft 
Sorry, but I don't understand your comment.  
We have more than 500 items with the same poll interval and at least more of them will be added to the instance for sure. The instance will never live as empty. I don't see any reason to modify the code. 
Please explain more detailed.

qiluo-msft: Let's assume in this use case:
1. You have one instance for 100 ms, and one instance for 200 ms.
2. You call getInstance(1), then it will create a new instance for 1, and add it to g_flex_counters_map. This new instance is never released.

andrii-savka: The g_flex_counters_map contains a shared pointers to a flex counter instances. 
The instances will be released when syncd process will finishing work. 
Stop syncd -> destroy g_flex_counters_map  -> destroy Flex counter instances.

qiluo-msft: This is too late. It is strange that a 'get*' function will accumulate memory garbage.

syncd/syncd_flex_counter.cpp
```diff
@@ -74,11 +157,12 @@ void FlexCounter::setQueueCounterList(
 void FlexCounter::setQueueAttrList(
         _In_ sai_object_id_t queueVid,
         _In_ sai_object_id_t queueId,
+        _In_ uint32_t pollInterval,
         _In_ const std::vector<sai_queue_attr_t> &attrIds)
 {
     SWSS_LOG_ENTER();
 
-    FlexCounter &fc = getInstance();
+    FlexCounter &fc = getInstance(pollInterval);
```
qiluo-msft: >getInstance [](start = 22, length = 11)

remove instance?

syncd/syncd_flex_counter.cpp
```diff
@@ -143,19 +229,20 @@ void FlexCounter::removeQueue(
...
 
 void FlexCounter::addPortCounterPlugin(
-        _In_ std::string sha)
+        _In_ std::string sha,
+        _In_ uint32_t pollInterval)
 {
     SWSS_LOG_ENTER();
 
-    FlexCounter &fc = getInstance();
+    FlexCounter &fc = getInstance(pollInterval);
```
qiluo-msft: >getInstance [](start = 22, length = 11)

remove instance?

syncd/syncd_flex_counter.cpp
```diff
@@ -168,11 +255,12 @@ void FlexCounter::addPortCounterPlugin(
...
 
 void FlexCounter::addQueueCounterPlugin(
-        _In_ std::string sha)
+        _In_ std::string sha,
+        _In_ uint32_t pollInterval)
 {
     SWSS_LOG_ENTER();
 
-    FlexCounter &fc = getInstance();
+    FlexCounter &fc = getInstance(pollInterval);
```
qiluo-msft: >getInstance [](start = 22, length = 11)

remove instance?

syncd/syncd_flex_counter.cpp
```diff
@@ -185,11 +273,12 @@ void FlexCounter::addQueueCounterPlugin(
...
 
 void FlexCounter::removeCounterPlugin(
-        _In_ std::string sha)
+        _In_ std::string sha,
+        _In_ uint32_t pollInterval)
 {
     SWSS_LOG_ENTER();
 
-    FlexCounter &fc = getInstance();
+    FlexCounter &fc = getInstance(pollInterval);
```
qiluo-msft: >getInstance [](start = 22, length = 11)

remove instance?

syncd/syncd_flex_counter.cpp
andrii-savka: Why do we need to remove the instance after creating? 
The port/queue list will be added to the instance. 
At least, we don't know if instance is just created or it has been already created.

syncd/syncd_flex_counter.cpp
andrii-savka: Why do we need to remove the instance after creating? 
The port/queue list will be added to the instance. 
At least, we don't know if instance is just created or it has been already created.

syncd/syncd_flex_counter.cpp
andrii-savka: No need to remove instance here.
We need to remove it only if port counters, queue counters or attr maps are empty according to the original code.

syncd/syncd_flex_counter.cpp
andrii-savka: No need to remove instance here.
We need to remove it only if port counters, queue counters or attr maps are empty according to the original code.

syncd/syncd_flex_counter.cpp
qiluo-msft: I mean if you create a new instance during getInstance, and find nothing inside, the unused instance should be garbage collected.

---
In reply to: [149030630](https://github.com/Azure/sonic-sairedis/pull/230#discussion_r149030630) [](ancestors = 149030630)

andrii-savka:
Rebased to the latest

qiluo-msft:
>     {

If you create an empty instance for pollInterval, find nothing then it will return here. You should remove the new empty instance.

---
Refers to: syncd/syncd_flex_counter.cpp:169 in 1d13c3b. [](commit_id = 1d13c3b62e2e1a920fcab02f412c50415d1b9a1a, deletion_comment = False)

qiluo-msft:
>         SWSS_LOG_ERROR("Trying to remove nonexisting port counter Ids 0x%lx", portVid);

If you create an empty instance for pollInterval, find nothing then it will return here. You should remove the new empty instance.

---
Refers to: syncd/syncd_flex_counter.cpp:193 in 1d13c3b. [](commit_id = 1d13c3b62e2e1a920fcab02f412c50415d1b9a1a, deletion_comment = False)

qiluo-msft:
>         SWSS_LOG_ERROR("Trying to remove nonexisting queue counter Ids 0x%lx", queueVid);

If you create an empty instance for pollInterval, find nothing then it will return here. You should remove the new empty instance.

---
Refers to: syncd/syncd_flex_counter.cpp:217 in 1d13c3b. [](commit_id = 1d13c3b62e2e1a920fcab02f412c50415d1b9a1a, deletion_comment = False)

qiluo-msft:
>         SWSS_LOG_ERROR("Trying to remove nonexisting queue attr Ids 0x%lx", queueVid);

If you create an empty instance for pollInterval, find nothing then it will return here. You should remove the new empty instance.

---
Refers to: syncd/syncd_flex_counter.cpp:226 in 1d13c3b. [](commit_id = 1d13c3b62e2e1a920fcab02f412c50415d1b9a1a, deletion_comment = False)

andrii-savka:
New remarks are fixed except the first one - no need to remove in that case because the list will be assigned in the any case.

PR #232, opened 2017-09-27
Enabled syncd ports statistic collection.

nazarii-gnydyn:


kcudnik:
Seems like error is related to sai version used in build system branch

syncd/scripts/syncd_init_common.sh
```diff
@@ -80,7 +80,7 @@ config_syncd_centec()
 
 config_syncd_cavium()
 {
-    CMD_ARGS+=" -p /etc/ssw/AS7512/profile.ini -N"
+    CMD_ARGS+=" -p /etc/ssw/AS7512/profile.ini"
```
stcheng: just curious. when is this file get installed? why not using `-p $HWSKU_DIR/sai.profile`

PR #240, opened 2017-10-10
[saiserialize]: Make delete[] match new[]

marian-pritsak:


lguohan:
@kcudnik , please review.

prsunny:
looks good to me since fdbevent is allocated as new []

PR #241, opened 2017-10-10
[pfcwd]: Add support to poll queue attr in syncd

sihuihan88:


sihuihan88:
Depends on https://github.com/Azure/sonic-swss-common/pull/141

syncd/syncd_pfc_watchdog.cpp
```diff
@@ -64,6 +71,30 @@ void PfcWatchdog::setQueueCounterList(
...
+
+    auto it = wd.m_queueAttrIdsMap.find(queueVid);
+    if (it != wd.m_queueAttrIdsMap.end())
+    {
+        (*it).second->queueAttrIds = attrIds;
+        return;
+    }
+
+    auto queueAttrIds = std::make_shared<QueueAttrIds>(queueId, attrIds);
+    wd.m_queueAttrIdsMap.emplace(queueVid, queueAttrIds);
```
marian-pritsak: Need to erase IDs on delete

sihuihan88: Good catch. Updated

syncd/syncd_pfc_watchdog.cpp
```diff
@@ -176,7 +216,6 @@ void PfcWatchdog::collectCounters(
 {
     SWSS_LOG_ENTER();
 
-    std::lock_guard<std::mutex> lock(g_mutex);
```
stcheng: remove one more empty line

PR #244, opened 2017-10-12
Add vslib support for tap device

kcudnik:


vslib/src/sai_vs_hostintf.cpp
```diff
@@ -29,7 +67,414 @@ sai_status_t vs_send_hostif_packet(
...
+    }
+
+    // TODO what about FDB entries notifications, they also should
+    // be generated if new mac addres will show up on the interface/arp table
+
+    // TODO IP address should be assigned when router interface is created
+
+    SWSS_LOG_INFO("created tap interface %s", name.c_str());
+
+    return vs_generic_create(object_type,
```
qiluo-msft: >return [](start = 4, length = 6)

Close tapfd before teturn?

vslib/src/sai_vs_hostintf.cpp
```diff
@@ -29,7 +67,414 @@ sai_status_t vs_send_hostif_packet(
...
+    if (tapfd < 0)
+    {
+        SWSS_LOG_ERROR("failed to create TAP device for %s", name.c_str());
+
+        return SAI_STATUS_FAILURE;
+    }
+
+    SWSS_LOG_INFO("created TAP device for %s, fd: %d", name.c_str(), tapfd);
+
+    // XXX currently tapfd is ignored, it should be closed on hostif_remove
```
qiluo-msft: >XXX [](start = 7, length = 3)

XXX is confusing. Use TODO?

vslib/src/sai_vs_hostintf.cpp
```diff
@@ -29,7 +67,414 @@ sai_status_t vs_send_hostif_packet(
...
+
+        close(fd);
+
+        return err;
+    }
+
+    return fd;
+}
+
+int vs_set_dev_mac_address(const char *dev, const sai_mac_t mac)
```
qiluo-msft: >sai_mac_t [](start = 50, length = 9)

sai_mac_t&

vslib/src/sai_vs_hostintf.cpp
```diff
@@ -29,7 +67,414 @@ sai_status_t vs_send_hostif_packet(
...
+    {
+        SWSS_LOG_ERROR("ioctl SIOCSIFFLAGS on socket %d %s failed, err %d", s, dev, err);
+    }
+
+    close(s);
+
+    return err;
+}
+
+void veth2tap_fun(std::shared_ptr<hostif_info_t> info)
```
qiluo-msft: >std::shared_ptr<hostif_info_t> [](start = 18, length = 30)

"hostif_info_t* info"
is good enough.

vslib/src/sai_vs_hostintf.cpp
```diff
@@ -29,7 +67,414 @@ sai_status_t vs_send_hostif_packet(
...
+
+        if (write(info->tapfd, buffer, size) < 0)
+        {
+            SWSS_LOG_ERROR("failed to write to tap device %d", info->tapfd);
+            break;
+        }
+    }
+}
+
+void tap2veth_fun(std::shared_ptr<hostif_info_t> info)
```
qiluo-msft: >std::shared_ptr<hostif_info_t> [](start = 18, length = 30)

The same

vslib/src/sai_vs_hostintf.cpp
```diff
@@ -29,7 +67,414 @@ sai_status_t vs_send_hostif_packet(
...
+    }
+
+    if (strnlen(attr_name->value.chardata, sizeof(attr_name->value.chardata)) >= MAX_INTERFACE_NAME_LEN)
+    {
+        SWSS_LOG_ERROR("interface name is too long: %.*s", MAX_INTERFACE_NAME_LEN, attr_name->value.chardata);
+
+        return SAI_STATUS_FAILURE;
+    }
+
+    if (strncmp(attr_name->value.chardata, "Ethernet", 8) != 0)
```
qiluo-msft: >8 [](start = 55, length = 1)

sizeof("Ethernet") -1 ?

vslib/src/sai_vs_hostintf.cpp
qiluo-msft: >XXX [](start = 7, length = 3)

The same

vslib/src/sai_vs_interfacequery.cpp
```diff
@@ -92,6 +93,13 @@ sai_status_t sai_api_initialize(
         return SAI_STATUS_FAILURE;
     }
 
+    const char *use_tap_dev = service_method_table->profile_get_value(0, SAI_KEY_VS_HOSTIF_USE_TAP_DEVICE);
```
qiluo-msft: >use_tap_dev [](start = 16, length = 11)

The validity or lifetime of returned string.c_str() is problematic. Suggest capture it in a string, or make profile_get_value() return a string directly.

vslib/src/sai_vs_hostintf.cpp
kcudnik: no, tap descriptor is used in thread to write packets

vslib/src/sai_vs_hostintf.cpp
kcudnik: XXX is used to get programmers attention, but i will change to TODO

vslib/src/sai_vs_hostintf.cpp
kcudnik: im using shard ptr since i want to avoid using "new" operator, and i use std::make_shared which will perform delete when needed

vslib/src/sai_vs_interfacequery.cpp
kcudnik: profile_get_value is sai api header defined, it needs to reurn char* since its ansi c compiant, i will change to string

vslib/src/sai_vs_hostintf.cpp
```diff
@@ -29,7 +67,414 @@ sai_status_t vs_send_hostif_packet(
...
+        return -1;
+    }
+
+    struct ifreq ifr;
+
+    memset(&ifr, 0, sizeof(ifr));
+
+    ifr.ifr_flags = (short int)flags;  // IFF_TUN or IFF_TAP, IFF_NO_PI
+
+    strncpy(ifr.ifr_name, dev, IFNAMSIZ);
```
prsunny: Just thinking if you should verify strlen(dev) < IFNAMSIZ

vslib/src/sai_vs_hostintf.cpp
```diff
@@ -29,7 +67,414 @@ sai_status_t vs_send_hostif_packet(
...
+        return -1;
+    }
+
+    struct ifreq ifr;
+
+    memset(&ifr, 0, sizeof ifr);
+
+    strncpy(ifr.ifr_name, dev , IFNAMSIZ);
+
+    ifr.ifr_flags |= IFF_UP;
```
prsunny: Since this flags is memset to 0, just want to let you know that in kernel, any other flag that is previously set for this interface would be overwritten and only IFF_UP would be set.

vslib/src/sai_vs_hostintf.cpp
```diff
@@ -29,7 +67,414 @@ sai_status_t vs_send_hostif_packet(
...
+            break;
+        }
+    }
+}
+
+void tap2veth_fun(std::shared_ptr<hostif_info_t> info)
+{
+    SWSS_LOG_ENTER();
+
+    unsigned char buffer[0x4000];
```
prsunny: Just a note - 16K bytes allocated from stack.

vslib/src/sai_vs_hostintf.cpp
prsunny: I see that you are checking this in the calling function.. {closed}

vslib/src/sai_vs_hostintf.cpp
qiluo-msft: I know the benefits of shared pointer. I mean in this function, raw pointer is good enough.

---
In reply to: [144568446](https://github.com/Azure/sonic-sairedis/pull/244#discussion_r144568446) [](ancestors = 144568446)

vslib/src/sai_vs_hostintf.cpp
qiluo-msft: Also in above function.

---
In reply to: [144614375](https://github.com/Azure/sonic-sairedis/pull/244#discussion_r144614375) [](ancestors = 144614375)

vslib/src/sai_vs_hostintf.cpp
qiluo-msft: I still suggest to protect for the boundary case in function local scope.

---
In reply to: [144618322](https://github.com/Azure/sonic-sairedis/pull/244#discussion_r144618322) [](ancestors = 144618322)

vslib/src/sai_vs_hostintf.cpp
prsunny: @qiluo-msft , I think if you want to serve the intended purpose of shared pointer since two threads are manipulating it separately, the pointer has to be passed as above. 
Having said that, since this is in the packet flow, any kind of extra overhead should ideally be avoided.

vslib/src/sai_vs_hostintf.cpp
prsunny: Agree

vslib/src/sai_vs_hostintf.cpp
qiluo-msft: Thanks Prince! Got the idea.

---
In reply to: [144668459](https://github.com/Azure/sonic-sairedis/pull/244#discussion_r144668459) [](ancestors = 144668459)

vslib/src/sai_vs_hostintf.cpp
```diff
@@ -29,7 +67,414 @@ sai_status_t vs_send_hostif_packet(
...
+
+    info->e2t->detach();
+    info->t2e->detach();
+
+    SWSS_LOG_NOTICE("setup forward rule for %s succeeded", tapname.c_str());
+
+    return true;
+}
+
+sai_status_t vs_create_hostif_int(
```
qiluo-msft: >vs_create_hostif_int [](start = 13, length = 20)

What is the use case for this function? Do you have a way to decently clean up? ie. stop the threads. I just feel the tricks to keep shared_pointer live is too smart...

vslib/src/sai_vs_hostintf.cpp
```diff
@@ -29,7 +67,414 @@ sai_status_t vs_send_hostif_packet(
...
+
+        return SAI_STATUS_FAILURE;
+    }
+
+    if (strncmp(attr_name->value.chardata, "Ethernet", 8) != 0)
+    {
+        SWSS_LOG_ERROR("interface name should start with EthernetX but is %s", attr_name->value.chardata);
+
+        return SAI_STATUS_FAILURE;
+    }
```
lguohan: why do we need this restriction?

vslib/src/sai_vs_hostintf.cpp
```diff
@@ -29,7 +67,414 @@ sai_status_t vs_send_hostif_packet(
...
+        SWSS_LOG_ERROR("failed to set MAC address %s for %s",
+                sai_serialize_mac(attr.value.mac).c_str(),
+                name.c_str());
+
+        close(tapfd);
+
+        return SAI_STATUS_FAILURE;
+    }
+
+    err = ifup(name.c_str());
```
lguohan: I think orchagent will have logic to bring it up, I am not sure we need to do this in the sai.

vslib/src/sai_vs_hostintf.cpp
lguohan: should be fine, the thread stack size is 8MB for 64-bit os and 2MB for 32-bit os

---
In reply to: [144627332](https://github.com/Azure/sonic-sairedis/pull/244#discussion_r144627332) [](ancestors = 144627332,144614375)

vslib/src/sai_vs_hostintf.cpp
```diff
@@ -29,7 +67,414 @@ sai_status_t vs_send_hostif_packet(
...
+    if (err < 0)
+    {
+        SWSS_LOG_ERROR("failed to bring ifup %s", name.c_str());
+
+        close(tapfd);
+
+        return SAI_STATUS_FAILURE;
+    }
+
+    if (!hostif_create_tap_veth_forwarding(name, tapfd))
```
lguohan: I think it is better to open the peer link netdev in the VS init functions. We can then assign each port a lane information, and return the lane information to the orchagent and a port id. In the create_hostif, we just attach the host interface (tap device) to the peer link based on the port id.

This work flow is more like the SAI flow.

vslib/src/sai_vs_hostintf.cpp
kcudnik: fixed

vslib/src/sai_vs_hostintf.cpp
kcudnik: fixed

vslib/src/sai_vs_hostintf.cpp
kcudnik: fixed

vslib/src/sai_vs_hostintf.cpp
kcudnik: fixed

vslib/src/sai_vs_hostintf.cpp
kcudnik: shouldbe fine

vslib/src/sai_vs_hostintf.cpp
kcudnik: threads should be stopped in remove function, this is not implemented yet, see XXX/TODO comment that you commented

vslib/src/sai_vs_hostintf.cpp
kcudnik: hmm, actually we dont

vslib/src/sai_vs_hostintf.cpp
kcudnik: i brought interface up since forwarding was not possible on interface that was down

vslib/src/sai_vs_hostintf.cpp
kcudnik: peer link is assigned when hostif is created, at VS init its not know yet, currenty all functions are asynchronous and orchagent has no idea about any return information, also lane information also can be obtained and assigned after hostif create using GET sai functions

lguohan: orchagent will query the lane information first, is that also asynchronous even for the get API?

kcudnik: get is synchronous, but why you want to create hostif at vs init ?

vslib/src/sai_vs_hostintf.cpp
kcudnik: i removed ifup but i dont see any reason why it can't be brought up at the first place

vslib/src/sai_vs_hostintf.cpp
lguohan: I do not mean create hostif at the vs init. I mean just open the virtual link at the vs init. The virtual link is the physical interface of the virtual switch. They should exist whether or not there is host interface. For example, you can get link status, notification, learn mac regardless of host interface creation or not.

I am thinking, maybe we should open the vpeer link at the vs init and start processing packet, maybe doing the mac learning and a few other things. Only when when create hostif then we can transferring the packets between the vpeer link and the tap devices.

For now, I am ok.

PR #247, opened 2017-10-20
Add get_stats SAI API support

marian-pritsak:


lguohan:
@kcudnik , @qiluo-msft , please review.

meta/sai_meta.cpp
```diff
@@ -4437,6 +4477,67 @@ sai_status_t meta_sai_get_oid(
...
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        return status;
+    }
+
+    if (get == NULL)
+    {
+        SWSS_LOG_ERROR("get function pointer is NULL");
+
+        return SAI_STATUS_FAILURE;
```
qiluo-msft: >SAI_STATUS_FAILURE [](start = 15, length = 18)

SAI_STATUS_INVALID_PARAMETER?

lib/src/sai_redis_generic_get_stats.cpp
```diff
@@ -0,0 +1,192 @@
...
+        _In_ swss::KeyOpFieldsValuesTuple &kco)
+{
+    SWSS_LOG_ENTER();
+
+    // key:         sai_status
+    // field:       stat_id
+    // value:       stat_value
+
+    const std::string &key = kfvKey(kco);
+    const std::vector<swss::FieldValueTuple> &values = kfvFieldsValues(kco);
```
qiluo-msft: >const std::vector<swss::FieldValueTuple> [](start = 4, length = 40)

suggest use 'auto' to make code readable.

lib/src/sai_redis_generic_get_stats.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+    if (status == SAI_STATUS_SUCCESS)
+    {
+        uint32_t i = 0;
+        for (const auto &v : values)
+        {
+            uint64_t value = 0;
+
+            value = stoull(fvValue(v));
+            counter_list[i] = value;
```
qiluo-msft: >i [](start = 25, length = 1)

check for index out of range at the loop beginning?

lib/src/sai_redis_generic_get_stats.cpp
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    return status;
+}
+
+// Another way to implement this would be to pass sai_serialize_ ## type ## stat as a parameter,
+// but the decision point of what function to pass is up in sai_redis_internal.h, which would make us
+// pass a callback through about five function calls, so for the sake of simplicity here's a macro that
+// generates overloaded functions
+#define DECLARE_SERIALIZE_COUNTER_ID_LIST(type)                                                               \
```
qiluo-msft: >DECLARE_SERIALIZE_COUNTER_ID_LIST [](start = 8, length = 33)

function macro is always unfriendly, suggest implement by template. You may need some more template function such as sai_serialize_stat<T> to implement this one.

lib/src/sai_redis_generic_get_stats.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+    // wait for response
+
+    swss::Select s;
+
+    s.addSelectable(g_redisGetConsumer.get());
+
+    while (true)
+    {
+        SWSS_LOG_DEBUG("wait for response");
```
qiluo-msft: >wait for response [](start = 24, length = 17)

wait for get_stats response

lib/src/sai_redis_generic_get_stats.cpp
```diff
@@ -0,0 +1,192 @@
...
+            sai_status_t status = internal_redis_get_stats_process(
+                    object_type,
+                    count,
+                    counter_list,
+                    kco);
+
+            if (g_record)
+            {
+                const std::string &str_status = kfvKey(kco);
+                const std::vector<swss::FieldValueTuple> &values = kfvFieldsValues(kco);
```
qiluo-msft: >const std::vector<swss::FieldValueTuple> [](start = 16, length = 40)

auto &

syncd/syncd.cpp
```diff
@@ -1570,6 +1570,111 @@ sai_status_t notifySyncd(
...
+
+    switch (object_type)
+    {
+        case SAI_OBJECT_TYPE_PORT:
+            status = getStatsGeneric<sai_port_stat_t>(
+                    rid,
+                    kco,
+                    result,
+                    sai_metadata_sai_port_api->get_port_stats,
+                    sai_deserialize_port_stat);
```
qiluo-msft: >sai_deserialize_port_stat [](start = 20, length = 25)

You may define a template function sai_deserialize_stat<T>, then simplify all the template functions

meta/sai_meta.cpp
marian-pritsak: Done

lib/src/sai_redis_generic_get_stats.cpp
marian-pritsak: Done

lib/src/sai_redis_generic_get_stats.cpp
marian-pritsak: Done

lib/src/sai_redis_generic_get_stats.cpp
marian-pritsak: Done

lib/src/sai_redis_generic_get_stats.cpp
marian-pritsak: Done

syncd/syncd.cpp
marian-pritsak: I cannot generalize over serialize_stat family of functions. The reason is that each one of them uses a different metadata enum that cannot be deferred from sai_*_stat_t. So I'll have to generate a set of overloaded functions instead.

lib/src/sai_redis_generic_get_stats.cpp
marian-pritsak: I cannot generalize over serialize_stat family of functions. The reason is that each one of them uses a different metadata enum that cannot be deferred from sai_*_stat_t. So I'll have to generate a set of overloaded functions instead.

qiluo-msft: Try something like
```
template<typename TYPE> struct sai_type_stat_t { };
template<> struct sai_type_stat_t<port> { typedef sai_port_stat_t type; };
template<> struct sai_type_stat_t<queue> { typedef sai_queue_stat_t type; };
```

---
In reply to: [147784356](https://github.com/Azure/sonic-sairedis/pull/247#discussion_r147784356) [](ancestors = 147784356)

marian-pritsak: It doesn't make a significant difference. I'll still need to define type traits for every type, and we will have a duplication in api because all other code uses old serialize/deserialize functions etc. and make metadata api inconsistent with other code. We already have similar defines in all sairedis_internal.h so I don't really know what's the benefit of doing such changes.

qiluo-msft: The goal here is to reduce the unfriendly function macro in the code, including existing one. So inconsistence is not a true problem here, and we have plan to readability and maintainable in future.

The review comment here is to ask your help to investigate the alternatives. This PR could be a good starting point.

---
In reply to: [147925839](https://github.com/Azure/sonic-sairedis/pull/247#discussion_r147925839) [](ancestors = 147925839)

lib/src/sai_redis_generic_get_stats.cpp
```diff
@@ -0,0 +1,201 @@
+#include "sai_redis.h"
+#include "meta/saiserialize.h"
+
+sai_status_t internal_redis_get_stats_process(
```
nikos-li: This is not an acceptable coding style.

You can write your function like so:

sai_status_t internal_redis_get_stats_process(_In_ sai_object_type_t object_type,
                                                                            _In_ uint32_t count,
                                                                            _Out_ uint64_t *counter_list,
                                                                            _In_ swss::KeyOpFieldsValuesTuple &kco)

or

sai_status_t internal_redis_get_stats_process(_In_ sai_object_type_t object_type, _In_ uint32_t count,
                                                                            _Out_ uint64_t *counter_list,  _In_ swss::KeyOpFieldsValuesTuple &kco)

but not what you have.

Same hold for other similar functions in your diffs.

marian-pritsak: Can you please explain what's the difference between functions in this file and https://github.com/Azure/sonic-sairedis/blob/master/lib/src/sai_redis_generic_create.cpp ?

nikos-li: Of course - you're just pointing out that different engineers have different styles, which is expected. So since that style wasn't corrected at commit time, eventually people will follow it and we should go ahead and adopt it, hence ending up with different styles inside sonic. Makes perfect sense to me :-)

marian-pritsak: Actually, I'm just trying to make it similar to what is already there, like

https://github.com/Azure/sonic-sairedis/blob/master/lib/src/sai_redis_buffer.cpp
https://github.com/Azure/sonic-sairedis/blob/master/lib/src/sai_redis_notifications.cpp
https://github.com/Azure/sonic-sairedis/blob/master/lib/src/sai_redis_policer.cpp
...

sairedis library has a common style across almost all its source files. There is, of course, some code that doesn't follow common style, but most of it follows the common pattern and I, in turn, follow it.

If you want some formal version of code style or you don't agree with how my function declarations look (which are same as almost all function declarations in sairedis), then I think that it's up to a discussion with the maintainer.

lib/src/sai_redis_generic_get_stats.cpp
marian-pritsak: Changed to template

lib/src/sai_redis_generic_get_stats.cpp
```diff
@@ -0,0 +1,220 @@
...
+        }
+    }
+
+    return status;
+}
+
+template <class T> class stat_traits {};
+
+template <>
+class stat_traits<sai_port_stat_t>
```
qiluo-msft: >class [](start = 0, length = 5)

better to use struct and remove public, so code will be cleaner.

lib/src/sai_redis_generic_get_stats.cpp
```diff
@@ -0,0 +1,220 @@
...
+        typedef std::string (*serialize_stat)(sai_port_stat_t);
+        static constexpr serialize_stat serialize_stat_fn = sai_serialize_port_stat;
+};
+
+template <>
+class stat_traits<sai_queue_stat_t>
+{
+    public:
+        typedef std::string (*serialize_stat)(sai_queue_stat_t);
+        static constexpr stat_traits<sai_queue_stat_t>::serialize_stat serialize_stat_fn = sai_serialize_queue_stat;
```
qiluo-msft: >stat_traits<sai_queue_stat_t> [](start = 25, length = 29)

It can be removed.

lib/src/sai_redis_generic_get_stats.cpp
```diff
@@ -0,0 +1,220 @@
...
+};
+
+template <>
+class stat_traits<sai_ingress_priority_group_stat_t>
+{
+    public:
+        typedef std::string (*serialize_stat)(sai_ingress_priority_group_stat_t);
+        static constexpr stat_traits<sai_ingress_priority_group_stat_t>::serialize_stat serialize_stat_fn = sai_serialize_ingress_priority_group_stat;
```
qiluo-msft: >stat_traits<sai_ingress_priority_group_stat_t>:: [](start = 25, length = 48)

the same

lib/src/sai_redis_generic_get_stats.cpp
```diff
@@ -0,0 +1,220 @@
...
+    return status;
+}
+
+template <class T> class stat_traits {};
+
+template <>
+class stat_traits<sai_port_stat_t>
+{
+    public:
+        typedef std::string (*serialize_stat)(sai_port_stat_t);
```
qiluo-msft: >typedef [](start = 8, length = 7)

typedef can be moved to the generic class. So only one is needed.

lib/src/sai_redis_generic_get_stats.cpp
marian-pritsak: Agree. Done.

lib/src/sai_redis_generic_get_stats.cpp
marian-pritsak: Done.

lib/src/sai_redis_generic_get_stats.cpp
marian-pritsak: Done

lib/src/sai_redis_generic_get_stats.cpp
marian-pritsak: It cannot. It is deferred from template parameter type, for which I define a specialization,

qiluo-msft: I see. Thanks!

PR #251, opened 2017-10-26
[syncd]: The cause of 10 minutes long reboot for SONiC project removed.

Mrak-IW:
This line causes to deletion of the database content AFTER configuration set.

Signed-off-by: Denis Maslov denis.maslov@cavium.com


stcheng:
thanks for fixing this issue.

PR #255, opened 2017-10-27
[254] Add port state change notification support to vslib

### LINKED ISSUE #254: Add port state change notification support to vslib
Notifications should be working when tap device is in use.

kcudnik:
Fixes #254

### DIFF
--- a/vslib/src/sai_vs_switch.cpp
+++ b/vslib/src/sai_vs_switch.cpp
@@ -89,7 +89,7 @@ class LinkMsg : public swss::NetMsg
             unsigned int    if_flags = rtnl_link_get_flags(link); // IFF_LOWER_UP and IFF_RUNNING
             const char*     if_name  = rtnl_link_get_name(link);
 
-            if (if_name[0] != 'v')
+            if (strncmp(if_name, SAI_VS_VETH_PREFIX, sizeof(SAI_VS_VETH_PREFIX) - 1) != 0)
             {
                 SWSS_LOG_INFO("skipping newlink for %s", if_name);
                 return;
--- a/vslib/src/sai_vs_switch.cpp
+++ b/vslib/src/sai_vs_switch.cpp
@@ -1,5 +1,287 @@
 #include "sai_vs.h"
 #include "sai_vs_internal.h"
+#include "sai_vs_state.h"
+
+#include "swss/selectableevent.h"
+#include "swss/netmsg.h"
+#include "swss/netlink.h"
+#include "swss/select.h"
+#include "swss/netdispatcher.h"
+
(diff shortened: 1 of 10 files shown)


kcudnik:
libnl-3 dev packages needs to be installed into jenkins build env to make project compile

```
cc1plus: error: /usr/include/libnl3: No such file or directory [-Werror]
```

lguohan:
retest this please

lguohan:
retest this please

lguohan:
retest this please

vslib/src/sai_vs_switch.cpp
```diff
@@ -1,5 +1,287 @@
...
+
+            // new link
+
+            struct rtnl_link *link = (struct rtnl_link *)obj;
+
+            int             if_index = rtnl_link_get_ifindex(link);
+            unsigned int    if_flags = rtnl_link_get_flags(link); // IFF_LOWER_UP and IFF_RUNNING
+            const char*     if_name  = rtnl_link_get_name(link);
+
+            if (if_name[0] != 'v')
```
lguohan: >'v' [](start = 30, length = 3)

this 'v' is used in multiple places, can we put this as a macro, so that if we want to change, we can change only one place.

vslib/src/sai_vs_switch_BCM56850.cpp
```diff
@@ -185,6 +199,11 @@ static sai_status_t create_ports()
 
         CHECK_STATUS(vs_generic_set(SAI_OBJECT_TYPE_PORT, port_id, &attr));
 
+        attr.id = SAI_PORT_ATTR_OPER_STATUS;
+        attr.value.s32 = SAI_PORT_OPER_STATUS_UP;
```
lguohan: >SAI_PORT_OPER_STATUS_UP [](start = 25, length = 23)

isn't the default value of port oper status to be down, and when you see netlink messages you will set it to be the correct value.

vslib/src/sai_vs_switch.cpp
```diff
@@ -1,5 +1,287 @@
...
+        _In_ sai_object_id_t switch_id)
+{
+    SWSS_LOG_ENTER();
+
+    if (g_vs_hostif_use_tap_device == false)
+    {
+        return;
+    }
+
+    auto sw = vs_get_switch_state(switch_id);
```
lguohan: what about initial value of port status? should we do a query for all vports and get their initial up/down status when we create our switch?

kcudnik:
I can set initial status to down, and when connecting to netlink we will receive actual status anyway

lguohan:
ok

PR #256, opened 2017-11-03
[syncd_flex_counter.cpp]: Move lock_guard to narrow scope

marian-pritsak:
lock_guard holds mutex during sleep, which does not allow other threads
to acquire it for most of time. It is moved to a scope that actually
needs this mutex.


marian-pritsak:
Since https://github.com/Azure/sonic-sairedis/pull/241

lguohan:
@qiluo-msft  to review.

PR #260, opened 2017-11-17
[rpc server]: Enable syncd-rpc target compilation.

oleksandrivantsiv:
Export switch ID variable required by SAI thrift.

### DIFF
--- a/syncd/syncd.cpp
+++ b/syncd/syncd.cpp
@@ -65,6 +65,13 @@ std::set<sai_object_id_t> initViewRemovedVidSet;
  */
 volatile bool g_asicInitViewMode = false;
 
+#ifdef SAITHRIFT
+/*
+ * SAI switch global needed for RPC server
+ */
+sai_object_id_t gSwitchId;
+#endif
+
 struct cmdOptions
 {
     int countersThreadIntervalInSeconds;
@@ -1641,6 +1648,10 @@ void on_switch_create_in_init_view(
             SWSS_LOG_THROW("failed to create switch in init view mode: %s",
                     sai_serialize_status(status).c_str());
         }
+
+#ifdef SAITHRIFT
+        gSwitchId = switch_rid;
+#endif
 
         /*
          * Object was created so new object id was generated we
--- a/debian/rules
+++ b/debian/rules
@@ -31,7 +31,7 @@ binary-syncd:
 	dh binary -N syncd-rpc -N syncd-rpc-dbg -N syncd-vs -N syncd-vs-dbg --with autotools-dev
 
 binary-syncd-rpc: | binary-syncd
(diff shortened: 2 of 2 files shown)


syncd/syncd.cpp
```diff
@@ -65,6 +65,13 @@ std::set<sai_object_id_t> initViewRemovedVidSet;
  */
 volatile bool g_asicInitViewMode = false;
 
+#ifdef SAITHRIFT
+/*
+ * SAI switch global needed for RPC server
+ */
+sai_object_id_t gSwitchId;
```
pavel-shirshov: What component does use the gSwitchId variable ?
I see you assign some value to it below, but how it is used?

oleksandrivantsiv: It is required by the latest SAI thrift implementation:
https://github.com/opencomputeproject/SAI/blob/master/test/saithrift/src/switch_sai_rpc_server.cpp#L80

pavel-shirshov: I'd better to make interface of SAI thrift implementation more explicit by introducing 
switch_id parameter to start_sai_thrift_rpc_server() function.
But anyway.

lguohan:
why PR build fails? what do I need to add?

andrii-savka:
We need to revert src/SAI to 1.0.2 version. The following commit can damage the build:
_commit aac4d6ecd6aef5d6a3f0805516bb3d515ec8291b
Author: Yakiv Huryk <yakivh@mellanox.com>
Date:   Sat Nov 18 12:35:41 2017 +0200

    PTF: alignment with 1.2 headers & fixes (#736)_

If I revert the src/SAI to following commit **_77b2ad28b8b6f3ad02105f209767dbda0fbba2a4_** then I have successfully built the image.

lguohan:
retest this please

andrii-savka:
@lguohan 
It was build and re-tested on the latest source code.

andrii-savka:
@lguohan 
The build is failing because of syncd can't include "switch_sai_rpc_server.h" from SAI. 
I don't have such error in my local build. 
Probably, we need to update SAI. 
I'm investigating how to fix it.

andrii-savka:
@lguohan 
The flow to build syncd with SAI thrift is the same. 
Could you check whether SAI which is used by Jenkins has a "switch_sai_rpc_server.h" thrift headers?

pavel-shirshov:
This build requires libsaithrift-dev. This library build was broken when we migrate from 0.9.4 to 1.0.2.
I'll try to fix the whole build process in the buildimage repo first.
I'll write results here as soon as I found anything.

pavel-shirshov:
Hi Guohan
You need to install libsaithrift-dev_1.0.4_amd64.deb
It should help you.

andrii-savka:
Tested and created rebased PR: #281

lguohan:
merged the #281

PR #261, opened 2017-11-20
[syncd]: Only query queue counters that are needed by the application

stcheng:
The previous hardcoded for loop to iterate items defined in the SAI
header file was not ideal. Update the logic here so that only needed
attributes are checked for the support.

Signed-off-by: Shu0T1an ChenG <shuche@microsoft.com>

### DIFF
--- a/syncd/syncd_flex_counter.cpp
+++ b/syncd/syncd_flex_counter.cpp
@@ -27,6 +27,11 @@ FlexCounter::QueueAttrIds::QueueAttrIds(
 {
 }
 
+/* The current implementation of 'setPortCounterList' and 'setQueueCounterList' are
+ * not the same. Need to refactor these two functions to have the similar logic.
+ * Either the full SAI attributes are queried once, or each of the needed counters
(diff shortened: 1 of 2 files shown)


stcheng:
This pull request is updated.

syncd/syncd_flex_counter.cpp
```diff
@@ -579,25 +580,29 @@ void FlexCounter::saiUpdateSupportedPortCounters(sai_object_id_t portId)
...
+            SWSS_LOG_WARN("Counter %s is not supported on queue %s, rv: %s",
                     sai_serialize_queue_stat(counter).c_str(),
                     sai_serialize_object_id(queueId).c_str(),
                     sai_serialize_status(status).c_str());
 
             continue;
         }
+        else
+        {
+            m_supportedQueueCounters.insert(counter);
```
lguohan: do we need to clear this counter?

stcheng: fixed.

PR #263, opened 2017-11-28
update latest v1.2

SAI-shared-user:


SAI-shared-user:
code review only, do not merge.

PR #264, opened 2017-11-28
update latest v1.2

lguohan:


lib/src/sai_redis_route.cpp
```diff
@@ -288,3 +280,16 @@ sai_status_t sai_bulk_get_route_entry_attribute(
...
 }
+
+REDIS_GENERIC_QUAD_ENTRY(ROUTE_ENTRY,route_entry);
+
+const sai_route_api_t redis_route_api = {
+
+    REDIS_GENERIC_QUAD_API(route_entry)
+
+    // TODO: upstream signiture fix to SAI repo
+    (sai_bulk_create_route_entry_fn)sai_bulk_create_route_entry,
```
kcudnik: why this is a cast ? is one of the params not matching ?

kcudnik:
cherrypick https://github.com/opencomputeproject/SAI/commit/121d107ee63b49c216b410cb01e065b29f0cb27e this commit from SAI master branch and it will fix stack protector issues

lib/inc/sai_redis.h
```diff
@@ -60,7 +60,7 @@ extern volatile bool g_useTempView;
 extern volatile bool g_asicInitViewMode;
 extern volatile bool g_logrotate;
 
-extern service_method_table_t                       g_services;
+extern sai_service_method_table_t                       g_services;
```
qiluo-msft: wrong indentation

lib/src/sai_redis_route.cpp
qiluo-msft: Yes, const for pointer of pointer

---
In reply to: [153569444](https://github.com/Azure/sonic-sairedis/pull/264#discussion_r153569444) [](ancestors = 153569444)

lib/inc/sai_redis.h
lguohan: fixed.

PR #273, opened 2018-01-05
Allow set readonly attributes in unittest mode for easier testing

kcudnik:


lguohan:
how can I enable this in my virtual switch based test?

https://github.com/Azure/sonic-swss/tree/master/tests

kcudnik:
swig/pyext needs to be created for saimeta that expose those methods

PR #275, opened 2018-01-10
Enable syncd-rpc target compilation

andrii-savka:
Enable syncd-rpc target compilation.
Export switch ID variable required by SAI thrift.


lguohan:
build still fails.

andrii-savka:
The rebased copy of #260

andrii-savka:
@lguohan 
The build is failing because of syncd can't include "switch_sai_rpc_server.h" from SAI. 
I don't have such error in my local build. 
Probably, we need to update SAI. 
I'm investigating how to fix it.

andrii-savka:
Close the duplication because it has the same problem.

Issue #270, opened 2017-12-19
vs error at the startup time


lguohan:
Dec 19 02:13:53.296659 bfa8f5cc60f2 ERR syncd: :- veth2tap_fun: failed to write to tap device 63
Dec 19 02:13:53.552463 bfa8f5cc60f2 ERR syncd: :- veth2tap_fun: failed to write to tap device 67
Dec 19 02:13:54.064588 bfa8f5cc60f2 ERR syncd: :- veth2tap_fun: failed to write to tap device 61
Dec 19 02:13:54.576561 bfa8f5cc60f2 ERR syncd: :- veth2tap_fun: failed to write to tap device 51
Dec 19 02:13:54.576630 bfa8f5cc60f2 ERR syncd: :- veth2tap_fun: failed to write to tap device 65
Dec 19 02:13:54.832563 bfa8f5cc60f2 ERR syncd: :- veth2tap_fun: failed to write to tap device 69
Dec 19 02:13:55.088500 bfa8f5cc60f2 ERR syncd: :- veth2tap_fun: failed to write to tap device 59
Dec 19 02:13:55.600660 bfa8f5cc60f2 ERR syncd: :- veth2tap_fun: failed to write to tap device 55
Dec 19 02:13:56.113016 bfa8f5cc60f2 ERR syncd: :- veth2tap_fun: failed to write to tap device 53
Dec 19 02:13:56.368590 bfa8f5cc60f2 ERR syncd: :- veth2tap_fun: failed to write to tap device 49
Dec 19 02:13:56.368636 bfa8f5cc60f2 ERR syncd: :- veth2tap_fun: failed to write to tap device 57
Dec 19 02:13:56.368710 bfa8f5cc60f2 ERR syncd: :- veth2tap_fun: failed to write to tap device 75
Dec 19 02:13:56.624558 bfa8f5cc60f2 ERR syncd: :- veth2tap_fun: failed to write to tap device 79
Dec 19 02:13:56.624768 bfa8f5cc60f2 ERR syncd: :- veth2tap_fun: failed to write to tap device 71
Dec 19 02:13:57.136521 bfa8f5cc60f2 ERR syncd: :- veth2tap_fun: failed to write to tap device 105
Dec 19 02:13:57.136628 bfa8f5cc60f2 ERR syncd: :- veth2tap_fun: failed to write to tap device 73
[...]

PR #278, opened 2018-01-11
Do not log error when write fails to tap device

kcudnik:
This is when one of the virtual ethernet interface
is down, then EIO is returned.

### DIFF
--- a/vslib/src/sai_vs_hostintf.cpp
+++ b/vslib/src/sai_vs_hostintf.cpp
@@ -34,6 +34,8 @@ typedef struct _hostif_info_t
 
     volatile bool run_thread;
 
+    std::string name;
+
 } hostif_info_t;
 
 std::map<std::string, std::shared_ptr<hostif_info_t>> hostif_info_map;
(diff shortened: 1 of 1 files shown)


vslib/src/sai_vs_hostintf.cpp
```diff
@@ -194,18 +196,32 @@ void veth2tap_fun(std::shared_ptr<hostif_info_t> info)
...
+            {
+                SWSS_LOG_ERROR("failed to write to tap device fd %d, errno(%d): %s",
+                        info->tapfd, errno, strerror(errno));
+            }
+
+            continue;
         }
     }
+
+    SWSS_LOG_NOTICE("engind thread proc for %s", info->name.c_str());
```
lguohan: ending? what is this engind?

vslib/src/sai_vs_hostintf.cpp
```diff
@@ -222,17 +238,22 @@ void tap2veth_fun(std::shared_ptr<hostif_info_t> info)
...
-            SWSS_LOG_ERROR("failed to write to socker %d", info->packet_socket);
-            break;
+            SWSS_LOG_ERROR("failed to write to socket fd %d, errno(%d): %s",
+                    info->packet_socket, errno, strerror(errno));
+
+            continue;
         }
     }
+
+    SWSS_LOG_NOTICE("engind thread proc for %s", info->name.c_str());
```
lguohan: what is engind?

kcudnik:
Fixes #270

PR #279, opened 2018-01-12
[syncd] Add support for a clear fdb cli

keboliu:
To implement a clear fdb cli,  added code in syncd.cpp and syncd_notifications.cpp.

for syncd_notifications.cpp
    -   revise funtion redisPutFdbEntryToAsicView to flush the fdb entries in ASIC_DB.

design doc see [https://github.com/Azure/SONiC/wiki/SONiC-Clear-FDB-CLI-Design](url)


syncd/syncd.cpp
```diff
@@ -2936,6 +2936,88 @@ bool handleRestartQuery(swss::NotificationConsumer &restartQuery)
     return false;
 }
 
+bool getSwitchRid(sai_object_id_t* switch_rid)
```
kcudnik: switch rid can be obtained in much easier way: global object "switches" contains all the switches where key is VID of the switch, take a look syncd.cpp at line 50, usually there will be 1 or 0 switches

this is not needed

syncd/syncd.cpp
```diff
@@ -3228,6 +3310,7 @@ int syncd_main(int argc, char **argv)
 
     std::shared_ptr<swss::ConsumerTable> asicState = std::make_shared<swss::ConsumerTable>(dbAsic.get(), ASIC_STATE_TABLE);
     std::shared_ptr<swss::NotificationConsumer> restartQuery = std::make_shared<swss::NotificationConsumer>(dbAsic.get(), "RESTARTQUERY");
+    std::shared_ptr<swss::NotificationConsumer> flushfdbRequest = std::make_shared<swss::NotificationConsumer>(dbAsic.get(), "FLUSHFDBREQUEST");
```
kcudnik: flush fdb request should be initiated via SAI api, and IMO it should be synchronous call similar to GET, and called as GET api, no need for another handler

syncd/syncd_notifications.cpp
```diff
@@ -92,19 +92,122 @@ void redisPutFdbEntryToAsicView(
...
+        g_redisClient->del(key);
+        return;
+    }
+
+    if (fdb->event_type == SAI_FDB_EVENT_FLUSHED)
+    {
+        sai_vlan_id_t vlan_id = fdb->fdb_entry.vlan_id;
+        sai_object_id_t port_oid = 0;
+        int port_oid_found = 0;
+        int attr_num = fdb->attr_count;
```
kcudnik: extra variable not needed, use direct attr_count in line 109

syncd/syncd_notifications.cpp
```diff
@@ -92,19 +92,122 @@ void redisPutFdbEntryToAsicView(
...
+        SWSS_LOG_DEBUG("remove fdb entry %s for SAI_FDB_EVENT_AGED",key.c_str());
+        g_redisClient->del(key);
+        return;
+    }
+
+    if (fdb->event_type == SAI_FDB_EVENT_FLUSHED)
+    {
+        sai_vlan_id_t vlan_id = fdb->fdb_entry.vlan_id;
+        sai_object_id_t port_oid = 0;
+        int port_oid_found = 0;
```
kcudnik: bool ?

syncd/syncd_notifications.cpp
```diff
@@ -92,19 +92,122 @@ void redisPutFdbEntryToAsicView(
...
+                \"vlan\":\"0\"}",
+            "fdb_event":"SAI_FDB_EVENT_FLUSHED",
+                "list":[
+                    {"id":"SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID","value":"oid:0x0"},
+                    {"id":"SAI_FDB_ENTRY_ATTR_TYPE","value":"SAI_FDB_ENTRY_TYPE_DYNAMIC"},
+                    {"id":"SAI_FDB_ENTRY_ATTR_PACKET_ACTION","value":"SAI_PACKET_ACTION_FORWARD"}
+                ]
+            }]
+            */
+            SWSS_LOG_NOTICE("received a flush all fdb event");
```
kcudnik: this behaviour is not described in SAI headers that when no port and no vlan id then 1 notification is send to notify that all have been flushed from the device, please put that description (and consult with SAI community?) whether that behavior is good 
for example, 
-all entries were flushed if "mac == all 0, vlanid == 0, bridge_id == 0"

this is fine by me, but please update SAI headers

syncd/syncd_notifications.cpp
```diff
@@ -92,19 +92,122 @@ void redisPutFdbEntryToAsicView(
...
+            
+        }
+        else if(port_oid && vlan_id)
+        {
+            SWSS_LOG_ERROR("received a flush port in a vlan fdb event, unsupported");
+        }
+        else
+        {
+            SWSS_LOG_ERROR("received a unknown flush fdb event, unsupported");
+        }
```
kcudnik: those 2 erros can be combined in 1 error message and print port_oid and vlan_id

syncd/syncd.cpp
keboliu: optimized as suggested.

syncd/syncd_notifications.cpp
keboliu: changed to bool.

syncd/syncd_notifications.cpp
keboliu: optimized as suggested.

syncd/syncd_notifications.cpp
keboliu: I will talk to SAI guys see whether they already have something similar if not may ask them to add.

syncd/syncd_notifications.cpp
keboliu: revised.

syncd/syncd.cpp
keboliu: When started the design my idea is that since "clear fdb" is a one-shot action, not like a normal configuration, should not change configurations in the DB, so implement it via publish/subscribe mechanism is a quite straightforward way.   Make it synchronous is a good suggestion, I am thinking next phase when adding "flush port" and "flush vlan" support,  also make cli utilities to wait for a response from syncd or somewhere.

lguohan: I agree with @kcudnik , the request should be initiated via SAI API. the sonic-clear command should invoke APP DB and let orchagent to issue corresponding SAI API.

lguohan: @keboliu, you probably did not understand "synchronous call similiar to GET api". You probably need to look at how GET api are implemented in syncd and sairedis.

keboliu: @lguohan yes, can you give some example for GET api if you have now?  And for your previous comments "regarding invoke APP DB and let orchagent to call SAI API", do you mean we need to add some new configuration in the APP DB and trigger orchagent with manipulating this new configuration?

kcudnik:
take a look at #288 i added support for sairedis to clear fdb entries, which request can come from orchagent directly

syncd/syncd_notifications.cpp
```diff
@@ -92,19 +92,130 @@ void redisPutFdbEntryToAsicView(
...
+        sai_object_id_t port_oid = 0;
+        bool port_oid_found = false;
+        
+        for (uint32_t i = 0; i < fdb->attr_count; i++)
         {
-            SWSS_LOG_DEBUG("remove fdb entry %s for SAI_FDB_EVENT_AGED",key.c_str());
+            if(fdb->attr[i].id == SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID)
+            {
+                port_oid = fdb->attr[i].value.oid;
+                port_oid_found = true;
```
prsunny: Just a note, you can break from here once port_oid_found is set to true

PR #284, opened 2018-01-17
[FlexCounter]: Use fine-grained locks.

sihuihan88:

### DIFF
--- a/syncd/syncd_flex_counter.cpp
+++ b/syncd/syncd_flex_counter.cpp
@@ -490,12 +490,16 @@ void FlexCounter::runPlugins(
     std::map<sai_object_id_t, std::shared_ptr<PortCounterIds>> portCounterIdsMap;
     std::map<sai_object_id_t, std::shared_ptr<QueueCounterIds>> queueCounterIdsMap;
     std::map<sai_object_id_t, std::shared_ptr<QueueAttrIds>> queueAttrIdsMap;
+    std::set<std::string> queuePlugins;
+    std::set<std::string> portPlugins;
 
     {
         std::lock_guard<std::mutex> lock(g_mutex);
         portCounterIdsMap = m_portCounterIdsMap;
         queueCounterIdsMap = m_queueCounterIdsMap;
         queueAttrIdsMap = m_queueAttrIdsMap;
+        queuePlugins = m_queuePlugins;
+        portPlugins = m_portPlugins;
     }
 
     const std::vector<std::string> argv = 
@@ -512,7 +516,7 @@ void FlexCounter::runPlugins(
         portList.push_back(sai_serialize_object_id(kv.first));
     }
 
-    for (const auto& sha : m_portPlugins)
+    for (const auto& sha : portPlugins)
     {
         runRedisScript(db, sha, portList, argv);
     }
@@ -524,7 +528,7 @@ void FlexCounter::runPlugins(
         queueList.push_back(sai_serialize_object_id(kv.first));
     }
 
-    for (const auto& sha : m_queuePlugins)
+    for (const auto& sha : queuePlugins)
     {
         runRedisScript(db, sha, queueList, argv);
     }
@@ -541,7 +545,6 @@ void FlexCounter::flexCounterThread(void)
     while (m_runFlexCounterThread)
     {
         auto start = std::chrono::steady_clock::now();
-        std::lock_guard<std::mutex> lock(g_mutex);
         collectCounters(countersTable);
         auto finish = std::chrono::steady_clock::now();
 
--- a/syncd/syncd_flex_counter.cpp
+++ b/syncd/syncd_flex_counter.cpp
@@ -354,8 +354,19 @@ void FlexCounter::collectCounters(
 {
     SWSS_LOG_ENTER();
 
+    std::map<sai_object_id_t, std::shared_ptr<PortCounterIds>> portCounterIdsMap;
+    std::map<sai_object_id_t, std::shared_ptr<QueueCounterIds>> queueCounterIdsMap;
+    std::map<sai_object_id_t, std::shared_ptr<QueueAttrIds>> queueAttrIdsMap;
+
+    {
+        std::lock_guard<std::mutex> lock(g_mutex);
+        portCounterIdsMap = m_portCounterIdsMap;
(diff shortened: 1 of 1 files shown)


sihuihan88:
depends on https://github.com/Azure/sonic-swss-common/pull/179

lguohan:
retest this please

lguohan:
Can you provide more rationale for your split. what I understand from your code, is you move counterTable.set functions out of collectCounters(). when the thread is not collecting pfc watch counters, you split the counterTable.set into three parts (portstat, queuestat, and queue attributes).

What I do not understand why you choose this approach versus others. For example, in collectCounters, you can only collect one port, and then lock. That is another way to break a single lock into multiple small locks.

syncd/syncd_flex_counter.cpp
```diff
@@ -522,9 +543,36 @@ void FlexCounter::flexCounterThread(void)
...
     {
         {
-            std::lock_guard<std::mutex> lock(g_mutex);
-            collectCounters(countersTable);
-            runPlugins(db);
+            std::map<std::string, std::vector<swss::FieldValueTuple>> portStats;
+            std::map<std::string, std::vector<swss::FieldValueTuple>> queueStats;
+            std::map<std::string, std::vector<swss::FieldValueTuple>> queueAttrs;
+            collectCounters(portStats, queueStats, queueAttrs);
+            if (m_pollInterval == PFC_WD_POLL_MSECS)
```
lguohan: we should probably be able to change the polling internal, use the polling interval to determine whether we should runPlugins or not is not very readable.

syncd/syncd_flex_counter.cpp
```diff
@@ -467,10 +465,33 @@ void FlexCounter::collectCounters(
...
-        // Write counters to DB
+        // Store queue attribute values for update
         std::string queueVidStr = sai_serialize_object_id(queueVid);
+        queueAttrsMap[queueVidStr] = values;
+    }
+}
+
+void FlexCounter::updateCounters(
+        _In_ swss::Table &countersTable,
+        _In_ std::map<std::string, std::vector<swss::FieldValueTuple>> &portStats,
```
lguohan: the variable name should really be portsStats since the key is the port, same for other variable names.

syncd/syncd_flex_counter.cpp
```diff
@@ -522,9 +543,36 @@ void FlexCounter::flexCounterThread(void)
     while (m_runFlexCounterThread)
     {
         {
-            std::lock_guard<std::mutex> lock(g_mutex);
-            collectCounters(countersTable);
-            runPlugins(db);
+            std::map<std::string, std::vector<swss::FieldValueTuple>> portStats;
+            std::map<std::string, std::vector<swss::FieldValueTuple>> queueStats;
+            std::map<std::string, std::vector<swss::FieldValueTuple>> queueAttrs;
+            collectCounters(portStats, queueStats, queueAttrs);
```
marian-pritsak: collectCounters has to be guarded by lock because it uses structures that can be updated from another thread

syncd/syncd_flex_counter.cpp
```diff
@@ -522,9 +543,36 @@ void FlexCounter::flexCounterThread(void)
...
+            {
+                std::lock_guard<std::mutex> lock(g_mutex);
+                updateCounters(countersTable, portStats, queueStats, queueAttrs);
+                runPlugins(db);
+            }
+            else
+            {
+                for (auto& portStat : portStats)
+                {
+                    std::lock_guard<std::mutex> lock(g_mutex);
```
marian-pritsak: Why is this guarded?
countersTable belongs only to this thread

sihuihan88: counterTable is the redis db shared by all threads and that's why we need the lock here.

syncd/syncd_flex_counter.cpp
sihuihan88: I am not sure there is some structures shared in collectCounters among threads. It's only querying sai attribute values here.

syncd/syncd_flex_counter.cpp
sihuihan88: countersTable are the redis table shared by different thread, and that's why we need the lock

syncd/syncd_flex_counter.cpp
sihuihan88: I am not sure there is some structure shared among threads within collectCounters; it's just querying sai attribute values.

sihuihan88:
Sorry for deleting @marian-pritsak's comment about "why use lock for counterTable.set in the flexcounter thread" by mistake.

From my point of view, different counterTable.set will write to the same redis db for same port/queue. Therefore I think we need to have lock for this.

syncd/syncd_flex_counter.cpp
sihuihan88: From my point of view, different counterTable.set will write to the same redis db for same port/queue. Therefore I think we need to have lock for this.

sihuihan88:
Sorry about deleting @marian-pritsak 's comment about "collectCounters requires lock"

I am not sure there is some structure shared among thread within collectCounters, as it only query sai attributes in this function

syncd/syncd_flex_counter.cpp
marian-pritsak: Redis handles that. It is single threaded and takes one request at a time.

sihuihan88:
@marian-pritsak Can you help to review?

sihuihan88:
I agree that when new oid is registered between 3 and 6, there is no counter available when plugin runs and we miss one round of result. That should be fine as in next poll interval we will have the information.
Even we have a giant lock to make collectCounter+runplugin atomic, we will still miss a round of result if syncd wants to add the new oid within the time period but wait for lock.

kcudnik:
Im not sure if this is the right solution, since between step 8 and 9 state of oid can change, and you can get errors while getting counters, if all those oids are ports then maybe we can get away with this since port oids dont change, but for orther objects states can change and oid numbers can change

marian-pritsak:
Nothing after step 5 is using SAI API, and if any OIDs changed, orchagent should know about it. So I would argue that we are safe here.

sihuihan88:
The oids are port oids and queue oids. I think queue oids also don't change. They are read-only, and only depends on port and queue index.

PR #286, opened 2018-01-20
[flex_counter]: Add delay correction to poll

marian-pritsak:
Measure the delay that it takes to read counters from hardware and
adjust sleep time by it.

### DIFF
--- a/syncd/syncd_flex_counter.cpp
+++ b/syncd/syncd_flex_counter.cpp
@@ -462,7 +462,8 @@ void FlexCounter::collectCounters(
 }
 
 void FlexCounter::runPlugins(
-        _In_ swss::DBConnector& db)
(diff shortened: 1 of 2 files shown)


sihuihan88:
It seems that you are counting the time of running thread and decrease this amount time from sleep. This might not work if the running time exceeds the polling interval occasionally. Can we as use kernel timer and   queue up the event if the running time exceeds?

marian-pritsak:
Sleep time is decreased by delay % poll_interval. And the actual time that it took to read counters + delay is passed to plugin to do the calculations.

For instance, if delay for reading counters was 120 ms, and poll interval is 100 ms, and previous delay was 40 ms, we pass (100 - 40 + 120) ms and sleep for (100 - 20) ms.

PR #287, opened 2018-01-22
Add support for FDB notification in virtual switch

kcudnik:


vslib/inc/sai_vs_state.h
```diff
@@ -103,6 +103,38 @@ typedef std::map<sai_object_type_t, std::map<std::string, AttrHash>> ObjectHash;
...
+        if (res < 0)
+            return true;
+
+        if (res > 0)
+            return false;
+
+        return fdb_entry.vlan_id < other.fdb_entry.vlan_id;
+    }
+
+    bool operator() (const _fdb_info_t& lhs, const _fdb_info_t & rhs) const
```
qiluo-msft: >operator [](start = 9, length = 8)

Why we need this operator?

vslib/src/sai_vs_hostintf.cpp
```diff
@@ -36,10 +37,279 @@ typedef struct _hostif_info_t
...
+        data.fdb_entry.bridge_id = SAI_NULL_OBJECT_ID;
+    }
+    else if (data.fdb_entry.bridge_type == SAI_FDB_ENTRY_BRIDGE_TYPE_1D)
+    {
+        // since this is only valid for 1Q
+        data.fdb_entry.vlan_id = 0;
+    }
+    else
+    {
+        SWSS_LOG_WARN("unknown bridge type %d", data.fdb_entry.bridge_type);
```
qiluo-msft: >SWSS_LOG_WARN [](start = 8, length = 13)

Should it be an error?

vslib/src/sai_vs_hostintf.cpp
```diff
@@ -36,10 +37,279 @@ typedef struct _hostif_info_t
...
+
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        return;
+    }
+
+    std::string s = sai_serialize_fdb_event_ntf(1, &data);
+
+    // TODO to debug
+    SWSS_LOG_NOTICE("calling user fdb event callback: %s", s.c_str());
```
qiluo-msft: >SWSS_LOG_NOTICE [](start = 4, length = 15)

SWSS_LOG_DEBUG?

vslib/inc/sai_vs_state.h
kcudnik: so it could be uses as function for comparison 2 fdb_info structs

vslib/src/sai_vs_hostintf.cpp
```diff
@@ -36,10 +37,278 @@ typedef struct _hostif_info_t
...
+        case ETH_P_8021Q:
+
+            {
+                // this is tagged frame, get vlan id from frame
+
+                uint16_t tci = htons(eh->h_proto);
+
+                vlan_id = tci & 0xfff;
+
+                if (vlan_id == 0 || vlan_id == 0xfff)
```
lguohan: >vlan_id == 0  [](start = 20, length = 13)

vlan id can be 0, when it is zero, it mean priority tag packet, should treat as the untagged packet, assigning the pvid to the packet as the vlan_id.

vslib/src/sai_vs_hostintf.cpp
```diff
@@ -36,10 +37,278 @@ typedef struct _hostif_info_t
...
+    uint16_t proto = htons(eh->h_proto);
+
+    uint16_t vlan_id = DEFAULT_VLAN_NUMBER;
+
+    switch (proto)
+    {
+        case ETH_P_IP:
+        case ETH_P_IPV6:
+        case ETH_P_ARP:
+        case ETH_P_RARP:
```
lguohan: we do not need to limit to these ether types, it can be 8021Q or else.

Inside 8021Q, we can later add QinQ packet process, but that is for the future.

vslib/src/sai_vs_hostintf.cpp
```diff
@@ -36,10 +37,278 @@ typedef struct _hostif_info_t
...
+
+            {
+                // IP frame, we need to get vlan if from port
+
+                sai_attribute_t attr;
+
+                attr.id = SAI_PORT_ATTR_PORT_VLAN_ID;
+
+                sai_status_t status = vs_generic_get(SAI_OBJECT_TYPE_PORT, info->portid, 1, &attr);
```
lguohan: >vs_generic_get(SAI_OBJECT_TYPE_PORT, [](start = 38, length = 36)

when the port belongs to a LAG, then, we need to get the 

SAI_LAG_ATTR_PORT_VLAN_ID

[https://github.com/opencomputeproject/SAI/blob/master/inc/sailag.h#L99](https://github.com/opencomputeproject/SAI/blob/master/inc/sailag.h#L99)

vslib/src/sai_vs_hostintf.cpp
```diff
@@ -36,10 +37,278 @@ typedef struct _hostif_info_t
...
+            }
+
+        default:
+            {
+                SWSS_LOG_WARN("unknown ethernet protocol: 0x%x on %s", proto, info->name.c_str());
+                return;
+            }
+    }
+
+    // we have vlan and mac addres which is KEY, so just see if that is already defined
```
lguohan: >addres [](start = 28, length = 6)

address

lguohan: >    // we have vlan and mac addres which is KEY, so just see if that is [](start = 0, length = 71)

if the port is port-based router interface, or lag-based router interface, we should skip mac address.

vslib/src/sai_vs_hostintf.cpp
```diff
@@ -36,10 +37,278 @@ typedef struct _hostif_info_t
...
+
+    sai_fdb_event_notification_fn ntf = (sai_fdb_event_notification_fn)attr.value.ptr;
+
+    if (ntf != NULL)
+    {
+        ntf(1, &data);
+    }
+}
+
+void findBridgeForPort(
```
lguohan: >findBridgeForPort [](start = 5, length = 17)

The bridge port lookup process is two steps:

1. use (vlan_id, phyiscal port_id) to match any .1D bridge port created. If there is match, then quit, found=true

2. use (physical port_id) to match any .1Q bridge created. if there is a match, the quite, found=true.

if found==true, generate fdb learn event on the .1D or .1Q bridge port. 

if not found, then do not generate fdb event. It means the packet is not received on the bridge port.

vslib/src/sai_vs_interfacequery.cpp
```diff
@@ -202,6 +206,73 @@ void unittestChannelThreadProc()
...
+
+    uint32_t current = (uint32_t)time(NULL);
+
+    // find aged fdb entries
+
+    for (auto it = g_fdb_info_set.begin(); it != g_fdb_info_set.end();)
+    {
+        sai_attribute_t attr;
+
+        attr.id = SAI_SWITCH_ATTR_FDB_AGING_TIME;
```
lguohan: >SAI_SWITCH_ATTR_FDB_AGING_TIME [](start = 18, length = 30)

since this is switch attribute, why not move it out of the for loop and skip the for loop if aging is not enable?

vslib/src/sai_vs_hostintf.cpp
kcudnik: what do you mean skip mac address?

vslib/src/sai_vs_interfacequery.cpp
kcudnik: this is obtained from switch id, in case of multiple switches where it can be different, or i can get this from first item on the list

vslib/src/sai_vs_hostintf.cpp
lguohan: skip mac address learning, sorry.

vslib/src/sai_vs_interfacequery.cpp
lguohan: ok

vslib/src/sai_vs_hostintf.cpp
kcudnik: pvid  ? port vlan id ?

vslib/src/sai_vs_hostintf.cpp
kcudnik: what do you mean skip mac ? just skip entire fdb notification processing?

vslib/src/sai_vs_hostintf.cpp
kcudnik: is there dfault 1D bridge created on the switch ? because from what i noticed and what i implemented in current vs is that there is default 1Q bridge, so for example for TOR usage we will need to create 1D brige and move ports from 1Q to 1D bridge? and create bridge ports for that ?

vslib/src/sai_vs_hostintf.cpp
kcudnik: this condition is getting complicated, do you mean https://github.com/opencomputeproject/SAI/blob/master/inc/sairouterinterface.h#L101
if this attribute is set to port or lag, that is based on port then we skip learning ?

vslib/src/sai_vs_hostintf.cpp
lguohan: yes, port vlan id.

vslib/src/sai_vs_hostintf.cpp
lguohan: no, there is no default 1d bridge created on the switch.

vslib/src/sai_vs_hostintf.cpp
kcudnik: fixed

vslib/src/sai_vs_hostintf.cpp
kcudnik: fixed

vslib/src/sai_vs_hostintf.cpp
kcudnik: fixdd

vslib/src/sai_vs_hostintf.cpp
kcudnik: removed, accepting all protocols

vslib/src/sai_vs_hostintf.cpp
kcudnik: fixed

vslib/src/sai_vs_hostintf.cpp
kcudnik: fixed

vslib/src/sai_vs_interfacequery.cpp
kcudnik: im leaving this as is

vslib/src/sai_vs_hostintf.cpp
kcudnik: fixed

vslib/src/sai_vs_hostintf.cpp
kcudnik: as we discussed on the im, im leaving this and just looking for any bridge since we will remove bridge when using router interface, and i put info in to the comment

vslib/inc/sai_vs_state.h
qiluo-msft: We could use operator< directly, right? This function is weird, since you compare 2 other object and ignore this object. This is a non static function.

kcudnik: this operator is used when inserting data to std::set, for sorting purpose, and operator < is added for simplicity here

PR #288, opened 2018-01-23
Add support for fdb flush in sairedis

kcudnik:


kcudnik:
test this please

meta/sai_meta.cpp
```diff
@@ -4737,3 +4737,154 @@ void meta_sai_on_fdb_event(
...
+    SWSS_LOG_ENTER();
+
+    if (flush_fdb_entries == NULL)
+    {
+        SWSS_LOG_ERROR("function pointer is NULL");
+
+        return SAI_STATUS_INVALID_PARAMETER;
+    }
+
+    if (attr_count > MAX_LIST_COUNT)
```
keboliu: should at least have one attribute in the list?

keboliu:
I tested this with my PRs, it works well.

meta/sai_meta.cpp
kcudnik: no, since list can be empty

kcudnik:
test this please

PR #289, opened 2018-01-25
[Flex_counter]: Remove poll interval from db key

sihuihan88:

### DIFF
--- a/syncd/syncd.cpp
+++ b/syncd/syncd.cpp
@@ -2535,8 +2535,62 @@ sai_status_t processEvent(
     return status;
 }
 
+void processFlexCounterGroupEvent(
+        _In_ swss::ConsumerTable &consumer)
+{
+    SWSS_LOG_ENTER();
+
+    std::lock_guard<std::mutex> lock(g_mutex);
+    swss::KeyOpFieldsValuesTuple kco;
+    consumer.pop(kco);
+
+    const auto &groupName = kfvKey(kco);
+    const auto &op = kfvOp(kco);
+    const auto values = kfvFieldsValues(kco);
+
+    if (op == DEL_COMMAND)
+    {
+        FlexCounter::removeCounterPlugin(groupName);
+        return;
+    }
+
+    for (const auto& valuePair : values)
+    {
+        const auto field = fvField(valuePair);
+        const auto value = fvValue(valuePair);
+
+        if (op == SET_COMMAND)
+        {
+            if (field == POLL_INTERVAL_FIELD)
+            {
+                FlexCounter::setPollInterval(stoi(value), groupName);
+            }
+            else if (field == QUEUE_PLUGIN_FIELD)
+            {
+                auto shaStrings = swss::tokenize(value, ',');
+                for (const auto &sha : shaStrings)
+                {
+                    FlexCounter::addQueueCounterPlugin(sha, groupName);
+                }
+            }
+            else if (field == PORT_PLUGIN_FIELD)
+            {
+                auto shaStrings = swss::tokenize(value, ',');
+                for (const auto &sha : shaStrings)
+                {
+                    FlexCounter::addPortCounterPlugin(sha, groupName);
+                }
+            }
+            else
+            {
+                SWSS_LOG_ERROR("Field is not supported %s", field.c_str());
+            }
+        }
+    }
+}
+
 void processFlexCounterEvent(
-        _In_ swss::ConsumerStateTable &consumer)
+        _In_ swss::ConsumerTable &consumer)
 {
     SWSS_LOG_ENTER();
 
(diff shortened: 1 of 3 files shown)


sihuihan88:
depends on https://github.com/Azure/sonic-swss-common/pull/182

sihuihan88:
retest this please

syncd/syncd.cpp
```diff
@@ -3292,14 +3260,14 @@ int syncd_main(int argc, char **argv)
...
-    std::shared_ptr<swss::DBConnector> dbFlexCounter = std::make_shared<swss::DBConnector>(PFC_WD_DB, swss::DBConnector::DEFAULT_UNIXSOCKET, 0);
+    std::shared_ptr<swss::DBConnector> dbFlexCounter = std::make_shared<swss::DBConnector>(FLEX_COUNTER_DB, swss::DBConnector::DEFAULT_UNIXSOCKET, 0);
 
     g_redisClient = std::make_shared<swss::RedisClient>(dbAsic.get());
 
     std::shared_ptr<swss::ConsumerTable> asicState = std::make_shared<swss::ConsumerTable>(dbAsic.get(), ASIC_STATE_TABLE);
     std::shared_ptr<swss::NotificationConsumer> restartQuery = std::make_shared<swss::NotificationConsumer>(dbAsic.get(), "RESTARTQUERY");
-    std::shared_ptr<swss::ConsumerStateTable> flexCounterState = std::make_shared<swss::ConsumerStateTable>(dbFlexCounter.get(), PFC_WD_STATE_TABLE);
-    std::shared_ptr<swss::ConsumerStateTable> flexCounterPlugin = std::make_shared<swss::ConsumerStateTable>(dbFlexCounter.get(), PLUGIN_TABLE);
+    std::shared_ptr<swss::ConsumerTable> pfcWdFlexCounter = std::make_shared<swss::ConsumerTable>(dbFlexCounter.get(), FLEX_PFC_WD_TABLE);
```
marian-pritsak: Flex counter was designed in a way that it's agnostic to its clients. In the case when we'll need to implement, for example, another watchdog, flex counter's code will need to be modified. I don't think that removing polling interval from key with a cost of introducing this dependency is a good thing to do.

sihuihan88:
retest this please

sihuihan88:
retest this please

sihuihan88:
@marian-pritsak Could you help to review? Thanks :)

lguohan:
>                 SWSS_LOG_ERROR("Object type not supported");

minor comments: better to print out the objectType and the field to give more debug information.

---
Refers to: syncd/syncd.cpp:2682 in 103a7f8. [](commit_id = 103a7f853b8e4fb966e6b4148bf73087c495a57b, deletion_comment = False)

syncd/syncd_flex_counter.cpp
```diff
@@ -30,6 +30,17 @@ FlexCounter::QueueAttrIds::QueueAttrIds(
...
 
+void FlexCounter::setPollInterval(
+        _In_ uint32_t pollInterval,
+        _In_ std::string instanceId)
+{
+    SWSS_LOG_ENTER();
+
+    FlexCounter &fc = getInstance(instanceId);
+    fc.m_pollInterval = pollInterval;
+    // fc.startFlexCounterThread();
```
lguohan: >   // fc.startFlexCounterThread(); [](start = 1, length = 34)

can we remove this?

syncd/syncd_flex_counter.h
```diff
@@ -82,9 +87,9 @@ class FlexCounter
             std::vector<sai_port_stat_t> portCounterIds;
         };
 
-        FlexCounter(uint32_t pollInterval);
-        static FlexCounter& getInstance(uint32_t pollInterval);
-        static void removeInstance(uint32_t pollInterval);
+        FlexCounter(std::string instanceId);
```
lguohan: >instanceId [](start = 32, length = 10)

can you explain this instanceId concept? I am not familiar with it.

lguohan: It looks like the instanceId is the flex counter group name, so maybe change to groupname is more clear?

---
In reply to: [166061744](https://github.com/Azure/sonic-sairedis/pull/289#discussion_r166061744) [](ancestors = 166061744)

sihuihan88: Each group will be an instance in flex counter and instance id is the group name.

syncd/syncd_flex_counter.cpp
```diff
@@ -208,53 +219,58 @@ void FlexCounter::removePort(
...
+    FlexCounter &fc = getInstance(instanceId);
 
     auto counterIter = fc.m_queueCounterIdsMap.find(queueVid);
-    if (counterIter != fc.m_queueCounterIdsMap.end())
+    if (counterIter == fc.m_queueCounterIdsMap.end())
     {
-        fc.m_queueCounterIdsMap.erase(counterIter);
-        found = true;
+        SWSS_LOG_ERROR("Trying to remove nonexisting queue counter Ids 0x%lx", queueVid);
+        if (fc.m_queueCounterIdsMap.empty() && fc.m_portCounterIdsMap.empty() && fc.m_queueAttrIdsMap.empty())
```
lguohan: >fc.m_queueCounterIdsMap.empty() && fc.m_portCounterIdsMap.empty() && fc.m_queueAttrIdsMap.empty( [](start = 12, length = 96)

lots of duplicated code to check empty counter id, suggest to have a function to check it.

lguohan: for example 

IsCounterIdsMapEmpty()

---
In reply to: [166065518](https://github.com/Azure/sonic-sairedis/pull/289#discussion_r166065518) [](ancestors = 166065518)

syncd/syncd_flex_counter.cpp
```diff
@@ -208,53 +219,58 @@ void FlexCounter::removePort(
     // Remove flex counter if counter IDs map is empty
     if (fc.m_queueCounterIdsMap.empty() && fc.m_portCounterIdsMap.empty() && fc.m_queueAttrIdsMap.empty())
     {
-        removeInstance(pollInterval);
+        removeInstance(instanceId);
     }
 }
 
 void FlexCounter::removeQueue(
```
lguohan: for removeQueue, I think you changed the logic.

In marian's code, he tries to remove the queue id both from m_queueCounterIdsMap and m_queueAttrIdsMap. 

In your code, if the first one is not found, you will return and will not remove the queueid from the second map.

Why do you change the logic here? Is this intended? IMO, marian code is more correct.

syncd/syncd.cpp
```diff
@@ -2535,41 +2535,83 @@ sai_status_t processEvent(
...
+            {
+                auto shaStrings = swss::tokenize(value, ',');
+                for (const auto &sha : shaStrings)
+                {
+                    FlexCounter::addPortCounterPlugin(sha, key);
+                }
+            }
+            else
+            {
+                SWSS_LOG_ERROR("Object type not supported");
```
lguohan: >SWSS_LOG_ERROR("Object type not supported"); [](start = 16, length = 44)

should be field not supported, also please print out the actual field for debugging purpose.

syncd/syncd.cpp
```diff
@@ -2535,41 +2535,83 @@ sai_status_t processEvent(
...
+        _In_ swss::ConsumerTable &consumer)
 {
     SWSS_LOG_ENTER();
 
     std::lock_guard<std::mutex> lock(g_mutex);
-
     swss::KeyOpFieldsValuesTuple kco;
     consumer.pop(kco);
 
     const auto &key = kfvKey(kco);
```
lguohan: >key [](start = 16, length = 3)

-> groupname

syncd/syncd_flex_counter.cpp
sihuihan88: Marian's new logic was added recently and I didn't rebase it before force push. Will update and sorry for the inconvenience

PR #290, opened 2018-01-30
Remove lane map divide restriction

kcudnik:


syncd/syncd_saiswitch.cpp
```diff
@@ -329,14 +329,12 @@ void SaiSwitch::helperCheckLaneMap()
     }
 
     /*
-     * We assume there are at least 32 ports and port number
-     * can be divided by 16 without rest.
+     * We assume there are at least 32 ports.
      */
 
-    if (laneMap.size() < 32 || (laneMap.size() % 16 != 0))
+    if (laneMap.size() < 32)
```
lguohan: it is interesting to have this check? any reason? if the switch has less than 32 ports, then we print error?

kcudnik: what do you mean?

lguohan: I mean why do check the lanemap size at all? there could be switch with less than 32 ports.

kcudnik: i added this at first to rule out some errors and mistakes with configuration file since we didnt used any other configuration, but can there be a switch with sonic with 5 ports? :P

stcheng:
this is still not merged?

pavel-shirshov:
I think all of the checks should be removed.
But current brcm SAI gives us wrong number of lanes, which is an error.

kcudnik:
You can merge this any time, and those checks were put there initially to prevent bugs on SAI from vendos, and at the beginning we only dealed with 1 type of switch so there was no issue, also i think right now we will not have less than 32 ports in switch and i would leave this as a safety check and remove when switch will less than 32 ports come to play

lguohan:
ok. can you check the ERROR to NOTICE?

kcudnik:
I removed this restriction at all

PR #297, opened 2018-02-18
[flex_counter]: Check if plugin lists are empty

marian-pritsak:
Check plugin lists along with counter IDs lists
before trying to remove FC instance

### DIFF
--- a/syncd/syncd_flex_counter.cpp
+++ b/syncd/syncd_flex_counter.cpp
@@ -75,7 +75,7 @@ void FlexCounter::setPortCounterList(
     {
         SWSS_LOG_ERROR("Port %s does not has supported counters", sai_serialize_object_id(portId).c_str());
 
-        // Remove flex counter if counter IDs map is empty
+        // Remove flex counter if all counter IDs and plugins are unregistered
         if (fc.isEmpty())
         {
             removeInstance(instanceId);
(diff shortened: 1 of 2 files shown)


syncd/syncd_flex_counter.cpp
```diff
@@ -341,7 +341,11 @@ bool FlexCounter::isCounterMapsEmpty()
 {
     SWSS_LOG_ENTER();
 
-    return m_queueCounterIdsMap.empty() && m_portCounterIdsMap.empty() && m_queueAttrIdsMap.empty();
+    return m_queueCounterIdsMap.empty() &&
+           m_portCounterIdsMap.empty() &&
+           m_queueAttrIdsMap.empty() &&
+           m_queuePlugins.empty() &&
+           m_portPlugins.empty();
```
lguohan: the function name does not quite match now. also if you look line 326, line 309 in this file, it checks m_queuePlugins/m_portPlugins, are those duplicated?

lguohan:
thanks marian, can you also fix the comments above the isEmpty() function?

marian-pritsak:
Yes, missed them. Done.

syncd/syncd_flex_counter.cpp
marian-pritsak: Done

lguohan:
@sihuihan88 , can you approved?

PR #299, opened 2018-02-22
[flex_counter]: Include plugins delay into poll correction

marian-pritsak:

### DIFF
--- a/syncd/syncd_flex_counter.cpp
+++ b/syncd/syncd_flex_counter.cpp
@@ -515,8 +515,7 @@ void FlexCounter::collectCounters(
 }
 
 void FlexCounter::runPlugins(
-        _In_ swss::DBConnector& db,
-        _In_ uint32_t pollInterval)
+        _In_ swss::DBConnector& db)
 {
     SWSS_LOG_ENTER();
 
@@ -539,7 +538,7 @@ void FlexCounter::runPlugins(
     {
         std::to_string(COUNTERS_DB),
         COUNTERS_TABLE,
-        std::to_string(pollInterval * 1000)
+        std::to_string(m_pollInterval * 1000)
     };
 
     std::vector<std::string> portList;
@@ -579,7 +578,7 @@ void FlexCounter::flexCounterThread(void)
         auto start = std::chrono::steady_clock::now();
 
         collectCounters(countersTable);
-        runPlugins(db, m_pollInterval);
+        runPlugins(db);
 
         auto finish = std::chrono::steady_clock::now();
         uint32_t delay = static_cast<uint32_t>(
--- a/syncd/syncd_flex_counter.cpp
+++ b/syncd/syncd_flex_counter.cpp
@@ -579,15 +579,19 @@ void FlexCounter::flexCounterThread(void)
     {
         auto start = std::chrono::steady_clock::now();
         collectCounters(countersTable);
-        auto finish = std::chrono::steady_clock::now();
+        auto countersFinish = std::chrono::steady_clock::now();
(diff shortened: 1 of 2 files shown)


marian-pritsak:
@sihuihan88 Can you please take a look? This change takes into account delay from running plugins.

syncd/syncd_flex_counter.cpp
```diff
@@ -579,15 +579,19 @@ void FlexCounter::flexCounterThread(void)
...
-        uint32_t delay = static_cast<uint32_t>(std::chrono::duration_cast<std::chrono::milliseconds>(finish - start).count());
-        uint32_t newCorrection = delay % m_pollInterval;
+        uint32_t countersDelay = static_cast<uint32_t>(
+                std::chrono::duration_cast<std::chrono::milliseconds>(countersFinish - start).count());
 
         // Run plugins with corrected interval
         // First we subtract correction from previous sleep and then add delay from current counters read
-        runPlugins(db, m_pollInterval - correction + delay);
-        correction = newCorrection;
+        runPlugins(db, m_pollInterval - correction + countersDelay);
```
sihuihan88: Currently we reduce timeleft by (m_pollInterval - correction + countersDelay) every time runplugin runs. However we also need to reduce timeleft by the time runplugin takes. I don' think it's easy to reduce the exact same value of runplugin takes for timeleft. It might be helpful to consider to count the number of runplugin runs instead rather than the real time in lua script

sihuihan88: Let me give an example. Let's first assume collect counter always takes 20ms, and runplugin always 30ms, poll 100ms. At the code start to run at 0ms
According to current code:
0: start to run
20ms: collectCOunter finishes, counterDelay 20ms, 
          start to runplugin, in lua script, timeleft -= m_pollInterval - correction + countersDelay= 100-0+20 = 80ms
50ms: runplgin finishes, correction = 50ms
 sleep for 100-50=50ms
100ms: thread start again and start to collectCounter
120ms: finished collectCounter, counterDelay 20ms,
            start to runplugin, in lua script, timeleft -= m_pollInterval - correction + countersDelay= 100-50+20 = 70ms

From previous example: you can see, runplugin function runs every 100ms. But in the script when calculating the timeleft it's wrong as the time of running plugin is not considered.

marian-pritsak: So counting number of run_plugins means the same as passing exact polling period to runPlugins, right?

sihuihan88: Yes, as long as we could pass the exact value.

marian-pritsak: Changed to polling period instead of calculated value

syncd/syncd_flex_counter.cpp
```diff
@@ -573,21 +573,18 @@ void FlexCounter::flexCounterThread(void)
...
     while (m_runFlexCounterThread)
     {
         auto start = std::chrono::steady_clock::now();
-        collectCounters(countersTable);
-        auto finish = std::chrono::steady_clock::now();
 
-        uint32_t delay = static_cast<uint32_t>(std::chrono::duration_cast<std::chrono::milliseconds>(finish - start).count());
-        uint32_t newCorrection = delay % m_pollInterval;
+        collectCounters(countersTable);
+        runPlugins(db, m_pollInterval);
```
sihuihan88: remove m_pollInterval parameter?

marian-pritsak: Yes, done

PR #303, opened 2018-03-13
[flexCounter]: Enable redis pipeline in flex counter table

sihuihan88:


syncd/syncd_flex_counter.cpp
```diff
@@ -517,6 +517,7 @@ void FlexCounter::collectCounters(
         std::string queueVidStr = sai_serialize_object_id(queueVid);
 
         countersTable.set(queueVidStr, values, "");
+        countersTable.flush();
```
qiluo-msft: >flush [](start = 22, length = 5)

If you flush after every set(), you gain nothing from pipeline.

PR #304, opened 2018-03-14
Support for platforms based on Barefoot Networks' device

rsunkad:
**- Incorporate support for platforms based on Barefoot Networks' device **

**- Added port config files and related scripts for the platforms **

**- Verified using SONiC tests **


saidiscovery/Makefile.am
```diff
@@ -11,8 +11,12 @@ endif
 if SAIVS
 SAILIB=-L$(top_srcdir)/vslib/src/.libs -lsaivs
 else
+if sonic_asic_platform_barefoot
+SAILIB=-L/opt/bfn/install/lib -lswitchsai -lswitchapi -lbf_switchd_lib -ltofinopdfixed_thrift -ldriver -lbfutils -lbfsys -lbfutils -L/opt/bfn/install/lib/tofinopd/switch -lpd -lpdcli -lpdthrift
```
lguohan: the libsai should be the only library that application link to. I will approve this for now, but you folks need to fix this.

PR #305, opened 2018-03-15
[syncd]: run switch->remove_switch() before exiting from syncd

pavel-shirshov:


syncd/syncd.cpp
```diff
@@ -3395,6 +3395,15 @@ int syncd_main(int argc, char **argv)
         }
     }
 
+    SWSS_LOG_NOTICE("Removing the switch gSwitchId=0x%lx", gSwitchId);
+    sai_switch_api_t *sai_switch_api = NULL;
+    sai_api_query(SAI_API_SWITCH, (void**)&sai_switch_api);
```
kcudnik: you dont need to query switch api, there is already populated sai_metadata_sai_switch_api global variable to access switch api's

PR #308, opened 2018-03-20
[src/sonic-sairedis] Change sai.profile directory

vharish02:
This commit changes the directory of sai.profile to point to the new
directory location for sai.profile /etc/sai.d/sai.profile file.
This is the location of sai.profile dynamic generation or for
copying the file from the hwsku directory.

Unit tested the code on S6100 for dynamic generation of sai.profile file for
both T0 and T1 and the file was created in /etc/sai.d/sai.profile
Similarly tested the sonic binary on S6000 to make sure that the sai.profile is
copied from the /usr/share/sonic/hwsku/sai.profile to /etc/sai.d/sai.profile.

Signed-off-by: Harish Venkatraman <Harish_Venkatraman@dell.com>

### DIFF
--- a/syncd/scripts/syncd_init_common.sh
+++ b/syncd/scripts/syncd_init_common.sh
@@ -32,7 +32,11 @@ esac
 
 config_syncd_bcm()
 {
-    CMD_ARGS+=" -p $HWSKU_DIR/sai.profile"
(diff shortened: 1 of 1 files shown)


vharish02:
@yxieca Can you please review the changes

PR #313, opened 2018-04-11
Use the refactored version of Selectables

pavel-shirshov:
**Please don't push it into master**

The changes in the pull request is required after the changes in the following PR: https://github.com/Azure/sonic-swss-common/pull/192

The changes below is removing tmp_fd, which is not used in the new Selectables


pavel-shirshov:
retest this please

PR #316, opened 2018-04-28
Change sai log level before discovery to prevent SAI ERR spam to the log

pavel-shirshov:
Change the level of sai api log to critical right before discovery and set it back to notice after the discovery to prevent error messages caused by discovery process


lguohan:
@kcudnik to check.

qiluo-msft:
@pavel-shirshov @lguohan This pull request will reset all the SAI components' loglevel back to NOTICE. It breaks loglevel for warm-reboot

Suggest store the previous loglevel, and restore back.

syncd/syncd_saiswitch.cpp
```diff
@@ -937,8 +937,9 @@ void SaiSwitch::helperDiscover()
 
     {
         SWSS_LOG_TIMER("discover");
-
+        set_sai_api_log_min_prio("SAI_LOG_LEVEL_CRITICAL");
```
kcudnik: why this is critical ?

pavel-shirshov: because otherwise sai will dump to syslog many thousands of the error messages "something not found" when syncd discover sai api". Output to syslog seriously hits cpu which increases fast-reboot time

kcudnik: there is nothing that produces critical message, so nothing would be logged except maybe something from internal sai
i would propose to log SAI_STATUS_NOT_IMPLEMENTED (which is the most errors probably) as info message in discovery method, instead of disabling all messages in this way

pavel-shirshov: Hi Kamil,

Yes. I made that level CRITICAL because no messages are produced by SAI on syncd discovery.
I had a patch making SAI_STATUS_NOT_IMPLEMENTED as INFO in SAI. I'm not sure it's still there.

kcudnik: yea that would be better approach, so other meaningfull errors if present they still would be logged

PR #321, opened 2018-05-24
[Syncd] make sure syncd can exit gracefully

yxieca:
- Stop notification thread before removing switch
- Stop flex counter queries before removing switch

### DIFF
--- a/syncd/syncd_notifications.cpp
+++ b/syncd/syncd_notifications.cpp
@@ -535,7 +535,6 @@ void ntf_process_function()
     while (runThread)
     {
         cv.wait(ulock);
-        if (! runThread) break;
 
         // this is notifications processing thread context, which is different
         // from SAI notifications context, we can safe use g_mutex here,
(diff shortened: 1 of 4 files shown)


syncd/syncd_notifications.cpp
```diff
@@ -535,6 +535,7 @@ void ntf_process_function()
     while (runThread)
     {
         cv.wait(ulock);
+        if (! runThread) break;
```
kcudnik: this is not needed, it will be checked after tryDequeue anyway

syncd/syncd_notifications.cpp
```diff
@@ -559,8 +560,6 @@ void startNotificationsProcessingThread()
     runThread = true;
 
     ntf_process_thread = std::make_shared<std::thread>(ntf_process_function);
-
-    ntf_process_thread->detach();
```
kcudnik: yea this was causing issue

yxieca:
This change has been tested with https://github.com/Azure/sonic-mgmt/pull/631 for more than 20 iterations. All passed.

PR #334, opened 2018-07-27
Suppress FDB events if notifier queue exceeds limit 

prsunny:
In the event of continuous mac-move, it is observed that the notifier-queue keeps growing consuming high memory. 
Current solution is to drop FDB events beyond a limit (set to `300k`, considering 256k MAC entries in a typical L2 deployment + some extra buffer for other events). A permanent approach would be to have the event stateful and notify only the latest event. 

Tested and observed that with this cap, the syncd memory usage is `~3.8%` and does not exceed beyond this.

### DIFF
--- a/syncd/syncd_notifications.cpp
+++ b/syncd/syncd_notifications.cpp
@@ -504,6 +504,8 @@ bool ntf_queue_t::tryDequeue(
 bool ntf_queue_t::enqueue(
         _In_ swss::KeyOpFieldsValuesTuple item)
 {
+    SWSS_LOG_ENTER();
+
     std::string notification = kfvKey(item);
 
     /*
--- a/syncd/syncd_notifications.cpp
+++ b/syncd/syncd_notifications.cpp
@@ -461,8 +461,68 @@ void processNotification(
 // that some notiffication arrived
 
 std::condition_variable cv;
-std::mutex queue_mutex;
-std::queue<swss::KeyOpFieldsValuesTuple> ntf_queue;
+
+class ntf_queue_t
+{
+public:
+    bool enqueue(swss::KeyOpFieldsValuesTuple msg);
+    bool tryDequeue (swss::KeyOpFieldsValuesTuple& msg);
+    size_t  queueStats()
+    {
+        return ntf_queue.size();
+    }
+
+private:
+    std::mutex queue_mutex;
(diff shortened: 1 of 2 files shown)


prsunny:
retest this please

syncd/syncd_notifications.cpp
```diff
@@ -461,8 +461,70 @@ void processNotification(
...
+    {
+        std::lock_guard<std::mutex> lock(queue_mutex);
+
+        return ntf_queue.size();
+    }
+
+private:
+    std::mutex queue_mutex;
+    std::queue<swss::KeyOpFieldsValuesTuple> ntf_queue;
+    const size_t limit = 300000;
```
lguohan: can you add some comments on why this number?

syncd/syncd_notifications.cpp
```diff
@@ -461,8 +461,70 @@ void processNotification(
...
+     */
+    if (queueStats() < limit || notification != "fdb_event")
+    {
+        std::lock_guard<std::mutex> lock(queue_mutex);
+
+        ntf_queue.push(item);
+        return true;
+    }
+
+    SWSS_LOG_INFO("Too many messages in queue (%ld), dropping FDB events!", queueStats());
```
lguohan: this should be info, can we add a counter (static) and print out message once every 100 packets.

syncd/syncd_notifications.cpp
prsunny: I specified it in PR description. Please check. I'll add as comment as well.

syncd/syncd_notifications.cpp
lguohan: sorry, I mean the level should be notice

syncd/syncd_notifications.cpp
```diff
@@ -461,8 +461,85 @@ void processNotification(
...
+
+        ntf_queue.push(item);
+        return true;
+    }
+
+    static uint32_t log_count = 0;
+
+    if (!log_count)
+    {
+        SWSS_LOG_NOTICE("Too many messages in queue (%ld), dropping FDB events!", queueStats());
```
lguohan: should also log how many messages we dropped.

prsunny:
Updated Message:
`NOTICE syncd/syncd: :- enqueue: Too many messages in queue (300000), dropped 253001 FDB events!`

PR #335, opened 2018-08-09
sairedis: add while loop to make syncd processEvent() handle as many entries as possible in one event call

dzhangalibaba:
…ction

* add while() loop to handle entries when buffer is Not empty

### DIFF
--- a/syncd/syncd.cpp
+++ b/syncd/syncd.cpp
@@ -2478,226 +2478,234 @@ sai_status_t processEvent(
 
     swss::KeyOpFieldsValuesTuple kco;
 
-    if (isInitViewMode())
-    {
-        /*
-         * In init mode we put all data to TEMP view and we snoop.  We need to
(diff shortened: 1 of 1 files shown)


dzhangalibaba:
could we make some progress on this, it is pending for a long time. Thanks

kcudnik:
@dzhangalibaba its waiting for my review, and i was on vacation, im back now :)

dzhangalibaba:
@kcudnik This is part of routing performance optimization. We want to handle as many events as possible already in the consumer buffer in one process function instead of waiting the while loop for each event. This was discussed about one month ago in the 4 way meeting. In this way, the performance improved ~200-300 routes per second.

qiluo-msft:
Retest this please

dzhangalibaba:
@qiluo-msft   is this PR OK to be merged now? Who can Help? Thanks

PR #336, opened 2018-08-09
sairedis: add RedisPipeline instance to construct producer g_asicState

dzhangalibaba:
* create RedisPipeline obj by default constructor , which enabled redispipeline with parameter 128
* create producer g_asicState using redisPipeline instance with redispipeline enabled
* with the changes in swss and swss-common, route performance improved by 200~300 routes/sec

* this changes depends on swss-common#219


dzhangalibaba:
could we make some progress on this, it is pending for a long time. Thanks

dzhangalibaba:
@kcudnik could you help take a look ? The whole function is related to https://github.com/Azure/sonic-swss-common/pull/219/ as well. When programming routes , it is using op "create".

PR #339, opened 2018-08-25
[saimeta]: Fix typo: > then -> >; don't exists -> doesn't exist

stcheng:
Signed-off-by: Shu0T1an ChenG <shuche@microsoft.com>

### DIFF
--- a/meta/sai_meta.cpp
+++ b/meta/sai_meta.cpp
@@ -457,7 +457,7 @@ void set_object(
 
     if (!object_exists(key))
     {
-        SWSS_LOG_THROW("FATAL: object %s don't exists", key.c_str());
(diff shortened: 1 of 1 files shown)


meta/sai_meta.cpp
```diff
@@ -533,7 +533,7 @@ sai_status_t meta_generic_validation_objlist(
 
     if (count > MAX_LIST_COUNT)
     {
-        META_LOG_ERROR(md, "object list count %u is > then max list count %u", count, MAX_LIST_COUNT);
+        META_LOG_ERROR(md, "object list count %u is > than max list count %u", count, MAX_LIST_COUNT);
```
qiluo-msft: Replace 
```
is (too) > than
```
with
```
>
```
is better.

stcheng:
thanks! updated and fixed more typos 💃

PR #347, opened 2018-09-25
[syncd] initiate warm shutdown when requested

yxieca:
When shutdown is requested, notify SAI when removing switch so that SAI would save states.

### DIFF
--- a/syncd/syncd.cpp
+++ b/syncd/syncd.cpp
@@ -3528,12 +3528,12 @@ int syncd_main(int argc, char **argv)
 
     SWSS_LOG_NOTICE("syncd started");
 
-    syncd_restart_type_t restartType = SYNCD_RESTART_TYPE_COLD;
+    syncd_restart_type_t shutdownType = SYNCD_RESTART_TYPE_COLD;
 
     try
     {
         SWSS_LOG_NOTICE("before onSyncdStart");
-        onSyncdStart(options.startType == SAI_WARM_BOOT);
+        onSyncdStart(false);
         SWSS_LOG_NOTICE("after onSyncdStart");
 
         startNotificationsProcessingThread();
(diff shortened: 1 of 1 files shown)


yxieca:
Jipang,

I could be wrong. But I am not sure we need to do that. My understanding is that SAI will initiate the save and syncd will simply discard all in-memory states. Because syncd is quitting after all.

I am not sure if you were asking about timing, I think remove_switch blocks until the SAI data has been saved to file system. So when we return from remove_switch(), we are ready to quit.

Let's talk tomorrow to sort this out. You might be talking about something I totally missed.

Cheers,
Ying

jipanyang:
Ying,

Sorry I didn't put it clear enough.  I was thinking about the ConsumerTable pops implementation:
https://github.com/Azure/sonic-swss-common/blob/master/common/consumer_table_pops.lua#L61
Before storing the data to m_buffer of ConsumerTable, the same data has been saved to ASIC DB.  I assume the data in ASICDB will be used to during syncd warm start.

syncd uses ConsumerTable for processing request from orchagent and other processes.  If m_buffer happens to have unprocessed data the moment syncd breaks from the while loop upon 
SYNCD_RESTART_TYPE_WARM shutdown request, there could be data inconsistency between ASIC DB and ASIC.

After previous post,  I realized that this PR https://github.com/Azure/sonic-sairedis/pull/335 should have eliminated the problem with asicState.  Not sure about the cases for flexCounter and flexCounterGroup,  they use ConsumerTable too.

I might have missed some internal processing of select() which possibly makes that a non-issue.

Thanks,
Jipan

yxieca:
Jipan,

All that you said is very valuable. I am still learning syncd warm shutdown and restore expectations from SAI. The things you listed would be things that I need to validate once I have basic warm restore procedure sorted out.

jipanyang:
The current change is a good starting point.  I think you may make further change later as needed with syncd restore processing.

yxieca:
I got verbal approval from Jipan and Kamil for this PR. I'll close the revert PR.

PR #349, opened 2018-09-26
[watarmark] add watermark support

mykolaf:
Add code for Watermark feature
[High Level Design Document](https://github.com/Azure/SONiC/blob/694754ac31c77724f43458e2779b552223791974/doc/watermarks_HLD.md)

*What I did*

- Extended Flex counter to collect PG stats
- Extended Flex counter to generate PG MAPS(PG to port, PG to index, PG to name)
- Read and clear stats mode support

### DIFF
--- a/syncd/syncd.cpp
+++ b/syncd/syncd.cpp
@@ -2748,6 +2748,14 @@ void processFlexCounterGroupEvent(
                     FlexCounter::addQueueCounterPlugin(sha, groupName);
                 }
             }
+            else if (field == PG_PLUGIN_FIELD)
+            {
+                auto shaStrings = swss::tokenize(value, ',');
+                for (const auto &sha : shaStrings)
+                {
+                    FlexCounter::addPriorityGroupCounterPlugin(sha, groupName);
+                }
+            }
             else if (field == PORT_PLUGIN_FIELD)
             {
                 auto shaStrings = swss::tokenize(value, ',');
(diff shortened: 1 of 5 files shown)


mykolaf:
Builds failed due to outdated schema. Dependency on sonic-swss-common.

lguohan:
retest this please

syncd/syncd.cpp
```diff
@@ -2857,6 +2873,29 @@ void processFlexCounterEvent(
...
+                std::vector<sai_ingress_priority_group_stat_t> pgCounterIds;
+                for (const auto &str : idStrings)
+                {
+                    sai_ingress_priority_group_stat_t stat;
+                    sai_deserialize_ingress_priority_group_stat(str, stat);
+                    pgCounterIds.push_back(stat);
+                }
+                FlexCounter::setPriorityGroupCounterList(vid, rid, groupName, pgCounterIds);
+            }
+            else if (objectType == SAI_OBJECT_TYPE_INGRESS_PRIORITY_GROUP && field == PG_ATTR_ID_LIST)
```
wendani: Is PG_ATTR_ID_LIST field used by watermark?

mykolaf: It's not used by watermark, but the other objects types (queue, port) have both  SAI_STAT and SAI_ATTR handlers, so implemented it for priority-groups as well.

PR #351, opened 2018-10-10
Add special comparison logic for LAG

kcudnik:


jipanyang:
What if the lag member had been removed temporarily?   Due to port state down or some configuration state.
https://github.com/Azure/sonic-swss/blob/master/orchagent/portsorch.cpp#L2087

kcudnik:
if lag was removed and there is no match for it we still try to match at random

jipanyang:
Does it mean once the member port for each lag comes back, they may be put into the wrong lag?

PR #353, opened 2018-10-15
Add next hop group specific comparison logic

kcudnik:


syncd/syncd_applyview.cpp
```diff
@@ -2302,6 +2387,21 @@ std::shared_ptr<SaiObj> findCurrentBestMatchForGenericObjectUsingHeuristic(
             return candidateLag;
     }
 
+    if (temporaryObj->getObjectType() == SAI_OBJECT_TYPE_NEXT_HOP_GROUP)
+    {
+        /*
+         * For next hop group, let's try find matching NHG which will be based on
+         * NHG set as SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID in route_entry.
+         * We assume that each class IPv4 and IPv6 will have different NGH, and
+         * each IP prefix will only be assigned to one NHG.
```
jipanyang: For the case of multiple VRFs,   one IP prefix may be assigned to more than one NHG?

lguohan: you are correct, we are not handling multiple vrf case now.

PR #357, opened 2018-10-18
Fix optimized remove in comparison logic

kcudnik:


syncd/syncd_applyview.cpp
```diff
@@ -1369,6 +1410,11 @@ class AsicView
                  * If operation is found then we need to insert this op after
                  * current iterator and forward iterator.
                  *
+                 * But here is a catch here, if taht last operation was REMOVE,
```
lguohan: >taht [](start = 48, length = 4)

that

syncd/syncd_applyview.cpp
```diff
@@ -1330,6 +1350,27 @@ class AsicView
...
+                 *
+                 * this logic could result with wrong order:
+                 * - remove D
+                 * - remove C (will break reference count on C, since used in B)
+                 *   since D is considered last OP releasing C reference
+                 * - create A
+                 * - set object B with A
+                 *
+                 * This is fixed by checking if last operation to release reference was REMOVE
+                 */
```
lguohan: If i understand correctly, with your fix, the optimized order is as follows?

remove D
create A
set object B with A
remove C

syncd/syncd_applyview.cpp
```diff
@@ -1381,6 +1427,23 @@ class AsicView
                             sai_serialize_object_id(op.vid).c_str());
                 }
 
+                if (itr->isRemove)
+                {
+                    SWSS_LOG_INFO("previous operation to release reference %s was also REMOVE, we push current REMOVE op at the list end",
+                            sai_serialize_object_id(mit->first).c_str());
+
+                    /*
+                     * If previous operation was REMOVE (it could be pused to
```
lguohan: pushed

meta/sai_meta.cpp
```diff
@@ -232,6 +232,41 @@ static std::unordered_map<sai_object_id_t,int32_t> ObjectReferences;
...
+
+            case SAI_OBJECT_TYPE_LAG:
+            case SAI_OBJECT_TYPE_NEXT_HOP:
+            case SAI_OBJECT_TYPE_NEXT_HOP_GROUP:
+            case SAI_OBJECT_TYPE_NEXT_HOP_GROUP_MEMBER:
+            case SAI_OBJECT_TYPE_ROUTE_ENTRY:
+            case SAI_OBJECT_TYPE_ROUTER_INTERFACE:
+                break;
+
+                // skip object we have no intrest into
```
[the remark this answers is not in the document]
jleveque: Also change "into" to "in"

syncd/syncd_applyview.cpp
```diff
@@ -13,6 +13,13 @@
  * parameters to every function.
  */
 
+/**
+ * When set to true extra logging will be added for tracking references.  This
+ * is useful for duugging, but for production operations this will produce too
```
stcheng: debugging

PR #358, opened 2018-10-18
Add support for ACL table group specific logic

kcudnik:


jipanyang:
How about ACL table group on VLAN,  Router?  And EGRESS_ACL?

lguohan:
is there a unit test for this matching?

kcudnik:
no, i will add unittests later for all objects at once

jipanyang:
What do you mean " vlan and egress acl, we cover current usecases"?    In your changes, the cases for port and lag are handled, but vlan/router interface and egress acl group are left out.

lguohan:
@jipanyang , I do not think we support binding vlan and router interface in swss. maybe I am missing something here. https://github.com/Azure/sonic-swss/blob/master/orchagent/aclorch.cpp#L1077

@for egress acl, it is for pfcwd. I will provide test case for kamil to verify that.

jipanyang:
@lguohan We have the need to support ACL on VLAN/RouterInterface and egress ACL on the objects too, the work is in progress. 

View comparison was supposed to be generic enough,  at least at SAI level. But it has become a key limiting factor for various functions and usage scenarios.

PR #360, opened 2018-10-22
Add specific comparison logic for more objects

kcudnik:


syncd/syncd_applyview.cpp
```diff
@@ -2629,54 +2677,342 @@ std::shared_ptr<SaiObj> findCurrentBestMatchForAclTableGroup(
...
+    SWSS_LOG_ENTER();
+
+    /*
+     * For policer we can see on which hostif trap group is set, and on which
+     * hostif trap.  Hostif trap have SAI_HOSTIF_TRAP_ATTR_TRAP_TYPE attribute
+     * which is KEY and there can be only one trap of that type. we can use
+     * that to match hostif trap group and later on policer.
+     *
+     * NOTE: policer can be set on default hostif trap group. In this case we
+     * are getting processed and not processed objects into account.
```
jipanyang: Policer could be applied on port and ACL too.  It looks the logic here limits the processing to hostif only.

syncd/syncd_applyview.cpp
```diff
@@ -6334,6 +6670,44 @@ sai_status_t asic_process_event(
...
+
+    std::ofstream log("applyview.log");
+
+    if (log.is_open())
+    {
+        log << ss.str();
+
+        log.close();
+
+        SWSS_LOG_NOTICE("wrote apply_view asic operations to applyview.log");
```
qiluo-msft: >applyview.log [](start = 61, length = 13)

Log the absolute path so we can easily find it.

syncd/syncd_applyview.cpp
```diff
@@ -6348,6 +6722,9 @@ void executeOperationsOnAsic(
 
         SWSS_LOG_TIMER("asic apply");
 
+        if (enableUnittests())
```
qiluo-msft: >enableUnittests [](start = 12, length = 15)

Name it as 'dumpComparision' or 'debugMode'?

syncd/syncd_applyview.cpp
kcudnik: correct, for current configuration this is only supported, but it will be expanded in the future

syncd/syncd_applyview.cpp
kcudnik: maybe not, since it maybe used in many other places in the future not only for comparsion

syncd/syncd_applyview.cpp
kcudnik: its in local directory, i can get pwd

syncd/syncd_applyview.cpp
```diff
@@ -2414,6 +2451,13 @@ std::shared_ptr<SaiObj> findCurrentBestMatchForNextHopGroup(
 {
     SWSS_LOG_ENTER();
 
+    /*
+     * For next hop group, let's try find matching NHG which will be based on
+     * NHG set as SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID in route_entry.  We assume
+     * that each class IPv4 and IPv6 will have different NGH, and each IP
+     * prefix will only be assigned to one NHG.
```
lguohan: this is true as v4/v6 route will point to v4/v6 next hop.

kcudnik: what do you mean ?

kcudnik:
test this please

PR #361, opened 2018-10-23
Add option to generate current dep graph to saidump

kcudnik:


saidump/saidump.cpp
```diff
@@ -150,6 +160,226 @@ void print_attributes(size_t indent, const TableMap& map)
     }
 }
 
+#define SAI_OBJECT_TYPE_PREFIX_LEN 16
```
qiluo-msft: >16 [](start = 35, length = 2)

Can we get the number programmatically?

qiluo-msft: >SAI_OBJECT_TYPE_PREFIX_LEN [](start = 8, length = 26)

Is it a local const?

saidump/saidump.cpp
```diff
@@ -150,6 +160,226 @@ void print_attributes(size_t indent, const TableMap& map)
...
+                    // skip link of SWITCH to non object id object types, since
+                    // all of them contain switch_id
+                    continue;
+                }
+
+                std::stringstream ss;
+
+                ss << std::string(member_info->objecttypename + SAI_OBJECT_TYPE_PREFIX_LEN) << " -> "
+                << std::string(info->objecttypename + SAI_OBJECT_TYPE_PREFIX_LEN)
+                << "[color=\"0.650 0.700 0.700\", style = dashed, penwidth=2]";
```
qiluo-msft: Define some color constants, so it's easy to adjust in future?

saidump/saidump.cpp
```diff
@@ -199,6 +429,13 @@ int main(int argc, char ** argv)
         }
     }
 
+    if (g_cmdOptions.dumpGraph)
+    {
+        dumpGraph(dump);
+
+        return EXIT_SUCCESS;
```
qiluo-msft: >EXIT_SUCCESS [](start = 15, length = 12)

How about no exit, and continue to dump everything?

saidump/saidump.cpp
kcudnik: lenght is "SAI_OBJECT_TYPE_" for this how to get size of it ? sizeof("SAI_OBJECT_TYPE_") -1 ?

kcudnik: yes

saidump/saidump.cpp
kcudnik: no, since graph is generated on the stdout so it could be dumped to *.gv file, and if other stuff would be generated then it would mess output

PR #367, opened 2018-10-30
Refactor saidump graph generator

kcudnik:


saidump/saidump.cpp
```diff
@@ -162,6 +164,11 @@ void print_attributes(size_t indent, const TableMap& map)
 
 #define SAI_OBJECT_TYPE_PREFIX_LEN 16
```
qiluo-msft: move into local scope?

qiluo-msft: sizeof("SAI_OBJECT_TYPE_") -1

kcudnik: fixed

PR #368, opened 2018-10-30
[syncd] re-establish notifications after warm boot

yxieca:
Signed-off-by: Ying Xie <ying.xie@microsoft.com>

### DIFF
--- a/syncd/syncd_hard_reinit.cpp
+++ b/syncd/syncd_hard_reinit.cpp
@@ -1253,10 +1253,25 @@ void performWarmRestart()
     sai_object_id_t orig_rid = translate_vid_to_rid(switch_vid);
 
     sai_object_id_t switch_rid;
-    sai_attribute_t switch_attr;
-    switch_attr.id             = SAI_SWITCH_ATTR_INIT_SWITCH;
-    switch_attr.value.booldata = true;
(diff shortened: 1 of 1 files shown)


syncd/syncd_hard_reinit.cpp
```diff
@@ -1253,10 +1253,25 @@ void performWarmRestart()
     sai_object_id_t orig_rid = translate_vid_to_rid(switch_vid);
 
     sai_object_id_t switch_rid;
-    sai_attribute_t switch_attr;
-    switch_attr.id             = SAI_SWITCH_ATTR_INIT_SWITCH;
-    switch_attr.value.booldata = true;
-    sai_status_t status = sai_metadata_sai_switch_api->create_switch(&switch_rid, 1, &switch_attr);
+    sai_attr_id_t   notifs[] = {
```
kcudnik: in total there is 8 notifications right now, i need to add some auto  generated metadata for all of them so that could be used in generic way

yxieca: I see. Do you want me to wait for your change?

kcudnik: no, since that would require also SAI update and cherry pick on branch 1.3, we can change this later

yxieca: cool! thanks!

PR #372, opened 2018-11-07
Ignore order when compare QOS MAP list entries

kcudnik:


jipanyang:
Sorry,  I just noticed this change.  One question, would all attributes value of list type have this issue potentially?    

Will it prevent such problem from happening if they are sorted when performing serialization before passing the request down to asicDB, so syncd may rely on the ordering property inherently for deserialization? Maybe I missed something,  I feel a little worried that the issue with list type might be a generic problem for comparison.

PR #374, opened 2018-11-10
Add warm boot support for virtual switch

kcudnik:


lguohan:
you need to create the linux host interface upon warm boot if it has been created before.

kcudnik:
scenario with working host interfaces is not supported yet, this is first PR of warm boot support, currently only pure virtual warm boot is supported, if necessary i can add linux hostif support in the future

kcudnik:
test this please

kcudnik:
Agree, but this also requires to dump more stuff from hostif and vs local fdb table, i will think how to do that. There may be some challenges here, since currently we have tap devices and fd descriptors opened, and they will change during system boot. But it's definitely doable.

PR #380, opened 2018-11-15
Rebuild VS local metadata on warm boot

kcudnik:


syncd/syncd.cpp
```diff
@@ -3427,9 +3425,9 @@ int syncd_main(int argc, char **argv)
 
     swss::Logger::getInstance().setMinPrio(swss::Logger::SWSS_NOTICE);
 
-    set_sai_api_loglevel();
+    //set_sai_api_loglevel();
```
lguohan: remove the comments?

syncd/syncd.cpp
```diff
@@ -1313,7 +1313,7 @@ sai_status_t handle_generic(
 
                         sai_object_id_t switch_vid = redis_sai_switch_id_query(object_id);
 
-                        if (switches.at(switch_vid)->isDefaultCreatedRid(rid))
+                        if (switches.at(switch_vid)->isDiscoveredRid(rid))
```
lguohan: is this about vs switch? does not seem fit into this PR?

kcudnik: this is because this change is on top of coldvids branch in other PR, since it will not compile without that

syncd/syncd.cpp
kcudnik: removed

PR #381, opened 2018-11-15
Add buffer pool logic and unittests

kcudnik:


lguohan:
can you remove the cold discovered oids pr?

PR #383, opened 2018-11-17
ASIC temperature sensors support

padmanarayana:
**Please do not merge yet.**

Recently (https://github.com/opencomputeproject/SAI/pull/880), new switch attributes were added to retrieve the temperature readings from the ASIC's internal sensors. 

This is a preliminary commit (pending SAI support from vendors) for temperature monitoring. The max, average and the entire list of temperatures are now added to the flex counters DB so that platform sensors scripts may query and use these values in thermal control algorithms.

### DIFF
--- a/meta/sai_serialize.h
+++ b/meta/sai_serialize.h
@@ -138,6 +138,9 @@ std::string sai_serialize_ipmc_entry_type(
 std::string sai_serialize_qos_map_item(
         _In_ const sai_qos_map_t& qosmap);
 
+std::string sai_serialize_switch_attr(
+        _In_ const sai_switch_attr_t switch_attr);
+
 // serialize notifications
 
 std::string sai_serialize_fdb_event_ntf(
@@ -286,5 +289,9 @@ void sai_deserialize_ingress_priority_group_attr(
 void sai_deserialize_queue_attr(
         _In_ const std::string& s,
         _Out_ sai_queue_attr_t& attr);
+
+void sai_deserialize_switch_attr(
+        _In_ const std::string& s,
+        _Out_ sai_switch_attr_t& attr);
 
 #endif // __SAI_SERIALIZE__
--- a/meta/saiserialize.cpp
+++ b/meta/saiserialize.cpp
@@ -1636,6 +1636,14 @@ std::string sai_serialize_object_meta_key(
     return key;
 }
 
+std::string sai_serialize_switch_attr(
+        _In_ const sai_switch_attr_t switch_attr)
+{
+    SWSS_LOG_ENTER();
+
+    return sai_serialize_enum(switch_attr, &sai_metadata_enum_sai_switch_attr_t);
+}
+
 // deserialize
 
 void sai_deserialize_bool(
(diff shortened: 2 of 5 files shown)


meta/sai_serialize.h
```diff
@@ -138,6 +138,9 @@ std::string sai_serialize_ipmc_entry_type(
 std::string sai_serialize_qos_map_item(
         _In_ const sai_qos_map_t& qosmap);
 
+std::string sai_serialize_switch_attr(
```
kcudnik: please remove we already have generic function to serialize attributes for all objects

meta/sai_serialize.h
```diff
@@ -287,4 +290,8 @@ void sai_deserialize_queue_attr(
         _In_ const std::string& s,
         _Out_ sai_queue_attr_t& attr);
 
+void sai_deserialize_switch_attr(
```
kcudnik: please remove we already have generic function to deserialize attributes for all objects

meta/saiserialize.cpp
```diff
@@ -1636,6 +1636,14 @@ std::string sai_serialize_object_meta_key(
     return key;
 }
 
+std::string sai_serialize_switch_attr(
```
kcudnik: please remove we already have generic function to serialize attributes for all objects

meta/saiserialize.cpp
```diff
@@ -2976,3 +2984,12 @@ void sai_deserialize_queue_attr(
 
     sai_deserialize_enum(s, &sai_metadata_enum_sai_queue_attr_t, (int32_t&)attr);
 }
+
+void sai_deserialize_switch_attr(
```
kcudnik: please remove we already have generic function to deserialize attributes for all objects

syncd/syncd_flex_counter.cpp
```diff
@@ -8,6 +8,7 @@ static std::map<std::string, std::shared_ptr<FlexCounter>> g_flex_counters_map;
 static std::set<sai_port_stat_t> supportedPortCounters;
 static std::set<sai_queue_stat_t> supportedQueueCounters;
 static std::set<sai_ingress_priority_group_stat_t> supportedPriorityGroupCounters;
+static std::set<sai_switch_attr_t> supportedSwitchSensors;
```
kcudnik: why temperature sensors are in counters ? i think we dont need this implementation here at all, seems like sensors are just another attribute and we can definitely query them

syncd/syncd.cpp
```diff
@@ -2907,6 +2907,17 @@ void processFlexCounterEvent(
 
                 FlexCounter::setPriorityGroupAttrList(vid, rid, groupName, pgAttrIds);
             }
+            else if (objectType == SAI_OBJECT_TYPE_SWITCH && field == SWITCH_SENSOR_ID_LIST)
```
kcudnik: why we need this here as special case ?
also this solution for those cases here is getting bigger and bigger we should think of some generic approach

syncd/syncd_flex_counter.cpp
kcudnik: as script i dont know, i had in mind orchagent which would query those sensors using sai apis

lguohan: since the load is low, ~8 values in maybe one minute. Moving to orchagent is better. agree with @kcudnik

AmitKaushik7:
Hi Padamanarayan,
    I see that these changes were not merged. Any reason for cancelling this.  Are you coming up with some generic approach to add temperature sensors support?
Thanks
Amit

padmanarayana:
Hi Amit,
Yes, we're planning to add a poller for the ASIC internal sensors and provide platform APIs that thermal algorithms may use. Will raise HLD PR for review at the earliest.

AmitKaushik7:
Thanks Padamanarayana,

Will the poller for ASIC internal sensors be able to handle other ASIC controls like LED ? 

As the higher speed ports can support various speeds(400G,100G,25G, 40G,10G) and multi-color LED reflects the color based on the port speed configured which needs to be updated whenever the port is configured for a specific speed. Can the platform API be provided to control the ASIC LED as well.

Could you please confirm about the tentative timelines for adding poller support?

PR #387, opened 2018-11-22
Add asic operations count log by object type

kcudnik:


syncd/syncd_applyview.cpp
```diff
@@ -7338,12 +7338,22 @@ void executeOperationsOnAsic(
...
 
             SWSS_LOG_NOTICE("%s: %s", opp.c_str(), key.c_str());
+
+            // count operations by object type
+            opByObjectType[key.substr(0, key.find(":"))]++;
+        }
+
+        for (auto kvp: opByObjectType)
+        {
+            SWSS_LOG_NOTICE("operations on %s: %d", kvp.first.c_str(), kvp.second);
```
lguohan: only print out objtype with non zero operations.

kcudnik: yea, this will print only non zero operations look at line 7343 and 7351

PR #388, opened 2018-11-22
Drain asic queue before processing shutdown request

kcudnik:


yxieca:
Kamil,

Can you rebase to the top of the current sonic-sairedis repo and add the even draining before pre-shutdown as well?

Thanks,
Ying

yxieca:
Kamil,

Maybe make it a helper function so that it could be called from both pre-shutdown and shutdown?

Also, maybe SWSS_LOG_TIMER it?

If you want to merge this one and create another PR. that would be fine too.

Thanks,
Ying

PR #390, opened 2018-11-26
[SAI header] upgrade SAI header to version v1.3.6

yxieca:
- Pick up SAI switch attribute SAI_SWITCH_ATTR_PRE_SHUTDOWN definition

Signed-off-by: Ying Xie <ying.xie@microsoft.com>


arkadiyshapiro:
will this be final SAI version for Nov SONiC release? Should that release still work with 1.3.3?

PR #391, opened 2018-11-26
[sycnd] add 2 stage shutdown support for warm reboot

yxieca:
Adding 2 stage shutdown support for warm reboot.

Stage-1: pre-shutdown, back up SAI/SDK data, detach almost all data plane, but leave CPU port active.
Stage-2: warm shutdown, fully shutdown SAI.

### DIFF
--- a/syncd/syncd.cpp
+++ b/syncd/syncd.cpp
@@ -1937,10 +1937,11 @@ void on_switch_create_in_init_view(
         sai_object_id_t switch_rid;
 
         sai_status_t status;
-        do {
+
+        {
             SWSS_LOG_TIMER("cold boot: create switch");
             status = sai_metadata_sai_switch_api->create_switch(&switch_rid, attr_count, attr_list);
-        } while (0);
+        }
 
         if (status != SAI_STATUS_SUCCESS)
         {
@@ -3726,10 +3727,11 @@ int syncd_main(int argc, char **argv)
     // Stop notification thread before removing switch
     stopNotificationsProcessingThread();
 
-    do {
+    {
         SWSS_LOG_TIMER("remove switch");
         status = sai_switch_api->remove_switch(gSwitchId);
-    } while (0);
+    }
+
     if (status != SAI_STATUS_SUCCESS)
     {
         SWSS_LOG_NOTICE("Can't delete a switch. gSwitchId=0x%lx status=%s", gSwitchId,
--- a/syncd/syncd.cpp
+++ b/syncd/syncd.cpp
@@ -6,6 +6,7 @@
 #include <limits.h>
 
 #include "swss/warm_restart.h"
+#include "swss/table.h"
 
 extern "C" {
 #include <sai.h>
@@ -3170,6 +3171,8 @@ typedef enum _syncd_restart_type_t
 
     SYNCD_RESTART_TYPE_FAST,
 
+    SYNCD_RESTART_TYPE_PRE_SHUTDOWN,
+
 } syncd_restart_type_t;
 
 syncd_restart_type_t handleRestartQuery(swss::NotificationConsumer &restartQuery)
@@ -3200,6 +3203,12 @@ syncd_restart_type_t handleRestartQuery(swss::NotificationConsumer &restartQuery
     {
         SWSS_LOG_NOTICE("received FAST switch shutdown event");
         return SYNCD_RESTART_TYPE_FAST;
+    }
+
+    if (op == "PRE")
+    {
+        SWSS_LOG_NOTICE("received PRE_SHUTDOWN switch event");
+        return SYNCD_RESTART_TYPE_PRE_SHUTDOWN;
     }
 
     SWSS_LOG_WARN("received '%s' unknown switch shutdown event, assuming COLD", op.c_str());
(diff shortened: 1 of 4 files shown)


syncd/syncd.cpp
```diff
@@ -3575,7 +3593,54 @@ int syncd_main(int argc, char **argv)
...
+                    continue;
+                }
+
+                attr.id = SAI_SWITCH_ATTR_PRE_SHUTDOWN;
+                attr.value.booldata = true;
+
+                status = sai_switch_api->set_switch_attribute(gSwitchId, &attr);
+                if (status == SAI_STATUS_SUCCESS)
+                {
+                    warmRestartTable->hset("warm-shutdown", "state", "pre-shutdown-succeeded");
```
jipanyang: Once the switch is in pre-shutdown-succeeded state, is there need to prevent syncd from fetching events other than SYNCD_RESTART_TYPE_WARM,  in case there is any pending request data in asicDB?

syncd/syncd.cpp
```diff
@@ -1935,7 +1936,11 @@ void on_switch_create_in_init_view(
 
         sai_object_id_t switch_rid;
 
-        sai_status_t status = sai_metadata_sai_switch_api->create_switch(&switch_rid, attr_count, attr_list);
+        sai_status_t status;
+        do {
+            SWSS_LOG_TIMER("cold boot: create switch");
+            status = sai_metadata_sai_switch_api->create_switch(&switch_rid, attr_count, attr_list);
+        } while (0);
```
qiluo-msft: >while (0); [](start = 10, length = 10)

I prefer only { }

syncd/syncd.cpp
```diff
@@ -3202,6 +3209,12 @@ syncd_restart_type_t handleRestartQuery(swss::NotificationConsumer &restartQuery
         return SYNCD_RESTART_TYPE_FAST;
     }
 
+    if (op == "PRE")
```
qiluo-msft: >PRE [](start = 15, length = 3)

PRE may be not that meaningful. How about PRE-WARM?

syncd/syncd.cpp
yxieca: The draining event logic is likely more important in syncd warm restart case. Since we paused orchagent before pre-shutdown, and also stopped event notification from SAI. We have slim chance running into issue in system level warm reboot.

With that said, what you pointed out is a valid issue.

syncd/syncd.cpp
```diff
@@ -3575,7 +3593,54 @@ int syncd_main(int argc, char **argv)
...
+                // event
+
+                SWSS_LOG_TIMER("warm pre-shutdown");
+
+                FlexCounter::removeAllCounters();
+                stopNotificationsProcessingThread();
+
+                sai_attribute_t attr;
+
+                attr.id = SAI_SWITCH_ATTR_RESTART_WARM;
```
qiluo-msft: >SAI_SWITCH_ATTR_RESTART_WARM [](start = 26, length = 28)

What is current supporting status of different SAI implementations? If not supported, what will happen?

syncd/syncd_request_shutdown.cpp
```diff
@@ -16,7 +16,9 @@ int main(int argc, char **argv)
     static struct option long_options[] =
     {
         { "cold", no_argument, 0, 'c' },
-        { "warm", no_argument, 0, 'w' }
+        { "warm", no_argument, 0, 'w' },
+        { "fast", no_argument, 0, 'f' },
+        { "pre",  no_argument, 0, 'p' }, // Requesting pre shutdown
```
qiluo-msft: >pre [](start = 11, length = 3)

Can we implement bash function wait_for_pre_shutdown_complete_or_fail inside syncd_request_shutdown? so when the process return, it's done.

syncd/syncd.cpp
yxieca: Done.

syncd/syncd.cpp
yxieca: Expanded to PRE-SHUTDOWN

syncd/syncd.cpp
yxieca: Both 3.1 and 3.3 support SAI_SWITCH_ATTR_RESTART_WARM. 3.1 support is not officially complete.

syncd/syncd_request_shutdown.cpp
yxieca: That is certainly possible. Take it in another review?

syncd/syncd.cpp
```diff
@@ -1935,7 +1936,12 @@ void on_switch_create_in_init_view(
 
         sai_object_id_t switch_rid;
 
-        sai_status_t status = sai_metadata_sai_switch_api->create_switch(&switch_rid, attr_count, attr_list);
+        sai_status_t status;
+
+        {
```
prsunny: Just curios, why do we need this grouping "{ }" for create/remove switch?

syncd/syncd.cpp
```diff
@@ -3575,7 +3594,54 @@ int syncd_main(int argc, char **argv)
                  */
 
                 shutdownType = handleRestartQuery(*restartQuery);
-                break;
+                if (shutdownType != SYNCD_RESTART_TYPE_PRE_SHUTDOWN) {
```
prsunny: `{ `to next line

syncd/syncd.cpp
yxieca: Kamil already has a PR outstanding to solve this issue. The code could extend to drain event queue before pre-shutdown.

Since pre-shutdown is only called after pausing orchagent,  adding that should solve the queue draining issue.

PR #392, opened 2018-11-27
Add support for recreate host interfaces tap devices on warm start

kcudnik:


lguohan:
as the initial version, it looks fine to me. there are some cases I do not see, for example, the oper status handling for the tap interface, but it can be in further PR.

PR #393, opened 2018-11-29
Add check ASIC vs DB state after comparison logic

kcudnik:


lguohan:
@kcudnik , can you take a look?

kcudnik:
yea, i send email about this, it's jenkins issue
`+ sudo service rsyslog start
rsyslog: unrecognized service
Build step 'Execute shell' marked build as failure
Archiving artifacts`

kcudnik:
test this please

yxieca:
Looks like your DB/SAI comparison is working! The only problem is that it surely triggers Broadcom SAI segment fault while comparing.

Can you make it optional?

I think the best way to trigger it might be reuse the syncd_request_shutdown program to send an event to check. What do you think?

Alternative would be if you receive a signal then start checking. like USR1 :-)

kcudnik:
i would rather like to avoid using linux signals to make make something like this

PR #394, opened 2018-12-04
Allow dst/src MAC set to fail on mirror session (temporary workaround)

kcudnik:


kcudnik:
why it's not correct ?

PR #395, opened 2018-12-04
Process only shutdown requests after pre shutdown

kcudnik:


syncd/syncd.cpp
```diff
@@ -3659,7 +3659,6 @@ int syncd_main(int argc, char **argv)
                 SWSS_LOG_TIMER("warm pre-shutdown");
 
                 FlexCounter::removeAllCounters();
-                stopNotificationsProcessingThread();
```
yxieca: Looks like you added this diff by accident. You removed this line but never put pit back.

kcudnik: as we spoke previously we want to enable notifications in pre shutdown stage

yxieca: We we handle this change in the next PR.

PR #398, opened 2018-12-07
[flex counter] use get_stats_ext SAI calls for queue and PG counters

mykolaf:
closes a TODO.

Use a get_stats_ext instead of separate get and clear calls.
Should boost performance.

This is related to watermark read and clear behavior.

Signed-off-by: Mykola Faryma <mykolaf@mellanox.com>

### DIFF
--- a/syncd/syncd_flex_counter.cpp
+++ b/syncd/syncd_flex_counter.cpp
@@ -700,26 +700,12 @@ void FlexCounter::collectCounters(
         std::vector<uint64_t> queueStats(queueCounterIds.size());
 
         // Get queue stats
-        sai_status_t status = -1;
-//        TODO: replace if with get_queue_stats_ext() call when it's fully supported
-//        Example:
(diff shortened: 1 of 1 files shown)


lguohan:
@pavel-shirshov , i think not.

wendani:
@mykolaf ext call is not yet implemented in brcm sai

mykolaf:
> @mykolaf ext call is not yet implemented in brcm sai

@wendani does the new Broadcom SAI 3.3.4.3-1 support it?

wendani:
@mykolaf no. I think your pr will stand for a while.

https://github.com/Azure/sonic-sairedis/pull/398#issuecomment-461857834

lguohan:
@mykolaf, can you first query whether the extended api is implemented or not. if it is, then use it, if it is not, then do not use.

kcudnik:
compilation issues are related to "nm" command displaying errors, and compilation of saimetadata not placing all the symbols in read only section of binary

liat-grozovik:
retest please

PR #401, opened 2018-12-11
Move stop notifications thread after remove switch

kcudnik:


yxieca:
You should also change the pre-shutdown code path to also move stop notification thread stop to after switch shutdown?

PR #407, opened 2019-01-09
Add support for any number of ports in virtual switch using lane map

kcudnik:


yxieca:

PR #408, opened 2019-01-11
Update SAI pointer to latest master

marian-pritsak:
Signed-off-by: Marian Pritsak <marianp@mellanox.com>

### DIFF
--- a/lib/inc/sai_redis_internal.h
+++ b/lib/inc/sai_redis_internal.h
@@ -145,7 +145,7 @@
     sai_status_t redis_get_ ## object_type ## _stats(               \
             _In_ sai_object_id_t object_type ## _id,                \
             _In_ uint32_t number_of_counters,                       \
-            _In_ const sai_ ## object_type ## _stat_t *counter_ids, \
+            _In_ const sai_stat_id_t *counter_ids, \
             _Out_ uint64_t *counters)                               \
     {                                                               \
         MUTEX();                                                    \
@@ -164,7 +164,7 @@
     sai_status_t redis_get_ ## object_type ## _stats_ext(           \
             _In_ sai_object_id_t object_type ## _id,                \
             _In_ uint32_t number_of_counters,                       \
-            _In_ const sai_ ## object_type ## _stat_t *counter_ids, \
+            _In_ const sai_stat_id_t *counter_ids, \
             _In_ sai_stats_mode_t mode,                             \
             _Out_ uint64_t *counters)                               \
     {                                                               \
@@ -184,7 +184,7 @@
     sai_status_t redis_clear_ ## object_type ## _stats(             \
             _In_ sai_object_id_t object_type ## _id,                \
             _In_ uint32_t number_of_counters,                       \
-            _In_ const sai_ ## object_type ## _stat_t *counter_ids) \
+            _In_ const sai_stat_id_t *counter_ids) \
     {                                                               \
         MUTEX();                                                    \
         SWSS_LOG_ENTER();                                           \
--- a/lib/inc/sai_redis_internal.h
+++ b/lib/inc/sai_redis_internal.h
@@ -3,64 +3,64 @@
 
 // object id
 
-#define REDIS_CREATE(OBJECT_TYPE,object_type)           \
-    sai_status_t redis_create_ ## object_type(          \
-            _Out_ sai_object_id_t *object_type ##_id,   \
-            _In_ sai_object_id_t switch_id,             \
-            _In_ uint32_t attr_count,                   \
-            _In_ const sai_attribute_t *attr_list)      \
-    {                                                   \
-        MUTEX();                                        \
-        SWSS_LOG_ENTER();                               \
-        return meta_sai_create_oid(                     \
-                SAI_OBJECT_TYPE_ ## OBJECT_TYPE,        \
-                object_type ## _id,                     \
-                switch_id,                              \
-                attr_count,                             \
-                attr_list,                              \
-                &redis_generic_create);                 \
+#define REDIS_CREATE(OBJECT_TYPE,object_type)                       \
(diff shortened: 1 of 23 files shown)


marian-pritsak:
Depends on https://github.com/opencomputeproject/SAI/pull/909

kcudnik:
overall those changes looks good, please corect the error:
saimetadatatest.c:2140:16: error: array subscript is above array bounds [-Werror=array-bounds]
         checked[(int)SAI_OBJECT_TYPE_TABLE_VNET_ENTRY] = SAI_OBJECT_TYPE_TABLE_VNET_ENTRY;
                ^
saimetadatatest.c:2160:16: error: array subscript is above array bounds [-Werror=array-bounds]
         checked[(int)SAI_OBJECT_TYPE_TABLE_TUNNEL_ROUTE_ENTRY] = SAI_OBJECT_TYPE_TABLE_TUNNEL_ROUTE_ENTRY;
                ^

lib/inc/sai_redis_internal.h
```diff
@@ -3,64 +3,64 @@
...
+    sai_status_t redis_create_ ## object_type(                      \
+            _Out_ sai_object_id_t *object_type ##_id,               \
+            _In_ sai_object_id_t switch_id,                         \
+            _In_ uint32_t attr_count,                               \
+            _In_ const sai_attribute_t *attr_list)                  \
+    {                                                               \
+        MUTEX();                                                    \
+        SWSS_LOG_ENTER();                                           \
+        return meta_sai_create_oid(                                 \
+                (sai_object_type_t)SAI_OBJECT_TYPE_ ## OBJECT_TYPE, \
```
kcudnik: is cast needed because extensions object types?

lib/inc/sai_redis_internal.h
```diff
@@ -184,7 +184,7 @@
     sai_status_t redis_clear_ ## object_type ## _stats(             \
             _In_ sai_object_id_t object_type ## _id,                \
             _In_ uint32_t number_of_counters,                       \
-            _In_ const sai_ ## object_type ## _stat_t *counter_ids) \
+            _In_ const sai_stat_id_t *counter_ids) \
```
kcudnik: i preffered to have separate type for each stats, this gave extra check for type, so user would not pass different stats for given objects at compile time
is there any reason you are changing this ?

syncd/syncd_flex_counter.cpp
```diff
@@ -799,13 +799,13 @@ void FlexCounter::collectCounters(
         status = sai_metadata_sai_buffer_api->get_ingress_priority_group_stats(
                         priorityGroupId,
                         static_cast<uint32_t>(priorityGroupCounterIds.size()),
-                        priorityGroupCounterIds.data(),
+                        (const sai_stat_id_t *)priorityGroupCounterIds.data(),
```
kcudnik: from my perspective it would be better to have each stats accepting separate stat's enum but i guess this is already changed on the SAI api's

vslib/src/tests.cpp
```diff
@@ -297,7 +297,7 @@ void test_set_stats_via_redis()
     sai_port_stat_t ids[1] = { SAI_PORT_STAT_IF_IN_ERRORS };
     uint64_t counters[1] = { 0 } ;
 
-    SUCCESS(sai_metadata_sai_port_api->get_port_stats(ports.at(0), 1, ids, counters));
+    SUCCESS(sai_metadata_sai_port_api->get_port_stats(ports.at(0), 1, (const sai_stat_id_t *)ids, counters));
```
kcudnik: previously we had each type for each stats, and cast was not necessary

lib/inc/sai_redis_internal.h
marian-pritsak: I'm afraid it's not possible to keep the old type, because this function then cannot be casted to a function pointer of the SAI API, that is using sai_stat_id_t.

syncd/syncd_flex_counter.cpp
marian-pritsak: Yes, SAI APIs were changed, so I had to do this cast.

vslib/src/tests.cpp
marian-pritsak: Yes, but I'm adapting to the new SAI APIs.
I don't know what was the reason behind changing all of them to the same type, but now this is a necessary change.

lib/inc/sai_redis_internal.h
kcudnik: yes, each new function pointer needs to be created for each stat, and this is how it was so far

marian-pritsak: Now in SAI stat functions look like this:

```
typedef sai_status_t (*sai_clear_bfd_session_stats_fn)(
        _In_ sai_object_id_t bfd_session_id,
        _In_ uint32_t number_of_counters,
        _In_ const **sai_stat_id_t** *counter_ids);
```

So if I leave macros as is, it will fail to compile with following error:

`error: invalid conversion from 'sai_status_t (*)(sai_object_id_t, uint32_t, const sai_bfd_session_stat_t*) {aka int (*)(long unsigned int, unsigned int, const _sai_bfd_session_stat_t*)}' to 'sai_clear_bfd_session_stats_fn {aka int (*)(long unsigned int, unsigned int, const unsigned int*)}' [-fpermissive]`

Maybe I'm missing your point, but doesn't redis SAI API have to look exactly like SAI api?
These changes to sairedis are done with an assumption that SAI API is taken as is.

marian-pritsak:
@kcudnik I agree with you that changes related to sai_stat_id_t seem odd, but they are necessary because sairedis compilation is constrained by SAI heders definition. I would suggest revisiting this uniform declaration of stat functions using sai_stat_id_t because it breaks a lot of code.

kcudnik:
yes, agree

PR #411, opened 2019-01-23
Initialize notification queue pointer before switch create

kcudnik:


syncd/syncd_notifications.cpp
```diff
@@ -667,6 +665,15 @@ void startNotificationsProcessingThread()
 {
     SWSS_LOG_ENTER();
 
+    /*
+     * Make sure that notification queue pointer is populated before we start
+     * thread, and before we create_switch, since at switch_create we can start
+     * receiving fdb_notifications which will arrive on differet thread and
```
yxieca: s/differet/different/

syncd/syncd_notifications.cpp
```diff
@@ -667,6 +665,15 @@ void startNotificationsProcessingThread()
...
+     * will call queueStats() when queue pointer could be null (this=0x0).
+     */
+
+    ntf_queue_hdlr = std::unique_ptr<ntf_queue_t>(new ntf_queue_t);
```
yxieca: I don't think you can use unique_ptr here. It will be destroyed when returning from startNotificationsProcessingThread() because it is a scope-life-span object.

syncd/syncd_notifications.cpp
kcudnik: fixed

yxieca:

PR #413, opened 2019-01-24
Add WRED specific comparison logic

kcudnik:
Add specific logic for matching WRED objects on temporary view and current view. In this scenario Queue will be used as anchor object to match corresponding WRED.


kcudnik:
done

yxieca:

PR #420, opened 2019-02-11
Add support for fdb_event MOVE and check fdb event oids

kcudnik:


syncd/syncd_notifications.cpp
```diff
@@ -240,18 +240,71 @@ void redisPutFdbEntryToAsicView(
...
+     * inconsistent DB state and possible failure on apply view after cold or
+     * warm boot.
+     *
+     * On switch init we do discover phase, and we should discover all objects
+     * so we should not get any of those messages if SAI is in consistent
+     * state.
+     */
+
+    if (!check_rid_exists(data.fdb_entry.bv_id))
+        SWSS_LOG_ERROR("bv_id RID 0x%lx is not present on local ASIC DB", data.fdb_entry.bv_id);
```
lguohan: add fdb_entry in the message for clarity.

syncd/syncd_notifications.cpp
```diff
@@ -240,18 +240,71 @@ void redisPutFdbEntryToAsicView(
...
+
+    if (!check_rid_exists(data.fdb_entry.switch_id) || data.fdb_entry.switch_id == SAI_NULL_OBJECT_ID)
+        SWSS_LOG_ERROR("switch_id RID 0x%lx is not present on local ASIC DB", data.fdb_entry.bv_id);
```
lguohan: add fdb_entry in the message for clarity.

syncd/syncd_notifications.cpp
```diff
@@ -240,18 +240,71 @@ void redisPutFdbEntryToAsicView(
...
+            SWSS_LOG_ERROR("unable to get metadata for fdb_entry attr.id = %d", attr.id);
+            continue;
+        }
+
+        // skip non oid attributes
+        if (meta->attrvaluetype != SAI_ATTR_VALUE_TYPE_OBJECT_ID)
+            continue;
+
+        if (!check_rid_exists(attr.value.oid))
+            SWSS_LOG_WARN("RID 0x%lx on %s is not present on local ASIC DB", attr.value.oid, meta->attridname);
```
lguohan: add fdb_entry in the message for clarity.

lguohan: why line 268 and 271 are error level, but this one is warning level? shouldn't them be same level?

saiplayer/saiplayer.cpp
```diff
@@ -1273,6 +1309,8 @@ void printUsage()
     std::cout << "        Enable temporary view between init and apply" << std::endl << std::endl;
     std::cout << "    -i --inspectAsic:" << std::endl;
     std::cout << "        Inspect ASIC by ASIC DB" << std::endl << std::endl;
+    std::cout << "    -s --sleep:" << std::endl;
+    std::cout << "        Sleep after success reply, to notice fdb notifications" << std::endl << std::endl;
```
lguohan: is this only for fdb notification, could be any notification, e.g. port oper status notification?

meta/sai_meta.cpp
```diff
@@ -6024,6 +6063,35 @@ void meta_sai_on_fdb_event_single(
...
+     * create bridge/vlan/bridge port we should snoop here new OIDs and put
+     * them in local DB.
+     *
+     * Unfortunately we don't have a way to check whether those OIDs are correct
+     * or whether there maybe some bug in vendor SAI and for example is sending
+     * invalid OIDs in those event's. Also sai_object_type_query can return
+     * valid object type for OID, but this does not guarantee that this OID is
+     * valid, for example one of existing bridge ports that orch agent didn't
+     * query yet.
+     */
```
lguohan: this PR add visibility to this issue, in such case you can print out warning message. but it does not solve the issue.

lguohan:
I have some minor comments, but ideally this should be separated into three PR. fdb move, check fdb evnt oid, and add sleep for saiplayer.

meta/sai_meta.cpp
kcudnik: yes, that was intention of this PR

saiplayer/saiplayer.cpp
kcudnik: actually no, you will see all notifications since it will just sleep and sairedis will receive all

syncd/syncd_notifications.cpp
kcudnik: because if received switch id in fdb notification is not existing in db already then this is a bug same for bridge vlan/id
and here is warning since user could not query yet all the bridge ports, or - thats depending on our discussion, depends how we want to handle ports, whether should we allow SAI to generate new ports on the fly

syncd/syncd_notifications.cpp
kcudnik: done

syncd/syncd_notifications.cpp
kcudnik: done

saiplayer/saiplayer.cpp
kcudnik: changed to any

yxieca:
Made to 201811 branch on 2/21/2019

meta/sai_meta.cpp
```diff
@@ -6024,6 +6063,35 @@ void meta_sai_on_fdb_event_single(
 
     std::string key_fdb = sai_serialize_object_meta_key(meta_key_fdb);
 
+    /*
+     * Because we could receive fdb event's before orch agent will query or
+     * create bridge/vlan/bridge port we should snoop here new OIDs and put
+     * them in local DB.
```
jipanyang: The comment is a little confusing.
In what scenario  fdb event may be received before orch agent  "create" bridge/vlan/bridge port?  If that does happen,  the VID that comes with FDB will be different from VID assigned to the object during "create" operation.  The integrity of meta reference will be messed up.

kcudnik: there are bridge ports creted by default on the switch and between create switch and removing them it may happen that fdb_event arrive

jipanyang: Ok,  I thought you were talking about FDB event for the non-default bridge/vlan/bridge port object.

If it does happen that fdb_event arrive before removing default bridge ports, will it cause validation error since the bridge ports has non-zero reference count? Orchagent initialization fails at this case?

PR #421, opened 2019-02-13
Add more specific logic for ingress ACL and buffer profile

kcudnik:
Add more specific comparison logic for matching ingress ACL and matching buffer profile on queues and ingress priority groups

### DIFF
--- a/syncd/syncd_applyview.cpp
+++ b/syncd/syncd_applyview.cpp
@@ -3431,7 +3431,7 @@ std::shared_ptr<SaiObj> findCurrentBestMatchForBufferProfile(
      * For buffer profile we will try using SAI_INGRESS_PRIORITY_GROUP_ATTR_BUFFER_PROFILE
      * or SAI_QUEUE_ATTR_BUFFER_PROFILE_ID for matching.
      *
-     * We we are here, and buffer profile has assigned buffer pool, then buffer
+     * If we are here, and buffer profile has assigned buffer pool, then buffer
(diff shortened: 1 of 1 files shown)


syncd/syncd_applyview.cpp
```diff
@@ -3412,6 +3419,102 @@ std::shared_ptr<SaiObj> findCurrentBestMatchForBufferPool(
...
+        _In_ const std::shared_ptr<const SaiObj> &temporaryObj,
+        _In_ const std::vector<sai_object_compare_info_t> &candidateObjects)
+{
+    SWSS_LOG_ENTER();
+
+    /*
+     * For buffer profile we will try using SAI_INGRESS_PRIORITY_GROUP_ATTR_BUFFER_PROFILE
+     * or SAI_QUEUE_ATTR_BUFFER_PROFILE_ID for matching.
+     *
+     * We we are here, and buffer profile has assigned buffer pool, then buffer
```
lguohan: remove we

lguohan:
do we need unit test for this comparison logic?

lguohan:
code looks good to me, but I think we need to add unit tests for these two cases.

kcudnik:
yes, i will add unittest with full log that Wenda provided which will also cover the empty Lag case, i will send  that in next pr

syncd/syncd_applyview.cpp
kcudnik: fixed

yxieca:
Made to 201811 branch on 2/14/2019

PR #423, opened 2019-02-14
Add pre match to comparison logic and unittests

kcudnik:


syncd/syncd_applyview.cpp
```diff
@@ -2563,7 +2569,7 @@ std::shared_ptr<SaiObj> findCurrentBestMatchForLag(
         }
     }
 
-    SWSS_LOG_WARN("failed to find best candidate for LAG using LAG member and port id: tmp %s", temporaryObj->str_object_id.c_str());
+    SWSS_LOG_NOTICE("failed to find best candidate for LAG using lag member and port");
```
jleveque: Suggest keeping "LAG" all upper-case.

kcudnik: done

PR #425, opened 2019-02-17
[syncd]: support query port with 8 lanes (400G)

lguohan:
Signed-off-by: Guohan Lu <gulv@microsoft.com>

### DIFF
--- a/syncd/syncd_saiswitch.cpp
+++ b/syncd/syncd_saiswitch.cpp
@@ -148,7 +148,7 @@ std::unordered_map<sai_uint32_t, sai_object_id_t> SaiSwitch::saiGetHardwareLaneM
      * addressed in future.
      */
 
-    const int lanesPerPort = 4;
+    const int lanesPerPort = 8;
 
     for (const auto &port_rid : portList)
     {


syncd/syncd_saiswitch.cpp
```diff
@@ -148,7 +148,7 @@ std::unordered_map<sai_uint32_t, sai_object_id_t> SaiSwitch::saiGetHardwareLaneM
      * addressed in future.
      */
 
-    const int lanesPerPort = 4;
+    const int lanesPerPort = 8;
```
kcudnik: this only change is not full, since there are more places in this repo, for example virtual switch only supports 1,2 or 4 lanes:
./vslib/src/sai_vs_hostintf.cpp:  line 948  uint32_t lanes[4];
line 952   attr.value.u32list.count = 4;

./vslib/src/sai_vs_interfacequery.cpp
575         if (n != 1 && n != 2 && n != 4)

actually this PR is a good opportunity to make 1 global define for all  those files and add extra method  to allow only number of lanes to be power of 2 

would you be able to do that ?

lguohan: where do I put the common header?

kcudnik: yea, there is no good place to put it :/ ok lets leave this as is, and i will add later some common lib inside this repo, since there are more stuff that could be moved to common which could be used in saiplyer, saidump, syncd, vslib and sairedis lib

PR #426, opened 2019-02-22
Fix compilation issues in stretch docker with gcc-6.3

stepanblyschak:
As part of upgrading syncd docker to stretch


yxieca:
201811 also hit compiler issue.

yxieca:
This PR also changed SAI. I will cherry-pick only the syncd_applyView.cpp change.

PR #427, opened 2019-02-25
 Watermark: test SAI clear api if the stats mode is READ_AND_CLEAR

wendani:


kcudnik:
In the title you have "test clear api", and you are adding some log messages with error/notice level, so they will be always visible in the logs, wont that produce a lot of noise in syslog if some of the platform don't support some features ?

wendani:
Hi Kamil
saiUpdateSupportedQueueCounters
saiUpdateSupportedPriorityGroupCounters

will probe the get_stat and clear_stat function call capability, and remove the stats if they are not supported on certain platforms. This ensures that it is less likely the polling will return error during the actual polling phase.


https://github.com/Azure/sonic-sairedis/pull/427#issuecomment-467345516

PR #428, opened 2019-02-26
Drop FDB notifications if they contain invalid OIDs

kcudnik:


syncd/syncd_notifications.cpp
```diff
@@ -301,11 +331,13 @@ void process_on_fdb_event(
...
     SWSS_LOG_INFO("fdb event count: %u", count);
 
+    bool sendntf = true;
+
     for (uint32_t i = 0; i < count; i++)
     {
         sai_fdb_event_notification_data_t *fdb = &data[i];
 
-        check_fdb_event_notification_data(*fdb);
+        sendntf &= check_fdb_event_notification_data(*fdb);
```
qiluo-msft: >check_fdb_event_notification_data [](start = 19, length = 33)

If any check failed, should we break the for-loop immediately?

kcudnik: i just wanted to print all the failures if there is more than one

PR #429, opened 2019-03-01
[bfn|syncd] Ged rid of ld_preload. Link against libsai only

NStetskovych:
- Remove hard coded libs in saidiscovery for bfn platform
- Remove ignoring unresolved-symbols
- Get rid of LD_PRELOAD in syncd

Signed-off-by: Nadiya.Stetskovych <nstetskovych@barefootnetworks.com>

### DIFF
--- a/saidiscovery/Makefile.am
+++ b/saidiscovery/Makefile.am
@@ -12,7 +12,8 @@ if SAIVS
 SAILIB=-L$(top_srcdir)/vslib/src/.libs -lsaivs
 else
 if sonic_asic_platform_barefoot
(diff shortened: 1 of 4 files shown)


NStetskovych:
@lguohan pls review
@stcheng

lguohan:
actually, I'd like you to get rid of the barefoot compiling option completely, can you do that?

NStetskovych:
> actually, I'd like you to get rid of the barefoot compiling specific option completely, can you do that?

@lguohan Could we do it in two stages to speed up? as it could require changes in bfn deb packages as well.

lguohan:
ok.

PR #438, opened 2019-03-28
 Make object list deterministic when iterating

kcudnik:


wendani:
What is the difference to https://github.com/Azure/sonic-sairedis/pull/433

only the last commits of your 3 commits in this pr?

kcudnik:
yes, since this is on top of pre match logic, and i was waiting to pre match to be merged

PR #439, opened 2019-03-28
Dynamic transceiver tuning support

dgsudharsan:
**- What I did**
Dynamic transceiver tuning support
**- How I did it**
Dynamically program pre-emphasis and other settings based on media detected in xcvrd. The various supported media are defined in media_settings.json

**- How to verify it**
Perform OIR of different media types. Dump the pre-emphasis values from hardware and verify that they got reflected

**- Description for the changelog**

Support for dynamically programming pre-emphasis

**- A picture of a cute animal (not mandatory but encouraged)**

### DIFF
--- a/meta/sai_serialize.h
+++ b/meta/sai_serialize.h
@@ -12,6 +12,7 @@ extern "C" {
 #include <streambuf>
 #include <iomanip>
 #include <map>
+#include <vector>
 #include <tuple>
 #include <string.h>
 #include "swss/logger.h"
@@ -280,5 +281,9 @@ void sai_deserialize_ingress_priority_group_attr(
 void sai_deserialize_queue_attr(
         _In_ const std::string& s,
         _Out_ sai_queue_attr_t& attr);
+
+void sai_deserialize_port_serdes_attr(
+        _In_ const std::string& s,
+        _Out_ std::vector<uint32_t> &lane_values);
 
 #endif // __SAI_SERIALIZE__
--- a/meta/saiserialize.cpp
+++ b/meta/saiserialize.cpp
@@ -2976,3 +2976,21 @@ void sai_deserialize_queue_attr(
 
     sai_deserialize_enum(s, &sai_metadata_enum_sai_queue_attr_t, (int32_t&)attr);
 }
+
+void sai_deserialize_port_serdes_attr(
+        _In_ const std::string& s,
+        _Out_ std::vector<uint32_t> &lane_values)
+{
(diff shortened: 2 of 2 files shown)


meta/saiserialize.cpp
```diff
@@ -2968,3 +2968,21 @@ void sai_deserialize_queue_attr(
 
     sai_deserialize_enum(s, &sai_metadata_enum_sai_queue_attr_t, (int32_t&)attr);
 }
+
+void sai_deserialize_media_change_ntf(
```
lguohan: which attribute? do we already have the serialize for the vector<uint32_t>?

meta/sai_serialize.h
```diff
@@ -278,4 +279,8 @@ void sai_deserialize_queue_attr(
         _In_ const std::string& s,
         _Out_ sai_queue_attr_t& attr);
 
+void sai_deserialize_media_change_ntf(
```
lguohan: I am not sure if we have such ntf? can you point out the sai function for this?

meta/saiserialize.cpp
dgsudharsan: Hi Guohan
The data is already serialized in media_settings.json in json format. Now the task is to de-serialize the data from the json format to the port attributes(SAI_PORT_ATTR_SERDES_PREEMPHASIS, SAI_PORT_ATTR_SERDES_IDRIVER, SAI_PORT_ATTR_SERDES_IPREDRIVER)

meta/sai_serialize.h
dgsudharsan: I think the name of the API might be little misleading here compared to other APIs present. The ntf stands for media notification. The intent of the API is to deserialize the notification received from xcvrd into list of SAI attributes. There are three attributes handled currently as mentioned in the above comment and in future many can be added. Thus i intended to keep the name of API as deserialize media notificaiton.
Please let me know if you want to rename the API

stevenlu99:
The change looks ok for me.

meta/saiserialize.cpp
lguohan: I do not see how this SAI serialize is related to media_settings.json. they are parsed by xrcvd, send to orchagent app db, then go to sai.

dgsudharsan:
> can you drop this PR? does not feel necessary.

Hi Guohan.. Sure. I am moving this functionality to sonic-swss
+    void sai_get_port_serdes_values (const std::string& s,
+                                                std::vector<uint32_t> &lane_values);

Issue #440, opened 2019-03-29
FDB meta reference validation error, orchagent crash


jipanyang:
meta_fdb_event_snoop_oid() is adding reference for FDB entries with objects bvid and SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID.   https://github.com/Azure/sonic-sairedis/pull/420/files

While before "status = sai_bridge_api->remove_bridge_port(port.m_bridge_port_id);"
https://github.com/Azure/sonic-swss/blob/master/orchagent/portsorch.cpp#L2609
no guarantee that all fdb entries have been flushed.  meta data validation will fail and cause orchagent to crash.

jipanyang:
A more general question,  why are we having void handle_fdb_event(_In_ const std::string &data) in ntf_thread() handling fdb meta data, and fdborch in orchagent processing fdb data.  Having them in two threads creates a lot of room for race condition. 

Could the fdb meta data handling be moved to fdborch context?

PR #441, opened 2019-03-29
Handle FDB meta data together with FDB data in fdborch to avoid race …

jipanyang:
…condition

Signed-off-by: Jipan Yang <jipan.yang@alibaba-inc.com>

Move FDB meta data data processing to fdborch.  This a first step try to solve issue https://github.com/Azure/sonic-sairedis/issues/440

### DIFF
--- a/lib/src/sai_redis_notifications.cpp
+++ b/lib/src/sai_redis_notifications.cpp
@@ -102,27 +102,29 @@ void handle_fdb_event(
 
     SWSS_LOG_DEBUG("data: %s", data.c_str());
 
-    uint32_t count;
-    sai_fdb_event_notification_data_t *fdbevent = NULL;
-
-    sai_deserialize_fdb_event_ntf(data, count, &fdbevent);
-
-    {
-        std::lock_guard<std::mutex> lock(g_apimutex);
-
-        // NOTE: this meta api must be under mutex since
-        // it will access meta DB and notification comes
-        // from different thread
-
-        meta_sai_on_fdb_event(count, fdbevent);
-    }
-
     if (sn.on_fdb_event != NULL)
     {
-        sn.on_fdb_event(count, fdbevent);
-    }
+        uint32_t count;
+        sai_fdb_event_notification_data_t *fdbevent = NULL;
 
-    sai_deserialize_free_fdb_event_ntf(count, fdbevent);
+        sai_deserialize_fdb_event_ntf(data, count, &fdbevent);
+        sn.on_fdb_event(count, fdbevent);
+        sai_deserialize_free_fdb_event_ntf(count, fdbevent);
+    }
+}
+
+void handle_meta_fdb_event(
+        _In_ uint32_t count,
+        _In_ sai_fdb_event_notification_data_t *fdbevent)
+{
+    SWSS_LOG_ENTER();
+
+    std::lock_guard<std::mutex> lock(g_apimutex);
+    // NOTE: this meta api must be under mutex since
+    // it will access meta DB and notification comes
+    // from different thread
+
+    meta_sai_on_fdb_event(count, fdbevent);
 }
 
 void handle_port_state_change(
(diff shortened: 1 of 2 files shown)


lib/src/sai_redis_notifications.cpp
```diff
@@ -102,27 +102,29 @@ void handle_fdb_event(
...
+        uint32_t count;
+        sai_fdb_event_notification_data_t *fdbevent = NULL;
+
+        sai_deserialize_fdb_event_ntf(data, count, &fdbevent);
         sn.on_fdb_event(count, fdbevent);
+        sai_deserialize_free_fdb_event_ntf(count, fdbevent);
     }
+}
+
+void handle_meta_fdb_event(
```
kcudnik: where this is called ?

kcudnik:
can you give scenario where you get race condition ?

lguohan:
issue #440

kcudnik:
you moved some functionality to handle_meta_fdb_event, but this function is not called anywhere, basically you remove meta_ call which is required, and issues in OA are caused because OA is listening on redis notifications for fdb, and not handling them in SAI FDB callback ! https://github.com/Azure/sonic-swss/blob/master/orchagent/notifications.cpp#L8

which is totally wrong !, "concurency" in the DB can be solved other way! OA should use SAI callback to handle fdb notifications, this is only valid way to handle them

lguohan:
@jipanyang , can you check the comments here?

jipanyang:
handle_meta_fdb_event() is used in https://github.com/Azure/sonic-swss/pull/824

jipanyang:
> you moved some functionality to handle_meta_fdb_event, but this function is not called anywhere, basically you remove meta_ call which is required, and issues in OA are caused because OA is listening on redis notifications for fdb, and not handling them in SAI FDB callback ! https://github.com/Azure/sonic-swss/blob/master/orchagent/notifications.cpp#L8
> 
> which is totally wrong !, "concurency" in the DB can be solved other way! OA should use SAI callback to handle fdb notifications, this is only valid way to handle them

@kcudnik Not sure if I got your point.  The current ntf_thread() is also a thread in OA, both fdborch in the main orchagent thread and ntf_thread operate on fdb meta data, there is race condition between them.

kcudnik:
there always will be race condition on 2 indepentent processes. interprocess synchronization could be used here.
fdb notification should be handled by using SAI notification callback, not by .listening to db directly.

you can use lock on ntf_callback from SAI when it will be used, but this does not guarantee anything, since more notifications can be in the redis queue waiting to be consumed by OA
because notifications are asynchronous the same issue also exist's when you would use vendor SAI directly instead of sairedis SAI. in vendor case, they jsut don't care about any reference count, bhats why we need, and this lead to finding a log of vendor bugs in the past, and prevent users writing OA to make simple mistakes like removing objects which is still in use

and again, you moved some functionality and not using using which opens possible race condition

jipanyang:
@kcudnik I think you are talking about the notification callback in syncd process.  We are not touching that part.

kcudnik:
no, im talking about notification callback in OA, take a look here: 
https://github.com/Azure/sonic-swss/blob/master/orchagent/notifications.cpp#L8

i already pointed this in my previous comment

wendani:
retest this please

jipanyang:
> no, im talking about notification callback in OA, take a look here:
> https://github.com/Azure/sonic-swss/blob/master/orchagent/notifications.cpp#L8
> 
> i already pointed this in my previous comment

The comment is correct.  While the change in meta_fdb_event_snoop_oid() simply conflicts with what is stated in the comment,  in orchangent, now you have two threads handling the FDB related meta data reference.  Please check https://github.com/Azure/sonic-sairedis/issues/440 .   

fdb is asynchronous.  Having them processed in one thread in orchagent doesn't prevent possible pending notification in syncd or the db channel, but that should not be the excuse for introducing further race condition at OA module.

kcudnik:
i already mentioned the same issue, it does not matter whether those pending notifications are in syncd or in internal vendors sai, always is issue of having them come ant any time.
please fix OA using correct notifications procedure, and please don't break existing stuff since it will create more issues

kcudnik:
closing this as obsolete, code drifted to cpp objects instead of global functions

PR #443, opened 2019-04-04
Ignore ACL_COUNTER bytes and packets during comparison logic

kcudnik:


wendani:
How about other counter-like attributes in the ASIC_DB?

kcudnik:
what are the other counter like attributes ?

PR #448, opened 2019-04-19
Add metadata serialization support for buffer pool stats

wendani:
Signed-off-by: Wenda Ni <wenni@microsoft.com>


lguohan:
retest this please

kcudnik:
i think this was already added some time ago, but since stats were converted to use stats_id then this was removed if i remember correctly

PR #452, opened 2019-04-24
[debian/rules] Change build order in target binary

marian-pritsak:
Make binary-syncd the last target, so the non-rpc requisites will be available after the first build
It will allow doing incremental builds later on for the main binary type

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -22,7 +22,7 @@ clean:
 build:
 	echo build stage is skipped. Please use binary to generate debian packages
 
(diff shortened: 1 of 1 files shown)


kcudnik:
why order here matters? for me seems like non-rpc binaries should be built first, and then rpc

marian-pritsak:
This way the non-RPC binaries will be available after the build, making it possible to use them and to do incremental build for non-RPC if needed, which I think is a more common case than for RPC binaries.

lguohan:
retest this please

PR #453, opened 2019-04-30
Modify sai_create_port for virtual switch

chiourung:
create some necessary attributes when do create port.
After do this, it can test port breakout in virtual switch.

### DIFF
--- a/vslib/src/sai_vs_port.cpp
+++ b/vslib/src/sai_vs_port.cpp
@@ -1,5 +1,6 @@
 #include "sai_vs.h"
 #include "sai_vs_internal.h"
+#include "sai_vs_state.h"
 
 sai_status_t vs_clear_port_all_stats(
         _In_ sai_object_id_t port_id)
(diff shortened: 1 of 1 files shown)


vslib/src/sai_vs_port.cpp
```diff
@@ -11,7 +12,82 @@ sai_status_t vs_clear_port_all_stats(
...
+    attr.id = SAI_PORT_ATTR_QOS_QUEUE_LIST;
+    attr.value.objlist.count = port_qos_queues_count;
+    attr.value.objlist.list = queues.data();
+
+    CHECK_STATUS(vs_generic_set(SAI_OBJECT_TYPE_PORT, *port_id, &attr));
+
+    return SAI_STATUS_SUCCESS;
+}
+
+VS_REMOVE(PORT,port);
```
lguohan: if you do all these things in the create_port, then you need also do reverse things in the remove_port.

kcudnik:
actual description of the PR is not telling why this PR is done at the first place

vslib/src/sai_vs_port.cpp
```diff
@@ -11,7 +12,82 @@ sai_status_t vs_clear_port_all_stats(
...
+    SWSS_LOG_ENTER();
+
+    sai_attribute_t attr;
+
+    /* create port */
+    CHECK_STATUS(meta_sai_create_oid((sai_object_type_t)SAI_OBJECT_TYPE_PORT,
+                port_id,switch_id,attr_count,attr_list,&vs_generic_create));
+
+    attr.id = SAI_PORT_ATTR_ADMIN_STATE;
+    attr.value.booldata = false;     /* default admin state is down as defined in SAI */
```
kcudnik: move this to each sai_vs_switch_xx.cpp

vslib/src/sai_vs_port.cpp
```diff
@@ -11,7 +12,82 @@ sai_status_t vs_clear_port_all_stats(
...
+
+    CHECK_STATUS(vs_generic_set(SAI_OBJECT_TYPE_PORT, *port_id, &attr));
+
+    /* create priority groups */
+    const uint32_t port_pgs_count = 8;
```
kcudnik: PGS are already created in VS in each class vslib/src/sai_vs_switch_xx.cpp
please remove

kcudnik:
sai_vs_port.cpp (and any other apis vs file) should only be modified to support system related stuff that can't be simulated by single attributes, example here is fdb learning or creating host interfaces, and those are independent from sai_vs_switch_xxx.cpp

vslib/src/sai_vs_port.cpp
chiourung: It just creates for default ports in sai_vs_switch_xx.cpp. For breakout port, it needs to remove port and create port dynamically. I add a test case for breakout port, and it needs this patch. I will move this code to sai_vs_switch_xx.cpp, and create new pull request. Thanks.

PR #454, opened 2019-05-13
Modify sai_create_port to breakout a port for virtual switch

chiourung:
For breakout a port, it needs to support create port and remove port.
When create a port, it needs to init some attributes.

Signed-off-by: chiourung_huang <chiourung_huang@edge-core.com>

### DIFF
--- a/vslib/src/sai_vs_port.cpp
+++ b/vslib/src/sai_vs_port.cpp
@@ -34,21 +34,7 @@ sai_status_t vs_create_port(
     return SAI_STATUS_SUCCESS;
 }
 
-sai_status_t vs_remove_port(
-            _In_ sai_object_id_t port_id)
-{
-    MUTEX();
-    SWSS_LOG_ENTER();
-
-    if (g_vs_switch_type == SAI_VS_SWITCH_TYPE_BCM56850)
-    {
-	vs_remove_port_BCM56850(port_id);
-    }
-
-    return meta_sai_remove_oid((sai_object_type_t)SAI_OBJECT_TYPE_PORT,
-                port_id,&vs_generic_remove);
-}
-
+VS_REMOVE(PORT,port);
 VS_SET(PORT,port);
 VS_GET(PORT,port);
 VS_GENERIC_QUAD(PORT_POOL,port_pool);
--- a/vslib/src/sai_vs_port.cpp
+++ b/vslib/src/sai_vs_port.cpp
@@ -1,5 +1,7 @@
 #include "sai_vs.h"
 #include "sai_vs_internal.h"
+#include "sai_vs_state.h"
+#include "sai_vs_switch_BCM56850.h"
 
 sai_status_t vs_clear_port_all_stats(
         _In_ sai_object_id_t port_id)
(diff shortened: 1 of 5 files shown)


chiourung:
The virtual switch test (Azure/sonic-swss#866) needs this patch to pass breakout port test.

vslib/src/sai_vs_switch_BCM56850.cpp
```diff
@@ -1210,3 +1210,119 @@ sai_status_t refresh_read_only_BCM56850(
 
     return SAI_STATUS_NOT_IMPLEMENTED;
 }
+
+sai_status_t vs_create_port_BCM56850(
```
kcudnik: please reuse logic that creates ports in init function, extract that code to separate function and reuse

vslib/src/sai_vs_switch_BCM56850.cpp
```diff
@@ -1210,3 +1210,119 @@ sai_status_t refresh_read_only_BCM56850(
...
+    attr.id = SAI_PORT_ATTR_QOS_QUEUE_LIST;
+    attr.value.objlist.count = port_qos_queues_count;
+    attr.value.objlist.list = queues.data();
+
+    CHECK_STATUS(vs_generic_set(SAI_OBJECT_TYPE_PORT, port_id, &attr));
+
+    return SAI_STATUS_SUCCESS;
+}
+
+sai_status_t vs_remove_port_BCM56850(
```
kcudnik: this needs very close validation what' happens here, is this reflecting reverse operation on create port that happens in vs init function ?

vslib/src/sai_vs_switch_BCM56850.cpp
```diff
@@ -1210,3 +1210,119 @@ sai_status_t refresh_read_only_BCM56850(
...
+{
+    SWSS_LOG_ENTER();
+
+    sai_attribute_t attr;
+
+    /* remove priority groups */
+    uint32_t port_pgs_count;
+    std::vector<sai_object_id_t> pgs;
+
+    attr.id = SAI_PORT_ATTR_NUMBER_OF_INGRESS_PRIORITY_GROUPS;
```
kcudnik: also, does this logic here reflects what happens in the actual BRCM SAI ?

lguohan:
@chiourung, can you revise the code based on the comments?

vslib/src/sai_vs_port.cpp
```diff
@@ -11,7 +13,30 @@ sai_status_t vs_clear_port_all_stats(
...
+            _In_ const sai_attribute_t *attr_list)
+{
+    MUTEX();
+    SWSS_LOG_ENTER();
+
+    /* create port */
+    CHECK_STATUS(meta_sai_create_oid((sai_object_type_t)SAI_OBJECT_TYPE_PORT,
+                port_id,switch_id,attr_count,attr_list,&vs_generic_create));
+
+    if (g_vs_switch_type == SAI_VS_SWITCH_TYPE_BCM56850)
```
kcudnik: what about mlnx chipset implementation ?

lguohan:
@kcudnik , anything good now? can you review?

vslib/src/sai_vs_switch_MLNX2700.cpp
```diff
@@ -350,111 +350,121 @@ static sai_status_t create_default_trap_group()
...
+    for (auto &port_id : copy)
+    {
+        create_qos_queues_per_port(switch_id, port_id);
     }
 
     return SAI_STATUS_SUCCESS;
 }
 
-static sai_status_t create_ingress_priority_groups()
+static sai_status_t create_ingress_priority_groups_per_port(sai_object_id_t switch_id, sai_object_id_t port_id)
```
kcudnik: we keep 1 param per line and adding "_in_" "_out"_ prefixes, please keep formating

lguohan:
hi kamil, instead of asking people on each and every format issue, can you add a similiar clang-format file and we can add the format test in the pr test.

https://github.com/Azure/sonic-swss/commit/9d69dd5f8c90ccaa19c9c4dadd991a564b94a329#diff-1c67ffda5f59314d6dabb50a1d45186c

kcudnik:
is this automatically taken care if we just add the file ?

lguohan:
no, you need to run the clang-format. to validate. I can add the script in the pr build, but I need you to come up with the file and first run this for the first time.

lguohan:
https://electronjs.org/docs/development/clang-format

PR #455, opened 2019-05-13
There is a version mismatch in BRCM SAI. SONIC was

stevenlu99:
- What I did
There is a version mismatch in BRCM SAI. SONIC was integrated with 3.3.4.x version of Broadcom SAI library. But, the master branch now has new features added that includes new SAI
definitions (say some of the new RIF counters APIs)

- How I did it
Added NULL checking for new API get_router_interface_stats
to avoid crash for now. This checking should be removed after SAI lib upgraded. And we need a better solution to avoid crash due to such situation.

- How to verify it
The syncd used crash after system reboot. With the code change, make sure sycnd is not crashing and running fine.

### DIFF
--- a/syncd/syncd_flex_counter.cpp
+++ b/syncd/syncd_flex_counter.cpp
@@ -976,6 +976,7 @@ void FlexCounter::collectCounters(
     // Collect stats for every registered router interface
     for (const auto &kv: rifCounterIdsMap)
     {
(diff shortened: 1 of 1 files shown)


lguohan:
the master is using SAI 3.5 which fixes this problem. We do not need this PR anymore.

PR #461, opened 2019-05-24
[debian] increment debian compatibility to 10 to enable parallel package build

stepanblyschak:


lguohan:
retest this please

liat-grozovik:
retest this please

kcudnik:
probably because of parallelism this will not pass unittests, since mlnx.pl and brcm.pl are not designed to work in parallel, they need to be executed separately, since each of those instances uses vssyncd process to perform tests, and when executed in parallel, it will conflict

lguohan:
how can we solve this issue?

kcudnik:
not easy, since single test requires a vssyncd to be present as a process, a clean redis db and so on,
we can run each test in separate docker XD, then it should work

configure.ac
```diff
@@ -3,7 +3,7 @@ AC_CONFIG_MACRO_DIR([m4])
 AC_CONFIG_SRCDIR([])
 AC_CONFIG_AUX_DIR(config)
 AM_CONFIG_HEADER(config.h)
-AM_INIT_AUTOMAKE([subdir-objects foreign])
+AM_INIT_AUTOMAKE([subdir-objects foreign serial-tests])
```
kcudnik: great

PR #462, opened 2019-05-25
Full sleep wait flex counter polling thread when POLL_COUNTER_STATUS is disable

wendani:
In the current implementation, when  FLEX_COUNTER_STATUS is set to disable, which can be issued by counterpoll disable cli, a flex counter thread will still wake up every polling interval, though skipping the actual collectCounters() and runPlugins() operations. 

The main drive of the PR is to put the polling thread to a complete sleep in the above situation to save CPU cycles, and hence to improve the scalability of flex counter polling architecture in general.

To that end, the following changes are made to pave the way and for further optimization purpose:

1\) Introduce a per-thread lock, together with a condition variable, to enable the full sleep behavior when FLEX_COUNTER_STATUS is set to disable, the status of which is maintained by data member m_enable per flex counter object. 

2\) Remove acquiring the global lock g_mutex to configure a flex counter object in the syncd main thread context. This is because each flex counter object maintains its own configuration setting, counter/attribute id list, plugin list, etc. to be accessed only in either the syncd main thread context or its own flex counter polling thread context.

With the introduction of per-thread lock, the scope of the data access serialization can be reduced from the unnecessarily global wise, among all flex counter polling threads and the syncd main thread, to per flex counter object wise, between only the syncd main thread and the corresponding flex counter polling thread.

3\) Remove acquiring the global lock g_mutex and remove copying the plugin list and/or the counter/attribute id list in the flex counter polling thread context as the acquisition of the per-thread lock before and during calling collectCounters() and runPlugins() ensures the correct synchronization to data member access.

4\) Change the memory management entity of the thread object from shared_ptr to unique_ptr in the flex counter object to avoid its join() function to be called more than once in the destruction phase. Because each thread can only be joined once, it does not make sense to use shared_ptr to manage the thread object memory, but merely creates the opportunity to have join() function being called multiple times if the thread object happens to be referenced by multiple shared_ptr objects.

5\) Change m_runFlexCounterThread from atomic_bool to a regular bool type.

Validation:

counterpoll watermark enable

May 26 23:38:04.968123 str-dx010-acs-1 ERR swss#orchagent: :- generateBufferPoolWatermarkCounterIdList: generateBufferPoolCounterIdList: Buffer pool COUNTER_ID_LIST already generated
...
May 26 23:38:07.129572 str-dx010-acs-1 ERR syncd#syncd: brcm_sai_get_buffer_pool_stats:718 Ingress buffer pool 0x1800000001: bst value: 416
May 26 23:38:07.134184 str-dx010-acs-1 ERR syncd#syncd: :- flexCounterThread: End of flex counter thread FC BUFFER_POOL_WATERMARK_STAT_COUNTER, took 47 ms
May 26 23:38:07.641090 str-dx010-acs-1 ERR syncd#syncd: :- flexCounterThread: End of flex counter thread FC QUEUE_WATERMARK_STAT_COUNTER, took 702 ms
...
May 26 23:38:08.452268 str-dx010-acs-1 ERR syncd#syncd: :- flexCounterThread: End of flex counter thread FC PG_WATERMARK_STAT_COUNTER, took 1512 ms
...
May 26 23:38:17.094576 str-dx010-acs-1 ERR syncd#syncd: brcm_sai_get_buffer_pool_stats:718 Ingress buffer pool 0x1800000001: bst value: 208
May 26 23:38:17.097077 str-dx010-acs-1 ERR syncd#syncd: :- flexCounterThread: End of flex counter thread FC BUFFER_POOL_WATERMARK_STAT_COUNTER, took 9 ms
...
May 26 23:38:17.765136 str-dx010-acs-1 ERR syncd#syncd: :- flexCounterThread: End of flex counter thread FC QUEUE_WATERMARK_STAT_COUNTER, took 825 ms
May 26 23:38:18.413790 str-dx010-acs-1 ERR syncd#syncd: :- flexCounterThread: End of flex counter thread FC PG_WATERMARK_STAT_COUNTER, took 1473 ms
...

counterpoll watermark disable
May 26 23:38:26.940287 str-dx010-acs-1 ERR syncd#syncd: :- flexCounterThread: Flex counter thread FC QUEUE_WATERMARK_STAT_COUNTER, condition wait
May 26 23:38:26.940802 str-dx010-acs-1 ERR syncd#syncd: :- flexCounterThread: Flex counter thread FC PG_WATERMARK_STAT_COUNTER, condition wait
May 26 23:38:27.088245 str-dx010-acs-1 ERR syncd#syncd: :- flexCounterThread: Flex counter thread FC BUFFER_POOL_WATERMARK_STAT_COUNTER, condition wait
...

counterpoll watermark enable
May 26 23:43:07.336258 str-dx010-acs-1 ERR swss#orchagent: :- generateBufferPoolWatermarkCounterIdList: generateBufferPoolCounterIdList: Buffer pool COUNTER_ID_LIST already generated
May 26 23:43:07.564374 str-dx010-acs-1 ERR syncd#syncd: brcm_sai_get_buffer_pool_stats:718 Ingress buffer pool 0x1800000001: bst value: 416
May 26 23:43:07.567735 str-dx010-acs-1 ERR syncd#syncd: :- flexCounterThread: End of flex counter thread FC BUFFER_POOL_WATERMARK_STAT_COUNTER, took 280479 ms
...
May 26 23:43:07.960222 str-dx010-acs-1 ERR syncd#syncd: :- flexCounterThread: End of flex counter thread FC QUEUE_WATERMARK_STAT_COUNTER, took 281019 ms
May 26 23:43:08.776857 str-dx010-acs-1 ERR syncd#syncd: :- flexCounterThread: End of flex counter thread FC PG_WATERMARK_STAT_COUNTER, took 281836 ms
...
May 26 23:43:17.157729 str-dx010-acs-1 ERR syncd#syncd: brcm_sai_get_buffer_pool_stats:718 Ingress buffer pool 0x1800000001: bst value: 208
May 26 23:43:17.163423 str-dx010-acs-1 ERR syncd#syncd: :- flexCounterThread: End of flex counter thread FC BUFFER_POOL_WATERMARK_STAT_COUNTER, took 74 ms
May 26 23:43:17.732528 str-dx010-acs-1 ERR syncd#syncd: :- flexCounterThread: End of flex counter thread FC QUEUE_WATERMARK_STAT_COUNTER, took 791 ms
...
May 26 23:43:18.381651 str-dx010-acs-1 ERR syncd#syncd: :- flexCounterThread: End of flex counter thread FC PG_WATERMARK_STAT_COUNTER, took 1440 ms

### DIFF
--- a/syncd/syncd.cpp
+++ b/syncd/syncd.cpp
@@ -2783,7 +2783,10 @@ void processFlexCounterGroupEvent(
     SWSS_LOG_ENTER();
 
     swss::KeyOpFieldsValuesTuple kco;
-    consumer.pop(kco);
+    {
+        std::lock_guard<std::mutex> lock(g_mutex);
+        consumer.pop(kco);
+    }
 
     const auto &groupName = kfvKey(kco);
     const auto &op = kfvOp(kco);
@@ -2860,7 +2863,10 @@ void processFlexCounterEvent(
     SWSS_LOG_ENTER();
 
     swss::KeyOpFieldsValuesTuple kco;
-    consumer.pop(kco);
+    {
+        std::lock_guard<std::mutex> lock(g_mutex);
+        consumer.pop(kco);
+    }
 
     const auto &key = kfvKey(kco);
     const auto &op = kfvOp(kco);
--- a/syncd/syncd.cpp
+++ b/syncd/syncd.cpp
@@ -2782,7 +2782,6 @@ void processFlexCounterGroupEvent(
 {
     SWSS_LOG_ENTER();
 
-    std::lock_guard<std::mutex> lock(g_mutex);
     swss::KeyOpFieldsValuesTuple kco;
     consumer.pop(kco);
 
@@ -2851,8 +2850,6 @@ void processFlexCounterEvent(
         _In_ swss::ConsumerTable &consumer)
 {
     SWSS_LOG_ENTER();
-
-    std::lock_guard<std::mutex> lock(g_mutex);
 
     swss::KeyOpFieldsValuesTuple kco;
     consumer.pop(kco);
--- a/syncd/syncd_flex_counter.cpp
+++ b/syncd/syncd_flex_counter.cpp
@@ -383,7 +383,7 @@ void FlexCounter::setPriorityGroupAttrList(
 
     FlexCounter &fc = getInstance(instanceId);
 
-    std::unique_lock<std::mutex> lkMgr(fc.m_mtx);
+    std::lock_guard<std::mutex> lkMgr(fc.m_mtx);
 
     auto it = fc.m_priorityGroupAttrIdsMap.find(priorityGroupVid);
     if (it != fc.m_priorityGroupAttrIdsMap.end())
@@ -494,6 +494,8 @@ void FlexCounter::setBufferPoolCounterList(
 
         return;
     }
+
+    std::lock_guard<std::mutex> lkMgr(fc.m_mtx);
 
     auto it = fc.m_bufferPoolCounterIdsMap.find(bufferPoolVid);
     if (it != fc.m_bufferPoolCounterIdsMap.end())
@@ -665,6 +667,8 @@ void FlexCounter::removeBufferPool(
     bool found = false;
     FlexCounter &fc = getInstance(instanceId);
 
+    std::unique_lock<std::mutex> lkMgr(fc.m_mtx);
+
     auto it = fc.m_bufferPoolCounterIdsMap.find(bufferPoolVid);
     if (it != fc.m_bufferPoolCounterIdsMap.end())
     {
@@ -681,6 +685,7 @@ void FlexCounter::removeBufferPool(
     // Remove flex counter if all counter IDs and plugins are unregistered
     if (fc.isEmpty())
     {
+        lkMgr.unlock();
         removeInstance(instanceId);
     }
 }
@@ -755,6 +760,8 @@ void FlexCounter::addBufferPoolCounterPlugin(
     SWSS_LOG_ENTER();
 
     FlexCounter &fc = getInstance(instanceId);
+
+    std::lock_guard<std::mutex> lkMgr(fc.m_mtx);
 
     if (fc.m_bufferPoolPlugins.find(sha) != fc.m_bufferPoolPlugins.end())
     {
@@ -1169,7 +1176,7 @@ void FlexCounter::collectCounters(
     }
 
     // Collect stats for every registered buffer pool
-    for (const auto &it : bufferPoolCounterIdsMap)
+    for (const auto &it : m_bufferPoolCounterIdsMap)
     {
         const auto &bufferPoolVid = it.first;
         const auto &bufferPoolId = it.second->bufferPoolId;
@@ -1272,12 +1279,12 @@ void FlexCounter::runPlugins(
     }
 
     std::vector<std::string> bufferPoolVids;
-    bufferPoolVids.reserve(bufferPoolCounterIdsMap.size());
-    for (const auto& it : bufferPoolCounterIdsMap)
+    bufferPoolVids.reserve(m_bufferPoolCounterIdsMap.size());
+    for (const auto& it : m_bufferPoolCounterIdsMap)
     {
         bufferPoolVids.push_back(sai_serialize_object_id(it.first));
     }
-    for (const auto& sha : bufferPoolPlugins)
+    for (const auto& sha : m_bufferPoolPlugins)
     {
         runRedisScript(db, sha, bufferPoolVids, argv);
     }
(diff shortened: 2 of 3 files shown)


lguohan:
can you resolve the conflict? 

also please have a more description on what changes you add and why.

syncd/syncd_flex_counter.cpp
```diff
@@ -982,8 +1015,9 @@ void FlexCounter::collectCounters(
 
         countersTable.set(priorityGroupVidStr, values, "");
     }
+
     // Collect stats for every registered router interface
-    for (const auto &kv: rifCounterIdsMap)
+    for (const auto &kv: m_rifCounterIdsMap)
```
lguohan: what is this fix? should we separate into one pr?

wendani: No, it is in the scope of pr only. Because locally I changed and validated aginast the 1130 branch, which does not include the rif counters. When cherry-picking to master, I need to take additional care exclusively for rif counters.

https://github.com/Azure/sonic-sairedis/pull/462#discussion_r287582193

syncd/syncd.cpp
```diff
@@ -2782,7 +2782,6 @@ void processFlexCounterGroupEvent(
 {
     SWSS_LOG_ENTER();
 
-    std::lock_guard<std::mutex> lock(g_mutex);
```
kcudnik: can''t be removed, consumer must be under lock

syncd/syncd.cpp
```diff
@@ -2860,8 +2859,6 @@ void processFlexCounterEvent(
...
```
kcudnik: can''t be removed, consumer must be under lock

syncd/syncd_flex_counter.cpp
```diff
@@ -587,6 +620,7 @@ void FlexCounter::removePriorityGroup(
     // Remove flex counter if all counter IDs and plugins are unregistered
     if (fc.isEmpty())
     {
+        lkMgr.unlock();
```
kcudnik: no no no .. dont' use manual unlock, we have RAII i c++ to take care of that, you will definetly lost track at some point where to unlock/lock

syncd/syncd_flex_counter.h
```diff
@@ -206,11 +206,13 @@ class FlexCounter
         std::set<std::string> m_priorityGroupPlugins;
         std::set<std::string> m_bufferPoolPlugins;
 
-        std::atomic_bool m_runFlexCounterThread = { false };
-        std::shared_ptr<std::thread> m_flexCounterThread = nullptr;
+        bool m_runFlexCounterThread = false;
+        std::unique_ptr<std::thread> m_flexCounterThread = nullptr;
```
kcudnik: why unique ? this is not needed

syncd/syncd.cpp
wendani: Is it because of the processNotification thread in syncd/syncd_notifications.cpp?

syncd/syncd.cpp
wendani: Is it because of the processNotification thread in syncd/syncd_notifications.cpp?

syncd/syncd_flex_counter.cpp
wendani: The removeInstance() that follows will erase() the FlexCounter object from the map, which has the possibility to call ~FlexCounter() destructor, which will call endFlexCounterThread(), so the syncd main thread has to release the lock so that the flex counter thread can be waken up with the lock in hold to return and join() to proceed endFlexCounterThread() to return.

syncd/syncd_flex_counter.h
wendani: From the FlexCounter object perspective, if it is destroyed, it will call endFlexCounterThread() to join() the thread object. If the thread object is referenced by multiple FlexCounter objects, join() can be called multiple times on each FlexCounter destruction. As a thread object can only be joined once, I think we should eliminate such a possibility. From the memory management perspective, it makes no difference using either shared_ptr or unique_ptr

kcudnik: are there multiple threads per counters? problem with unique pointer is that it can be hard to track where actual pointer is held if something wrong will happen

wendani: Changed to shared_ptr

syncd/syncd_flex_counter.cpp
```diff
@@ -1311,36 +1337,42 @@ void FlexCounter::startFlexCounterThread(void)
 {
     SWSS_LOG_ENTER();
 
-    if (m_runFlexCounterThread.load() == true)
+    if (m_runFlexCounterThread == true)
     {
+        m_pollCond.notify_all();
```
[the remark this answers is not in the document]
wendani: When we do counterpoll enable cli, "FLEX_COUNTER_STATUS" is changed from "disable" to "enable"; we will get notification to start the not-running thread. 

If the polling thread is currently running. This notify_all() will not have any effect on the execution flow of the polling thread. Is this your concern?

PR #464, opened 2019-06-03
Check if port VID exists in db on flex counter query

kcudnik:


syncd/syncd.cpp
```diff
@@ -2880,8 +2898,17 @@ void processFlexCounterEvent(
...
-    sai_object_type_t objectType = sai_object_type_query(rid);
+    sai_object_id_t rid;
+
+    if (!try_translate_vid_to_rid(vid, rid))
+    {
+        SWSS_LOG_WARN("port VID %s, was not found (probably port was removed/splitted) and will remove from counters now", 
+                sai_serialize_object_id(vid).c_str());
+
+        op = DEL_COMMAND;
+    }
```
lguohan: will there be another case, where flex counter request come first and by that time, the port vid has not been added to the mapping table, which end up with the case, the flex counter is not added at all?

kcudnik: yes, this is possible, but no harm here, since FlexCounter::removeX() functios checks internally wheter vid is in the map, and only if then remove it

jipanyang:
There might still be issue here:
<1> at intforch,  rif flex counter is removed first, then rif:
https://github.com/Azure/sonic-swss/blob/master/orchagent/intfsorch.cpp#L563
https://github.com/Azure/sonic-swss/blob/master/orchagent/intfsorch.cpp#L565

but rif operation goes through asicDB channel,  flex counter op uses flexcounterDB,  no receiving order at syncd is guaranteed.

<2> rif vid/rid mapping is removed in syncd with rif delete request,  flex counter removal to be handled if the operation request comes later from flexcounterDB channel.

<3> flexCounterThread as a separate thread may keep collecting counter data for that RIF though it has been deleted.

Did I miss anything here?

kcudnik:
i think it will be no collecting counters, since if rif ws removed, queued counters are also cleared,  look here:
https://github.com/Azure/sonic-sairedis/pull/464/files#diff-91b3e380b7f6f4a0a048ca8296a12e0dR2908
i dont know exact internals of collecting counters but this seems ok to me

jipanyang:
I see.  Having some stale counters in the counterIdsMap temporarily probably is not fatal with this change, but just results in some error log or stale VID counters inside counterDB (due to the asynchronous processing in flexCounterThread).

syncd/syncd.cpp
jipanyang: I think the original question remains unanswered:  "which end up with the case, the flex counter is `not added at all`?"   

Or we treat it as a compromise?  Maybe it is time to unify flexcounter DB channel with asicDB channel to avoid the race condition?

kcudnik:
not following, what stale counters? they will be removed if vid is not found

syncd/syncd.cpp
kaoyusean: I ran into an issue that caused Orchagent to crash as follows:
1. swss added an ACL rule, and an associated new flex counter was passed to syncd to be added.
2. Flex counter somehow moved faster than the ACL rule, so sycnd decided that instead of adding this flex counter without finding the ACL rule sitting in the database, syncd did nothing, which is the intended behavior of this code change. 
3. Orchagent finished adding this ACL rule and thought that everything is fine.
4. Later on orchagent tried to look up the flex counter value but syncd reported that there is nothing there.
5. Orchagent then did abort().

PR #465, opened 2019-06-04
Full sleep wait change for PFC watchdog

wendani:
When issuing 'pfcwd stop' cli to stop PFC watchdog, it only clears the counter id list for physical ports under the hood, leaving lua script plugins still installed and ieEmpty() returns false.

So even when PFC watchdog is stopped, the flex counter thread would wake up every 200 ms.

Jun  4 17:58:51.273413 str-a7050-acs-1 ERR syncd#syncd: :- flexCounterThread: End of flex counter thread FC PFC_WD, took 33 ms
Jun  4 17:58:51.471288 str-a7050-acs-1 ERR syncd#syncd: :- flexCounterThread: End of flex counter thread FC PFC_WD, took 30 ms
Jun  4 17:58:51.672930 str-a7050-acs-1 ERR syncd#syncd: :- flexCounterThread: End of flex counter thread FC PFC_WD, took 31 ms
Jun  4 17:58:51.882972 str-a7050-acs-1 ERR syncd#syncd: :- flexCounterThread: End of flex counter thread FC PFC_WD, took 40 ms
Jun  4 17:58:52.074862 str-a7050-acs-1 ERR syncd#syncd: :- flexCounterThread: End of flex counter thread FC PFC_WD, took 31 ms

This PR is to address the above observation:

Fine granularize the isEmpty() to two functions---isIdsEmpty() and isPluginsEmpty(), and use the emptiness check of the counter id list, isIdsEmpty(), as the criteria for full sleep wait. This approach is generally valid because current flex counter use cases would have the presence of non-empty counter id list to be meaningful, followed by an optional execution of the Lua script plugins if they are installed.

Validated on brcm dut

Signed-off-by: Wenda Ni <wenni@microsoft.com>

### DIFF
--- a/syncd/syncd_flex_counter.cpp
+++ b/syncd/syncd_flex_counter.cpp
@@ -836,16 +836,29 @@ bool FlexCounter::isEmpty()
 {
     SWSS_LOG_ENTER();
 
+    return isIdsEmpty() && isPluginsEmpty();
+}
+
+bool FlexCounter::isIdsEmpty()
+{
+    SWSS_LOG_ENTER();
+
     return m_priorityGroupCounterIdsMap.empty() &&
            m_priorityGroupAttrIdsMap.empty() &&
-           m_priorityGroupPlugins.empty() &&
            m_queueCounterIdsMap.empty() &&
+           m_queueAttrIdsMap.empty() &&
            m_portCounterIdsMap.empty() &&
            m_rifCounterIdsMap.empty() &&
-           m_queueAttrIdsMap.empty() &&
+           m_bufferPoolCounterIdsMap.empty();
+}
+
+bool FlexCounter::isPluginsEmpty()
+{
+    SWSS_LOG_ENTER();
+
+    return m_priorityGroupPlugins.empty() &&
            m_queuePlugins.empty() &&
            m_portPlugins.empty() &&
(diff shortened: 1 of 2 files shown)


syncd/syncd_flex_counter.cpp
```diff
@@ -836,16 +836,29 @@ bool FlexCounter::isEmpty()
 {
     SWSS_LOG_ENTER();
 
+    return isIdsEmpty() && isPluginsEmpty();
+}
+
+bool FlexCounter::isIdsEmpty()
```
kcudnik: areIdsEmpty ? since multipe ids maps we check here ?

wendani:
retest this please

wendani:
retest this please

syncd/syncd_flex_counter.cpp
jleveque: Agree with @kcudnik. Same for the plugins. Since the functions return true if **all** IDs/plugins are empty, I suggest `allIdsEmpty()` and `allPluginsEmpty()`

wendani:
retest this please

wendani:
retest this please

wendani:
retest this please

wendani:
retest this please

wendani:
retest this please

wendani:
retest this please

PR #468, opened 2019-06-11
ARM32 bit fixes, for 64bit printf format specifier

antony-rheneus:
Signed-off-by: Antony Rheneus <arheneus@marvell.com>

### DIFF
--- a/lib/src/sai_redis_generic_create.cpp
+++ b/lib/src/sai_redis_generic_create.cpp
@@ -1,6 +1,7 @@
 #include "sai_redis.h"
 #include "meta/sai_serialize.h"
 #include "meta/saiattributelist.h"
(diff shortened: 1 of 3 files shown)


kcudnik:
why arm is not accepting standard prefixes ? this actually makes code much more messy :/

jleveque:
@kcudnik: The "PRI..." constants which get defined differently depending on the architecture you're compiling for, thus making the code more portable. Some archs define a 64-bit int as a `long`, some as a `long long`. Withe this change, we never have to worry about the underlying type.

kcudnik:
yea grepped via headers, more portable less readability  XD

jleveque:
Yes, unfortunately we sacrifice readability for portability.

PR #469, opened 2019-06-13
Add TimerWatchdog for monitoring long execution apis

kcudnik:


lguohan:
the new file name (TimeWatchdog.{h, cpp}) is not consistent with existing file names?

kcudnik:
Yes, it not follow current convention, but i took this opportunity to move to a different scheme:
- one class per file
- files names as classes names with Capital letters
- files with classes not prefixed as syncd_*.cpp

i will be doing this refactoring as new changes will be added to syncd/sairedis
this will make more sense in long run and it should keep the project more clean
especially if we plan to support multiple asics in the future, then everything related to single unit switch will be encloses to 1 class representing this unit

PR #470, opened 2019-06-17
Full sleep wait flex counter polling thread when POLL_COUNTER_STATUS is disable

wendani:
A PR tailored for 201811 based on https://github.com/Azure/sonic-sairedis/pull/462   

In 201811, we do not have buffer pool watermark and rif counters. So this version resolves conflicts raised herein when backporting to 201811.

### DIFF
--- a/syncd/syncd.cpp
+++ b/syncd/syncd.cpp
@@ -2782,7 +2782,6 @@ void processFlexCounterGroupEvent(
 {
     SWSS_LOG_ENTER();
 
-    std::lock_guard<std::mutex> lock(g_mutex);
     swss::KeyOpFieldsValuesTuple kco;
     consumer.pop(kco);
 
@@ -2851,8 +2850,6 @@ void processFlexCounterEvent(
         _In_ swss::ConsumerTable &consumer)
 {
     SWSS_LOG_ENTER();
-
-    std::lock_guard<std::mutex> lock(g_mutex);
 
     swss::KeyOpFieldsValuesTuple kco;
     consumer.pop(kco);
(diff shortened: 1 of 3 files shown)


syncd/syncd_flex_counter.cpp
```diff
@@ -548,13 +587,16 @@ void FlexCounter::removeCounterPlugin(
...
+    std::unique_lock<std::mutex> lkMgr(fc.m_mtx);
+
     fc.m_queuePlugins.clear();
     fc.m_portPlugins.clear();
     fc.m_priorityGroupPlugins.clear();
 
     // Remove flex counter if all counter IDs and plugins are unregistered
     if (fc.isEmpty())
     {
+        lkMgr.unlock();
```
kcudnik: please don't do manual unlocking, this is really bad design, use scope lock/unlock

syncd/syncd.cpp
```diff
@@ -2782,9 +2782,11 @@ void processFlexCounterGroupEvent(
 {
     SWSS_LOG_ENTER();
 
-    std::lock_guard<std::mutex> lock(g_mutex);
     swss::KeyOpFieldsValuesTuple kco;
-    consumer.pop(kco);
+    {
```
kcudnik: why you moved this here? please revert this change, since g_mutex is not protecting only consumer.pop but other variables too!!

wendani:
Discussed offline with Renuka. The problem is in the signal handling path. Not the normal path that we have thought of.

kcudnik:
there is no signal handling in syncd, also there is no way to handle sigkill

renukamanavalan:
SIGKILL can't be handled in any case.
 The easiest fix could be, add an exit handler and all that the exit handler does is call abort()

PR #472, opened 2019-06-19
Reduce the timeout (GET_RESPONSE_TIMEOUT) from 6 minutes to 1 minute.

renukamanavalan:


lib/inc/sai_redis.h
```diff
@@ -43,7 +43,7 @@ void check_notifications_pointers(
 
 // if we don't receive response from syncd in 60 seconds
 // there is something wrong and we should fail
-#define GET_RESPONSE_TIMEOUT (6*60*1000)
```
kcudnik: just fyi i remember that this was changed on purpose but i cant remember why ;/

renukamanavalan: Got it. Let us face it again. If we face, I would prefer to fix the other end.

yxieca:
Removed from 201811 branch until further validation is done.

PR #473, opened 2019-06-21
Add default bridge id for bridge port id of type PORT in virtual switch

kcudnik:


vslib/src/sai_vs_switch_BCM56850.cpp
```diff
@@ -436,9 +438,17 @@ static sai_status_t create_qos_queues_per_port(
         CHECK_STATUS(vs_generic_create(SAI_OBJECT_TYPE_QUEUE, &queue_id, switch_object_id, 0, NULL));
 
         queues.push_back(queue_id);
-    }
 
-    sai_attribute_t attr;
+        attr.id = SAI_QUEUE_ATTR_TYPE;
+        attr.value.s32 = SAI_QUEUE_TYPE_ALL;
```
lguohan: >SAI_QUEUE_TYPE_ALL [](start = 25, length = 18)

why type all. Usually, the first 10 are type unicast and the latter 10 are type multicast.

kcudnik:
sure will update

vslib/src/sai_vs_switch_BCM56850.cpp
kcudnik: done

kcudnik:
test this please

PR #476, opened 2019-06-24
Make sairedis/syncd synchronous

kcudnik:


lib/src/sai_redis_generic_get.cpp
```diff
@@ -205,7 +205,7 @@ sai_status_t internal_redis_generic_get(
             const std::string &op = kfvOp(kco);
             const std::string &opkey = kfvKey(kco);
 
-            SWSS_LOG_DEBUG("response: op = %s, key = %s", opkey.c_str(), op.c_str());
+            SWSS_LOG_INFO("response: op = %s, key = %s", opkey.c_str(), op.c_str());
```
qiluo-msft: too verbose?

lib/src/sai_redis_generic_remove.cpp
```diff
@@ -118,10 +118,10 @@ sai_status_t internal_redis_bulk_generic_remove(
         }
 
         /*
-         * Capital 'C' stands for bulk CREATE operation.
+         * Capital 'R' stands for bulk CREATE operation.
          */
 
-        recordLine("C|" + str_object_type + joined);
+        recordLine("R|" + str_object_type + joined);
```
qiluo-msft: >R [](start = 20, length = 1)

Is there a wiki page to summarize all the special char used in recording? If not, please create one.

Alternatively, how about centralized the definition in one header file and use constants or macros in code?

lib/src/sai_redis_switch.cpp
```diff
@@ -267,7 +267,7 @@ sai_status_t redis_set_switch_attribute(
                 return SAI_STATUS_SUCCESS;
 
             case SAI_REDIS_SWITCH_ATTR_USE_PIPELINE:
-                g_asicState->setBuffered(attr->value.booldata);
+                // g_asicState->setBuffered(attr->value.booldata); // this is ignored in sync mode
                 return SAI_STATUS_SUCCESS;
```
qiluo-msft: >SAI_STATUS_SUCCESS [](start = 23, length = 18)

SAI_STATUS_NOT_SUPPORTED ?

meta/saiserialize.cpp
```diff
@@ -1,7 +1,11 @@
 #include "sai_serialize.h"
 #include "meta/sai_meta.h"
 #include "swss/tokenize.h"
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wshadow"
 #include "swss/json.hpp"
```
qiluo-msft: Add pragma into header file directly so others could also benefit?

lib/src/sai_redis_generic_remove.cpp
jipanyang: Could we set such a rule that,  small letter means request, capital letter is for response? 
 Ex:  
```
g/G(get request/Get response), 
c/C (create request/create reponse).  
r/R(remove request/Remove response).  
s/S(set request/Set reponse)
b/B(buld create/Bulk create response)
e/E(bulk remove request/Bulk remove response)
m/M(bulk set request/Buld set reponse)
```

lib/src/sai_redis_generic_get.cpp
kcudnik: maybe not, normally its a NOTICE anyway

lib/src/sai_redis_generic_remove.cpp
kcudnik: actually m/M should be removed since it's not SAI api call, its's request for counters subscription now, and recording should log only sai call's
also single 'b' is not good enough since you can have bulk remove, bulk create and so on

lib/src/sai_redis_switch.cpp
kcudnik: yes

meta/saiserialize.cpp
kcudnik: this is in swss-common repo, also new update to json would need to keep this but sure i can add

kcudnik:
@jipanyang  yes i have performance, send in email

vslib/src/sai_vs_hostintf.cpp
```diff
@@ -21,7 +21,11 @@
 #include <net/if_arp.h>
 #include <linux/if_ether.h>
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wshadow"
```
stcheng: it's separated from this PR;
meanwhile, i'm wondering if we want to upgrade the version of this `json.hpp` file to a newer one.

kcudnik: if i remember there were some compiling issues with newer one thats why we stick with this one but you can take a look, im adding this since gcc 9.1 is complaining

meta/saiserialize.cpp
qiluo-msft: any follow up?

kcudnik: working on it

kcudnik: https://github.com/Azure/sonic-swss-common/pull/292

lguohan:
can you make this optional so that we do not need to turn it on by default for now?

kcudnik:
Do you mean compile time optional or dynamic optional ?
notice that there are 2 independent components changed there syncd and sairedis
so if dynamic on/off then they would need to be switched independently and both of them in syncd and OA

syncd/syncd.cpp
```diff
@@ -3150,6 +3194,8 @@ void printUsage()
     std::cout << "    -m --portmap"             << std::endl;
     std::cout << "        Specify port map file" << std::endl;
 #endif // SAITHRIFT
+    std::cout << "    -s --syncMode"             << std::endl;
```
lguohan: if we have option in SAI redis to pass syncmode from orchagent, why do you add option in syncd to enable/disable syncd mode?

kcudnik: as mentioned those are 2 independent modules, you need to enable sync mode in both of them independently

lib/src/sai_redis_generic_remove.cpp
qiluo-msft: Thanks @jipanyang ! I still suggest to maintain it as any of the below options:
1. code comment or Github wiki page and referred by code comment.
2. centralized the definition in one header file and use constants/macros/enums in code?

Nowadays it is hard to find, and easy to conflict.

PR #480, opened 2019-06-25
[syncd_init_common.sh] fix fast reboot backwards compatibility

stepanblyschak:
We should handle both cases for backward-compatible with 201803:
 - fast-reboot
 - SONIC_BOOT_TYPE=fast-reboot

Signed-off-by: Stepan Blyschak <stepanb@mellanox.com>

### DIFF
--- a/syncd/scripts/syncd_init_common.sh
+++ b/syncd/scripts/syncd_init_common.sh
@@ -24,16 +24,14 @@ fi
 # Use temporary view between init and apply
 CMD_ARGS+=" -u"
 
-BOOT_TYPE="$(cat /proc/cmdline | grep -o 'SONIC_BOOT_TYPE=\S*' | cut -d'=' -f2)"
-
-case "$BOOT_TYPE" in
-  fast-reboot)
-     FAST_REBOOT='yes'
-    ;;
-  fastfast)
+case "$(cat /proc/cmdline)" in
+  *fastfast*)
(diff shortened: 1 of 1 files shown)


syncd/scripts/syncd_init_common.sh
```diff
@@ -24,17 +24,15 @@ fi
...
 
-BOOT_TYPE="$(cat /proc/cmdline | grep -o 'SONIC_BOOT_TYPE=\S*' | cut -d'=' -f2)"
-
-case "$BOOT_TYPE" in
-  fast-reboot)
-     FAST_REBOOT='yes'
-    ;;
-  fastfast)
+case "$(cat /proc/cmdline)" in
+  *fastfast*)
```
qiluo-msft: >fastfast [](start = 3, length = 8)

Be more specific `SONIC_BOOT_TYPE=fastfast`

syncd/scripts/syncd_init_common.sh
```diff
@@ -24,17 +24,15 @@ fi
...
     if [ -e /var/warmboot/warm-starting ]; then
         FASTFAST_REBOOT='yes'
     fi
     ;;
+  *fast*)
```
qiluo-msft: >fast [](start = 3, length = 4)

This one is also too general. We can match for 2 patterns:

1. `*SONIC_BOOT_TYPE=fast*`
2. `*fast-reboot*`

PR #483, opened 2019-07-01
Process flex counters requests in separate thread

kcudnik:


lguohan:
@wendani , can you take a look?

syncd/syncd.cpp
```diff
@@ -2954,25 +2957,98 @@ void processFlexCounterGroupEvent(
...
+    {
+        swss::KeyOpFieldsValuesTuple kco;
+
+        if (tryPopFlexCounterEvent(kco))
+        {
+            if (!processFlexCounterEvent(kco))
+            {
+                // event was not successfully processed, put it again to the queue
+
+                pushFlexCounterEvent(kco);
```
wendani: This may upset the event sequence in the processing failure case. The processing failure case, however, is the problem you would like to solve.

For the flex counter table in particular, it typically contains the counter id list that is expected to remain static and unchanged after SET. We should not see any two operations on the same key in the queue at the same time, although there leaves the possibility of running into such a sequence problem.

If we enqueue at the front and retry, we can avoid the event sequence problem. However, we may experience head of line blocking until the event processing retry succeeds.

syncd/syncd.cpp
```diff
@@ -2954,25 +2957,98 @@ void processFlexCounterGroupEvent(
...
+            }
+        }
+
+        sleep(1);
```
wendani: Can we wake up every 1 second only if there is a processing event failure?  In other cases, we wake up on an enqueue event. Flex counter table event should be very rare after the initialization.

kcudnik: this is actually overkill to make other thread and condition variable to wait for specific event, also making that event, also don't guarantee that this event will be processed after second select will create rif port, so there must be wait somewhere anyway

syncd/syncd.cpp
kcudnik: i don't get the problem you described here, all those issues with counters are present here which we want to workaround because they are executed on the same channel and can be picked up by select  independently, and i stated at the beginning that subscribe to the counter is broken design and of course you can came up with scenario that can break even this workaround, since you need to wait for RIF to be created and for example potentially not destroyed and created again, which all would be solved when syncd would become synchronous on which i already send  PR

lguohan:
@kcudnik , can you resolve the conflict?

PR #488, opened 2019-07-18
Use pointer to member function to reduce the footprint of collectCounters()

wendani:
Each time a new counter/attribute type is added to flex counter infrastructure, collectCounters() is tailed with the logic of collecting the stats for the new type. As more and more possible id lists are being added to the flex counter, the collectCounters() member function is becoming bulky to check against all possible id lists, both existing and newly added ones.

Current list is already significantly long---port counters, queue counters, queue attributes, pg counters, pg attributes, and buffer pool counters. And most of them are irrelevant to a particular flex counter instance thread. E.g., For buffer pool watermark thread, only buffer pool counters matter, not the rest, port counters, queue counters, queue attributes, pg counters, pg attributes.

Improvement:
1. Separate each counter/attribute type to be a member function of FlexCounter with the same function prototype.

2. Have a data member (unordered_map) to maintain pointers to member functions for each flex counter instance, and install only the essential counter collection member functions at the counter/attribute list set operation.

- Choice of data member structure: Originally considered unordered_set with element T to be a pointer to member function. However, the element type does not support a certain operations, and thus can not be taken as a key, but only as a value.

In doing so, we can avoid checking against irrelevant counter/attribute id lists, and keep a fixed size for collectCounters function.

Test on dut with debugging messages:
```
Jul 18 19:45:23.127582 str-a7050-acs-1 ERR syncd#syncd: :- collectCounters: Flex counter instance: QUEUE_STAT_COUNTER
Jul 18 19:45:23.127582 str-a7050-acs-1 ERR syncd#syncd: :- collectCounters: Collect type: QUEUE_COUNTER_ID_LIST
Jul 18 19:45:23.128843 str-a7050-acs-1 ERR syncd#syncd: :- collectCounters: Flex counter instance: QUEUE_WATERMARK_STAT_COUNTER
Jul 18 19:45:23.128843 str-a7050-acs-1 ERR syncd#syncd: :- collectCounters: Collect type: QUEUE_COUNTER_ID_LIST
Jul 18 19:45:23.134331 str-a7050-acs-1 ERR syncd#syncd: :- collectCounters: Flex counter instance: PG_WATERMARK_STAT_COUNTER
Jul 18 19:45:23.134331 str-a7050-acs-1 ERR syncd#syncd: :- collectCounters: Collect type: PG_COUNTER_ID_LIST
```

### DIFF
--- a/syncd/syncd_flex_counter.cpp
+++ b/syncd/syncd_flex_counter.cpp
@@ -132,6 +132,16 @@ void FlexCounter::updateFlexCounterStatsMode(
     }
  }
 
+void FlexCounter::addCollectCountersHandler(const std::string &key, collect_counters_handler_t)
+{
+    m_collectCountersHandlers.emplace(key, collect_counters_handler_t);
+}
+
+void FlexCounter::removeCollectCountersHandler(const std::string &key)
+{
+    m_collectCountersHandlers.erase(key);
+}
+
 /* The current implementation of 'setPortCounterList' and 'setQueueCounterList' are
  * not the same. Need to refactor these two functions to have the similar logic.
  * Either the full SAI attributes are queried once, or each of the needed counters
@@ -188,7 +198,7 @@ void FlexCounter::setPortCounterList(
     auto portCounterIds = std::make_shared<PortCounterIds>(portId, supportedIds);
     fc.m_portCounterIdsMap.emplace(portVid, portCounterIds);
 
-    fc.m_collectCountersHandlers.emplace(PORT_COUNTER_ID_LIST, &FlexCounter::collectPortCounters);
+    fc.addCollectCountersHandler(PORT_COUNTER_ID_LIST, &FlexCounter::collectPortCounters);
 
     // Start flex counter thread in case it was not running due to empty counter IDs map
     if (fc.m_pollInterval > 0)
@@ -265,7 +275,7 @@ void FlexCounter::setQueueCounterList(
     auto queueCounterIds = std::make_shared<QueueCounterIds>(queueId, supportedIds);
     fc.m_queueCounterIdsMap.emplace(queueVid, queueCounterIds);
 
-    fc.m_collectCountersHandlers.emplace(QUEUE_COUNTER_ID_LIST, &FlexCounter::collectQueueCounters);
+    fc.addCollectCountersHandler(QUEUE_COUNTER_ID_LIST, &FlexCounter::collectQueueCounters);
 
     // Start flex counter thread in case it was not running due to empty counter IDs map
     if (fc.m_pollInterval > 0)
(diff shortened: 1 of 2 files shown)


wendani:
Can you point a reference? https://github.com/Azure/sonic-sairedis/pull/488#pullrequestreview-264131993

syncd/syncd_flex_counter.cpp
```diff
@@ -554,6 +569,10 @@ void FlexCounter::removePort(
     }
 
     fc.m_portCounterIdsMap.erase(it);
+    if (!fc.m_portCounterIdsMap.size())
```
qiluo-msft: >!fc.m_portCounterIdsMap.size() [](start = 8, length = 30)

m_portCounterIdsMap.empty()

syncd/syncd_flex_counter.cpp
```diff
@@ -188,6 +188,8 @@ void FlexCounter::setPortCounterList(
     auto portCounterIds = std::make_shared<PortCounterIds>(portId, supportedIds);
     fc.m_portCounterIdsMap.emplace(portVid, portCounterIds);
 
+    fc.m_collectCountersHandlers.emplace(PORT_COUNTER_ID_LIST, &FlexCounter::collectPortCounters);
```
kcudnik: if you look from this line to below it's look like copy paste a lot of statements
maybe it could be optimized/merged to some array or function

wendani: Not quite get the way for generalization. Wrap the emplace/erase inside a member function is what I can think of. Please provide more details about your idea.

syncd/syncd_flex_counter.h
```diff
@@ -198,6 +199,20 @@ class FlexCounter
...
+        typedef void (FlexCounter::*collect_counters_handler_t)(_In_ swss::Table &countersTable);
+        typedef std::unordered_map<std::string, collect_counters_handler_t> collect_counters_handler_unordered_map_t;
+
+        void collectPortCounters(_In_ swss::Table &countersTable);
+        void collectQueueCounters(_In_ swss::Table &countersTable);
+        void collectQueueAttrs(_In_ swss::Table &countersTable);
+        void collectPriorityGroupCounters(_In_ swss::Table &countersTable);
+        void collectPriorityGroupAttrs(_In_ swss::Table &countersTable);
+        void collectRifCounters(_In_ swss::Table &countersTable);
+        void collectBufferPoolCounters(_In_ swss::Table &countersTable);
```
jleveque: Along the lines of Kamil's request to try to generalize things, would it make sense to have a generic collectCounters() method and pass in the counter type key (e.g., `void collectPortCounters(_In_ const std::string &key, _In_ swss::Table &countersTable);`, then incorporate logic in that single method to collect the appropriate counter type based on the key provided. This would eliminate a log of duplicate code, but depending on how much code can't be shared, the method could potentially become long and hard to follow. That requires a judgement call.

wendani: This can be our ultimate goal. We can achieve it step by step. https://github.com/Azure/sonic-sairedis/pull/488#pullrequestreview-265057977

PR #496, opened 2019-07-31
[syncd] Add support for Innovium platform

tonytitus:
Signed-off-by: Tony Titus ttitus@innovium.com

### DIFF
--- a/syncd/scripts/syncd_init_common.sh
+++ b/syncd/scripts/syncd_init_common.sh
@@ -158,6 +158,15 @@ config_syncd_vs()
     CMD_ARGS+=" -p $HWSKU_DIR/sai.profile"
 }
 
+config_syncd_innovium()
+{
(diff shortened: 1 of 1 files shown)


syncd/scripts/syncd_init_common.sh
```diff
@@ -158,6 +158,15 @@ config_syncd_vs()
     CMD_ARGS+=" -p $HWSKU_DIR/sai.profile"
 }
 
+config_syncd_innovium()
+{
+    CMD_ARGS+=" -p $HWSKU_DIR/sai.profile"
+    ulimit -s 65536
```
lguohan: why do need this?

tonytitus: We see stack usage going beyond 8K in certain scenario causing syncd crash.

PR #497, opened 2019-07-31
Sairedis nulloid derefence prevention

antony-rheneus:
Related to Handling of Unsupported SAI OID/attributes
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!searchin/sonicproject/oid$20!$3D$20SAI_NULL_OBJECT_ID%7Csort:date/sonicproject/VU7dTazZB2U/FY0dm2KQIQAJ

### DIFF
--- a/syncd/syncd_saiswitch.cpp
+++ b/syncd/syncd_saiswitch.cpp
@@ -936,6 +936,12 @@ void SaiSwitch::saiDiscover(
             {
                 sai_object_id_t oid = attr.value.objlist.list[i];
 
+                if(SAI_NULL_OBJECT_ID == oid)
+                {
+                    SWSS_LOG_DEBUG("oid is NULL for %s index %u", md->attridname, i);
+                    continue;
+                }
+
                 ot = sai_object_type_query(oid);
 
                 if (ot == SAI_OBJECT_TYPE_NULL)


syncd/syncd_saiswitch.cpp
```diff
@@ -936,6 +936,12 @@ void SaiSwitch::saiDiscover(
             {
                 sai_object_id_t oid = attr.value.objlist.list[i];
 
+                if(SAI_NULL_OBJECT_ID == oid)
+                {
+                    SWSS_LOG_DEBUG("oid is NULL for %s index %u", md->attridname, i);
+                    continue;
+                }
+
```
kcudnik: there is no dereference prevention here, if you look at line 939:
 ot = sai_object_type_query(oid); this should return SAI_OBJECT_TYPE_NULL if oid is null,
and SAI_NULL_OBJECT_ID  is not expected value on any attr.value.objlist, so at this point throw should be made ad line 943, indicating that this is vendor bug
and this entire check you added is not necessary

syncd/syncd_saiswitch.cpp
```diff
@@ -936,6 +936,12 @@ void SaiSwitch::saiDiscover(
             {
                 sai_object_id_t oid = attr.value.objlist.list[i];
 
+                if(SAI_NULL_OBJECT_ID == oid)
+                {
+                    SWSS_LOG_DEBUG("oid is NULL for %s index %u", md->attridname, i);
```
kcudnik: is so then warning

syncd/syncd_saiswitch.cpp
antony-rheneus: Thanks for the comment. Will decline this PR.

PR #498, opened 2019-07-31
For sflow feature, SAI changes needed for sonic-vs platform

rakeshdatta:
- _What I did_

Add SAI library changes to support sflow in sonic-vs.

- _How I did it_

When sflow is configured using CLI or other supported mechanisms, the orch-agent invokes 
the SAI API set_port_attribute(). For the sflow feature, his API is used to set the SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE attribute of a netdev port. The value stored for this attribute is a samplepacket object, which essentially contains various sampling attributes (sampling rate, sampler group etc.) associated with the port.

When sampling is disabled on a port, the  SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE attribute  is set to a null object. 
When sampling is enabled on a port, the  SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE attribute is set to a valid samplepacket object. 

For sonic-vs, in the absence of a real ASIC programming, the 'tc' command is used instead to configure the sampling parameters inside the kernel. In this PR we invoke the appropriate 'tc' command, based on the requested config actions (disable or enable or update). 

- _How to verify it_

  - Using https://github.com/Azure/sonic-buildimage/pull/3251 changes, build a sonic image
  - Using https://github.com/Azure/sonic-utilities/pull/592 changes, use the CLIs to configure sflow globally and on a port. Thereafter, use the show CLI to verify the changes made. 
  - Use the logs to verify the invocation of the proposed change in this  PR.

- _Description for the changelog_

   - Update: Added vslib changes to support sflow in sonic-vs
 
- _Additional Information_
  Please **DO NOT MERGE** until these PRs are merged:
  https://github.com/Azure/sonic-buildimage/pull/3251

Signed-off-by: Rakesh Datta <rakesh.datta@dell.com>

### DIFF
--- a/vslib/inc/sai_vs_state.h
+++ b/vslib/inc/sai_vs_state.h
@@ -253,11 +253,34 @@ class SwitchState
         return it->second;
     }
 
+    void setTapNameToPortId(
+            _In_ const std::string& tapname,
+            _In_ sai_object_id_t port_id)
+    {
+        SWSS_LOG_ENTER();
+
+        m_port_id_to_tapname[port_id] = tapname;
+    }
+
+    bool  getTapNameFromPortId(
+            _In_ const sai_object_id_t port_id,
+            _Out_ std::string& if_name)
+    {
+        SWSS_LOG_ENTER();
+
+        if (m_port_id_to_tapname.find(port_id) != m_port_id_to_tapname.end())
+        {
+            if_name = m_port_id_to_tapname[port_id];
+            return true;
+        }
+        return false;
+    }
+
     private:
 
         sai_object_id_t m_switch_id;
 
-        std::map<std::string, sai_object_id_t> m_ifname_to_port_id;
+        std::map<sai_object_id_t, std::string> m_port_id_to_tapname;
 
         swss::SelectableEvent m_link_thread_event;
 
--- a/vslib/src/sai_vs_switch.cpp
+++ b/vslib/src/sai_vs_switch.cpp
@@ -14,6 +14,8 @@
 
 #include <thread>
 
+sai_object_id_t g_vs_switch_id = SAI_NULL_OBJECT_ID;
+
 /**
  * @brief Get SwitchState by switch id.
  *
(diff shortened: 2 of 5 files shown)


vslib/inc/sai_vs_state.h
```diff
@@ -253,11 +253,34 @@ class SwitchState
...
+            return true;
+        }
+        return false;
+    }
+
     private:
 
         sai_object_id_t m_switch_id;
 
-        std::map<std::string, sai_object_id_t> m_ifname_to_port_id;
```
kcudnik: why this is removed ?

vslib/src/sai_vs_switch.cpp
```diff
@@ -14,6 +14,8 @@
 
 #include <thread>
 
+sai_object_id_t g_vs_switch_id = SAI_NULL_OBJECT_ID;
```
kcudnik: this is not needed, vs_switch_id can be obtained from object_id and checked whether is in objects map

vslib/inc/sai_vs_state.h
rakeshdatta: m_ifname_to_port_id is unused in the code. Instead, m_ifname_to_port_id_map is used every where.
This is bit of a cleanup.

vslib/src/sai_vs_switch.cpp
rakeshdatta: We need the switch_id inside vs_set_port_attribute(). 
Unfortunately, only two arguments are passed to it, namely port_id and attr. 
We are not able to determine switch_id from any of these arguments. Hence, we need a global.
Please let me know your views.

kcudnik: yes you are able to determine that even using sai_switch_id_query api

rakeshdatta: Sure, required modifications are made.

PR #500, opened 2019-08-01
Add support for port remove and port create (no warm boot)

kcudnik:
- added support in syncd
- added support in sairedis
- added added support in virtual switch
- added basic port remove/create unittests

warm boot scenario is not yet supported


meta/sai_meta.cpp
```diff
@@ -1705,6 +1725,151 @@ sai_status_t meta_generic_validation_create(
...
+        _In_ sai_object_id_t oid)
+{
+    SWSS_LOG_ENTER();
+
+    if (oid == SAI_NULL_OBJECT_ID)
+        SWSS_LOG_THROW("not expected NULL object id");
+
+    //sai_object_type_t object_type = sai_object_type_query(oid);
+
+    // TODO for each oid - check reference count if it's safe to remove
```
lguohan: is this still TODO?

kcudnik: forgot to remove that, this is checked in line 1831 and 1849

PR #508, opened 2019-08-29
Add to filter out switch MAC address FDB entry from FDB notification

vincentchiang-ec:
Signed-off-by: vincent_chiang <vincent_chiang@edge-core.com>

This modification is to fix issue [#507](https://github.com/Azure/sonic-sairedis/issues/507).

- What I did
Add to filter out switch MAC address FDB entry from FDB notification

- How I did it
Skip to notify FDB entry when MAC address is the same as the switch MAC address.

- How to verify it
There is no FDB error message occurred on system boot up.

- Description for the changelog
N/A

### DIFF
--- a/syncd/syncd_notifications.cpp
+++ b/syncd/syncd_notifications.cpp
@@ -334,9 +334,28 @@ void process_on_fdb_event(
 
     bool sendntf = true;
 
+    std::vector<sai_fdb_event_notification_data_t> ntf_data;
+
+    sai_object_meta_key_t meta_key;
+    meta_key.objecttype = SAI_OBJECT_TYPE_SWITCH;
+    meta_key.objectkey.key.object_id = data[0].fdb_entry.switch_id;
+    sai_attribute_t attr;
+    attr.id = SAI_SWITCH_ATTR_SRC_MAC_ADDRESS;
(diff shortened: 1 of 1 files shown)


syncd/syncd_notifications.cpp
```diff
@@ -334,10 +334,29 @@ void process_on_fdb_event(
...
     bool sendntf = true;
 
+    std::vector<sai_fdb_event_notification_data_t> ntf_data;
+
+    sai_object_meta_key_t meta_key;
+    meta_key.objecttype = SAI_OBJECT_TYPE_SWITCH;
+    meta_key.objectkey.key.object_id = data[0].fdb_entry.switch_id;
+    sai_attribute_t attr;
+    attr.id = SAI_SWITCH_ATTR_SRC_MAC_ADDRESS;
+    auto info = sai_metadata_get_object_type_info(SAI_OBJECT_TYPE_SWITCH);
```
kcudnik: please don't do any SAI queries in this functions, this is processed in separate thread, use the switch mac from SaiSwitch objects which has switch mac queried at instantiating swtich

also your approach will do query every time new notification will come which is not necessary

vincentchiang-ec:
Got it, should I finish the modification? Or just close this PR?

kcudnik:
You can go and close it

PR #510, opened 2019-09-08
Layer 2 Forwarding Enhancements

anilkpandey:
Added code changes related to Layer 2 Forwarding Enhancement:

As part of L2 enhancements, per port, per vlan and per port-vlan fdb flush support has been added. The fdb flush response from SAI comes as individual mac delete as mentioned in the HLD. The change here is to follow the same way for VS as well. 

Regarding SAI_FDB_EVENT_FLUSHED, it is handled in sync only for 'flush all' case and handling SAI_FDB_EVENT_FLUSHED for per port and per port-vlan will require hgetall on all fdb keys. So, individual mac delete response is preferred.

Another change is to move fdb handling to fdborch. As the fdb reference count is now maintained in fdborch as well, need to move the sai redis fdb handling also there to avoid the reference counts to go out of sync between the two.


lguohan:
@anilkpandey , thanks for the enhancement. can you provide full description of the change?

lib/src/sai_redis_notifications.cpp
```diff
@@ -216,7 +216,11 @@ void handle_notification(
     }
     else if (notification == "fdb_event")
     {
-        handle_fdb_event(data);
+        /* The fdb handling is moved to fdbOrch so that both
+         * reference count updates (sai_redis and portsOrch)
+         * happen in same context to avoid any mismatch.
+         */
+        /* handle_fdb_event(data);*/
```
kcudnik: no no no, this must be handled here, its counting references and all other stuff, please revert

vslib/src/sai_vs_fdb.cpp
```diff
@@ -156,10 +206,21 @@ sai_status_t internal_vs_flush_fdb_entries(
...
+
+            SWSS_LOG_INFO("FDBDelNotification: fdb_entry:%s bpid:%s", sai_serialize_fdb_entry(fi.fdb_entry).c_str(), sai_serialize_object_id(fi.bridge_port_id).c_str());
+
             it = fdbs.erase(it);
+
+            sendFdbDelNotification(fi);
         }
     }
 
+    /* Sending notification for each fdb entry individually above, as a single
```
kcudnik: please revert those changes, this remove in a loop can be enhanced adding lua script if necessary

lib/src/sai_redis_notifications.cpp
anilkpandey: > no no no, this must be handled here, its counting references and all other stuff, please revert

This function call has been moved to fdborch and is called from the orchagent main thread when fdb event is received. It was needed to avoid inconsistent reference count between fdborch and sai_redis, as now we maintain a per vlan and per port reference count in portsorch/fdborch which can be referenced before deleting bridge port or vlan in ASIC_DB.
Is it absolutely necessary to call this function here from the notification thread? There is no change in functionality except that it is being called from a different thread.

If it is absolutely needed to call this function from notification thread, then I plan to add additional data structure in fdborch to maintain the reference count instead of using the existing m_portlist and then call fdborch fdb event handling from on_fdb_event() callback, which is invoked  by sai_redis notification thread. 
This approach will also require a new semaphore for access to this new data structure as it will be updated by both notification thread and orchagent main thread (the main thread will update it when bridge port/vlan is added/deleted and also it will check the reference count before add/delete bridge port/vlan from ASIC_DB).

Let me know whether you agree with this approach.

vslib/src/sai_vs_fdb.cpp
anilkpandey: Sure, I will revert this.

lib/src/sai_redis_notifications.cpp
kcudnik: i get that, but it cant be moved to orchagent, since sairedis must be selfcontaing library, cant be splitted apart,  please revert

anilkpandey: > i get that, but it cant be moved to orchagent, since sairedis must be selfcontaing library, cant be splitted apart, please revert

Done

vslib/src/sai_vs_fdb.cpp
anilkpandey: Hi Kamil, with this change reverted, per port and per vlan fdb flush will not work on VS.

vslib/src/sai_vs_fdb.cpp
```diff
@@ -156,10 +206,21 @@ sai_status_t internal_vs_flush_fdb_entries(
...
+     * response causes syncd to delete ASIC_DB entries in a loop  and
+     * stuck for long time when a large number of FDB entries are present.
+     */
+#if 0
```
kcudnik: still dont know why this is disabled

vslib/src/sai_vs_fdb.cpp
kcudnik: is that because your change ?

anilkpandey: > is that because your change ?

yes, per port and per vlan fdb flush support has been added in orchagent. As syncd does not support SAI_FDB_EVENT_FLUSHED for per port and per vlan, individual mac delete response is sent from VS instead of SAI_FDB_EVENT_FLUSHED.

vslib/src/sai_vs_fdb.cpp
anilkpandey: sendFdbDelNotification() will notify sai redis lib notification thread and orchagent, so there is no need to do meta_sai_on_fdb_event().

kcudnik: there is need to call that, please analyse whats happens in that function meta_sai_on_fdb_event

anilkpandey: Got it. This meta sai update is for the vs sai. I just disabled the flush notification part now, as notification is added for each individual fdb. Thanks.

vslib/src/sai_vs_fdb.cpp
```diff
@@ -328,18 +338,21 @@ sai_status_t internal_vs_flush_fdb_entries(
         attrs[0].value.s32 = SAI_FDB_ENTRY_TYPE_DYNAMIC;
 
         meta_sai_on_fdb_event(1, &data); // update metadata
-
+        
+        /* notification has been sent for individual mac delete, so 
+         * not sending the bulk flush notification
+         */
+#if 0
```
kcudnik: this is still an issue, notification callback must be called if it was set

anilkpandey: As per my understanding, notification sent here if for syncd and meta_sai_on_fdb_event() is to update local SAI meta data (meta_sai_on_fdb_event(1, &data); // update metadata). The notification is already being sent in sendFdbDelNotification(), sending it again will cause duplicate notifications to syncd. Is my understanding correct?

anilkpandey: Actually, the new function sendFdbDelNotification() that I added, calls meta_sai_on_fdb_event() for individual mac and then send notification. So, we don't have to do it again. My previous change was correct. Please take a look at this in at end of function internal_vs_flush_fdb_entries():

fi.bridge_port_id = it->second.at("SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID")->getAttr()->value.oid;
SWSS_LOG_INFO("FDBDelNotification: fdb_entry:%s bpid:%s", sai_serialize_fdb_entry(fi.fdb_entry).c_str(), sai_serialize_object_id(fi.bridge_port_id).c_str());
it = fdbs.erase(it);
sendFdbDelNotification(fi);

And in function sendFdbDelNotification():

    // update metadata DB
    meta_sai_on_fdb_event(1, &data)

    if (ntf != NULL)
    {
        ntf(1, &data);
    }

I will change the code back to my previous change.

kcudnik: still heave some doubts on that, major change is that in previous implementation flush notifications are consolidated into "1" notifications, for example if you flash for all for 1 specific bridge id (static/dynamic) you willl get only 1 notification (even for 1000 macs), which is correct (consolidation is correct, take a look at fdb headers in SAI repo about flush api comments) and here your change is adding 1 notification per 1 single entry, which will flood receiver and we probably want to avoid that

anilkpandey: > still heave some doubts on that, major change is that in previous implementation flush notifications are consolidated into "1" notifications, for example if you flash for all for 1 specific bridge id (static/dynamic) you willl get only 1 notification (even for 1000 macs), which is correct (consolidation is correct, take a look at fdb headers in SAI repo about flush api comments) and here your change is adding 1 notification per 1 single entry, which will flood receiver and we probably want to avoid that

Hi Kamil, if though VS sent one notification for all mac, syncd does not support the cases where mac are flushed on either one bridge port or one vlan (per port, per vlan or per port-vlan). 
In function redisPutFdbEntryToAsicView(), 

        else if (port_oid && !bv_id)
        {
            /* we got a flush port fdb event here        */
            /* not supported yet, this is a place holder */
            /* example of a flush port fdb event         */
            /*
        }
        else if (!port_oid && bv_id)
        {
            /* we got a flush vlan fdb event here        */
            /* not supported yet, this is a place holder */
            /* example of a flush vlan event             */
            /*
         }
        else
        {
            SWSS_LOG_ERROR("received a flush fdb event, port_oid = 0x%lx, bv_id = 0x%lx, unsupported", port_oid, bv_id);
        }

Even if we try to handle this consolidated message in syncd, it will require reading the data of all the mac entries and match the bridge port or vlan and then delete it, which is a very very slow operation. 
Sending 1 notification per mac from VS can be handled in syncd irrespective of all/per port, per vlan and per port-vlan flush. It is same as what is being done from broadcom SAI, for an example.

kcudnik: its not matter of supporting this on syncd, syncd that should be supported on vendor SAI to delete all mac's. Syncd should only support removing those from redis DB, and as this im not aware if it supported since this part code you pointers is not code added by me. But i believe both scenarios should be supported anyway, consolidated and 1 by 1. If you really want to send this from VS 1 by 1, please add option on virtual switch profile so that can be switched on/off allowing user to choose which scenario should be uses. In our case we are not dealing as much L2 so its not issues for us.

kcudnik: also, if you are pointing out that this is expensive operation, then  matching 1 by 1 also can be slower operation, for example you could receive 1000 notifications to remove 1k fdb entries, and each of them would need to be processed separately, in consolidated message, probably a lua scrip would be better to achieve goal since just getting all mac addresses and them removing them could indeed be slow

anilkpandey: Hi Kamil, yes LUA script seems to be the best option to follow to handle consolidated event, I will work on that. For VS, it seems to be better than 1 by 1 event. 
For broadcom SAI 1 by 1 seems to be better as even though processing takes time, syncd notification thread is not blocked.

kcudnik: thanks, brcm sai internally sends event 1 by 1, so there is no workaround here

anilkpandey: I reverted to single notification and added support for all types of flush in syncd using lua script.
The lua script is in the following pull request:
https://github.com/Azure/sonic-swss-common/pull/303

kcudnik:
of course this needs to be build/checked again after swss-common lua fix

anilkpandey:
retest this please

anilkpandey:
build passed. please merge.

lguohan:
please remove .DS_Store

anilkpandey:
> please remove .DS_Store

Done.

kcudnik:
@anilkpandey Hey, since you worked on this PR i want to ask about special case, since we are moving forward with supporting syncd multiple switch instances in the same process, then for example there can be possibility that we do a flush with no attributes, just for a single switch, then we want all fdb entries to be removed from that single switch, this is the case when port=0 and bv_id = 0, but your script (the one in swss-common fdb_flush.lua) will remove all the fdb entries from possible all switches. 
To make this work, an extra param needs to be pushed and checked which is switchId, and cehcking each fdb_entry.switch_id in each entry, before actually making remove.
This switchId must be passed to the lua script and inside checked if fdb entry to be removed comes from that particular switch.
Ideally this swtich Id would need to be checked in other cases too (for example when bv_id is specified and non zero). Since we won't cross pass objects between switches (for example case when passing one bv_id from one switch to another is forbidden) to be extra sure, always switch Id should be checked.
Can you make an update to your script and submit PR ?

anilkpandey:
Yes, I will create a PR for this change.

kcudnik:
thanks !

PR #513, opened 2019-09-10
Modify the lanes of port in SYNCD

tiantianlv:
What I did
Support 8 lanes for port
Why I did it
In 201811 branch, lanes of port are for 4, it results syncd exception for 32x400G of TH3


lguohan:
we have an existing pr to cherry-pick

PR #514, opened 2019-09-17
Fix c_str() in printf for string

kcudnik:


kcudnik:
@rakeshdatta Please pay more attention in the future

PR #520, opened 2019-09-25
[flex_counter] Add sairedis support for drop counters

daall:
Add sairedis support for drop counters.
- Create two new flex counter types for handling port and switch level debug counters
- Add serialization functions to handle debug counter attributes
- Add stubs for querying device capabilities

Depends on:
~- Azure/sonic#434~
~- Azure/sonic-swss-common#308~

Signed-off-by: Danny Allen daall@microsoft.com


syncd/syncd_flex_counter.cpp
```diff
@@ -597,6 +645,44 @@ void FlexCounter::removePort(
...
+
+    auto it = fc.m_portDebugCounterIdsMap.find(portVid);
+    if (it == fc.m_portDebugCounterIdsMap.end())
+    {
+        SWSS_LOG_NOTICE("Trying to remove nonexisting port debug counter Ids 0x%" PRIx64, portVid);
+
+        // Remove flex counter if all counter IDs and plugins are unregistered
+        if (fc.isEmpty())
+        {
+            lkMgr.unlock();
```
kcudnik: i would suggest to avoid manual unlocking locks, since this leads to some complicated scenarios, when for example exception is throw, or someone else starts modifying the code and he don't know about whats protected by lock and where it should be released. Its better to create samaller method that will have scope lock on mutex protecting the only thing that needs to be protected. But this entire flex counters needs to refactored anyway, locks are used here in more than 20 places which is overkill

daall: I completely agree. I'll go ahead and refactor the new methods for debug counters to follow this convention, and make a note to go back and clean-up the lock usage at some point in the future.

daall: There's actually a much bigger problem here that I did not notice before - none of the collectCounters methods acquires the lock before iterating over the counter ID map. I'll probably raise a separate PR for this.

vslib/src/sai_vs_debug_counter.cpp
```diff
@@ -1,7 +1,81 @@
...
+    CHECK_STATUS(meta_sai_create_oid(
+                (sai_object_type_t)SAI_OBJECT_TYPE_DEBUG_COUNTER,
+                debug_counter_id,
+                switch_id,
+                attr_count,
+                attr_list,
+                &vs_generic_create));
+
+    sai_attribute_t attr;
+    attr.id = SAI_DEBUG_COUNTER_ATTR_INDEX;
```
kcudnik: why this is actually needed ?

daall: The virtual switch was complaining about the read only attribute not being implemented. I'll see if I can find the exact error message.

syncd/syncd_flex_counter.cpp
daall: So after taking a closer look at this issue, it seems that right now the manual locking/unlocking is necessary because of this line here: https://github.com/Azure/sonic-sairedis/blob/89b4dc013070b1951daf22d26b3ca025da3146c0/syncd/syncd_flex_counter.cpp#L1164

Because this method is returning the raw pointer instead of a shared pointer the call to removeInstance will delete the flex counter, which may cause issues if we're still holding the lock.

@kcudnik I think maybe we should continue to follow what the other flex counter methods are doing for this particular change and then proceed with a refactor to fix the locking and shared pointer issues in a later PR. What do you think?

vslib/src/sai_vs_debug_counter.cpp
```diff
@@ -1,7 +1,83 @@
...
+    attr.value.u32 = get_index();
```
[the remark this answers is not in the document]
lguohan: also if the index already exist, we should return already exist error. I think.

vslib/src/sai_vs_debug_counter.cpp
daall: I was finally able to re-test this and it looks like as long as the case is handled in refresh_read_only then we don't actually need to include this for the virtual switch tests to work. I'll go ahead and remove this part.

syncd/syncd_flex_counter.h
```diff
@@ -163,6 +180,20 @@ class FlexCounter
             std::vector<sai_port_stat_t> portCounterIds;
         };
 
+        struct SwitchCounterIds
+        {
+            SwitchCounterIds(
+                _In_ sai_object_id_t oid,
+                _In_ const std::vector<sai_switch_stat_t> &counterIds)
+                    : switchId(oid), switchCounterIds(counterIds)
```
kcudnik: implementation should be in cpp file

kcudnik: also if initializer list used, please place 1 member per line

syncd/syncd_flex_counter.h
```diff
@@ -190,6 +221,12 @@ class FlexCounter
         void saiUpdateSupportedBufferPoolCounters(sai_object_id_t bufferPoolId,
                                                   const std::vector<sai_buffer_pool_stat_t> &counterIds,
                                                   sai_stats_mode_t statsMode = SAI_STATS_MODE_READ_AND_CLEAR);
+
+        std::vector<sai_port_stat_t> saiCheckSupportedPortDebugCounters(sai_object_id_t portId,
+                                                                        const std::vector<sai_port_stat_t> &counterIds);
```
kcudnik: missing _In_ tag on each param, also dont put param in the first line, but on next when function defined

PR #521, opened 2019-09-26
Improved the initialization time by a few minutes by optimizing the flex counter queue configurations

Prasanth-KV:
Synd gets busy collecting counters when flex counter config replay happens during init.
Each queue list update for flex counter results in waking-up the counter collection thread.
This makes syncd busy for a few minutes.
Waking up the stats collection thread is required when there is a change in polling interval or when a flex counter is enabled.

How to Test?
Measure the initialization time with this change during normal boot-up and config-reload. 

Signed-off-by: Prasanth Kunjum Veettil <prasanth.kunjumveettil@broadcom.com>


kcudnik:
closing as obsolete, flex counters code changed dramatically and was redesigned since this post

PR #522, opened 2019-10-12
Copp changes for supporting genetlink in vs

dgsudharsan:
**What I did**
Added support in swss for SFLOW. It includes sfloworch and sflowmgr changes.

**Why I did it**
Added support for supporting genetlink attribute in vs
**How I verified it**
Loaded vs image. Enabled sflow and checked the redis-db

**Details if related**
These changes are required for supporting sflow. When sflow is enabled, the genetlink attributes are programmed and vslib shouldn't return any errors which might crash orchagent.

### DIFF
--- a/vslib/src/sai_vs_hostintf.cpp
+++ b/vslib/src/sai_vs_hostintf.cpp
@@ -1122,6 +1122,9 @@ sai_status_t vs_create_hostif_tap_interface(
         return SAI_STATUS_FAILURE;
     }
 
+    /* The genetlink host interface is created to associate trap group to genetlink family and multicast group
+     * created by driver. It does not create any netdev interface. Hence skipping tap interface creation
+     */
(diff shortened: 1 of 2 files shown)


vslib/src/sai_vs_hostintf.cpp
```diff
@@ -1122,6 +1122,13 @@ sai_status_t vs_create_hostif_tap_interface(
         return SAI_STATUS_FAILURE;
     }
 
+    if (attr_type->value.s32 == SAI_HOSTIF_TYPE_GENETLINK)
```
lguohan: can you add comment why we are skipping creating SAI_HOSTIF_TYPE_GENETLINK here?

dgsudharsan: Done

PR #524, opened 2019-10-21
Fix learn fdb events after fdb flush event

kcudnik:


prsunny:
@kcudnik , Can you give some details on the fix in description? AFAIK, there were few issues with flush/learn events.

kcudnik:
added description as requested

PR #525, opened 2019-10-24
[sai_redis_interface_query] Add sairedis support for sai_query_attribute_enum_values_capability

daall:
[sai_redis_interface_query] Add sairedis support for sai_query_attribute_enum_values_capability
* Add sai_query_attribute_enum_values_capability to libsairedis
* Add handler for sai_query_attribute_enum_values_capability to syncd
* Add sai_query_attribute_enum_values_capability implementation to virtual switch

Signed-off-by: Danny Allen <daall@microsoft.com>

Depends on: Azure/sonic-swss-common#313


vslib/src/sai_vs_interfacequery.cpp
```diff
@@ -921,3 +921,44 @@ sai_status_t sai_api_query(
...
+    SWSS_LOG_ENTER();
+
+    if (object_type == SAI_OBJECT_TYPE_DEBUG_COUNTER && attr_id == SAI_DEBUG_COUNTER_ATTR_IN_DROP_REASON_LIST)
+    {
+        if (enum_values_capability->count < 3)
+        {
+            return SAI_STATUS_BUFFER_OVERFLOW;
+        }
+
+        enum_values_capability->count = 3;
```
kcudnik: capabilities should come from saimetadata, since there is a way to add capability file when generating saimetadata from SAI/meta repo, but i think here we can skip it since it would be too hard for this pr to include that an refactor generating metadata independently for virtual switch

syncd/syncd.cpp
```diff
@@ -2867,6 +2867,66 @@ sai_status_t processFdbFlush(
...
+    sai_status_t status = sai_query_attribute_enum_values_capability(switch_rid, object_type, attr_id, &enum_values_capability);
+
+    std::vector<swss::FieldValueTuple> response_payload;
+
+    if (status == SAI_STATUS_SUCCESS)
+    {
+        std::string serialized_enum_capabilities;
+        for (uint32_t i = 0; i < enum_values_capability.count; i++)
+        {
+            serialized_enum_capabilities += ',';
```
kcudnik: first comma will always be added

vslib/src/sai_vs_interfacequery.cpp
daall: I'll go ahead and make a note for this.

daall:
Opened #526 and #527 to track testing/debugging improvements for these new query APIs.

Walker917:
a error
./syncd/syncd.cpp:2969: undefined reference to `sai_object_type_get_availability'

wendani:
What is the purpose of having this support?

PR #529, opened 2019-11-12
Enable FastReboot if we have key "FAST_REBOOT|system" in State db

pavel-shirshov:
Enable fast-reboot only if "FAST_REBOOT|system" key exists in STATE db.
This PR is required to be included into https://github.com/Azure/sonic-buildimage/pull/3741 to fix https://github.com/Azure/sonic-buildimage/issues/3697


yxieca:
Backed out from 201811 branch until the whole change is ready.

PR #542, opened 2019-12-02
[syncd] Move port map and port map parser to proper class

kcudnik:
This PR is on top of https://github.com/Azure/sonic-sairedis/pull/541, please check that PR first


syncd/syncd.cpp
```diff
@@ -3569,174 +3548,6 @@ void processFlexCounterEvent(
...
-#endif // SAITHRIFT
-            { "syncMode",                no_argument,       0, 's' },
-            { 0,                         0,                 0,  0  }
-        };
-
-        int option_index = 0;
-
-        int c = getopt_long(argc, argv, optstring, long_options, &option_index);
-
-        if (c == -1)
```
qiluo-msft: >gPortMap [](start = 8, length = 8)

I search the code and no one is using gPortMap. Can you remove it at all?

kcudnik: no i can't, gPortMap is used in sai thirft rpc server and it;s linked agains this object here, i sent email to Pavel to maybe figure out some other to make this like call function to set actual port map or set parh to it to saithirft server, also there is similar object gSwitchId which is also uses in saithirt and it's global object providing functionality for multiple switches, then this approach will not  work, i will forward you email

syncd/PortMap.cpp
```diff
@@ -0,0 +1,51 @@
...
+ */
+std::map<std::set<int>, std::string> gPortMap;
+
+void PortMap::insert(
+        _In_ const std::set<int>& laneSet,
+        _In_ const std::string& name)
+{
+    SWSS_LOG_ENTER();
+
+    m_portMap.insert(std::make_pair(laneSet, name));
```
pavel-shirshov: `m_portMap.emplace(laneSet, name);`
for my taste looks better

kcudnik: that was copy paste from original code :)

kcudnik:
test this please

PR #544, opened 2019-12-03
Move api mutex to global class and add sairedis namespace

kcudnik:


lib/inc/Globals.h
```diff
@@ -0,0 +1,30 @@
...
+             * All SAI interface API should be protected by this mutex to make
+             * sure there can be only one API executed at a given time.  This
+             * will also make sure that underlying metadata database will be
+             * also accessed only from one place at a given time.
+             *
+             * Since SAI interface is global for entire sairedis library there
+             * is no need to have more than one mutex protecting API access, so
+             * a global API mutex is declared.
+             */
+            static std::mutex apimutex;
```
qiluo-msft: >std::mutex apimutex [](start = 19, length = 19)

since everyone is not using it directly, how about move MUTEX() inside the class as a public function, and change mutex member to private?

kcudnik: it can't be function, since MUTEX macto is locking on the current executing scope
unless you suggesting to make a global function like getApiMutex() but this seems to me to have not sense

kcudnik:
test this please

PR #552, opened 2019-12-16
[sairedis] Introduce VirtualObjectIdManager class for managing VIDs

kcudnik:
Add VirtualObjectIdManager class for easier handling generating virtual OIDs for sairedis library and for syncd


lguohan:
is this still needed? @kcudnik, if not, can you close it?

kcudnik:
not needed any more

PR #553, opened 2020-01-02
[meta] Return OBJECT_IN_USE when removing port is in use

kcudnik:


zhenggen-xu:
@kcudnik Looks good.  A question, does it make sense to make this generic for different objects remove case? I,E, whenever we remove an object and if if it is still reference, we return "SAI_STATUS_OBJECT_IN_USE"

kcudnik:
in different PR

PR #561, opened 2020-02-11
[vs] Use MAC addr of the default gw interface for the default switch …

ishidawataru:
…address

11:22:33:44:55:66 is a multicast mac address and can't be assigned to
TAP interfaces.

Use MAC addr of the default gw interface instead.
If failed to get the address, use 22:33:44:55:66:77 which is not a
multicast mac address.

Signed-off-by: Wataru Ishida <ishida@nel-america.com>

### DIFF
--- a/vslib/inc/sai_vs_switch_common.h
+++ b/vslib/inc/sai_vs_switch_common.h
@@ -1,0 +1,9 @@
+#ifndef __SAI_VS_SWITCH_COMMON__
+#define __SAI_VS_SWITCH_COMMON__
+
+#include "sai_vs_state.h"
+#include <memory>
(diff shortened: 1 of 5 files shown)


kcudnik:
I also pulled this change to my rebase branch

lguohan:
do we need this now? @kcudnik , it seems that you have pull this change in your rebase branch, so we do not need this any more?

kcudnik:
yes, this is pulled when i was doing rebase https://github.com/Azure/sonic-sairedis/blob/master/vslib/src/SwitchStateBase.cpp#L649 so this PR is not needed any more

PR #330, opened 2018-06-14
[syncd]: Clean-up log messages

volodymyrsamotiy:
Signed-off-by: Volodymyr Samotiy <volodymyrs@mellanox.com>

PR #562, opened 2020-02-19
[flex counter] add rif plugins

mykolaf:
According to [HLD](https://github.com/Azure/SONiC/blob/master/doc/rates-and-utilization/Rates_and_utilization_HLD.md)

Depends on [sonic-swws-common #330](https://github.com/Azure/sonic-swss-common/pull/330)

Signed-off-by: Mykola Faryma <mykolaf@mellanox.com>


kcudnik:
ca you resolve conflicts ?

liat-grozovik:
retest this please

liat-grozovik:
@abdosi where do we stands with this review? can it be approved?

kcudnik:
You have compilation errors:
FlexCounter.cpp:871:27: error: 'RIF_PLUGIN_FIELD' was not declared in this scope
12:02:50           else if (field == RIF_PLUGIN_FIELD)

mykolaf:
@kcudnik The error is caused by dependency https://github.com/Azure/sonic-swss-common/pull/330
which is not yet merged

syncd/FlexCounter.cpp
```diff
@@ -1478,7 +1478,9 @@ void FlexCounter::runPlugins(
     {
         std::to_string(COUNTERS_DB), // TODO
```
kcudnik, line 1479: this needs to co updated to m_dbCounters, and if you need other ASIC_DB and COUNFIG_DB they also needs to be passed to class, since this is hard coding now, and it needs  to be able to support multiple databases

syncd/FlexCounter.cpp
```diff
@@ -1478,7 +1478,9 @@ void FlexCounter::runPlugins(
...
         COUNTERS_TABLE,
-        std::to_string(m_pollInterval * 1000)
+        std::to_string(m_pollInterval * 1000),
+        std::to_string(ASIC_DB),
```
kcudnik, line 1482: also why you need ASIC_DB and CONFIG_DB ? oncly counters_db should be utilized here

mykolaf: is this hard requirement? I'm utilising asic db to get port speed. I see that pfc scripts do the same.

kcudnik: hard requirement it to not use hardcoded databases indexes/names and use those which are loaded from config file, also why do you need port speed here?

mykolaf: It's for calculating port utilization, port byte rate is divided by port speed.

mykolaf: @kcudnik Decided to avoid passing multiple db ids to the lua plugins and use only counters DB.
Removed hardcoded db ids.

kcudnik:
please fix errors
15:45:08  FlexCounter.cpp:871:27: error: 'RIF_PLUGIN_FIELD' was not declared in this scope
15:45:08           else if (field == RIF_PLUGIN_FIELD)
15:45:08                             ^~~~~~~~~~~~~~~~
15:45:08  FlexCounter.cpp: In member function 'void syncd::FlexCounter::runPlugins(swss::DBConnector&)':
15:45:08  FlexCounter.cpp:1506:24: error: 'db' was not declared in this scope
15:45:08           runRedisScript(db, sha, rifList, argv);
15:45:08                          ^~
15:45:08  FlexCounter.cpp:1506:46: error: 'runRedisScript' was not declared in this scope
15:45:08           runRedisScript(db, sha, rifList, argv);

PR #563, opened 2020-02-21
[meta/sai_meta]Fix the issue that fdb learnt during create switch fails removeDefaultBridgePorts

stephenxs:
In normal operation orch agent should query or create all bridge, vlan and bridge port, so we should not get this message. Let's put it as warning for better visibility. Most likely if this happen  there is a vendor bug in SAI and we should also see warnings or errors reported from syncd in logs.
Originally the solution was to create the object and then continue to learn fdb_entry, which causes other issue on SONiC. Consider the following scenario:
1. SWITCH ASIC reports fdb learnt on a certain vlan/port
2. Create object for the vlan/port
3. Learn the mac address, which will increase the reference count of port
4. On SONiC, all port will be removed from bridge during creating switch, which is done by calling removeDefaultBridgePorts. It contains a logic of checking the reference number of ports being removed and fails if not zero. For the ports have fdb learnt ahead of this, their were referenced by the fdb and thus fails.

This issue can be reproduced on a normal switch.
As eventually we will remove ports from the bridge, we don't need to learn the fdb here.


liat-grozovik:
retest vs please

liat-grozovik:
retest this please

stephenxs:
retest this please

kcudnik:
please refactor on latest origin/master

kcudnik:
please refactor on latest origin/master to not pull over 350 commits to this PR

stephenxs:
Close this one due to refactoring. Open [580](https://github.com/Azure/sonic-sairedis/pull/580)

stephenxs:
> please refactor on latest origin/master

Hi, I move the PR to [580](https://github.com/Azure/sonic-sairedis/pull/580). Please check. Thanks.

PR #565, opened 2020-02-25
Return correct error code

vasant17:
What I did:
Changed the error code from SAI_STATUS_INVALID_PARAMETER to SAI_STATUS_OBJECT_IN_USE in meta_generic_validation_remove(). Also, deleted the similar check in port object validation as that is NOT required.

Why I did:
When an object has 1 or more references, and it's been requested to delete the object, orchagent expects SAI_STATUS_OBJECT_IN_USE error code. For now, this is true for port object. I believe this should be the case for all objects as well. 

How I tested it:
While testing dynamic port breakout, port object was being deleted and it was being referenced by router interface object. Orchagent was expecting SAI_STATUS_OBJECT_IN_USE but receiving SAI_STATUS_INVALID_PARAMETER, hence it was throwing an exception.. After the fix, correct error code is being returned and orchagent waits for the dependencies to be removed before deleting the port.

### DIFF
--- a/meta/Meta.cpp
+++ b/meta/Meta.cpp
@@ -2958,7 +2958,7 @@ sai_status_t Meta::meta_generic_validation_remove(
 
         SWSS_LOG_ERROR("object 0x%" PRIx64 " reference count is %d, can't remove", oid, count);
 
-        return SAI_STATUS_INVALID_PARAMETER;
+        return SAI_STATUS_OBJECT_IN_USE;
     }
 
     if (meta_key.objecttype == SAI_OBJECT_TYPE_PORT)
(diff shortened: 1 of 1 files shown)


meta/sai_meta.cpp
```diff
@@ -1824,14 +1824,6 @@ sai_status_t meta_port_remove_validation(
         return SAI_STATUS_SUCCESS;
     }
 
-    if (object_reference_count(port_id) != 0)
```
kcudnik: why this is removed ?

vasant17: Isn't the line 1917 does the same thing?

kcudnik: yes, but i have a big changes and this code would need to be pulled in to my new branch: https://github.com/Azure/sonic-sairedis/compare/master...kcudnik:bigrebase?expand=1 and this branch will be rebased to master this or next week, and it still contains this code but in different files so you could then make this  PR on new code

vasant17: how difficult is it to pull this into you branch?

vasant17: OK. I got it. Let me wait and make these changes when your code is merged.

kcudnik: please rebase, new origin/master is laready present

lguohan:
is this pr still needed?

kcudnik:
yes, still needed, as asked author to rebase this change to new origin/master

PR #568, opened 2020-03-10
Extend response timeout from 60s to 120s

Junchao-Mellanox:
Extend response timeout from 60s to 120s for Mellanox spectrum 3.

### DIFF
--- a/lib/src/RedisChannel.cpp
+++ b/lib/src/RedisChannel.cpp
@@ -12,7 +12,7 @@ using namespace sairedis;
 /**
  * @brief Get response timeout in milliseconds.
  */
-#define REDIS_ASIC_STATE_COMMAND_GETRESPONSE_TIMEOUT_MS (60*1000)
+#define REDIS_ASIC_STATE_COMMAND_GETRESPONSE_TIMEOUT_MS (120*1000)
 
 RedisChannel::RedisChannel(
         _In_ const std::string& dbAsic,


liat-grozovik:
@Junchao-Mellanox  before we merge it i would like to understand first if this is correct for all other ASICs and not just for SPC3.

Junchao-Mellanox:
Hi @liat-grozovik , this change will apply to all ASIC, not only for SPC3.

kcudnik:
this is response for redis database on GET api, why its increased ??

Junchao-Mellanox:
Hi @kcudnik, now in spectrum 3 platform, firmware start slower than spectrum 2 and it may take more than 60 seconds for the "sending syncd INIT view" and wait for notify response. So I extend the response time for Mellanox spectrum 3 platform, and it only affect the startup time. There is also an internal firmware version which already improve the time usage, and we can change the response time back to 60 seconds once that firmware version release and upstream.

kcudnik:
Is device at this point configured ? what about warm boot ?

lguohan:
i suggest to add an command line option so that we can customize for different asic.

kcudnik:
@Junchao-Mellanox if you want to do it please add this as custom parameter in sairedis similar fashion  as SAI_REDIS_SWITCH_ATTR_CONTEXT

Junchao-Mellanox:
> @Junchao-Mellanox if you want to do it please add this as custom parameter in sairedis similar fashion as SAI_REDIS_SWITCH_ATTR_CONTEXT

I will close this PR first and discuss with Liat and Kebo about this issue. Thanks for your suggestion.

PR #580, opened 2020-03-22
[meta/sai_meta]Fix the issue that fdb learnt during create switch fails removeDefaultBridgePorts

stephenxs:
In normal operation orch agent should query or create all bridge, vlan and bridge port, so we should not get this message. Let's put it as warning for better visibility. Most likely if this happen  there is a vendor bug in SAI and we should also see warnings or errors reported from syncd in logs.
Originally the solution was to create the object and then continue to learn fdb_entry, which causes other issue on SONiC. Consider the following scenario:
1. SWITCH ASIC reports fdb learnt on a certain vlan/port
2. Create object for the vlan/port
3. Learn the mac address, which will increase the reference count of port
4. On SONiC, all port will be removed from bridge during creating switch, which is done by calling removeDefaultBridgePorts. It contains a logic of checking the reference number of ports being removed and fails if not zero. For the ports have fdb learnt ahead of this, their were referenced by the fdb and thus fails.

This issue can be reproduced on a normal switch.
As eventually we will remove ports from the bridge, we don't need to learn the fdb here.

### DIFF
--- a/meta/Meta.cpp
+++ b/meta/Meta.cpp
@@ -7006,31 +7006,16 @@ void Meta::meta_sai_on_fdb_flush_event_consolidated(
     }
 }
 
-void Meta::meta_fdb_event_snoop_oid(
+int Meta::meta_fdb_event_snoop_oid(
         _In_ sai_object_id_t oid)
 {
     SWSS_LOG_ENTER();
 
     if (oid == SAI_NULL_OBJECT_ID)
-        return;
+        return -1;
 
-    if (m_oids.objectReferenceExists(oid))
-        return;
-
-    sai_object_type_t ot = objectTypeQuery(oid);
-
-    if (ot == SAI_OBJECT_TYPE_NULL)
-    {
-        SWSS_LOG_ERROR("failed to get object type on fdb_event oid: 0x%" PRIx64 "", oid);
-        return;
-    }
-
-    sai_object_meta_key_t key = { .objecttype = ot, .objectkey = { .key = { .object_id = oid } } };
-
-    m_oids.objectReferenceInsert(oid);
-
-    if (!m_saiObjectCollection.objectExists(key))
-        m_saiObjectCollection.createObject(key);
+    if (m_oids.object_reference_exists(oid))
+        return 0;
 
     /*
      * In normal operation orch agent should query or create all bridge, vlan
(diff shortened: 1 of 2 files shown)


meta/Meta.cpp
```diff
@@ -7067,7 +7067,11 @@ void Meta::meta_sai_on_fdb_event_single(
      * query yet.
      */
 
-    meta_fdb_event_snoop_oid(data.fdb_entry.bv_id);
+    if (meta_fdb_event_snoop_oid(data.fdb_entry.bv_id) < 0)
+    {
+        SWSS_LOG_WARN("fdb_entry oid (snoop): ignore the mac address learning during creating switch");
```
kcudnik: how do you know this is during create switch ? it can be in different scenarios

meta/Meta.cpp
```diff
@@ -7006,43 +7006,43 @@ void Meta::meta_sai_on_fdb_flush_event_consolidated(
...
-
-    if (ot == SAI_OBJECT_TYPE_NULL)
-    {
-        SWSS_LOG_ERROR("failed to get object type on fdb_event oid: 0x%" PRIx64 "", oid);
-        return;
-    }
-
-    sai_object_meta_key_t key = { .objecttype = ot, .objectkey = { .key = { .object_id = oid } } };
-
-    m_oids.objectReferenceInsert(oid);
```
kcudnik: inserting object  reference must happen to lean the oid.

meta/Meta.cpp
```diff
@@ -7006,43 +7006,43 @@ void Meta::meta_sai_on_fdb_flush_event_consolidated(
...
      * vendor bug in SAI and we should also see warnings or errors reported
      * from syncd in logs.
+     *
+     * Originally the solution was to create the object and then continue to
+     * learn fdb_entry, which causes other issue on SONiC. Consider the following
+     * scenario:
+     * 1. SWITCH ASIC reports fdb learnt on a certain vlan/port
+     * 2. Create object for the vlan/port
+     * 3. Learn the mac address, which will increase the reference count of port
+     * 4. On SONiC, all port will be removed from bridge during creating switch,
```
kcudnik: learning fdb entry during switch create i perfectly normal scenario, so learn should happen even during switch create. If some fdb entiies were learned during that time, and reference was increased, then is up to user to remove those fdb entries before removing port

kcudnik:
also your change causes unittests to fail

meta/Meta.cpp
stephenxs: I debugged this issue and found it occurred during create switch.
However, I can't tell whether it is reproducible in other flows.

meta/Meta.cpp
stephenxs: Do you mean this issue should be handled in orchagent where the port is removed?

kcudnik: yes, that is my intention

stephenxs:
Close this PR for now. Will address the issue in SONiC side.

PR #581, opened 2020-03-25
[meta] Flush fdb entries after flush api success

kcudnik:


meta/Meta.cpp
```diff
@@ -6954,7 +6976,8 @@ void Meta::meta_sai_on_fdb_flush_event_consolidated(
             continue;
         }
 
-        if (fdbTypeAttr->getSaiAttr()->value.s32 != type->value.s32)
+        if ((type->value.s32 != SAI_FDB_FLUSH_ENTRY_TYPE_ALL) &&
```
vasant17: How about this?
```
if (type->value.s32 == SAI_FDB_FLUSH_ENTRY_TYPE_ALL) {
    // add fdb to remove vector and continue;
}

// check type match
// check bridge_port_id matches
// check vlan ID matches
```

If you dont like this, then when someone specified type == SAI_FDB_FLUSH_ENTRY_TYPE_ALL, and also provides vlan_id or bridge_port_id or both, we are NOT going to flush all. Is that the expected behavior? Or we should honor the type attribute?

meta/Meta.cpp
```diff
@@ -6954,7 +6976,8 @@ void Meta::meta_sai_on_fdb_flush_event_consolidated(
...
+            (fdbTypeAttr->getSaiAttr()->value.s32 != type->value.s32))
```
vasant17: Aren't these values from different enum types? one from ..fdb_flush_attr_type.. and another one from ...fdb_attr_type..?

vasant17: May be we want this?
```
sai_attribute_t *fdb_type_attr_p = fdbTypeAttr->getSaiAttr();
        if ((flush_type_attr_p->value.s32 == SAI_FDB_FLUSH_ENTRY_TYPE_DYNAMIC &&
             fdb_type_attr_p->value.s32 != SAI_FDB_ENTRY_TYPE_DYNAMIC) ||
             (flush_type_attr_p->value.s32 == SAI_FDB_FLUSH_ENTRY_TYPE_STATIC &&
             fdb_type_attr_p->value.s32 != SAI_FDB_ENTRY_TYPE_STATIC))
        {
            // entry type is not matching on this fdb entry
            continue;
        }
```

vasant17: Also at line number 6962
Instead of this
```
SWSS_LOG_NOTICE("processing consolidated fdb flush event of type: %s",
            sai_metadata_get_fdb_entry_type_name((sai_fdb_entry_type_t)type->value.s32));
```
shall we have
```
SWSS_LOG_NOTICE("processing consolidated fdb flush event of type: %s",
            sai_metadata_get_fdb_flush_entry_type_name((sai_fdb_flush_entry_type_t)type->value.s32));
```

meta/Meta.cpp
```diff
@@ -1683,7 +1683,29 @@ sai_status_t Meta::flushFdbEntries(
...
+        sai_attribute_t list[2];
+
+        list[0].id = SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID;
+        list[0].value.oid = bp_id ? bp_id->value.oid : SAI_NULL_OBJECT_ID;
+
+        list[1].id = SAI_FDB_ENTRY_ATTR_TYPE;
+        list[1].value.s32 = et ? et->value.s32 : SAI_FDB_FLUSH_ENTRY_TYPE_ALL;
+
+        data.event_type = SAI_FDB_EVENT_FLUSHED;
+        data.fdb_entry.switch_id = switch_id;
```
vasant17: Should we have fdb_flush_entry as well in sai_fdb_event_notification_data_t and use it when event is SAI_FDB_EVENT_FLUSHED. Otherwise, we are converting FDB_FLUSH_ENTRY attributes to FDB_ENTRY attributes and in-fact if you look at the second attribute here, attribute type is SAI_FDB_ENTRY_ATTR_TYPE and  value is SAI_FDB_FLUSH_ENTRY_TYPE_ALL. I personally do NOT like this. what do you say?

meta/Meta.cpp
vasant17: Also, what do we print on line number 6963 when value is SAI_FDB_FLUSH_ENTRY_TYPE_ALL? Because we can casting from FLUSH_ENTRY_TYPE to FDB_ENTRY_TYPE enum

kcudnik: cant be like that , since if it's ALL (static and dynamic, other attributes also needs to be checked, like vlan id, or bridge port, ALL here appllies only to static or dynamic

meta/Meta.cpp
kcudnik: yes you are right, those are different entry types, it happens to work, since ALL has different enum value than STATIC and DYNAMIC, but it works here :P, i will update that

meta/Meta.cpp
vasant17: Got it. SAI_FDB_FLUSH_ENTRY_TYPE_ALL means BOTH(STATIC and DYNAMIC) type! NOT really ALL FDB entries!

meta/Meta.cpp
```diff
@@ -1685,27 +1685,45 @@ sai_status_t Meta::flushFdbEntries(
...
 
-        sai_fdb_event_notification_data_t data = {};
+        std::vector<int32_t> types;
 
-        auto *bv_id = sai_metadata_get_attr_by_id(SAI_FDB_FLUSH_ATTR_BV_ID, attr_count, attr_list);
-        auto *bp_id = sai_metadata_get_attr_by_id(SAI_FDB_FLUSH_ATTR_BRIDGE_PORT_ID, attr_count, attr_list);
         auto *et = sai_metadata_get_attr_by_id(SAI_FDB_FLUSH_ATTR_ENTRY_TYPE, attr_count, attr_list);
 
-        sai_attribute_t list[2];
+        if (et)
```
vasant17: What is someone specifies the type but sets it to TYPE_ALL?

kcudnik: already fixed in my next commit

meta/Meta.cpp
kcudnik: yes, already addressed in below commit

meta/Meta.cpp
kcudnik: there are no other entries except static and dynamic, so ALL means static and dynamic together

zhenggen-xu:
@kcudnik  Can we backport this to 201911 branch?

vasant17:
> @kcudnik Can we backport this to 201911 branch?

I did it. @kcudnik or @qiluo-msft could you please approve and merge.

PR #586, opened 2020-04-01
[syncd] Fix crash during stats polling

vsenchyshyn:
The crash is observed during pfc_wd CT execution on BFN platform.

Null pointer is passed to VendorSai::getStats() function due to data() method call for empty "result"  vector. Then NULL pointer is passed to SAI handler where it causes crash.

In order to fix this "result" vector size is set to the size of counter_ids vector. In addition to this checks for NULL pointers are added to VendorSai::getStats().


vsenchyshyn:
@lguohan Please review

kcudnik:
thanks for fixing this

PR #603, opened 2020-04-14
[vs] Make default port admin state to false

kcudnik:


vslib/src/SwitchStateBase.cpp
```diff
@@ -378,6 +378,13 @@ sai_status_t SwitchStateBase::setPort(
 
             if (kvp.second->m_portId == portId)
             {
+                if (ifup(tapname.c_str(), portId, up, false))
+                {
+                    SWSS_LOG_ERROR("if admin %s failed on %s failed: %s", (up ? "UP" : "DOWN"), tapname.c_str(),  strerror(errno));
+
+                    return SAI_STATUS_FAILURE;
+                }
```
lguohan, line 386: this 381 - 386 should be removed. not neccessary.

vslib/src/SwitchStateBaseHostif.cpp
```diff
@@ -611,15 +611,34 @@ sai_status_t SwitchStateBase::vs_create_hostif_tap_interface(
...
+
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("failed to get admin state for port %s",
+                sai_serialize_object_id(obj_id).c_str());
+
+        return false;
+    }
+
+    if (ifup(name.c_str(), obj_id, attr.value.booldata, false))
```
lguohan, line 628: this should the vname.

lguohan:
this does not work. check hostif_create_tap_veth_forwarding(), you always bring the link up there. Can you fix that function?

kcudnik:
I fixed that, but that was introduced also on your request to have notifications when switch is instantiated

vslib/src/SwitchStateBaseHostif.cpp
```diff
@@ -479,7 +479,7 @@ bool SwitchStateBase::hostif_create_tap_veth_forwarding(
 
     SWSS_LOG_NOTICE("interface index = %d, %s\n", sock_address.sll_ifindex, vethname.c_str());
 
-    if (ifup(vethname.c_str(), port_id, true, true))
+    if (ifup(vethname.c_str(), port_id, false, false))
```
lguohan, line 482: why do you need this line at all? you have set the admin state right before enter this function.

kcudnik:
you are right, not needed

PR #605, opened 2020-04-16
Fix the classical if (x=y) bug

vasant17:
Fixing a bug introduced by me while cherry-pickinhg one of the commits from master to 201911
https://github.com/Azure/sonic-sairedis/pull/585


lguohan:
why include sai header update?

PR #606, opened 2020-04-16
Fixing classical if (x==y) bug introduced while cherry-pick merging

vasant17:


meta/sai_meta.cpp
```diff
@@ -6819,6 +6819,7 @@ sai_status_t meta_sai_flush_fdb_entries(
             meta_sai_on_fdb_flush_event_consolidated(data);
         }
     }
+    return status;
```
lguohan, line 6822: we need a space above this line.

PR #607, opened 2020-04-20
[py] Add sairedis python module skeleton

kcudnik:


python/test.py
```diff
@@ -0,0 +1,27 @@
...
+print r
+
+swid = r["oid"]
+
+r = get_switch_attribute(swid, "SAI_SWITCH_ATTR_PORT_LIST")
+print r
+
+args = dict()
+args["SAI_VLAN_ATTR_VLAN_ID"] = "11"
+r = create_vlan(swid, args)
```
lguohan, line 18: is this always synchronous call?

kcudnik: no, this is wrapper on sairedis library

lguohan: can you give example how to enable synchronous call for the python client?

lguohan: In this case, would it make sense to use synchronous call since the client is very much interactive, and would like to see if sai function calls fail or not.

kcudnik: currently synchronous mode is not implemented yet in py client, i would need to add it, and my proposal is to make it in separate PR

kcudnik: correction, this can't be synchronous, because if you have orchagent that runs asynchronous, then syncd was also started in asynchronous mode, do you can't suddenly speak to syncd from side channel.
also we spoke that you can't create new objects, or set new information, since OA will be not aware of those, so basically this tool will be like a inspector, and using GETs, which are always synchronous

lguohan: no, I am talking about standard alone mode where orchagent is not running. only this client.

kcudnik: for synchronous mode in case of failure syncd code needs to be updated, and probably lua scripts, since if for example create api will fail, today "pop" on consumer table puts data to redis database, but if api will fail, that data needs to be removed, similar in case of SET and opposite in case of remove api, lets sync on this online

daall:
@kcudnik can you check https://lgtm.com/projects/g/Azure/sonic-sairedis/logs/rev/pr-0807948cce856d1cd5a6dc9c5df6c819746c1bf9/lang:cpp/stage:Build%20merge_2a19632df2ff0aa98fb77d04a176f4cdd4e84a18 and fix the missing include? I'm not sure how it's being included in the other directories.

PR #610, opened 2020-05-06
Change log level 

vasant17:
Change log level for not implemented port attributes and skipped mac learning
What I did:

Change log level from warning/error to debug.

Why I did:

Due to execessive logging we were observing that, orchagent or syncd or bot get stuck in send() system call of logging. This was happening because rsyslog was NOT reading from socket buffer. Verified this by sending SIGSTOP and SIGCONT to rsyslogd. This problem happened everytime we ran Dynamic Port Brteakout scale test case. (test_port_dpb.py)

How I verified:

Ran DPB test case again in Jenkins, the problem did NOT occur


vasant17:
retest this please

PR #612, opened 2020-05-06
Revert "[py] Add sairedis python module skeleton"

daall:
Reverts Azure/sonic-sairedis#607


daall:
retest vs please

daall:
retest this please

daall:
retest this please

daall:
The LGTM merge build is working, but the checker is failing because it can't build master. I think this is OK. Still looking into the vs failures.

daall:
It's not clear to my why the revert is breaking the sai player tests. @kcudnik I don't know whether it's easier to fix this or fix the build issue from the original PR, just give me a heads up when you're able to take a look at this and we can figure out what to do next.

kcudnik:
i would like to fix master, not sure why its failing on that include, since that include file is present in swss-common

lguohan:
the problem is that python directory is not using Makefile.am, so that configure.ac is not generate the correct Makefile.

lguohan:
fixed in #615 

https://lgtm.com/logs/6fbd60b61c86ef989bf1a7034fb3b378773f38f7/lang:cpp

@kcudnik, let's make sure we merge the pr that has pass the lgtm test. it is a little bit difficult to fix it later.

PR #613, opened 2020-05-08
[syncd] Use steady clock for TimerWatchdog

tahmed-dev:
Clock can rollback in big jumps and this causes TimerWatchdog to through
and crashes syncd. This code uses steady clock instead. Steady clock is
guaranteed to be monotonically increasing clock.

signed-of-by: Tamer Ahmed <tamer.ahmed@microsoft.com>


yxieca:
retest LGTM please

abdosi:
@tahmed-dev Can you please create PR 201911. Cherry-pick has conflict

PR #617, opened 2020-05-26
Add synchronous mode to sairedis library

kcudnik:
Will require https://github.com/Azure/sonic-swss-common/pull/344 to work

### DIFF
--- a/saiplayer/CommandLineOptions.cpp
+++ b/saiplayer/CommandLineOptions.cpp
@@ -17,6 +17,7 @@ CommandLineOptions::CommandLineOptions()
     m_skipNotifySyncd = false;
     m_enableDebug = false;
     m_sleep = false;
+    m_syncMode = false;
 }
 
 std::string CommandLineOptions::getCommandLineString() const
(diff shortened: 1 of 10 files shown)


meta/saiserialize.cpp
```diff
@@ -1934,6 +1934,13 @@ void sai_deserialize_enum(
         }
     }
 
+    // for backward compatibility from SAI v1.6
+    if (s == "SAI_NEXT_HOP_GROUP_TYPE_ECMP")
```
lguohan, line 1938: i can see see this in the SAI header, why do we need this?

slogan621: This does in fact squash a test failure that I was observing. Not an argument for or against, just observation.

kcudnik:
retest this please

meta/saiserialize.cpp
kcudnik: this is fix for sai header 1.6 since that enum was marked as ignore, and is there just for backward compatibility to compile, but is no longer added to metadata to check for deserialize

kcudnik:
retest this please

PR #618, opened 2020-05-29
[MultiDB] use get API to obtain dbid instead of hardcode value

dzhangalibaba:
* use get API to obtain dbid instead of hardcode value

### DIFF
--- a/syncd/FlexCounter.cpp
+++ b/syncd/FlexCounter.cpp
@@ -1454,7 +1454,7 @@ void FlexCounter::runPlugins(
 
     const std::vector<std::string> argv =
     {
-        std::to_string(COUNTERS_DB), // TODO
+        std::to_string(db.getDbId()),
         COUNTERS_TABLE,
         std::to_string(m_pollInterval * 1000)
     };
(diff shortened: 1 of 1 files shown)


dzhangalibaba:
@kcudnik   please take a look , we can get DB id from db connector instead of hardcode COUNTERS_DB(2) . Please make sure the db is always a counters db connector, my understanding this is true.

dzhangalibaba:
@qiluo-msft  please take a look as well

syncd/FlexCounter.cpp
```diff
@@ -1454,7 +1454,7 @@ void FlexCounter::runPlugins(
 
     const std::vector<std::string> argv =
     {
-        std::to_string(COUNTERS_DB), // TODO
+        std::to_string(db.getDbId()),
```
lguohan, line 1457: it db is always counter db, can you rename it to cdb?

dzhangalibaba: sure. updated

dzhangalibaba:
@kcudnik  please take a look

PR #624, opened 2020-06-17
sonic-sairedis: Add support to sonic-sairedis for gearbox phys

sydlogan:
* builds on support for multiple switches in sonic-sairedis
* new vslib switch BCM81724 implements a virtual gearbox phy.
* support for launching second (BCM81724 is supported by its own syncd)
* simple refactoring of tests to support switches by part number, still working with
  sairedis to support multiple switch in tests so BCM81724 will be a separate pull request)
* changed example context_config.json to reflect renaming of phy REDIS tables (see sonic-swss-common
  commit 292b08a3a80b24b23663020b37e6260039a311c0)

Note that a future commit to sonic-buildimage will be required to trigger launch of physyncd
(launching is based on device config files which are currently not present in sonic-buildimage).

Testing done in multiple environments (broadcom fork and pure upstream). Example CLI output
based on changes pushed to sonic-utilities (commit a6c4456f6965b79bf9d02ff1962070a5eae6ea55)
running in VS switch supporting BCM81724:

root@sonic:/home/admin# show gearbox interfaces  status
  PHY Id    Interface    MAC Lanes    MAC Lane Speed    PHY Lanes    PHY Lane Speed    Line Lanes    Line Lane Speed    Oper    Admin
--------  -----------  -----------  ----------------  -----------  ----------------  ------------  -----------------  ------  -------
       1    Ethernet0  25,26,27,28               10G      200,201               20G           206                40G      up       up
       1    Ethernet4  29,30,31,32               10G      202,203               20G           207                40G      up       up
       1    Ethernet8  33,34,35,36               10G      204,205               20G           208                40G      up       up

HLD is located at https://github.com/Azure/SONiC/blob/b817a12fd89520d3fd26bbc5897487928e7f6de7/doc/gearbox/gearbox_mgr_design.md

  Signed-off-by: syd.logan@broadcom.com


lib/src/RedisChannel.cpp
```diff
@@ -175,7 +175,7 @@ sai_status_t RedisChannel::wait(
 
         swss::Selectable *sel;
 
-        int result = s.select(&sel, REDIS_ASIC_STATE_COMMAND_GETRESPONSE_TIMEOUT_MS);
+        int result = s.select(&sel, REDIS_ASIC_STATE_COMMAND_GETRESPONSE_TIMEOUT_MS * 10);
```
kcudnik, line 178: please not use * 10

syncd/SaiSwitch.cpp
```diff
@@ -59,13 +59,19 @@ SaiSwitch::SaiSwitch(
 
     helperInternalOids();
 
-    helperCheckLaneMap();
+    if (getSwitchType() == SAI_SWITCH_TYPE_NPU) 
+    {
+      helperCheckLaneMap();
```
kcudnik, line 64: please fix ident

syncd/SaiSwitch.cpp
```diff
@@ -131,6 +137,26 @@ void SaiSwitch::getDefaultMacAddress(
...
+
+    sai_attribute_t attr;
+
+    attr.id = SAI_SWITCH_ATTR_TYPE;
+
+    sai_status_t status = m_vendorSai->get(SAI_OBJECT_TYPE_SWITCH, m_switch_rid, 1, &attr);
+
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("failed to get switch type");
```
kcudnik, line 152: in case of failure we should throw here, or return explicitly default switch type

vslib/inc/saivs.h
```diff
@@ -5,6 +5,10 @@ extern "C" {
 }
 
 #define SAI_KEY_VS_SWITCH_TYPE              "SAI_VS_SWITCH_TYPE"
+#define SAI_KEY_VS_SAI_SWITCH_TYPE          "SAI_VS_SAI_SWITCH_TYPE"
+
+#define SAI_VALUE_SAI_SWITCH_TYPE_NPU       "SAI_SWITCH_TYPE_NPU"
```
kcudnik, line 10: why you need this ? if i understand correctly this is attribute on switch_create

vslib/src/Makefile.am
```diff
@@ -45,6 +45,7 @@ libSaiVS_a_SOURCES = \
 					  SelectableFd.cpp \
 					  SwitchState.cpp \
 					  SwitchBCM56850.cpp \
+            SwitchBCM81724.cpp \
```
kcudnik, line 48: fix ident

vslib/src/Sai.cpp
```diff
@@ -94,6 +94,18 @@ sai_status_t Sai::initialize(
...
     }
 
+    auto sai_switch_type = service_method_table->profile_get_value(0, SAI_KEY_VS_SAI_SWITCH_TYPE);
+    sai_switch_type_t saiSwitchType;
+
+    if (sai_switch_type == NULL)
+    {
+        SWSS_LOG_NOTICE("failed to obtain service method table value: %s", SAI_KEY_VS_SAI_SWITCH_TYPE);
+        saiSwitchType = SAI_SWITCH_TYPE_NPU;
+    } else if (!SwitchConfig::parseSaiSwitchType(sai_switch_type, saiSwitchType))
```
kcudnik, line 104: new line for else, also inside fix ident

vslib/src/Sai.cpp
```diff
@@ -94,6 +94,18 @@ sai_status_t Sai::initialize(
         return SAI_STATUS_FAILURE;
     }
 
+    auto sai_switch_type = service_method_table->profile_get_value(0, SAI_KEY_VS_SAI_SWITCH_TYPE);
+    sai_switch_type_t saiSwitchType;
+
+    if (sai_switch_type == NULL)
+    {
+        SWSS_LOG_NOTICE("failed to obtain service method table value: %s", SAI_KEY_VS_SAI_SWITCH_TYPE);
+        saiSwitchType = SAI_SWITCH_TYPE_NPU;
```
kcudnik, line 103: why you need this, i though this is switch_create attribute

vslib/src/Sai.cpp
```diff
@@ -161,7 +174,11 @@ sai_status_t Sai::initialize(
 
     startUnittestThread();
 
-    startFdbAgingThread();
+    if (saiSwitchType == SAI_SWITCH_TYPE_NPU) 
+    {
+
```
kcudnik, line 179: empty line and ident fix

vslib/src/SwitchConfig.cpp
```diff
@@ -18,6 +19,35 @@ SwitchConfig::SwitchConfig():
...
 
+bool SwitchConfig::parseSaiSwitchType(
+        _In_ const char* saiSwitchTypeStr,
+        _Out_ sai_switch_type_t& saiSwitchType)
+{
+    SWSS_LOG_ENTER();
+
+    std::string st = (saiSwitchTypeStr == NULL) ? "unknown" : saiSwitchTypeStr;
+
+    if (st == SAI_VALUE_SAI_SWITCH_TYPE_NPU)
```
kcudnik, line 30: same here, why this is needed, this is switch_create attribute

vslib/src/SwitchConfig.cpp
```diff
@@ -30,14 +60,19 @@ bool SwitchConfig::parseSwitchType(
...
+        switchType = SAI_VS_SWITCH_TYPE_BCM81724;
+    }
     else if (st == SAI_VALUE_VS_SWITCH_TYPE_MLNX2700)
     {
         switchType = SAI_VS_SWITCH_TYPE_MLNX2700;
     }
     else
     {
-        SWSS_LOG_ERROR("unknown switch type: '%s', expected (%s|%s)",
+        SWSS_LOG_ERROR("unknown switch type: '%s', expected (%s|%s)(%s)",
```
kcudnik, line 73: (%s|%s|%s)

lib/src/RedisChannel.cpp
sydlogan: This was not ideal approach, should have changed context, but this was debug code - at one point, we timed out. Just tried some experiments with current default of 6000 ms, seems ok, so will remove.

syncd/SaiSwitch.cpp
sydlogan: good catch

syncd/SaiSwitch.cpp
sydlogan: Ok, will throw here

vslib/src/Sai.cpp
sydlogan: We need, at this early state, to know if this is NPU, which supports the starting of fdb agent thread, or is PHY, which does not. We experienced failures if we start this thread for phy, and it is not desireable anyway. Introducing the switch type in the ini file as in this example, is how we determine type.

SAI_WARM_BOOT_READ_FILE=./sai_warmboot.bin
SAI_WARM_BOOT_WRITE_FILE=./sai_warmboot.bin
SAI_VS_SWITCH_TYPE=SAI_VS_SWITCH_TYPE_BCM81724
SAI_VS_SAI_SWITCH_TYPE=SAI_SWITCH_TYPE_PHY
SAI_VS_HOSTIF_USE_TAP_DEVICE=false
SAI_VS_USE_BCMSIM_LINK_MON=true
SAI_VS_INTERFACE_LANE_MAP_FILE=BCM81724/lanemapphy.ini

Another option would be to hard code a mapping of part number to type somewhere, but that hardly seems like a good solution (might as well put it here, where the switch resource is defined).

The field is considered optional (default is NPU) since that is the dominant switch type, and for backwards compatibility. Effectively, at this point, making it a boolean "isPhy" but if you consider that downstream there might be a third or fourth class of "switch", better to call it out explicitly.

vslib/src/SwitchConfig.cpp
sydlogan: Good catch.

lib/src/RedisChannel.cpp
kcudnik: currently timeout is 60 seconds, and you increases this to 10 minutes, please remove

sydlogan: Not sure I understand, did I not address this already?

https://github.com/sydlogan/sonic-sairedis/commit/8695a23e13596d4f7bec1f2b7ce40745f54dc921

vslib/src/SwitchConfig.cpp
slogan621: So that we know chip function and can avoid launch of fdb thread, which is not valid for phy.

syncd/scripts/physyncd_init_common.sh
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env bash
```
lguohan, line 1: can we reuse syncd_init_common.sh?

slogan621: Probably, so, an approach would be to pass an argument to the common script to indicate if we were launched by syncd_start.sh or physyncd_start.sh, and then put if statements around the code that does not make sense for a phy (for example, we would not be checking platform type and calling functions like config_syncd_vs, we would instead look at the argument we passed and call config_physyncd). Would that be acceptable?

slogan621: Addressed, please review

slogan621: @lguohan Addressed, thanks for the catch.

syncd/scripts/syncd_init_common.sh
```diff
@@ -1,6 +1,20 @@
...
 
-CMD_SYNCD=/usr/bin/syncd
+CHIP_TYPE="npu"
+
+while getopts p option
+do
+case "${option}"
+in
+p) CHIP_TYPE="phy";;
+esac
```
lguohan, line 10: spacing?

syncd/scripts/syncd_init_common.sh
```diff
@@ -1,6 +1,20 @@
...
+done
+
+if [ x"$CHIP_TYPE" == x"phy" ]; then
+    CMD_SYNCD=/usr/bin/physyncd
```
lguohan, line 14: where do you build physyncd? I do not see it. I suggest both use syncd name. then just provide different cmdline option.

syncd/scripts/physyncd_startup.py
```diff
@@ -0,0 +1,48 @@
...
+
+def main():
+
+    # Only privileged users can execute this command
+    if os.geteuid() != 0:
+        sys.exit("Root privileges required for this operation")
+
+    """ Loads json file """
+    try:
+        with open('/usr/share/sonic/hwsku/gearbox_config.json') as file_object:
```
lguohan, line 38: I would put this logic in a gbsyncd container. The reason is that the gb sai is managed differently from the switch sai. in the real switch, we would like to create two syncd containers one for gb sai, another for switch sai. Then, different vendor can update the syncd container them self. also, depending on platforms, we may have to build different gb containers into one image so different platform can choose to load them.

for now, I think it is ok, but we will need to refactor the code here.

syncd/scripts/syncd_init_common.sh
sydlogan: Yeah, something looks really wrong here on closer inspection, let me cycle back to the developer who wrote this and get some clarity. It looks like the script itself isn't even being used.

sydlogan: I've reverted the changes to syncd_init_common.sh. The intent here is for syncd to be launched via a python script, not via bash. The physyncd start script is no longer sourcing syncd_init_common, because it does not use the resulting CMD (and as you pointed out, there is no /usr/bin/physyncd). It was judged more straightforward to launch from python given its support for parsing JSON. Note we may have multiple syncds running in the physyncd docker, one per gearbox phy that is defined in (JSON) configuration.

sydlogan: @lguohan please re-review, re-approve, and, if possible, merge. Thanks!

lguohan:
lgtm, @kcudnik, please merge if you are ok with it.

sydlogan:
@kcudnik  and @lguohan thanks for the careful reviews, approval, merge.

PR #626, opened 2020-06-23
[submodule update] SAI submodule update to 1.6.3

smaheshm:
Change log:
```
Update SAI VoQ support (#1107) …
Voq system (#1081) …
[meta] Add support for ignored enum values (#1099)
TPID SAI proposal (#1089) …
ACL GRE key match (#1076) …
Add IPv6 NS and NA Traps (#1092) …
MACsec flow list attribute added in MACsec object (#1095) …
Add Enterprise Number for IPFIX Report Type (#1072) …
Provide TTL and QoS treatment during MPLS encap and decap (#1079)
Create and Set for Tunnel Attributes (#1086) …
```


smaheshm:
needs rebase

PR #633, opened 2020-07-02
Fixing hostif For Genetlink host interfaces

dgsudharsan:
**What I did**
Fixed the delete API for Hostif type genetlink in vs

**Why I did it**
The hostif type genetlink does not create any tap interfaces. Hence during remove there is no tap interfaces to be removed. Without the check to skip removing tap interfaces the remove hostif API would fail

**How I verified it**
The verification is done using manual testing and by pytests simluating create and remove genetlink hostif scenarios.

**Details if related**

### DIFF
--- a/vslib/src/SwitchStateBaseHostif.cpp
+++ b/vslib/src/SwitchStateBaseHostif.cpp
@@ -722,9 +722,30 @@ sai_status_t SwitchStateBase::vs_remove_hostif_tap_interface(
 
     sai_attribute_t attr;
 
+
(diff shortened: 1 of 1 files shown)


kcudnik:
just wondering if such check for getlink should be also added when creating tap interfaces ?

dgsudharsan:
> just wondering if such check for getlink should be also added when creating tap interfaces ?

The check is already present during create hostif. However it was not added during remove which might cause problems.

PR #634, opened 2020-07-04
Check sairedis submodule SAI headers against vendor SAI headers

kcudnik:


lguohan:
can this be part of build instead of test?

kcudnik:
yes, but why you want this as a build part ?

lguohan:
in some build cases we bypass the test. it is better to integrate with
autoconfigure.

kcudnik:
ok, will update

kcudnik:
retest this please

kcudnik:
retiring

PR #635, opened 2020-07-07
[syncd] Fix notification on shutdown request

kcudnik:
Fix notification on shutdown request

### DIFF
--- a/syncd/Syncd.cpp
+++ b/syncd/Syncd.cpp
@@ -3156,6 +3156,12 @@ void Syncd::sendShutdownRequest(
 {
     SWSS_LOG_ENTER();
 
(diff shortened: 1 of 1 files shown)


syncd/Syncd.cpp
```diff
@@ -3170,6 +3170,8 @@ void Syncd::sendShutdownRequest(
         return;
     }
 
+    s = sai_serialize_switch_shutdown_request(switchVid);
```
renukamanavalan, line 3173: This looks good. 
Can you please move the check, if (m_notifications == nullptr) to the start of the function, so we log the line, "sending switch_shutdown..." only when we could send.

PR #642, opened 2020-07-20
Add Break Before Make mechanism to syncd comparison logic

kcudnik:
* added break before make (BBM) mechanism
* added virtual switch resource limiter (to simulate real device sai object resource limit)
* added syncd comparison logic break config, for allowing only certain sai objects to take BBM path
* added unitttests for all 3 functionalities


syncd/ComparisonLogic.cpp
```diff
@@ -1915,6 +1932,192 @@ void ComparisonLogic::processObjectForViewTransition(
...
+        return;
+    }
+
+    if (currentBestMatch == nullptr)
+    {
+        // it can happen that current best match is not found for temporary
+        // object then we need to find best object for removal from existing
+        // ones the most suitable should be the one with most same attributes,
+        // since maybe only one read only attribute has been changed, and this
+        // will automatically result in null best match
```
daall, line 2101: Let's say we have two objects with identical attributes (a good example in SONiC would be an L3 ACL table and a MIRROR ACL table), and then between versions we add one attribute to the MIRROR table. Are we guaranteed that this new attribute will be added to the MIRROR table and not the L3 table? How does that reconciliation happen?

syncd/BestCandidateFinder.cpp
```diff
@@ -2358,6 +2358,117 @@ std::shared_ptr<SaiObj> BestCandidateFinder::findCurrentBestMatch(
...
+
+    for (const auto &currentObj: notProcessedObjects)
+    {
+        // log how many attributes current object have
+
+        SWSS_LOG_INFO("current obj %s, attrs: %zu",
+                currentObj->m_str_object_id.c_str(),
+                currentObj->getAllAttributes().size());
+
+        sai_object_compare_info_t soci = { 0, currentObj };
```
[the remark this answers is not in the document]
kcudnik: first letters of sai_object_compare_info

syncd/BestCandidateFinder.cpp
```diff
@@ -2358,6 +2358,117 @@ std::shared_ptr<SaiObj> BestCandidateFinder::findCurrentBestMatch(
...
+        SWSS_LOG_INFO("eq attributes: %ld vs %ld",
+            candidateObjects.at(0).equal_attributes,
+            candidateObjects.at(1).equal_attributes);
+
+        return candidateObjects.begin()->obj;
+    }
+
+    SWSS_LOG_WARN("same number of attributes equal, selecting first");
+
+    return candidateObjects.begin()->obj;
```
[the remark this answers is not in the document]
kcudnik: sort does this :P

syncd/ComparisonLogic.cpp
kcudnik: not sure if i understand your question, if you have 2 different objects, and then add another attribute to one of them, comparison logic make sure that this mirror object will have this new attribute, and object id will be set on other objects as compiled by orch agent. entire comparison logic process is complicated to describe it here, if you want to find out more we can have online call to discuss this

syncd/BestCandidateFinder.cpp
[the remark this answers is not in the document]
daall: ok fair. :P

syncd/ComparisonLogic.cpp
daall: I think I get your point, but I would still like to pick your brain about this sometime. :)

kcudnik:
retest this please

syncd/ComparisonLogic.cpp
kcudnik: sure :)

kcudnik:
retest this please

kcudnik:
retest this please

kcudnik:
retest this please

kcudnik:
retest this please

kcudnik:
retest this please

PR #643, opened 2020-07-21
[syncd] ARM32 bit fixes, for 64bit printf format specifier.

rajkumar38:
Signed-off-by: Rajkumar Pennadam Ramamoorthy <rpennadamram@marvell.com>


lguohan:
we should really do an arm64 build for this project. any idea how to setup the environment for this arm build?

rajkumar38:
@kcudnik 
When DEBUG was enabled, syncd was crashing in 32-bit CPU because of the format specifier. Fixed it. Pls review.

PR #645, opened 2020-07-21
Adding support for new SAI API sai_query_attribute_capability()

gechiang:
This change adds the support of allowing application to invoking SAI API "sai_query_attribute_capability()" from the sonic application side. 
This change has been tested using the Mellanox platform by exercising this new change to query the port object attribute SAI_PORT_ATTR_TPID

    status = sai_query_attribute_capability(gSwitchId, SAI_OBJECT_TYPE_PORT, SAI_PORT_ATTR_TPID, &capability);

as well as LAG object attribute SAI_LAG_ATTR_TPID.

    status = sai_query_attribute_capability(gSwitchId, SAI_OBJECT_TYPE_LAG, SAI_LAG_ATTR_TPID, &capability);

These two calls were exercised at orchagent init time in order to test out the changes are working.
I could not test this with BRCM platform due to the required SAI API "sai_query_attribute_capability()" although is defined in the SAI header file but was not part of the BRCM SAI code in 3.7.5.1 and causes link issue during build.  So the final change that I am asking to check in comments out the final call to "sai_query_attribute_capability()".  When BRCM SAI 4.2 (which has this SAI API support) is finally integrated into the master branch, I will have this uncommented out, tested and raise another PR to enable this full functionality.

Please see attached for the syslog and sairedis record captured during my testing using the Mellanox platform.
[AttrCapabilityQueryTestLog.txt](https://github.com/Azure/sonic-sairedis/files/4954988/AttrCapabilityQueryTestLog.txt)

Please note that this PR has a dependency on https://github.com/Azure/sonic-swss-common/pull/362
If this PR is needed in other branches, one must ensure that the change for https://github.com/Azure/sonic-swss-common/pull/362 is already picked up in that branch as well.

### DIFF
--- a/lib/inc/Recorder.h
+++ b/lib/inc/Recorder.h
@@ -290,7 +290,7 @@ namespace sairedis
                     _In_ sai_status_t status,
                     _In_ const uint64_t *count);
 
-            void recordQueryAattributeCapability(
+            void recordQueryAttributeCapability(
                     _In_ sai_object_id_t switch_id,
                     _In_ sai_object_type_t object_type,
                     _In_ sai_attr_id_t attr_id,
--- a/lib/inc/Recorder.h
+++ b/lib/inc/Recorder.h
@@ -296,7 +296,7 @@ namespace sairedis
                     _In_ sai_attr_id_t attr_id,
                     _Out_ sai_attr_capability_t *capability);
 
-            void recordQueryAattributeCapabilityResponse(
+            void recordQueryAttributeCapabilityResponse(
                     _In_ sai_status_t status,
                     _In_ sai_object_type_t objectType,
                     _In_ sai_attr_id_t attrId,
--- a/lib/inc/Recorder.h
+++ b/lib/inc/Recorder.h
@@ -290,6 +290,18 @@ namespace sairedis
                     _In_ sai_status_t status,
                     _In_ const uint64_t *count);
 
+            void recordQueryAattributeCapability(
+                    _In_ sai_object_id_t switch_id,
+                    _In_ sai_object_type_t object_type,
+                    _In_ sai_attr_id_t attr_id,
+                    _Out_ sai_attr_capability_t *capability);
+
+            void recordQueryAattributeCapabilityResponse(
+                    _In_ sai_status_t status,
+                    _In_ sai_object_type_t objectType,
+                    _In_ sai_attr_id_t attrId,
+                    _In_ const sai_attr_capability_t* capability);
+
             void recordQueryAattributeEnumValuesCapability(
                     _In_ sai_object_id_t switch_id,
                     _In_ sai_object_type_t object_type,
@@ -303,6 +315,14 @@ namespace sairedis
                     _In_ const sai_s32_list_t* enumValuesCapability);
 
             // TODO move to private
+            void recordQueryAttributeCapability(
+                    _In_ const std::string& key,
+                    _In_ const std::vector<swss::FieldValueTuple>& arguments);
+
+            void recordQueryAttributeCapabilityResponse(
+                    _In_ sai_status_t status,
+                    _In_ const std::vector<swss::FieldValueTuple>& arguments);
+
             void recordQueryAttributeEnumValuesCapability(
                     _In_ const std::string& key,
                     _In_ const std::vector<swss::FieldValueTuple>& arguments);
--- a/lib/inc/RedisRemoteSaiInterface.h
+++ b/lib/inc/RedisRemoteSaiInterface.h
@@ -240,6 +240,12 @@ namespace sairedis
                     _In_ const sai_attribute_t *attrList,
                     _Out_ uint64_t *count) override;
 
+            virtual sai_status_t queryAattributeCapability(
+                    _In_ sai_object_id_t switch_id,
+                    _In_ sai_object_type_t object_type,
+                    _In_ sai_attr_id_t attr_id,
+                    _Out_ sai_attr_capability_t *capability) override;
+
             virtual sai_status_t queryAattributeEnumValuesCapability(
                     _In_ sai_object_id_t switch_id,
                     _In_ sai_object_type_t object_type,
@@ -385,6 +391,9 @@ namespace sairedis
             sai_status_t waitForFlushFdbEntriesResponse();
 
         private: // SAI API response
+
+            sai_status_t waitForQueryAattributeCapabilityResponse(
+                    _Out_ sai_attr_capability_t* capability);
 
             sai_status_t waitForQueryAattributeEnumValuesCapabilityResponse(
                     _Inout_ sai_s32_list_t* enumValuesCapability);
--- a/lib/inc/RedisRemoteSaiInterface.h
+++ b/lib/inc/RedisRemoteSaiInterface.h
@@ -240,7 +240,7 @@ namespace sairedis
                     _In_ const sai_attribute_t *attrList,
                     _Out_ uint64_t *count) override;
 
-            virtual sai_status_t queryAattributeCapability(
+            virtual sai_status_t queryAttributeCapability(
                     _In_ sai_object_id_t switch_id,
                     _In_ sai_object_type_t object_type,
                     _In_ sai_attr_id_t attr_id,
@@ -392,7 +392,7 @@ namespace sairedis
 
         private: // SAI API response
 
-            sai_status_t waitForQueryAattributeCapabilityResponse(
+            sai_status_t waitForQueryAttributeCapabilityResponse(
                     _Out_ sai_attr_capability_t* capability);
 
             sai_status_t waitForQueryAattributeEnumValuesCapabilityResponse(
--- a/lib/inc/SaiInterface.h
+++ b/lib/inc/SaiInterface.h
@@ -242,6 +242,12 @@ namespace sairedis
                     _In_ const sai_attribute_t *attrList,
                     _Out_ uint64_t *count) = 0;
 
+            virtual sai_status_t queryAattributeCapability(
+                    _In_ sai_object_id_t switch_id,
+                    _In_ sai_object_type_t object_type,
+                    _In_ sai_attr_id_t attr_id,
+                    _Out_ sai_attr_capability_t *capability) = 0;
+
             virtual sai_status_t queryAattributeEnumValuesCapability(
                     _In_ sai_object_id_t switch_id,
                     _In_ sai_object_type_t object_type,
--- a/lib/inc/SaiInterface.h
+++ b/lib/inc/SaiInterface.h
@@ -242,7 +242,7 @@ namespace sairedis
                     _In_ const sai_attribute_t *attrList,
                     _Out_ uint64_t *count) = 0;
 
-            virtual sai_status_t queryAattributeCapability(
+            virtual sai_status_t queryAttributeCapability(
                     _In_ sai_object_id_t switch_id,
                     _In_ sai_object_type_t object_type,
                     _In_ sai_attr_id_t attr_id,
--- a/lib/src/RedisRemoteSaiInterface.cpp
+++ b/lib/src/RedisRemoteSaiInterface.cpp
@@ -895,38 +895,12 @@ sai_status_t RedisRemoteSaiInterface::waitForQueryAttributeCapabilityResponse(
             return SAI_STATUS_FAILURE;
         }
 
-        const uint32_t create_implemented = std::stoi(fvValue(values[0]));
-        const uint32_t set_implemented = std::stoi(fvValue(values[1]));
-        const uint32_t get_implemented = std::stoi(fvValue(values[2]));
+        capability->create_implemented = (fvValue(values[0]) == "true" ? true : false);
+        capability->set_implemented    = (fvValue(values[1]) == "true" ? true : false);
+        capability->get_implemented    = (fvValue(values[2]) == "true" ? true : false);
 
-        SWSS_LOG_DEBUG("Received payload: create_implemented:%d, set_implemented:%d, get_implemented:%d",
-            create_implemented, set_implemented, get_implemented);
-
-        if(create_implemented)
-        {
-            capability->create_implemented = true;
-        }
-        else
-        {
-            capability->create_implemented = false;
-        }
-        if(set_implemented)
-        {
-            capability->set_implemented = true;
-        }
-        else
-        {
-            capability->set_implemented = false;
-        }
-
-        if(get_implemented)
-        {
-            capability->get_implemented = true;
-        }
-        else
-        {
-            capability->get_implemented = false;
-        }
+        SWSS_LOG_DEBUG("Received payload: create_implemented:%s, set_implemented:%s, get_implemented:%s",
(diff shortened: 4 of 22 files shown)


lib/src/RedisRemoteSaiInterface.cpp
```diff
@@ -827,6 +827,106 @@ sai_status_t RedisRemoteSaiInterface::waitForObjectTypeGetAvailabilityResponse(
...
+            capability->set_implemented = true;
+        else
+            capability->set_implemented = false;
+
+        if(get_implemented)
+            capability->get_implemented = true;
+        else
+            capability->get_implemented = false;
+    }
+    else if (status ==  SAI_STATUS_BUFFER_OVERFLOW)
```
kcudnik, line 920: overflow is not needed here, since there is no "count" element in capability

meta/Meta.cpp
```diff
@@ -1856,6 +1856,37 @@ sai_status_t Meta::objectTypeGetAvailability(
     return status;
 }
 
+sai_status_t Meta::queryAattributeCapability(
+        _In_ sai_object_id_t switchId,
+        _In_ sai_object_type_t objectType,
+        _In_ sai_attr_id_t attrId,
+        _Out_ sai_attr_capability_t *Capability)
```
kcudnik, line 1863: minor, Capability -> capability

lib/src/RedisRemoteSaiInterface.cpp
```diff
@@ -827,6 +827,106 @@ sai_status_t RedisRemoteSaiInterface::waitForObjectTypeGetAvailabilityResponse(
...
+    const std::vector<swss::FieldValueTuple> entry =
+    {
+        swss::FieldValueTuple("OBJECT_TYPE", object_type_str),
+        swss::FieldValueTuple("ATTR_ID", attr_id_str)
+    };
+
+    SWSS_LOG_DEBUG(
+            "Query arguments: switch %s, object type: %s, attribute: %s",
+            switch_id_str.c_str(),
+            object_type_str.c_str(),
```
kcudnik, line 860: in all other places we tray to keep format prefixed by str like "strObjectType", coult you keep that also in all your variables?

syncd/Syncd.cpp
```diff
@@ -310,6 +313,59 @@ sai_status_t Syncd::processSingleEvent(
...
+    if (status == SAI_STATUS_SUCCESS)
+    {
+        entry =
+        {
+            swss::FieldValueTuple("CREATE_IMPLEMENTED", std::to_string(capability.create_implemented)),
+            swss::FieldValueTuple("SET_IMPLEMENTED", std::to_string(capability.set_implemented)),
+            swss::FieldValueTuple("GET_IMPLEMENTED", std::to_string(capability.get_implemented))
+        };
+
+        SWSS_LOG_DEBUG("Sending response: create_implemented:%d, set_implemented:%d, get_implemented:%d",
```
kcudnik, line 360: you could change that to LOG_INFO

vslib/src/Sai.cpp
```diff
@@ -706,6 +706,23 @@ sai_status_t Sai::objectTypeGetAvailability(
...
+        _In_ sai_object_id_t switch_id,
+        _In_ sai_object_type_t object_type,
+        _In_ sai_attr_id_t attr_id,
+        _Out_ sai_attr_capability_t *capability)
+{
+    MUTEX();
+    SWSS_LOG_ENTER();
+    VS_CHECK_API_INITIALIZED();
+
+    return m_meta->queryAattributeCapability(
```
kcudnik, line 719: double "Aa" in name

vslib/src/VirtualSwitchSaiInterface.cpp
```diff
@@ -789,6 +789,37 @@ sai_status_t VirtualSwitchSaiInterface::objectTypeGetAvailability(
...
+        _In_ sai_object_type_t object_type,
+        _In_ sai_attr_id_t attr_id,
+        _Out_ sai_attr_capability_t *capability)
+{
+    SWSS_LOG_ENTER();
+
+    // TODO: We should generate this metadata for the virtual switch rather
+    // than hard-coding it here.
+
+    if (object_type == SAI_OBJECT_TYPE_PORT && attr_id == SAI_PORT_ATTR_TPID)
```
kcudnik, line 803: in virtual switch by default all apis are implemented for all objects, so you could just set true for all without if's

lib/inc/Recorder.h
```diff
@@ -290,6 +290,18 @@ namespace sairedis
                     _In_ sai_status_t status,
                     _In_ const uint64_t *count);
 
+            void recordQueryAattributeCapability(
```
kcudnik, line 293: double "Aa" in name

lib/inc/Recorder.h
```diff
@@ -290,6 +290,18 @@ namespace sairedis
...
+                    _In_ sai_object_id_t switch_id,
+                    _In_ sai_object_type_t object_type,
+                    _In_ sai_attr_id_t attr_id,
+                    _Out_ sai_attr_capability_t *capability);
+
+            void recordQueryAattributeCapabilityResponse(
```
kcudnik, line 299: double "Aa" in name

lib/inc/RedisRemoteSaiInterface.h
```diff
@@ -386,6 +392,9 @@ namespace sairedis
 
         private: // SAI API response
 
+            sai_status_t waitForQueryAattributeCapabilityResponse(
```
kcudnik, line 395: double "Aa" in name

lib/inc/SaiInterface.h
```diff
@@ -242,6 +242,12 @@ namespace sairedis
                     _In_ const sai_attribute_t *attrList,
                     _Out_ uint64_t *count) = 0;
 
+            virtual sai_status_t queryAattributeCapability(
```
kcudnik, line 245: double "Aa" in name

lib/src/RedisRemoteSaiInterface.cpp
```diff
@@ -827,6 +827,106 @@ sai_status_t RedisRemoteSaiInterface::waitForObjectTypeGetAvailabilityResponse(
...
+        const std::vector<swss::FieldValueTuple> &values = kfvFieldsValues(kco);
+
+        if (values.size() != 3)
+        {
+            SWSS_LOG_ERROR("Invalid response from syncd: expected 3 values, received %zu", values.size());
+
+            return SAI_STATUS_FAILURE;
+        }
+
+        const uint32_t create_implemented = std::stoi(fvValue(values[0]));
```
kcudnik, line 898: those are numbers ? instead of string "true"/"false" ?

vslib/src/Sai.cpp
```diff
@@ -706,6 +706,23 @@ sai_status_t Sai::objectTypeGetAvailability(
             count);
 }
 
+sai_status_t Sai::queryAattributeCapability(
+        _In_ sai_object_id_t switch_id,
+        _In_ sai_object_type_t object_type,
+        _In_ sai_attr_id_t attr_id,
+        _Out_ sai_attr_capability_t *capability)
```
kcudnik, line 713: try using name convention like switchId, objectType, attrId

rlhui:
Typo in the PR title.

lib/src/RedisRemoteSaiInterface.cpp
gechiang: I Agree.  This is now removed in my new change.

meta/Meta.cpp
gechiang: Renamed to use lower case for variable "capability" and renamed the function name to remove the extra "a" all across files that referenced this function name.

lib/src/RedisRemoteSaiInterface.cpp
gechiang: Renamed to follow the suggested naming convention for the variables introduced.

syncd/Syncd.cpp
gechiang: log level changed to INFO

vslib/src/Sai.cpp
gechiang: renamed the function name to remove the extra "a"

vslib/src/VirtualSwitchSaiInterface.cpp
gechiang: Added the comment as you suggested and removed the checking to always return success regardless of the Attribute involved and set them all implemented.

lib/src/RedisRemoteSaiInterface.cpp
```diff
@@ -827,6 +827,100 @@ sai_status_t RedisRemoteSaiInterface::waitForObjectTypeGetAvailabilityResponse(
...
+
+        if(set_implemented)
+            capability->set_implemented = true;
+        else
+            capability->set_implemented = false;
+
+        if(get_implemented)
+            capability->get_implemented = true;
+        else
+            capability->get_implemented = false;
```
daall, line 918: I would still add curly braces here just to be explicit.

meta/Meta.cpp
```diff
@@ -1856,6 +1856,37 @@ sai_status_t Meta::objectTypeGetAvailability(
...
+        _In_ sai_attr_id_t attrId,
+        _Out_ sai_attr_capability_t *capability)
+{
+    SWSS_LOG_ENTER();
+
+    PARAMETER_CHECK_OID_OBJECT_TYPE(switchId, SAI_OBJECT_TYPE_SWITCH);
+    PARAMETER_CHECK_OID_EXISTS(switchId, SAI_OBJECT_TYPE_SWITCH);
+    PARAMETER_CHECK_OBJECT_TYPE_VALID(objectType);
+
+    auto mdp = sai_metadata_get_attr_metadata(objectType, attrId);
```
daall, line 1871: What is `mdp`? I would make the name more specific.

syncd/VendorSai.cpp
```diff
@@ -941,6 +941,27 @@ sai_status_t VendorSai::objectTypeGetAvailability(
...
+
+#if 0 // Wait until BRCM fix their SAI issue then enable this
+    return sai_query_attribute_capability(
+            switchId,
+            objectType,
+            attrId,
+            capability);
+#else
+    return SAI_STATUS_NOT_IMPLEMENTED;
+#endif
```
daall, line 962: What is the behavior if this preprocessor guard is not here? Will this crash if the method is called, or will it crash on startup, or something else?

lib/inc/RedisRemoteSaiInterface.h
gechiang: extra "a" removed from function name.

lib/inc/SaiInterface.h
gechiang: Extra "a" removed from function name.

lib/src/RedisRemoteSaiInterface.cpp
gechiang: yes they are number strings "0 or 1" instead of "true/false"
See syncd/Syncd.cpp:
entry =
        {
            swss::FieldValueTuple("CREATE_IMPLEMENTED", std::to_string(capability.create_implemented)),
            swss::FieldValueTuple("SET_IMPLEMENTED", std::to_string(capability.set_implemented)),
            swss::FieldValueTuple("GET_IMPLEMENTED", std::to_string(capability.get_implemented))
        };

meta/Meta.cpp
gechiang: I think it stands for meta data pointer. I am merely follow the convention setup in this file with all previous existing functions that make the same call with same naming.

syncd/VendorSai.cpp
gechiang: BRCM did not implement this and it will cause build issue during link time.  So the agreement is to comment this actual Vendor SAI call until BRCM is ready in SAI 4.2 and patch in 3.7.5.1 in other branches if this API is needed there.
At that time I will have to raise another PR to re-enable this section of code.

lib/inc/Recorder.h
gechiang: removed the extra "a" from the function name.

lib/inc/Recorder.h
gechiang: Removed the extra 'a" from function name.

lib/src/RedisRemoteSaiInterface.cpp
gechiang: Braces added.

syncd/VendorSai.cpp
daall: At that point, what would happen if we tried to build this against some other vendor's SAI?

gechiang: Spoke with Kamil on this and currently there is no platform based compiler flags that differentiate the different platforms that we support.  If such compiler flags were available, we could make this common code to work for certain platforms while avoid build issue for those that did not implement the stub code for API they do not yet support.  The expectation is that when a SAI header is released, all the SAI vendors that agreed for such support should ensure that their SAI release should include the handling of all the defined SAI APIs whether it is fully implemented or not.  In this particular case although all the code were implemented in SAIREDIS by this PR and for platform such as Mellanox this was tested to be working fine, we are disabling this API for now at the lowest level until we incorporate a new SAI release drop from BRCM (4.2) before we can open it up for all platforms.  Alternatively is to define these platform based compiler flags now and incorporate them in the Makefiles involved so that we can have more control over which platform to turn on/off for a particular SAI API.

daall: Makes sense, thanks!

lib/src/RedisRemoteSaiInterface.cpp
kcudnik: could you make them as true false strings? all other bool that are serialized are strings like that

lib/src/RedisRemoteSaiInterface.cpp
```diff
@@ -903,19 +903,30 @@ sai_status_t RedisRemoteSaiInterface::waitForQueryAttributeCapabilityResponse(
             create_implemented, set_implemented, get_implemented);
 
         if(create_implemented)
+        {
```
kcudnik, line 906: you could use 
capability->create_implemented = (create_implemnted) ? true : false;

it would save like 20 lines of code

lib/src/RedisRemoteSaiInterface.cpp
gechiang: Agreed. I have made the change to not use "number string" to follow the common practice to use "true/false string".
In order to make this change I have also changed syncd/Syncd.cpp line 355-361 to write with "true/false string".
Please help review that portion of code as well in addition to this change here.

lib/src/RedisRemoteSaiInterface.cpp
gechiang: As part of the change to not use "number string" this part of the code has changed accordingly and it did simplify the coding.  please review my last change for this.

gechiang:
> Please address comments

All comments have been addressed. Please help take a look at my last change and see if anything else needs to be addressed.
Thanks!

PR #650, opened 2020-08-02
Fix synchronous with bulk API

shi-su:
- Why I did it
This PR contains the code changes so that the synchronous bulk API processes all events.

- How I did it
Let the syncUpdateRedisBulkQuadEvent function loop across all events in a bulk rather than returning after processing the first one.

- How to verify it
Verify route entries in ASIC_DB after loading multiple routes using swssconfig in vs tests with synchronous mode enabled.


kcudnik:
thanks for this fix!

PR #651, opened 2020-08-05
Re-enable support for new SAI API sai_query_attribute_capability()

gechiang:
Previously the code to support query SAI attribute capabilities were commented out due to discovered missing BRCM Vendor SAI API missing in their 3.7.5.1 debian package that caused linking error during image build.  
See https://github.com/Azure/sonic-sairedis/pull/645 for details on the SAI attribute capability query support.

Since BRCM has provided the patch for the missing API stub code and included them in BRCM SAI 3.7.5.1-2 (see https://github.com/Azure/sonic-buildimage/pull/5101), I am raising this PR to re-enable the calling to Vendor SAI API sai_query_attribute_capability().

I have tested this with the new BRCM SAI 3.7.5.1-2 debians and ensured that the linking issue was fixed by that patch.

### DIFF
--- a/syncd/VendorSai.cpp
+++ b/syncd/VendorSai.cpp
@@ -951,15 +951,11 @@ sai_status_t VendorSai::queryAttributeCapability(
     SWSS_LOG_ENTER();
     VENDOR_CHECK_API_INITIALIZED();
 
-#if 0 // Wait until BRCM fix their SAI issue then enable this
     return sai_query_attribute_capability(
             switchId,
             objectType,
             attrId,
             capability);
-#else
-    return SAI_STATUS_NOT_IMPLEMENTED;
-#endif
 }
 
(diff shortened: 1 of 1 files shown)


gechiang:
retest vs please

gechiang:
Looks like the VS build/test is broken.  The change I made has nothing to do with any of the failure.  Infact.  The change has no functional impact to VS at all...

gechiang:
Guohan Confirmed that the VS Testing was broken by another PR not related to this one.
Since this change is very limited in the area that is not VS related, If code review looks good we should go ahead with the merge.
If one of you can kindly do the 3 line change code review and grant merge permission that would be great.
Thanks!

gechiang:
retest vs please

gechiang:
retest vs please

PR #653, opened 2020-08-28
Prevent syncd from stopping swss in the presence of SAI failures in synchronous mode

shi-su:
**What I did**
Let syncd send statuses to orchagent rather than stopping swss in the presence of SAI failures in synchronous mode.

**Why I did it**
I expect syncd to send the failure status to orchagent and let orchagent process the failure in synchronous mode when SAI failures happen. Currently, syncd would throw an exception and stop swss before the failure information is sent to orchagent. 

**How I did it**
Let syncd log the error message rather than throw the exception. As such, the syncd would not trigger swss shutdown in the presence of SAI failures.

**How I verified it**
Test locally with a virtual switch.

### DIFF
--- a/syncd/Syncd.cpp
+++ b/syncd/Syncd.cpp
@@ -1244,15 +1244,6 @@ void Syncd::sendApiResponse(
 
     if (status != SAI_STATUS_SUCCESS)
     {
-        /*
-         * TODO: If api fill fail in sync mode, we need to take some actions to
-         * handle that, and those are not trivial tasks:
-         *
-         * - in case create fail, remove object from redis database
-         * - in case remove fail, bring back removed object to database
-         * - in case set fail, bring back previous value if it was set
-         */
-
         SWSS_LOG_ERROR("api %s failed in syncd mode: %s, FIXME",
                     sai_serialize_common_api(api).c_str(),
                     sai_serialize_status(status).c_str());
--- a/syncd/Syncd.cpp
+++ b/syncd/Syncd.cpp
@@ -1244,7 +1244,7 @@ void Syncd::sendApiResponse(
 
     if (status != SAI_STATUS_SUCCESS)
     {
-        SWSS_LOG_ERROR("api %s failed in syncd mode: %s, FIXME",
+        SWSS_LOG_ERROR("api %s failed in syncd mode: %s",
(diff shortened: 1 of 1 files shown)


qiluo-msft:
>          * TODO: If api fill fail in sync mode, we need to take some actions to

Is this TODO still valid in sync mode? If they are already implemented, we can remove TODO

---
Refers to: syncd/Syncd.cpp:1248 in 8c6b1a9. [](commit_id = 8c6b1a98abd8737ea2499e57878dab98842ffe97, deletion_comment = False)

shi-su:
> > ```
> >      * TODO: If api fill fail in sync mode, we need to take some actions to
> > ```
> 
> Is this TODO still valid in sync mode? If they are already implemented, we can remove TODO
> 
> Refers to: syncd/Syncd.cpp:1248 in [8c6b1a9](https://github.com/Azure/sonic-sairedis/commit/8c6b1a98abd8737ea2499e57878dab98842ffe97). [](commit_id = [8c6b1a9](https://github.com/Azure/sonic-sairedis/commit/8c6b1a98abd8737ea2499e57878dab98842ffe97), deletion_comment = False)



> > ```
> >      * TODO: If api fill fail in sync mode, we need to take some actions to
> > ```
> 
> Is this TODO still valid in sync mode? If they are already implemented, we can remove TODO
> 
> Refers to: syncd/Syncd.cpp:1248 in [8c6b1a9](https://github.com/Azure/sonic-sairedis/commit/8c6b1a98abd8737ea2499e57878dab98842ffe97). [](commit_id = [8c6b1a9](https://github.com/Azure/sonic-sairedis/commit/8c6b1a98abd8737ea2499e57878dab98842ffe97), deletion_comment = False)

According to my observations in tests, this is implemented to handle asic_db. I think I should also confirm with @kcudnik if this TODO is ready for removal and I will update it accordingly.

kcudnik:
Ok, this is already handled in syncUpdateRedisBulkQuadEvent method, i forgot to remove that check from the code

shi-su:
> Ok, this is already handled in syncUpdateRedisBulkQuadEvent method, i forgot to remove that check from the code

Thank you, Kamil @kcudnik ! Removed the TODO message.

kcudnik (approved):

PR #654, opened 2020-08-29
Fix Warmboot Issue when upgraded Image SAI return Switch Internal OID not accounted in previous image.

abdosi:
Why I did:
Fix is for the issue as reported in https://github.com/Azure/sonic-buildimage/issues/5274
Fix https://github.com/Azure/sonic-buildimage/issues/5274

Basically Switch Internal OID should always be accounted in Temp and
Current Logic Comparison. It should never trigger remove operation.

Changes is to always add Switch Internal OID to COLDVIDS (even in case of
warm-boot)

Signed-off-by: Abhishek Dosi <abdosi@microsoft.com>

How I verfiy:

a) After change above issue is resolved
b) Dump COLDVIDS/ASIC_VIEW/HIDDEN Table and Switch Internal OID's are accounted in all 3 tables

### DIFF
--- a/syncd/SaiSwitch.cpp
+++ b/syncd/SaiSwitch.cpp
@@ -517,7 +517,7 @@ sai_object_id_t SaiSwitch::helperGetSwitchAttrOid(
          * Redis value of this attribute is not present yet, save it!
          */
 
-        redisSetDummyAsicStateForRealObjectId(rid);
+        redisSaveInternalOids(rid);
 
         SWSS_LOG_INFO("redis %s id is not defined yet in redis", meta->attridname);
 
@@ -742,6 +742,25 @@ std::set<sai_object_id_t> SaiSwitch::getWarmBootDiscoveredVids() const
     SWSS_LOG_ENTER();
 
     return m_warmBootDiscoveredVids;
+}
+
+void SaiSwitch::redisSaveInternalOids(_In_ sai_object_id_t rid) const
+{
+    SWSS_LOG_ENTER();
+
+    std::set<sai_object_id_t> coldVids;
+
+    sai_object_id_t vid = m_translator->translateRidToVid(rid, m_switch_vid);
+
+    coldVids.insert(vid);
+
+    m_client->setDummyAsicStateObject(vid);
+
+    m_client->saveColdBootDiscoveredVids(m_switch_vid, coldVids);
+
+    SWSS_LOG_NOTICE("put switch internal discovered rid %s to Asic View and COLDVIDS", 
+                    sai_serialize_object_id(rid).c_str());
+
 }
 
 void SaiSwitch::redisSaveColdBootDiscoveredVids() const
--- a/syncd/SaiSwitch.cpp
+++ b/syncd/SaiSwitch.cpp
@@ -611,6 +611,20 @@ bool SaiSwitch::isNonRemovableRid(
         SWSS_LOG_THROW("NULL rid passed");
     }
 
+    /*
+     * Check for SAI_SWITCH_ATTR_DEFAULT_* oids like cpu, default virtual
+     * router.  Those objects can't be removed if user ask for it.
+     */
+
+    /* Here we are checking for isSwitchObjectDefaultRid first then ColdBootDiscoveredRid
+     * as it is possible we can discover switch Internal OID as part of warm-booot also especially
+     * when we are doing SAI upgrade as part of warm-boot.*/
+
+    if (isSwitchObjectDefaultRid(rid))
+    {
+        return true;
+    }
+
     if (!isColdBootDiscoveredRid(rid))
     {
         /*
@@ -618,16 +632,6 @@ bool SaiSwitch::isNonRemovableRid(
          */
 
         return false;
-    }
-
-    /*
-     * Check for SAI_SWITCH_ATTR_DEFAULT_* oids like cpu, default virtual
-     * router.  Those objects can't be removed if user ask for it.
-     */
-
-    if (isSwitchObjectDefaultRid(rid))
-    {
-        return true;
     }
 
     sai_object_type_t ot = m_vendorSai->objectTypeQuery(rid);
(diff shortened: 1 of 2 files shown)


kcudnik:
Could you explain this more ? i dont understand this issue here

syncd/SaiSwitch.cpp
```diff
@@ -744,6 +744,25 @@ std::set<sai_object_id_t> SaiSwitch::getWarmBootDiscoveredVids() const
...
+
+    std::set<sai_object_id_t> coldVids;
+
+    sai_object_id_t vid = m_translator->translateRidToVid(rid, m_switch_vid);
+
+    coldVids.insert(vid);
+
+    m_client->setDummyAsicStateObject(vid);
+
+    m_client->saveColdBootDiscoveredVids(m_switch_vid, coldVids);
```
kcudnik, line 759: when doing discovery on switch when switch is created, all those "default" objects are already discovered and put as cold vids to the database, why this is needed again ?

kcudnik:
@abdosi  when switch is created, there is discovery logic that discovers all the oids, it does not matter whether its cold or warm boot
so at new instance 3.7 SAI_SWITCH_ATTR_DEFAULT_STP_INST_ID  will be discovered, as it was not present on 3.5
and this should be done in different way, SAI_SWITCH_ATTR_DEFAULT_STP_INST_ID should be added to m_default_rid_map in SaiSwitch.cpp, and then when function isSwitchObjectDefaultRid() inside isNonRemovableRid(), and isNonRemovableRid is called in ComparisonLogic, then that object will not be removed, since that is default object

syncd/SaiSwitch.h
```diff
@@ -299,6 +299,8 @@ namespace syncd
 
             void helperInternalOids();
 
+            void redisSaveInternalOids(_In_ sai_object_id_t rid) const;
```
kcudnik, line 302: put _In_ in new line

syncd/SaiSwitch.cpp
```diff
@@ -744,6 +744,25 @@ std::set<sai_object_id_t> SaiSwitch::getWarmBootDiscoveredVids() const
     return m_warmBootDiscoveredVids;
 }
 
+void SaiSwitch::redisSaveInternalOids(_In_ sai_object_id_t rid) const
```
kcudnik, line 747: put _In_ in new line

syncd/SaiSwitch.cpp
kcudnik, line 759: put extensive comment on this line, why this needs to be done,  give example witht that default object

lguohan:
can we have some kinds of unit test for this?

abdosi:
> @abdosi when switch is created, there is discovery logic that discovers all the oids, it does not matter whether its cold or warm boot
> so at new instance 3.7 SAI_SWITCH_ATTR_DEFAULT_STP_INST_ID will be discovered, as it was not present on 3.5
> and this should be done in different way, SAI_SWITCH_ATTR_DEFAULT_STP_INST_ID should be added to m_default_rid_map in SaiSwitch.cpp, and then when function isSwitchObjectDefaultRid() inside isNonRemovableRid(), and isNonRemovableRid is called in ComparisonLogic, then that object will not be removed, since that is default object

@kcudnik : Thanks looks like logic is different in master vs 201911. I will rework this PR for 201911 branch. Master looks fine.

kcudnik:
Ok, so master works fine without any issues? just 201911 is affected?

as for unittests, this will be tricky to implement in current test mechanism, since it would require to have 2 different version of syncd (compiled with version A and then B) and cold boot from A to warm boot to B, or to have a new special configuration file for virtual switch that on warm boot it could be specified to "enable" new STP object to be existing on the VS

abdosi:
> Ok, so master works fine without any issues? just 201911 is affected?
> 
I revisited the code-path you mention and issue will be present both in master and 201911. I am verifying on 201911 as of now. Here are two issue in above approach -

Issue 1: 

In Function isNonRemovableRid() we first check for isColdBootDiscoveredRid() and then isSwitchObjectDefaultRid(). In This particular case DefaultRid was dicovered as part of Warm-boot so first check will fail and return. 
I think the order should be reverse.  Let me know.


Issue 2:
Even after we reverse the order (as mention in Issue 1 above ) the API ComparisonLogic::checkMap() will fail. This is because since this discovered Switch Internal OID never gets updated to tempView.

I think my initial change where I add Internal OID to COLDVIDS (irrespective of Cold or Warm boot)  make it clean because then populateExisting Object logic will take care of adding this to temp view with the existing logic.

> as for unittests, this will be tricky to implement in current test mechanism, since it would require to have 2 different version of syncd (compiled with version A and then B) and cold boot from A to warm boot to B, or to have a new special configuration file for virtual switch that on warm boot it could be specified to "enable" new STP object to be existing on the VS

abdosi:
@kcudnik can you check my comments above

kcudnik:
ok, thats sounds good, just put a those comments in that change why we populate those tables in redis

syncd/SaiSwitch.cpp
abdosi: @kcudnik  Done

syncd/SaiSwitch.cpp
abdosi: @kcudnik Closed as per below comments

abdosi: @kcudnik Done

abdosi: @kcudnik Done

syncd/SaiSwitch.cpp
abdosi: @kcudnik Done

syncd/SaiSwitch.h
abdosi: @kcudnik Done

kcudnik (approved):

PR #656, opened 2020-09-10
[syncd] Add supports of bulk api in syncd

dmytroxshevchuk:
- Why I did it
Currently, when syncd must process bulk APIs, syncd calls create/remove/get/set in the loop.
- How I did it
Added implementation of calls sai bulk APIs.
- How to verify it

- Which release branch to backport (provide reason below if selected)

- Description for the changelog

- A picture of a cute animal (not mandatory but encouraged)


syncd/Syncd.cpp
```diff
@@ -898,6 +959,122 @@ sai_status_t Syncd::processEntry(
...
+sai_status_t Syncd::processBulkOidCreate(_In_ sai_object_type_t objectType,
+            _In_ sai_bulk_op_error_mode_t mode,
+            _In_ const std::vector<std::string>& objectIds,
+            _In_ const std::vector<std::shared_ptr<saimeta::SaiAttributeList>>& attributes,
+            _Out_ std::vector<sai_status_t>& statuses)
+{
+    SWSS_LOG_ENTER();
+
+    /*
+     * Switch, Port objects doesn't have bulk API`s, so we no need additional
```
kcudnik, line 971: a lot of other objects also don't have bulk API, this check is not needed, please remove, should be done in meta/Meta.cpp anyway

syncd/Syncd.cpp
```diff
@@ -898,6 +959,122 @@ sai_status_t Syncd::processEntry(
...
+    {
+        sai_deserialize_object_id(objectIds[idx], objectVids[idx]);
+
+        attr_counts[idx] = attributes[idx]->get_attr_count();
+        attr_lists[idx] = attributes[idx]->get_attr_list();
+    }
+
+    sai_object_id_t switchRid = SAI_NULL_OBJECT_ID;
+
+    if (objectVids.size() > 0 && objectType != SAI_OBJECT_TYPE_SWITCH)
```
kcudnik, line 996: this check is not needed, objectVids() should be checked in meta/meta.cpp and it is should be required there that this should > 0, also != SWITCH is not reqirered eather, switchIdQuery from SWITCH will return itself anyway

syncd/Syncd.cpp
```diff
@@ -898,6 +959,122 @@ sai_status_t Syncd::processEntry(
...
+    {
+        sai_object_id_t switchVid = VidManager::switchIdQuery(objectVids.front());
```
kcudnik, line 998: you are getting switchId from First object on objectVids, thats fine, but there should be check in meta/meta.cpp  that all objectVids belong to the same switchId

syncd/Syncd.cpp
```diff
@@ -898,6 +959,122 @@ sai_status_t Syncd::processEntry(
...
+        switchRid = m_translator->translateVidToRid(switchVid);
+    }
+
+    std::vector<sai_object_id_t> objectRids(objectIds.size());
+
+    status = m_vendorSai->bulkCreate(objectType, switchRid, (uint32_t)objectIds.size(), attr_counts.data(),
```
kcudnik, line 1004: if line is too long, move each param to separate line

syncd/Syncd.cpp
```diff
@@ -898,6 +959,122 @@ sai_status_t Syncd::processEntry(
...
+                                        attr_lists.data(), mode, objectRids.data(), statuses.data());
+
+    if (status == SAI_STATUS_SUCCESS)
```
kcudnik, line 1007: this check is invalid, according to bulk API, success will be returned only when all create will succeed, if any fails, then return will be FAILURE, but even do general status is FAILURE, lest say 99 of 100 could be succeeded, and then you need to check statuses vector for each object to figure out which one failed, and insert to translator only those with success, and skip those without

syncd/Syncd.cpp
```diff
@@ -898,6 +959,122 @@ sai_status_t Syncd::processEntry(
...
+            SWSS_LOG_INFO("saved VID %s to RID %s",
+                    sai_serialize_object_id(objectVids[idx]).c_str(),
+                    sai_serialize_object_id(objectRids[idx]).c_str());
+        }
+    }
+
+    return status;
+}
+
+sai_status_t Syncd::processBulkOidRemove(_In_ sai_object_type_t objectType,
```
kcudnik, line 1025: first param to new line

syncd/Syncd.cpp
```diff
@@ -898,6 +959,122 @@ sai_status_t Syncd::processEntry(
...
+            _In_ const std::vector<std::string>& objectIds,
+            _Out_ std::vector<sai_status_t>& statuses)
+{
+    SWSS_LOG_ENTER();
+
+    /*
+     * Switch, Port objects doesn't have bulk API`s, so we no need additional
+     * functionality for process it
+     */
+    if (objectType == SAI_OBJECT_TYPE_SWITCH || objectType == SAI_OBJECT_TYPE_PORT)
```
kcudnik, line 1036: as above this check is no needed

syncd/Syncd.cpp
```diff
@@ -898,6 +959,122 @@ sai_status_t Syncd::processEntry(
...
+    std::vector<sai_object_id_t> objectVids(objectIds.size());
+    std::vector<sai_object_id_t> objectRids(objectIds.size());
+
+    for (size_t idx = 0; idx < objectIds.size(); idx++)
+    {
+        sai_deserialize_object_id(objectIds[idx], objectVids[idx]);
+        objectRids[idx] = m_translator->translateVidToRid(objectVids[idx]);
+    }
+
+    status = m_vendorSai->bulkRemove(objectType, (uint32_t)objectIds.size(),
```
kcudnik, line 1052: each param to new line

syncd/Syncd.cpp
```diff
@@ -898,6 +959,122 @@ sai_status_t Syncd::processEntry(
...
+                                objectVids.data(), mode, statuses.data());
+
+    if (status == SAI_STATUS_SUCCESS)
```
kcudnik, line 1055: similar logic should be as in create, on FAILURE you need to check each of statuses vector

syncd/Syncd.cpp
```diff
@@ -898,6 +959,122 @@ sai_status_t Syncd::processEntry(
...
+    {
+        /*
+         * remove all related objects from REDIS DB and also from existing
+         * object references since at this point they are no longer valid
+         */
+        sai_object_id_t switchVid;
+        for (size_t idx = 0; idx < objectIds.size(); idx++)
+        {
+            m_translator->eraseRidAndVid(objectRids[idx], objectVids[idx]);
```
kcudnik, line 1064: you need to check each status vector and then if success then execute for contents for that index

syncd/Syncd.cpp
```diff
@@ -920,6 +1097,35 @@ sai_status_t Syncd::processBulkOid(
 
     sai_status_t all = SAI_STATUS_SUCCESS;
 
+#ifdef SAI_SUPPORT_BULK_API
```
kcudnik, line 1100: as mentioned before, make this command line flag, and also  split processBulkOid to 2 methods 1st to use bulk vendor API and 2nd that will use unwrapped loop using single vendor api

syncd/Syncd.cpp
```diff
@@ -920,6 +1097,35 @@ sai_status_t Syncd::processBulkOid(
...
+        case SAI_COMMON_API_BULK_REMOVE:
+            all = processBulkOidRemove(objectType, mode, objectIds, statuses);
+            break;
+
+        default:
+            all = SAI_STATUS_NOT_SUPPORTED;
+            SWSS_LOG_THROW("api %d is not supported in bulk mode", api);
+    }
+
+    if (all != SAI_STATUS_NOT_SUPPORTED && all != SAI_STATUS_NOT_IMPLEMENTED)
```
kcudnik, line 1119: remove this check, since in syncd mode we also required to send response to sender, and in line 1116 you throw anyway

syncd/Syncd.cpp
```diff
@@ -920,6 +1097,35 @@ sai_status_t Syncd::processBulkOid(
...
+            all = processBulkOidCreate(objectType, mode, objectIds, attributes, statuses);
+            break;
+
+        case SAI_COMMON_API_BULK_REMOVE:
+            all = processBulkOidRemove(objectType, mode, objectIds, statuses);
+            break;
+
+        default:
+            all = SAI_STATUS_NOT_SUPPORTED;
+            SWSS_LOG_THROW("api %d is not supported in bulk mode", api);
```
kcudnik, line 1116: you can use %s and sai_serialize_common_api(api).c_str() to get text form of api instead of number

syncd/Syncd.h
```diff
@@ -197,6 +197,19 @@ namespace syncd
                     _In_ uint32_t attr_count,
                     _In_ sai_attribute_t *attr_list);
 
+        private: // process bulk oid
+
+            sai_status_t processBulkOidCreate(_In_ sai_object_type_t objectType,
```
kcudnik, line 202: first param always to new line

syncd/Syncd.h
```diff
@@ -197,6 +197,19 @@ namespace syncd
...
+                    _In_ sai_bulk_op_error_mode_t mode,
+                    _In_ const std::vector<std::string>& objectIds,
+                    _In_ const std::vector<std::shared_ptr<saimeta::SaiAttributeList>>& attributes,
+                    _Out_ std::vector<sai_status_t>& statuses);
+
+            sai_status_t processBulkOidRemove(_In_ sai_object_type_t objectType,
```
kcudnik, line 208: first param always to new line

syncd/VendorSai.cpp
```diff
@@ -663,7 +663,11 @@ sai_status_t VendorSai::bulkCreate(
 
         default:
             SWSS_LOG_ERROR("not implemented, FIXME");
-            return SAI_STATUS_FAILURE;
+            return SAI_STATUS_NOT_SUPPORTED;
+    }
+
+    if (!ptr) {
```
kcudnik, line 669: { to new line, and add error message that obtained api pointer is null

kcudnik: also, you are required to SET all statuses vector to failure, since external logic if see api failure, will also want to check statuses

syncd/VendorSai.cpp
```diff
@@ -663,7 +663,11 @@ sai_status_t VendorSai::bulkCreate(
 
         default:
             SWSS_LOG_ERROR("not implemented, FIXME");
-            return SAI_STATUS_FAILURE;
+            return SAI_STATUS_NOT_SUPPORTED;
```
kcudnik, line 666: inconsistent return code, error message says not implemented, and return code not supported
you can also extend error message by %s and sai_serialize_object_type(object_type).c_str()

syncd/VendorSai.cpp
```diff
@@ -716,7 +720,11 @@ sai_status_t VendorSai::bulkRemove(
...
```
kcudnik, line 723: same as in bulkCreate

syncd/VendorSai.cpp
```diff
@@ -716,7 +720,11 @@ sai_status_t VendorSai::bulkRemove(
...
+    }
+
+    if (!ptr) {
```
kcudnik, line 726: same as in bulkCreate

syncd/VendorSai.cpp
```diff
@@ -753,6 +761,10 @@ sai_status_t VendorSai::bulkCreate(
     SWSS_LOG_ENTER();
     VENDOR_CHECK_API_INITIALIZED();
 
+    if (!m_apis.route_api->create_route_entries) {
```
kcudnik, line 764: { to new line, keep consistent with code
and add error log message

syncd/VendorSai.cpp
```diff
@@ -774,9 +786,17 @@ sai_status_t VendorSai::bulkCreate(
     SWSS_LOG_ENTER();
     VENDOR_CHECK_API_INITIALIZED();
 
-    SWSS_LOG_ERROR("not supported by SAI");
+    if (!m_apis.fdb_api->create_fdb_entries) {
```
kcudnik, line 789: { to new line,
add error message log

syncd/VendorSai.cpp
```diff
@@ -811,6 +831,11 @@ sai_status_t VendorSai::bulkRemove(
     MUTEX();
     SWSS_LOG_ENTER();
     VENDOR_CHECK_API_INITIALIZED();
+
+    if (!m_apis.route_api->remove_route_entries) {
```
kcudnik, line 835: { + error message

syncd/VendorSai.cpp
```diff
@@ -829,9 +854,15 @@ sai_status_t VendorSai::bulkRemove(
     SWSS_LOG_ENTER();
     VENDOR_CHECK_API_INITIALIZED();
 
-    SWSS_LOG_ERROR("not supported by SAI");
+    if (!m_apis.fdb_api->remove_fdb_entries) {
```
kcudnik, line 857: { + error message

syncd/VendorSai.cpp
```diff
@@ -863,6 +894,10 @@ sai_status_t VendorSai::bulkSet(
     SWSS_LOG_ENTER();
     VENDOR_CHECK_API_INITIALIZED();
 
+    if (!m_apis.route_api->set_route_entries_attribute) {
```
kcudnik, line 897: { + error message

syncd/VendorSai.cpp
```diff
@@ -882,9 +917,17 @@ sai_status_t VendorSai::bulkSet(
     SWSS_LOG_ENTER();
     VENDOR_CHECK_API_INITIALIZED();
 
-    SWSS_LOG_ERROR("not supported by SAI");
 
-    return SAI_STATUS_NOT_SUPPORTED;
+    if (!m_apis.fdb_api->set_fdb_entries_attribute) {
```
kcudnik, line 921: { + error message

kcudnik (changes requested):
Please address comments

kcudnik:
Also since you are adding bulk support, an unittests would be appreciated:
* for this it would be be best to make this bulk support flag as commandline option, then you can easy support that in unittest
* virtual switch will need to be updated to support bulk API
* saiplayer will also need to be updated to support bulk API

lib/inc/SaiInterface.h
```diff
@@ -203,6 +203,30 @@ namespace sairedis
...
                     _Inout_ sai_attribute_t *attr_list);
 
+        public: // bulk QUAD meta key
+
+            virtual sai_status_t bulkCreate(
+                    _Inout_ sai_object_meta_key_t* metaKey,
+                    _In_ sai_object_id_t switch_id,
+                    _In_ uint32_t object_count,
+                    _In_ uint32_t *attr_count,
+                    _In_ const sai_attribute_t **attr_list,
```
kcudnik, line 213: also on pointers we prefer to use * on the like "sai_attribute_t** attr_list"

syncd/Syncd.cpp
```diff
@@ -802,6 +802,67 @@ sai_status_t Syncd::processBulkEntry(
...
+            all = m_vendorSai->bulkSet(metaKeys.data(), (uint32_t)objectIds.size(),
+                                             attr_lists.data(), mode, statuses.data());
+            break;
+
+        default:
+            all = SAI_STATUS_NOT_SUPPORTED;
+            SWSS_LOG_THROW("api %d is not supported in bulk mode", api);
+    }
+
+    if (all != SAI_STATUS_NOT_SUPPORTED && all != SAI_STATUS_NOT_IMPLEMENTED)
```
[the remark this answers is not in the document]
dmytroxshevchuk: But what if bulk functions is not implemented in SAI? then we skip sendResponse in 858 line and try to process this entries in loop from 866 line.
Also if bulk functions is supported and implemented, then we need send response in 858 line and exit from function in 862 line.
In both case response will be send in 858 line (if bulk functions are supported) or in 925 line (if bulk functions are not supported)
This check was added in case if bulk functions is not implemented and this case should not affect flow.

syncd/Syncd.cpp
dmytroxshevchuk: the same as above, in 856 line

syncd/Syncd.cpp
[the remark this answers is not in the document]
dmytroxshevchuk: SAI_STATUS_NOT_SUPPORTED or SAI_STATUS_NOT_IMPLEMENTED status can be returned from bulkCreate/bulkSet/bulkRemove functions

lib/inc/SaiInterface.h
```diff
@@ -203,6 +203,30 @@ namespace sairedis
                     _In_ uint32_t attr_count,
                     _Inout_ sai_attribute_t *attr_list);
 
+        public: // bulk QUAD meta key
+
+            virtual sai_status_t bulkCreate(
+                    _Inout_ sai_object_meta_key_t* metaKey,
```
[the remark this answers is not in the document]
dmytroxshevchuk: Are you mean we should use std::vector instead of pointers for metaKey and attr_lists(in comment below)?
The same for attr_count?

syncd/Syncd.cpp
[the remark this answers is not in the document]
kcudnik: that behavior depends whether we want to fallback to loop 1by1 if vendor bulk api is not available, this was not mendioned

lib/src/SaiInterface.cpp
```diff
@@ -173,3 +175,161 @@ sai_status_t SaiInterface::get(
...
+        return SAI_STATUS_FAILURE;
+    }
+
+    if (!object_count)
+    {
+        SWSS_LOG_WARN("Attributes array is empty in bulkCreate");
+        return SAI_STATUS_SUCCESS;
+    }
+
+    auto info = sai_metadata_get_object_type_info(metaKey[0].objecttype);
```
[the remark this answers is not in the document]
dmytroxshevchuk: I did not found "check" for checking array with metakeys, so added own realization of this check from 199 line, do I need move it to meta.cpp?

lib/inc/SaiInterface.h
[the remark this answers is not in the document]
kcudnik: i think only for vector, since all other are following SAI parameter convention

lib/src/SaiInterface.cpp
```diff
@@ -173,3 +173,275 @@ sai_status_t SaiInterface::get(
...
+
+        default:
+            SWSS_LOG_ERROR("object type %s not implemented, FIXME", info->objecttypename);
+            return SAI_STATUS_NOT_SUPPORTED;
+    }
+}
+
+sai_status_t SaiInterface::bulkSet(
+        _Inout_ std::vector<sai_object_meta_key_t> &metaKey,
+        _In_ const sai_attribute_t **attr_list,
```
kcudnik, line 357: why attr_list is double pointer? SET operation takes only 1 attribute, so for bulk operation only single list of attribute would be required

lib/src/SaiInterface.cpp
```diff
@@ -173,3 +173,275 @@ sai_status_t SaiInterface::get(
...
+
+    if (info->isobjectid)
+    {
+        std::vector<sai_object_id_t> objectIds(object_count);
+        std::vector<sai_attribute_t> attr_lists(object_count);
+
+        for (uint32_t idx = 0; idx < object_count; idx++)
+        {
+            objectIds[idx] = metaKey[idx].objectkey.key.object_id;
+            attr_lists[idx] = attr_list[idx][0];
```
kcudnik, line 393: no need to do that [0] if single list would be used, you could just pass that list directy, no need to attr_lists ether

syncd/Syncd.cpp
```diff
@@ -802,6 +802,81 @@ sai_status_t Syncd::processBulkEntry(
 
     sai_status_t all = SAI_STATUS_SUCCESS;
 
+    if (m_commandLineOptions->m_enableSaiBulkSupport)
```
kcudnik, line 805: please move this to separate method

syncd/Syncd.cpp
```diff
@@ -914,12 +1124,39 @@ sai_status_t Syncd::processBulkOid(
...
+                all = processBulkOidCreate(objectType, mode, objectIds, attributes, statuses);
+                break;
+
+            case SAI_COMMON_API_BULK_REMOVE:
+                all = processBulkOidRemove(objectType, mode, objectIds, statuses);
+                break;
+
+            default:
+                all = SAI_STATUS_NOT_SUPPORTED;
+                SWSS_LOG_ERROR("api %s is not supported in bulk mode", sai_serialize_common_api(api).c_str());
```
kcudnik, line 1147: missing break

syncd/VendorSai.cpp
```diff
@@ -716,7 +722,13 @@ sai_status_t VendorSai::bulkRemove(
 
         default:
             SWSS_LOG_ERROR("not implemented, FIXME");
-            return SAI_STATUS_FAILURE;
+            return SAI_STATUS_NOT_IMPLEMENTED;
+    }
+
+    if (!ptr)
+    {
+        SWSS_LOG_ERROR("remove bulk not supported from SAI, object_type = %u", object_type);
```
kcudnik, line 730: instead of %u, use string and serialize_object_type to get string representation

syncd/VendorSai.cpp
```diff
@@ -662,8 +662,14 @@ sai_status_t VendorSai::bulkCreate(
...
         default:
-            SWSS_LOG_ERROR("not implemented, FIXME");
-            return SAI_STATUS_FAILURE;
+            SWSS_LOG_ERROR("not implemented, FIXME changed!!");
+            return SAI_STATUS_NOT_IMPLEMENTED;
+    }
+
+    if (!ptr)
+    {
+        SWSS_LOG_ERROR("create bulk not supported from SAI, object_type = %u", object_type);
```
kcudnik, line 671: same here, use string repreentation

syncd/VendorSai.cpp
```diff
@@ -662,8 +662,14 @@ sai_status_t VendorSai::bulkCreate(
             break;
 
         default:
-            SWSS_LOG_ERROR("not implemented, FIXME");
-            return SAI_STATUS_FAILURE;
+            SWSS_LOG_ERROR("not implemented, FIXME changed!!");
```
kcudnik, line 665: just curious what " FIXME changed!!" means ?

lib/inc/SaiInterface.h
```diff
@@ -203,6 +205,27 @@ namespace sairedis
...
+                    _Out_ sai_status_t *object_statuses);
+
+            virtual sai_status_t bulkRemove(
+                    _Inout_ std::vector<sai_object_meta_key_t> &metaKey,
+                    _In_ sai_bulk_op_error_mode_t mode,
+                    _Out_ sai_status_t *object_statuses);
+
+            virtual sai_status_t bulkSet(
+                    _Inout_ std::vector<sai_object_meta_key_t> &metaKey,
+                    _In_ const sai_attribute_t **attr_list,
```
kcudnik, line 225: bulk set should get only single pointer * list

lib/src/SaiInterface.cpp
```diff
@@ -173,3 +173,275 @@ sai_status_t SaiInterface::get(
...
+            return bulkCreate(object_count, entries.data(), attr_count, attr_list, mode, object_statuses);
+        }
+
+        default:
+            SWSS_LOG_ERROR("object type %s not implemented, FIXME", info->objecttypename);
+            return SAI_STATUS_NOT_SUPPORTED;
+    }
+}
+
+sai_status_t SaiInterface::bulkRemove(
```
kcudnik, line 271: im also interested, since you added create/remove/set bulk APIS that handle meta_key, but we already added
170             SAIREDIS_SAIINTERFACE_DECLARE_BULK_CREATE_ENTRY(fdb_entry);
171             SAIREDIS_SAIINTERFACE_DECLARE_BULK_CREATE_ENTRY(nat_entry);
172             SAIREDIS_SAIINTERFACE_DECLARE_BULK_CREATE_ENTRY(route_entry);

and remove/set also for *_entry objects, so actually using vector of meta_key will require copy to new array of objects, like you are for example doing like in line 322, we can keep that, but from performance reason when we doing "deserialize" we already know what object type we are dealing with for example route_entry, so make no sense to copy that to meta key, and then copy from meta key to next vector to just call actual BulkRouteCreate api without meta key, it seem convenient, but notice that in case of single non bulk api where meta_key is used, inside that api let say:  sai_status_t SaiInterface::set( _In_ const sai_object_meta_key_t& metaKey, _In_ const sai_attribute_t *attr), there i switch/case statement for object type, and actual *_entry api is called for that particular object type, but there is no data copy there back and forth, if i remember correctly thats the reason i didnt put bulk meta_key apis at the first place in SaiInterface.h class

lib/inc/SaiInterface.h
dmytroxshevchuk: we have **attr_list definition in Meta for bulkCreate (src/sonic-sairedis/meta/Meta.cpp:2563)
```
sai_status_t Meta::bulkCreate(
        _In_ sai_object_type_t object_type,
        _In_ sai_object_id_t switchId,
        _In_ uint32_t object_count,
        _In_ const uint32_t *attr_count,
        _In_ const sai_attribute_t **attr_list,
        _In_ sai_bulk_op_error_mode_t mode,
        _Out_ sai_object_id_t *object_id,
        _Out_ sai_status_t *object_statuses)
{
```

kcudnik: ok

syncd/Syncd.cpp
dmytroxshevchuk: do we need break there? in any case we get out from switch

dmytroxshevchuk:
> Also since you are adding bulk support, an unittests would be appreciated:
> 
> * for this it would be be best to make this bulk support flag as commandline option, then you can easy support that in unittest
> * virtual switch will need to be updated to support bulk API
> * saiplayer will also need to be updated to support bulk API

kcudnik:
for unittests i was thinking on adding unittest replay/recordings in sairedis/tests/BCM56850.pl, take a look at this file

dmytroxshevchuk:
> for unittests i was thinking on adding unittest replay/recordings in sairedis/tests/BCM56850.pl, take a look at this file

done, please check

kcudnik:
@dmytroxshevchuk hey, for a past few weeks i was doing performance tests around route_entry object, and i recently checked that in, which also includes bulk api implementation in syncd/sairedis/saiplayer
can you rebase/merge master to this branch, and resolve conflicts, also checking if everything is working as you intended, since there are quite some changes, after that i will review again your whole pr
let me know if you need any help/explanation about my changes #666

dmytroxshevchuk:
> @dmytroxshevchuk hey, for a past few weeks i was doing performance tests around route_entry object, and i recently checked that in, which also includes bulk api implementation in syncd/sairedis/saiplayer
> can you rebase/merge master to this branch, and resolve conflicts, also checking if everything is working as you intended, since there are quite some changes, after that i will review again your whole pr
> let me know if you need any help/explanation about my changes #666

done, please review whole pr

syncd/Syncd.cpp
```diff
@@ -802,6 +802,67 @@ sai_status_t Syncd::processBulkEntry(
...
+    {
+        metaKeys[idx].objecttype = objectType;
+
+        switch (objectType)
+        {
+            case SAI_OBJECT_TYPE_ROUTE_ENTRY:
+                sai_deserialize_route_entry(objectIds[idx], metaKeys[idx].objectkey.key.route_entry);
+                break;
+
+            case SAI_OBJECT_TYPE_FDB_ENTRY:
```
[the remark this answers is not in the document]
kcudnik: move all of them over entire file

syncd/Syncd.cpp
```diff
@@ -797,6 +797,242 @@ sai_status_t Syncd::processBulkQuadEventInInitViewMode(
     }
 }
 
+sai_status_t Syncd::processBulkEntryWithSaiBulkSupport(
```
kcudnik, line 800: this function is too big for my taste, you can split to smaller functions to better digest this, we probably would like to have functions that fit in 1-2 screens

syncd/Syncd.cpp
```diff
@@ -930,6 +1179,140 @@ sai_status_t Syncd::processEntry(
...
+        _Out_ std::vector<sai_status_t>& statuses)
+{
+    SWSS_LOG_ENTER();
+
+    sai_status_t status = SAI_STATUS_SUCCESS;
+    uint32_t object_count = (uint32_t)objectIds.size();
+
+    if (!object_count)
+    {
+        return SAI_STATUS_FAILURE;
```
kcudnik, line 1268: missing log error/warning

saiplayer/SaiPlayer.cpp
```diff
@@ -1022,309 +1049,338 @@ std::vector<std::string> SaiPlayer::tokenize(
...
-        sai_deserialize_route_entry(object_ids[i], route_entry);
+        std::vector<uint32_t> attr_counts(object_count);
+        std::vector<const sai_attribute_t*> attr_lists(object_count);
 
-        route_entry.vr_id = translate_local_to_redis(route_entry.vr_id);
+        for (uint32_t idx = 0; idx < object_count; idx++)
+        {
+            attr_counts[idx] = attributes[idx]->get_attr_count();
+            attr_lists[idx] = attributes[idx]->get_attr_list();
+        }
```
kcudnik, line 1076: github diff is making this such a mess i'm not able to review it, every couple lines it switches from one to another file
could you move your function to some other place in SaiPlayer.cpp, like 200-300 lines above/below, so diff will not pick this up ? its really hard to view if lines are alternating so many times :( 
diff gets confused because of  { }
image

saiplayer/SaiPlayer.cpp
```diff
@@ -1022,309 +1049,338 @@ std::vector<std::string> SaiPlayer::tokenize(
...
+                std::vector<sai_route_entry_t> entries(object_count);
 
-        SWSS_LOG_DEBUG("route: %s", object_ids[i].c_str());
-    }
+                for (size_t it = 0; it < object_count; it++)
+                {
+                    sai_deserialize_route_entry(object_ids[it], entries[it]);
 
-    std::vector<sai_status_t> statuses;
+                    entries[it].switch_id = translate_local_to_redis(entries[it].switch_id);
```
kcudnik, line 1088: as for translation oids, there is generic function that translates automatically all oids in entry structure, but it must be meta_key, and this is trade off here, whether we copy meta_key and then extract actual route/nat/fdb to feed SAI api, or we manually translate each oid :(

kcudnik (changes requested):
please address comments

saiplayer/SaiPlayer.cpp
dmytroxshevchuk: as for me, more quickly its manually translate each oid

kcudnik (approved):

lguohan:
please merge?

PR #657, opened 2020-09-10
Implementation of System ports initialization, Interface & Neighbor Synchronization...

minionatwork:
This PR includes implementation of System ports initialization & orchestration, Interface & Neighbor Synchronization using GLOBAL_APP_DB, Inband Interface Configuration.

HLD: https://github.com/Azure/SONiC/pull/639
Dependency PR(sonic-swss-common): https://github.com/Azure/sonic-swss-common/pull/380


lib/src/RedisChannel.cpp
```diff
@@ -12,7 +12,7 @@ using namespace sairedis;
 /**
  * @brief Get response timeout in milliseconds.
  */
-#define REDIS_ASIC_STATE_COMMAND_GETRESPONSE_TIMEOUT_MS (60*1000)
+#define REDIS_ASIC_STATE_COMMAND_GETRESPONSE_TIMEOUT_MS (5*60*1000)
```
kcudnik, line 15: we would like to keep this as is, since there is scenario, where OA is requesting a lot of route updates, and then doing GET, those request are asynchronous, and syncd will be processing them much longer than 5sec, which will cause GET to timeout, 
only in ASYNC mode which is default right now

kcudnik, line 15: we would like to keep this as is, since there is scenario, where OA is requesting a lot of route updates, and then doing GET, those request are asynchronous, and syncd will be processing them much longer than 5sec, which will cause GET to timeout,
only in ASYNC mode which is default right now

kcudnik (changes requested):
Please address comments

lib/src/RedisChannel.cpp
vganesan-nokia: Before this change, the time out was 1 min. Processing longer than 5 sec but less than 1 min will not time out. The change is to make more waiting than 1 min before declaring "no response received" to OA. I guess increasing the time out to 5 min from 1 min actually helps the scenario mentioned.

kcudnik: Ahh sorry, i reinterpreted that as 5 seconds :P

kcudnik (approved):

lib/src/RedisChannel.cpp
vganesan-nokia, line 15: Increasing to 5 minutes may delay the recognition of failure from syncd. Try adjusting to lower levels, if possible, after syncd initialization is complete (after create switch).

minionatwork: Rita: Might  be even better if this 5min timeout is for chassis only.

vganesan-nokia: We'll undo this in this PR. In a separate PR, we'll change the time to 5 minutes only for switch_type = "voq" and the timeout will be changed back to 1 min after SAI initialization is complete.

kcudnik (approved):

PR #659, opened 2020-09-15
Add ZeroMQ communication channel between sairedis and syncd

kcudnik:
This channel is always synchronous, and when used will force synchronous mode


kcudnik:
16:27:23  ZeroMQChannel.cpp:10:10: fatal error: zmq.h: No such file or directory
16:27:23   #include <zmq.h>

zmq lib is needed in compiler  environment to make this compile

kcudnik:
retest this please

kcudnik:
retest this please

lib/inc/RedisChannel.h
```diff
@@ -17,49 +18,42 @@
...
 
             virtual ~RedisChannel();
 
         public:
 
             std::shared_ptr<swss::DBConnector> getDbConnector() const;
 
-            void setBuffered(
-                    _In_ bool buffered);
+            virtual void setBuffered(
```
qiluo-msft, line 36: >virtual [](start = 12, length = 7)

remove `virtual` and use `override` only in derived class.

kcudnik: not following?

qiluo-msft: ref: https://stackoverflow.com/a/39932616

kcudnik: i will still keep virtual and override keywords in child class, its consistent with all there is right now,
also for example in your case, if in child class function is marked as virtual foo() override, and base class foo() prototype changes, compiler generate error that there is no function to override from base class, so no mistake here if suddenly foo will be different function in child classed

kcudnik:
retest this please

kcudnik:
how to rerun LGTM analysis: C/C++ ?

kcudnik:
retest LGTM please

kcudnik:
[2020-09-28 21:01:14] [build] /usr/bin/ld: ./.libs/libsairedis.so: undefined reference to `zmq_connect'
[2020-09-28 21:01:14] [build] /usr/bin/ld: ./.libs/libsairedis.so: undefined reference to `zmq_send'
[2020-09-28 21:01:14] [build] /usr/bin/ld: ./.libs/libsairedis.so: undefined reference to `zmq_poll'
[2020-09-28 21:01:14] [build] /usr/bin/ld: ./.libs/libsairedis.so: undefined reference to `zmq_setsockopt'
[2020-09-28 21:01:14] [build] /usr/bin/ld: ./.libs/libsairedis.so: undefined reference to `zmq_errno'
[2020-09-28 21:01:14] [build] /usr/bin/ld: ./.libs/libsairedis.so: undefined reference to `zmq_recv'
[2020-09-28 21:01:14] [build] /usr/bin/ld: ./.libs/libsairedis.so: undefined reference to `zmq_close'
[2020-09-28 21:01:14] [build] /usr/bin/ld: ./.libs/libsairedis.so: undefined reference to `zmq_ctx_destroy'
[2020-09-28 21:01:14] [build] /usr/bin/ld: ./.libs/libsairedis.so: undefined reference to `zmq_ctx_new'
[2020-09-28 21:01:14] [build] /usr/bin/ld: ./.libs/libsairedis.so: undefined reference to `zmq_socket'

does LGTM has other build environment where libzmq5 needs to be added ?
but:

[2020-09-28 20:59:56] [build] The following NEW packages will be installed:
[2020-09-28 20:59:56] [build]   libnorm-dev libnorm1 libpgm-5.2-0 libpgm-dev libsodium-dev libsodium23
[2020-09-28 20:59:56] [build]   libzmq3-dev libzmq5

kcudnik:
doesn't work :/

kcudnik:
retest this please

lguohan (approved):

shishao7sxm:
Hi @kcudnik, Just wonder that ZMQ needs sync mode to be enabled, since in common understanding, sync mode will take performance down. Will appreciate if share some cases that turning on ZMQ could help the performance, Thanks a lot.

kcudnik:
image
take a look at performance test made on VS swtich

kcudnik:
also those stats are from Jan 2021 and code since then was improved so it's even faster, we currently have over 10k routes per second

shishao7sxm:
> also those stats are from Jan 2021 and code since then was improved so it's even faster, we currently have over 10k routes per second

Thanks Kcudnik. 
In above sheet, seems the bulk mode approximate rate is 3k per second end to end. But currently it can be up to 10k? Are there any other enhancements in SONiC for this?

kcudnik:
yes, we would need to run new test to get current numbers

shishao7sxm:
yes, it's big improvement. What's the differences between the sheet test and the ~16k/sec test? the new enhancement code or the test dev device?

kcudnik:
difference is running in January vs July, code improvements and dev machine vs local dev laptop

shishao7sxm:
Got it, Thanks Kcudnik.

shishao7sxm:
Hi Kcudnik, On current env, do you have async mode performance number? that would be helpful. Thanks a lot.

kcudnik:
for async mode:
```
$ time $PL -u -z redis_async 1000_100
real    0m2,993s
user    0m2,596s
sys     0m0,070s

$ time $PL -u -z redis_async 100_1000
real    0m2,953s
user    0m2,450s
sys     0m0,059s
```

PR #664, opened 2020-10-03
[syncd] Fix passing switch pointers during warm boot

kcudnik:
During warm boot we also need to update notification pointers when creating switch in warm boot mode, since syncd/vendor_sai_lib could be loaded at different address space, and this can lead to SIGSEGV, since previous address may not be available


shi-su (approved):
Thanks for this fix.

PR #667, opened 2020-10-06
[vs] Respect fdb learning mode on VS

kcudnik:
Respect fdb learning mode on virtual switch, mainly for disable learning mode on bridge port, so when a new packet will arrive, no fdb event notification will be sent, to prevent sairedis receive that notification and prevent metadata learning (snoop) bridge port oid. Notification will be sent only when learning is enabled at the first packet for that particular bridge port/mac/vlan.


lguohan:
can you add more description?

lguohan (approved):

PR #668, opened 2020-10-07
[syncd] Notify orchagent instead of shutting down swss when translateVidToRid fails in processClearStatsEvent

shi-su:
**What I did**
Let `processClearStatsEvent` notify orchagent instead of crashing orchagent when `TranslateVidToRid` fails.

**Why I did it**
In INIT_VIEW during warm-reboot, the FlexCounterOrch tries to clear stats for BUFFER_POOL_WATERMARK. This is expected to fail in INIT_VIEW in `TranslateVidToRid` since the object is not yet created. In such a scenario, the syncd should respond with a failure rather than crashing orchagent.

**How I verified it**
Verified using virtual switch.

### DIFF
--- a/syncd/Syncd.cpp
+++ b/syncd/Syncd.cpp
@@ -538,7 +538,13 @@ sai_status_t Syncd::processClearStatsEvent(
     sai_object_meta_key_t metaKey;
     sai_deserialize_object_meta_key(key, metaKey);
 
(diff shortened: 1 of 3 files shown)


qiluo-msft (commented):
LGTM. Please wait other @kcudnik 's approval.

kcudnik (approved):

PR #670, opened 2020-10-08
Fix SAI_FDB_EVENT_MOVE not supported error

vincentchiang-ec:
The event SAI_FDB_EVENT_MOVE should do the same operation like SAI_FDB_EVENT_LEARNED.

Signed-off-by: Vincent Chiang (vincent_chiang@edge-core.com)

### DIFF
--- a/syncd/NotificationProcessor.cpp
+++ b/syncd/NotificationProcessor.cpp
@@ -160,7 +160,7 @@ void NotificationProcessor::redisPutFdbEntryToAsicView(
         return;
     }
 
-    if (fdb->event_type == SAI_FDB_EVENT_LEARNED)
+    if (fdb->event_type == SAI_FDB_EVENT_LEARNED || fdb->event_type == SAI_FDB_EVENT_MOVE)
     {
         // currently we need to add type manually since fdb event don't contain type
         sai_attribute_t attr;


syncd/NotificationProcessor.cpp
```diff
@@ -160,7 +160,7 @@ void NotificationProcessor::redisPutFdbEntryToAsicView(
         return;
     }
 
-    if (fdb->event_type == SAI_FDB_EVENT_LEARNED)
+    if (fdb->event_type == SAI_FDB_EVENT_LEARNED || fdb->event_type == SAI_FDB_EVENT_MOVE)
```
kcudnik, line 163: this is assuming that fdb_entry didn't changed, just SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID attribute changed, so it went from 1 port to another port, so bridge_id or vlan_id stayed the same?

vincentchiang-ec: Yes, I think it is the meaning of SAI_FDB_EVENT_MOVE, right?

kcudnik: please check the tests, since i see that some assertions failed

vincentchiang-ec:
retest this please

kcudnik (approved):

PR #673, opened 2020-10-16
[vs] Add CRM SAI attributes to virtual switch interface

daall:
Signed-off-by: Danny Allen <daall@microsoft.com>

Currently the virtual switch reports errors when CRMOrch attempts to fetch the CRM counters because these are RO attributes that are not refreshed. Adding them to the `refresh_read_only` method to avoid these errors.


vslib/src/SwitchStateBase.cpp
```diff
@@ -1576,6 +1576,17 @@ sai_status_t SwitchStateBase::refresh_read_only(
...
+            case SAI_SWITCH_ATTR_AVAILABLE_IPV6_ROUTE_ENTRY:
+            case SAI_SWITCH_ATTR_AVAILABLE_IPV4_NEXTHOP_ENTRY:
+            case SAI_SWITCH_ATTR_AVAILABLE_IPV6_NEXTHOP_ENTRY:
+            case SAI_SWITCH_ATTR_AVAILABLE_IPV4_NEIGHBOR_ENTRY:
+            case SAI_SWITCH_ATTR_AVAILABLE_IPV6_NEIGHBOR_ENTRY:
+            case SAI_SWITCH_ATTR_AVAILABLE_NEXT_HOP_GROUP_MEMBER_ENTRY:
+            case SAI_SWITCH_ATTR_AVAILABLE_NEXT_HOP_GROUP_ENTRY:
+            case SAI_SWITCH_ATTR_AVAILABLE_FDB_ENTRY:
+            case SAI_SWITCH_ATTR_AVAILABLE_ACL_TABLE:
+            case SAI_SWITCH_ATTR_AVAILABLE_ACL_TABLE_GROUP:
```
kcudnik, line 1589: it's not that easy, please take a look at internal company mail, i posted some details how this should be done
then we can discuss whether at this stage this is good enough since i see already SNAT/DNAT entries are already here
but at least default values are set in:
 SwitchStateBase::set_switch_default_attributes() please check this method

vslib/inc/SwitchStateBase.h
```diff
@@ -82,6 +82,10 @@ namespace saivs
 
             virtual sai_status_t set_number_of_ecmp_groups();
 
+            virtual sai_status_t set_static_crm_values();
+
+            virtual sai_status_t set_static_acl_resource_list(sai_switch_attr_t acl_resource, int max_count);
```
kcudnik, line 87: 1 param per line

vslib/src/SwitchStateBase.cpp
```diff
@@ -746,6 +731,74 @@ sai_status_t SwitchStateBase::set_switch_default_attributes()
...
+        attr.value.u32 = resource.second;
+        CHECK_STATUS(set(SAI_OBJECT_TYPE_SWITCH, m_switch_id, &attr));
+    }
+
+    CHECK_STATUS(set_static_acl_resource_list(SAI_SWITCH_ATTR_AVAILABLE_ACL_TABLE, m_maxAclTables));
+
+    return set_static_acl_resource_list(SAI_SWITCH_ATTR_AVAILABLE_ACL_TABLE_GROUP, m_maxAclTableGroups);
+}
+
+sai_status_t SwitchStateBase::set_static_acl_resource_list(sai_switch_attr_t acl_resource, int max_count)
```
kcudnik, line 766: 1 param per line please

daall: Fixed! Forgot the `_In_` too. :)

kcudnik (approved):
lgtm

PR #674, opened 2020-10-16
[fast-reboot] fix boot type for fast boot in virtual switch

shi-su:
Fix `parseBootType()` function so that it could work with fast reboot in virtual switch.


qiluo-msft (approved):
Also check with @kcudnik

kcudnik (approved):

PR #676, opened 2020-10-21
[gbsyncd] exit with zero when platform has no gearbox

lguohan:

### DIFF
--- a/syncd/scripts/gbsyncd_startup.py
+++ b/syncd/scripts/gbsyncd_startup.py
@@ -19,6 +19,8 @@ You may obtain a copy of the License at
 import os
 import sys
 import json
+import time
+import syslog
 import subprocess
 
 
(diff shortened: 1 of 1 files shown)


syncd/scripts/gbsyncd_startup.py
```diff
@@ -39,7 +41,9 @@ def main():
         with open('/usr/share/sonic/hwsku/gearbox_config.json') as file_object:
             gearbox_config=json.load(file_object)
     except:
-        sys.exit("No external PHY / gearbox supported on this platform, existing physycd application")
+        syslog.syslog(syslog.LOG_NOTICE, "No external PHY / gearbox supported on this platform, existing gbsyncd application")
+        time.sleep(2)
+        sys.exit(0)
```
jleveque, line 46: We could choose a exit code other than 0 here and specify it in the supervisor.conf file as an expected exit code, if we want to differentiate the exit type.

lguohan: i think we can return 0 here. it is fine.

jleveque (approved):

PR #679, opened 2020-10-23
[syncd/FlexCounter]:Fix last remove bug

Pterosaur:
- The bug description
When the last object in FlexCounter was removed, the FlexCounter is also removed from FlexCounterManager. At the next time, when a new observed object is added and a new FlexCounter will be created. But the new FlexCounter cannot get the configuration. Because the configuration message from FlexCounterDB has been consumed at the first time.

- How to fix it
Add a flag to indicate whether the configuration also be removed from the FlexCounterDB. If not, don't delete the FlexCounter from FlexCounterManager even no observed objects in the FlexCounter.

### DIFF
--- a/syncd/FlexCounter.cpp
+++ b/syncd/FlexCounter.cpp
@@ -821,10 +821,7 @@ void FlexCounter::removeCounterPlugins()
     m_priorityGroupPlugins.clear();
     m_bufferPoolPlugins.clear();
 
-    if (isEmpty())
-    {
-        m_is_discard = true;
-    }
+    m_is_discard = true;
 }
 
 void FlexCounter::addCounterPlugin(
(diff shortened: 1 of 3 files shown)


Pterosaur:
retest vs please

lguohan:
@qiluo-msft , can you check why this pr is broken?

lguohan (dismissed):

syncd/FlexCounter.cpp
```diff
@@ -26,6 +26,7 @@ FlexCounter::FlexCounter(
     SWSS_LOG_ENTER();
 
     m_enable = false;
+    m_is_discard = false;
```
kcudnik, line 29: please rename to m_isDiscard to follow naming convention

qiluo-msft:
retest vs please

qiluo-msft:
The build should be fixed by https://github.com/Azure/sonic-sairedis/pull/681

kcudnik (approved):

PR #687, opened 2020-10-30
[syncd] Lower bulk missing api message level from error to info

kcudnik:
there is command line switch for syncd which was added recently "-l" i think, and it enables use for bulk vendor sai api, previously by default we used 1by1, so actually those error messages are kind a justified that they are in ERROR, since in syncd command line right now we explicitly ask for bulk api

### DIFF
--- a/syncd/VendorSai.cpp
+++ b/syncd/VendorSai.cpp
@@ -668,7 +668,7 @@ sai_status_t VendorSai::bulkCreate(
 
     if (!ptr)
     {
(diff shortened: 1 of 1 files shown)


kcudnik:
@dmytroxshevchuk take a look

daall (approved):

daall:
retest vs please

lguohan:
retest this please

kcudnik:
retest this please

lguohan:
the test failure is not related to this pr.

PR #690, opened 2020-11-02
[syncd] Use lua script to update db when using bulk api

kcudnik:
When using SAI bulk in syncd, like create/remove route_entry, we want to have ability to update database as fast as possible, so best way would be to execute every mutable db command inside lua script to make things faster, rather than execute 1 by 1 operation which is done currently inside syncd

depends on https://github.com/Azure/sonic-swss-common/pull/406

### DIFF
--- a/syncd/RedisClient.cpp
+++ b/syncd/RedisClient.cpp
@@ -593,7 +593,14 @@ void RedisClient::createAsicObjects(
     // we need to rewrite hash to add table prefix
     for (const auto& kvp: multiHash)
     {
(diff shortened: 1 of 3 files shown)


kcudnik:
retest vs please

lguohan (dismissed):

kcudnik:
retest this please

kcudnik:
also depends on https://github.com/Azure/sonic-swss-common/pull/410

lguohan (approved):

PR #691, opened 2020-11-03
MACsec Virtual SAI

Pterosaur:
This PR is about the MACsec in virtual SAI.
The document can be found at : https://github.com/Azure/SONiC/blob/9edcf67d0e0095e9f831a94ddf4fbb4a267adf7b/doc/macsec/MACsec_hld.md#345-virtual-macsec-sai

All changes has been separated to several commits and their purposes can be found at each commit description.

### DIFF
--- a/meta/saiserialize.cpp
+++ b/meta/saiserialize.cpp
@@ -298,6 +298,12 @@ sai_status_t transfer_attribute(
             transfer_primitive(src_attr.value.aclfield.data.s32, dst_attr.value.aclfield.data.s32);
             break;
 
+        case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_UINT64:
+            transfer_primitive(src_attr.value.aclfield.enable, dst_attr.value.aclfield.enable);
+            transfer_primitive(src_attr.value.aclfield.mask.u64, dst_attr.value.aclfield.mask.u64);
+            transfer_primitive(src_attr.value.aclfield.data.u64, dst_attr.value.aclfield.data.u64);
+            break;
+
         case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_MAC:
             transfer_primitive(src_attr.value.aclfield.enable, dst_attr.value.aclfield.enable);
             transfer_primitive(src_attr.value.aclfield.mask.mac, dst_attr.value.aclfield.mask.mac);
@@ -1369,6 +1375,9 @@ std::string sai_serialize_acl_field(
         case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_INT32:
             return sai_serialize_enum(field.data.s32, meta.enummetadata) + "&mask:" + sai_serialize_number(field.mask.s32, true);
 
+        case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_UINT64:
+            return sai_serialize_number(field.data.u64) + "&mask:" + sai_serialize_number(field.mask.u64, true);
+
         case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_MAC:
             return sai_serialize_mac(field.data.mac) +"&mask:" + sai_serialize_mac(field.mask.mac);
 
@@ -1404,6 +1413,29 @@ std::string sai_serialize_acl_capability(
     auto list = sai_serialize_enum_list(cap.action_list, &sai_metadata_enum_sai_acl_action_type_t, countOnly);
 
     return mandatory + ":" + list;
+}
+
+std::string sai_serialize_hex_binary(
+    _In_ const void *buffer,
+    _In_ size_t length)
+{
+    SWSS_LOG_ENTER();
+
+    std::string s;
+
+    if (buffer == NULL || length == 0)
+    {
+        return s;
+    }
+    s.resize(2 * length, '0');
+    const unsigned char *input = static_cast<const unsigned char *>(buffer);
+    char *output = &s[0];
+    for (size_t i = 0; i < length; i++)
+    {
+        snprintf(&output[i * 2], 3, "%02X", input[i]);
+    }
+
+    return s;
 }
 
 std::string sai_serialize_system_port_config(
@@ -1581,6 +1613,7 @@ std::string sai_serialize_attr_value(
         case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_INT16:
         case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_UINT32:
         case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_INT32:
+        case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_UINT64:
         case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_MAC:
         case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_IPV4:
         case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_IPV6:
@@ -1607,6 +1640,19 @@ std::string sai_serialize_attr_value(
 
         case SAI_ATTR_VALUE_TYPE_ACL_CAPABILITY:
             return sai_serialize_acl_capability(meta, attr.value.aclcapability, countOnly);
+
+            // MACsec Attributions
+
+        case SAI_ATTR_VALUE_TYPE_MACSEC_SAK:
+            return sai_serialize_hex_binary(attr.value.macsecsak);
+        case SAI_ATTR_VALUE_TYPE_MACSEC_AUTH_KEY:
+            return sai_serialize_hex_binary(attr.value.macsecauthkey);
+        case SAI_ATTR_VALUE_TYPE_MACSEC_SALT:
+            return sai_serialize_hex_binary(attr.value.macsecsalt);
+        case SAI_ATTR_VALUE_TYPE_MACSEC_SCI:
+            return sai_serialize_number(attr.value.u64);
+        case SAI_ATTR_VALUE_TYPE_MACSEC_SSCI:
+            return sai_serialize_number(attr.value.u32);
 
         case SAI_ATTR_VALUE_TYPE_SYSTEM_PORT_CONFIG:
             return sai_serialize_system_port_config(meta, attr.value.sysportconfig);
@@ -2539,6 +2585,11 @@ void sai_deserialize_acl_field(
             sai_deserialize_number(smask, field.mask.s32, true);
             return;
 
+        case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_UINT64:
+            sai_deserialize_number(sfield, field.data.u64);
+            sai_deserialize_number(smask, field.mask.u64, true);
+            return;
+
         case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_MAC:
             sai_deserialize_mac(sfield, field.data.mac);
             sai_deserialize_mac(smask, field.mask.mac);
@@ -2645,6 +2696,61 @@ void sai_deserialize_acl_capability(
 
     sai_deserialize_bool(mandatory_on_create_str, cap.is_action_list_mandatory);
     sai_deserialize_enum_list(list, &sai_metadata_enum_sai_acl_action_type_t, cap.action_list, false);
+}
+
+void sai_deserialize_hex_binary(
+    _In_ const std::string &s,
+    _Out_ void *buffer,
+    _In_ size_t length)
+{
+    SWSS_LOG_ENTER();
+    if (s.length() % 2 != 0)
+    {
+        SWSS_LOG_THROW("Invalid hex string %s", s.c_str());
+    }
+    if (s.length() > (length * 2))
+    {
+        SWSS_LOG_THROW("Buffer length isn't sufficient.");
+    }
+    size_t buffer_cur = 0;
+    size_t hex_cur = 0;
+    unsigned char *output = static_cast<unsigned char *>(buffer);
+    while (hex_cur < s.length())
+    {
+        if (!std::isxdigit(static_cast<std::uint8_t>(s[hex_cur])))
+        {
+            SWSS_LOG_THROW(
+                "Invalid hex string %s at %lu(%c)",
+                s.c_str(),
+                hex_cur,
+                s[hex_cur]);
+        }
+        if (!std::isxdigit(static_cast<std::uint8_t>(s[hex_cur + 1])))
+        {
+            SWSS_LOG_THROW(
+                "Invalid hex string %s at %lu(%c)",
+                s.c_str(),
+                hex_cur + 1,
+                s[hex_cur + 1]);
+        }
+        std::stringstream stream;
+        stream << std::hex;
+        stream << s[hex_cur++];
+        stream << s[hex_cur++];
+        std::uint32_t value;
+        stream >> value;
+        output[buffer_cur++] = static_cast<std::uint8_t>(value);
+    }
+}
+
+template<typename T>
+void sai_deserialize_hex_binary(
+    _In_ const std::string &s,
+    _Out_ T &value)
+{
+    SWSS_LOG_ENTER();
+
+    return sai_deserialize_hex_binary(s, &value, sizeof(T));
 }
 
 void sai_deserialize_system_port_config(
@@ -2829,6 +2935,7 @@ void sai_deserialize_attr_value(
         case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_INT16:
         case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_UINT32:
         case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_INT32:
+        case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_UINT64:
         case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_MAC:
         case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_IPV4:
         case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_IPV6:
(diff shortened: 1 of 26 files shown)


vslib/src/SwitchStateBaseMACsec.cpp
```diff
@@ -0,0 +1,859 @@
...
+}
+
+sai_status_t SwitchStateBase::findHostInterfaceInfoByPort(
+        _In_ sai_object_id_t &port_id,
+        _Out_ std::shared_ptr<HostInterfaceInfo> &info)
+{
+    SWSS_LOG_ENTER();
+
+    info.reset();
+    for (auto &kvp : m_hostif_info_map)
```
kcudnik, line 423: empty line

vslib/src/SwitchStateBaseMACsec.cpp
```diff
@@ -0,0 +1,859 @@
...
+    {
+        if (kvp.second->m_portId == port_id)
+        {
+            info = kvp.second;
+            break;
+        }
+    }
+    if (info == nullptr)
```
kcudnik, line 431: empty line

vslib/src/SwitchStateBaseMACsec.cpp
```diff
@@ -0,0 +1,859 @@
...
+    {
+        SWSS_LOG_ERROR(
+            "Cannot find corresponding port %s",
+            sai_serialize_object_id(port_id).c_str());
+        return SAI_STATUS_FAILURE;
+    }
+    return SAI_STATUS_SUCCESS;
```
kcudnik, line 438: empty line

vslib/src/SwitchStateBaseMACsec.cpp
```diff
@@ -0,0 +1,859 @@
...
+    _In_ sai_object_id_t object_id,
+    _In_ uint32_t attr_count,
+    _In_ const sai_attribute_t *attr_list,
+    _Out_ MACsecAttr &macsec_attr)
+{
+    SWSS_LOG_ENTER();
+
+    if (attr_list == nullptr || attr_count == 0)
+    {
+        SWSS_LOG_ERROR("Attribute is empty");
```
kcudnik, line 451: attribute list,

vslib/src/SwitchStateBaseMACsec.cpp
```diff
@@ -0,0 +1,859 @@
...
+        return SAI_STATUS_FAILURE;
+    }
+
+    const sai_attribute_t *attr = nullptr;
+    SAI_METADATA_GET_ATTR_BY_ID(attr, SAI_MACSEC_PORT_ATTR_MACSEC_DIRECTION, attr_count, attr_list);
```
kcudnik, line 456: empty line before/after macro

vslib/src/SwitchStateBaseMACsec.cpp
```diff
@@ -0,0 +1,859 @@
...
+    macsec_attr.m_direction = attr->value.s32;
+
+    SAI_METADATA_GET_ATTR_BY_ID(attr, SAI_MACSEC_PORT_ATTR_PORT_ID, attr_count, attr_list);
+    auto port_id = attr->value.oid;
+    CHECK_STATUS(findHostInterfaceInfoByPort(port_id, macsec_attr.m_info));
```
kcudnik, line 461: empty line before/after macro

vslib/src/SwitchStateBaseMACsec.cpp
```diff
@@ -0,0 +1,859 @@
...
+    _In_ sai_object_id_t object_id,
+    _In_ uint32_t attr_count,
+    _In_ const sai_attribute_t *attr_list,
+    _Out_ MACsecAttr &macsec_attr)
+{
+    SWSS_LOG_ENTER();
+
+    if (attr_list == nullptr || attr_count == 0)
+    {
+        SWSS_LOG_ERROR("Attribute is empty");
```
kcudnik, line 478: attr list

vslib/src/SwitchStateBaseMACsec.cpp
```diff
@@ -0,0 +1,859 @@
...
+
+    SAI_METADATA_GET_ATTR_BY_ID(attr, SAI_MACSEC_SC_ATTR_MACSEC_EXPLICIT_SCI_ENABLE, attr_count, attr_list);
+    macsec_attr.m_send_sci = attr->value.booldata;
+
+    SAI_METADATA_GET_ATTR_BY_ID(attr, SAI_MACSEC_SC_ATTR_FLOW_ID, attr_count, attr_list);
+    auto flow_id = attr->value.oid;
+
+    sai_object_id_t port_id = SAI_NULL_OBJECT_ID;
+    CHECK_STATUS(findPortByMACsecFlow(flow_id, port_id));
+    CHECK_STATUS(findHostInterfaceInfoByPort(port_id, macsec_attr.m_info));
```
kcudnik, line 501: empty line before/after macro

vslib/src/SwitchStateBaseMACsec.cpp
```diff
@@ -0,0 +1,859 @@
...
+    _In_ sai_object_id_t object_id,
+    _In_ uint32_t attr_count,
+    _In_ const sai_attribute_t *attr_list,
+    _Out_ MACsecAttr &macsec_attr)
+{
+    SWSS_LOG_ENTER();
+
+    if (attr_list == nullptr || attr_count == 0)
+    {
+        SWSS_LOG_ERROR("Attribute is empty");
```
kcudnik, line 518: attr list

vslib/src/SwitchStateBaseMACsec.cpp
```diff
@@ -0,0 +1,859 @@
...
+    std::vector<sai_attribute_t> attrs;
+
+    SAI_METADATA_GET_ATTR_BY_ID(attr, SAI_MACSEC_SA_ATTR_SC_ID, attr_count, attr_list);
+
+    // Find SCI and MACsec flow
+    attrs.clear();
+    attrs.emplace_back();
+    attrs.back().id = SAI_MACSEC_SC_ATTR_FLOW_ID;
+    attrs.emplace_back();
+    attrs.back().id = SAI_MACSEC_SC_ATTR_MACSEC_SCI;
```
kcudnik, line 532: std::vector<sai_attribute_t> attrs(2);

attrs[0].id = SAI_MACSEC_SC_ATTR_FLOW_ID;
attrs[1].id = SAI_MACSEC_SC_ATTR_MACSEC_SCI;

3 lines instead of 6, clear is not needed, back will get iterator when you can access directly by index

vslib/src/SwitchStateBaseMACsec.cpp
```diff
@@ -0,0 +1,859 @@
...
+    SAI_METADATA_GET_ATTR_BY_ID(attr, SAI_MACSEC_SA_ATTR_AN, attr_count, attr_list);
+    macsec_attr.m_an = attr->value.u8;
+
+    SAI_METADATA_GET_ATTR_BY_ID(attr, SAI_MACSEC_SA_ATTR_ENCRYPTION_ENABLE, attr_count, attr_list);
+    macsec_attr.m_encryption_enable = attr->value.booldata;
+
+    SAI_METADATA_GET_ATTR_BY_ID(attr, SAI_MACSEC_SA_ATTR_SAK, attr_count, attr_list);
+    macsec_attr.m_sak = sai_serialize_hex_binary(attr->value.macsecsak);
+
+    SAI_METADATA_GET_ATTR_BY_ID(attr, SAI_MACSEC_SA_ATTR_SAK_256_BITS, attr_count, attr_list);
```
kcudnik, line 564: empty lines before/after macros

vslib/src/SwitchStateBaseMACsec.cpp
```diff
@@ -0,0 +1,859 @@
...
+    CHECK_STATUS(
+        get(
+            SAI_OBJECT_TYPE_MACSEC_SC,
+            attr->value.oid,
+            static_cast<uint32_t>(attrs.size()),
+            attrs.data()));
+    auto flow_id = attrs[0].value.oid;
+    auto sci = attrs[1].value.u64;
+    std::stringstream sci_convert;
+    sci_convert << std::hex << bswap_64(sci);
```
kcudnik, line 542: is this bswap only in 1 way ? this could be dangerous if we dont know architecture

vslib/src/SwitchStateBaseMACsec.cpp
```diff
@@ -0,0 +1,859 @@
...
+    macsec_attr.m_an = attr->value.u8;
+
+    SAI_METADATA_GET_ATTR_BY_ID(attr, SAI_MACSEC_SA_ATTR_ENCRYPTION_ENABLE, attr_count, attr_list);
+    macsec_attr.m_encryption_enable = attr->value.booldata;
+
+    SAI_METADATA_GET_ATTR_BY_ID(attr, SAI_MACSEC_SA_ATTR_SAK, attr_count, attr_list);
+    macsec_attr.m_sak = sai_serialize_hex_binary(attr->value.macsecsak);
+
+    SAI_METADATA_GET_ATTR_BY_ID(attr, SAI_MACSEC_SA_ATTR_SAK_256_BITS, attr_count, attr_list);
+    if (!attr->value.booldata)
```
kcudnik, line 565: empty line before if

vslib/src/SwitchStateBaseMACsec.cpp
```diff
@@ -0,0 +1,859 @@
...
+    macsec_attr.m_auth_key = sai_serialize_hex_binary(attr->value.macsecauthkey);
+
+    if (macsec_attr.m_direction == SAI_MACSEC_DIRECTION_EGRESS)
+    {
+        SAI_METADATA_GET_ATTR_BY_ID(attr, SAI_MACSEC_SA_ATTR_XPN, attr_count, attr_list);
+    }
+    else
+    {
+        SAI_METADATA_GET_ATTR_BY_ID(attr, SAI_MACSEC_SA_ATTR_MINIMUM_XPN, attr_count, attr_list);
+    }
```
kcudnik, line 582: empty line after }

vslib/src/SwitchStateBaseMACsec.cpp
```diff
@@ -0,0 +1,859 @@
...
+            attr_count,
+            attr_list,
+            macsec_attr);
+    case SAI_OBJECT_TYPE_MACSEC_SA:
+        return loadMACsecAttrFromMACsecSA(
+            object_id,
+            attr_count,
+            attr_list,
+            macsec_attr);
+    default:
```
kcudnik, line 617: empty line before case/default

kcudnik:
also, seems like much of functionality should be only used when tap_device is enabled, but seems like macsec is always used despite that flag, please add unittests for setting/getting macsec related objects attributes in tests/BRCM and recording file to make sure that macsec will work without tap device enabled

meta/saiserialize.cpp
```diff
@@ -1414,6 +1423,29 @@ std::string sai_serialize_acl_capability(
...
+    if (buffer == NULL || length == 0)
+    {
+        return s;
+    }
+    s.resize(2 * length, '0');
+    const unsigned char *input = static_cast<const unsigned char *>(buffer);
+    char *output = &s[0];
+    for (size_t i = 0; i < length; i++)
+    {
+        snprintf(&output[i * 2], 3, "%02X", input[i]);
```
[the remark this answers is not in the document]
Pterosaur: I'm not sure whether the second argument `n` will include the last `\0`. According to the document, http://www.cplusplus.com/reference/cstdio/snprintf/ , I guess it include `\0`, which means I want to 2 valid chars and 1 `\0`

vslib/inc/MACsecAttr.h
```diff
@@ -0,0 +1,37 @@
...
+        bool m_encryption_enable;
+        sai_int32_t m_direction;
+        std::shared_ptr<HostInterfaceInfo> m_info;
+
+        // Explicitely declare constructor and destructor as non-inline functions
+        // to avoid 'call is unlikely and code size would grow [-Werror=inline]'
+        // in "-O3" optimization
+        MACsecAttr();
+
+        ~MACsecAttr();
```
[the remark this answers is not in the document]
Pterosaur: I wanted to declare them as  "= default", but it will raise a compile error 'call is unlikely and code size would grow'. Because '= default' will be the inline function by default.

In O3 optimization, a variable with inline ctr/dtr will be rejected if it's a local variable in a function.

 Here is some hint I found: https://stackoverflow.com/questions/36921744/gcc-o3-flag-causes-winline-call-is-unlikely-and-code-size-would-grow-warning . I don't know what is the graceful method to fix it. So I just choose to explicitly declare ctr/dtr to fix this issue. Do you have any suggestion?

vslib/src/MACsecForwarder.cpp
```diff
@@ -0,0 +1,191 @@
...
+            sizeof(sock_address)) < 0)
+    {
+        close(m_macsecfd);
+        SWSS_LOG_THROW(
+            "bind failed on %s",
+            m_macsec_interface_name.c_str());
+    }
+
+    m_forward_thread =
+        std::make_shared<std::thread>(&MACsecForwarder::forward, this);
```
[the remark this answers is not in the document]
Pterosaur: Here is our HLD https://github.com/Azure/SONiC/blob/9edcf67d0e0095e9f831a94ddf4fbb4a267adf7b/doc/macsec/MACsec_hld.md#345-virtual-macsec-sai .  As you can see, we need a thread to forward the traffic from MACsec device to Ethernet device. It looks like we cannot leverage the previous forwarding threads to do it.

vslib/src/HostInterfaceInfo.cpp
```diff
@@ -197,6 +231,17 @@ void HostInterfaceInfo::veth2tap_fun()
 
         async_process_packet_for_fdb_event(buffer, size);
 
+        auto ret = m_e2tFilters.execute(buffer, size);
```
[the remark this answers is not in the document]
Pterosaur: In MACsec scenario, we also need to write FDB into internal database. So I think to change the order doesn't make sense. How about we abstract `async_process_packet_for_fdb_event` as a filter and use the priority of filter to decide whether need to insert FDB. 
In the future, if any filter don't want to add FDB, just set a higher priority before the `async_process_packet_for_fdb_event` or delete this filter from filter group.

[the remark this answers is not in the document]
kcudnik: why you need to also write fdb in macsec scenarion?
i think we alwasy want to write fdb event in redis database if we are in execution of this path
i dont know the details on macse and it's pipeline, but seems wrong to me if we would juggle fdb event based on macsec priority, could you explain more here and give some scenarions ?

vslib/src/MACsecManager.cpp
```diff
@@ -0,0 +1,748 @@
...
+}
+
+std::string MACsecManager::shellquote(
+    _In_ const std::string &str) const
+{
+    SWSS_LOG_ENTER();
+
+    static const std::regex re("([$`\"\\\n])");
+
+    return "\"" + std::regex_replace(str, re, "\\$1") + "\"";
```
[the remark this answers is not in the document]
Pterosaur: This conversion is copied from https://github.com/Azure/sonic-swss/blob/85036dfbf22fb4b76016f115a377f77983ab6ab0/cfgmgr/shellcmd.h#L26-L30

This helper function was used with `swss::exec` in many position in SONiC project. So I guess it should consider that `bash/sh` situation.

Maybe a better practice is to use quoted in C++14(https://en.cppreference.com/w/cpp/io/manip/quoted), but the current C++ version looks like C++11

vslib/src/SwitchState.cpp
```diff
@@ -291,3 +293,55 @@ std::shared_ptr<saimeta::Meta> SwitchState::getMeta()
 
     return meta;
 }
+
+void SwitchState::findObjects(
+        _In_ sai_object_type_t object_type,
+        _In_ const sai_attribute_t &expect,
```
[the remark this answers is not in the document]
Pterosaur: I need a function to find out all objects that have specified attributes. Do you have any suggestion to design such function?

vslib/src/SwitchState.cpp
```diff
@@ -291,3 +293,55 @@ std::shared_ptr<saimeta::Meta> SwitchState::getMeta()
...
+            sai_object_id_t object_id = SAI_NULL_OBJECT_ID;
+            sai_deserialize_object_id(obj.first, object_id);
+            objects.push_back(object_id);
+        }
+
+    }
+
+}
+
+bool SwitchState::dumpObject(
```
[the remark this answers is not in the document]
Pterosaur: I want a function that can find out the attributes table according to object id, meanwhile I hope the type of attributes table is `std::vector<sai_attribute_t>`.
Because I have some functions as following, I use they to load MACsec attributes. I hope all of them can be used both in create, get, remove and set actions. 
The parameters of `create` and `set` actions include `sai_attribute_t *` which means I have opportunity to load MACsec Attr from their parameters. 
But parameters of `remove` and `get` actions only include object id, I have to find out those MACsec attributes  I needed according to the object ID. 
So I designed this function, `dumpObject`, to dump

https://github.com/Azure/sonic-sairedis/blob/d937d66426b3d62e10cb710dd192ad45b485ec1a/vslib/inc/SwitchStateBase.h#L473-L501

vslib/src/MACsecManager.cpp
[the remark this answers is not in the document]
kcudnik: ok

vslib/src/SwitchStateBaseMACsec.cpp
```diff
@@ -0,0 +1,859 @@
...
+        return SAI_STATUS_SUCCESS;
+    }
+
+    // Find all MACsec SCs that use this flow
+    std::vector<sai_object_id_t> macsec_scs;
+    attr.id = SAI_MACSEC_SC_ATTR_FLOW_ID;
+    attr.value.oid = flow_id;
+    findObjects(SAI_OBJECT_TYPE_MACSEC_SC, attr, macsec_scs);
+    if (object_type == SAI_OBJECT_TYPE_MACSEC_SC)
+    {
```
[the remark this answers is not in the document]
Pterosaur: I cannot move findObjects  into the `if`, because macsec_scs will be used at below code https://github.com/Azure/sonic-sairedis/pull/691/files?file-filters%5B%5D=.cpp#diff-254eea58a90a57d74818f0f06d67bcc09fa036c43af818c63988547bb9b99ae7R717

Pterosaur:
> also, seems like much of functionality should be only used when tap_device is enabled, but seems like macsec is always used despite that flag, please add unittests for setting/getting macsec related objects attributes in tests/BRCM and recording file to make sure that macsec will work without tap device enabled

I will submit the PR of unittests as a independent PR later. Because I'm not familiar about how to write the UnitTest of sairedis and I need some time to learn it.

vslib/src/HostInterfaceInfo.cpp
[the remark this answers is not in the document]
lguohan: in the asic macsec is processed before fdb learning, therefore if macsec decide to drop the packet, then fdb learning should not be generated.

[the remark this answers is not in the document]
Pterosaur: I synced this question with Guohan. You are right.
The MACsec should be processed before fdb learning, And if macsec decide to drop the packet, then fdb learning should not be generated.

kcudnik:
works for me

PR #692, opened 2020-11-03
test bulk api for vs tests

lguohan:
this is to test vs tests for the bulk api implementation. DO NOT merge this pr.

### DIFF
--- a/syncd/scripts/syncd_init_common.sh
+++ b/syncd/scripts/syncd_init_common.sh
@@ -29,7 +29,7 @@ CMD_ARGS+=" -u"
 
 # Use bulk api`s in SAI
 # currently disabled since most vendors don't support that yet
-# CMD_ARGS+=" -l"
+CMD_ARGS+=" -l"
 
 # Set synchronous mode if it is enabled in CONFIG_DB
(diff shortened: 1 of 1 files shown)


lguohan:
@kcudnik, i think the build api for syncd has some issue to pass the vs tests. can someone take a look?

kcudnik:
We recently talked that there is and issue and we want to disable "-l" flag, so why enable it in this pr ? just to test it ?

kcudnik:
@dmytroxshevchuk  you added bulk api support can you take a look why this is failing ?

dmytroxshevchuk:
@lguohan @kcudnik I added bulk API in syncd. As for me my mistake was with log level in case when bulk api is not implemented(its cause of much logs from syncd), for now it is fixed in PR #687, thanks for this. Regarding vs tests it was passed in my PR #656, when bulk was implemented in syncd, so its interesting why tests failing now.
I will check why the tests fall.

lguohan:
i do not know if it makes difference, but sync mode is now enabled by default on swss

dmytroxshevchuk:
WIP: #694

kcudnik:
this could be revisited, we already have fallback for bulk api in syncd even if vendors don't support that

PR #693, opened 2020-11-04
Generalizing config.bcm support for BRCM silicons

geans-pin:


kcudnik:
that's a lot of stuff added to config, does this needs to be executed every time ? and is this the right file to add this ?

geans-pin:
> that's a lot of stuff added to config, does this needs to be executed every time ? and is this the right file to add this ?

Yes, this is right place to add this which is only for BRCM and will be executed for every SONIC BOOT syncd docker init. Please check the HLD => https://github.com/Azure/SONiC/pull/699

syncd/scripts/syncd_init_common.sh
```diff
@@ -87,7 +87,56 @@ function set_start_type()
...
+                        echo "Keep the config $(grep $sedline $file) in $file"
+                     else
+                        echo $line >> $file
+                     fi
+                 fi
+             fi
+           done < $COMMON_CONFIG_BCM
+           echo "# End of chip common properties" >> $file         
+       done
+       echo "Merging $PLT_CONFIG_BCM with $COMMON_CONFIG_BCM, merge files stored in $CONFIG_BCM"
```
lguohan, line 127: i think the problem of this pr is that it does not provide a good transition path for all broadcom skus. for example, for existing sku, the config.bcm is working fine. how to you ensure your change will not affect existing skus?

lguohan: to just clarify, i am not saying i do not like this efforts, i think it is good to unify the config.bcm configuration, but it is better to provide a better transition path to every platform since you may not have tested this configuration on all platforms, and it is better to give an option for each platform to opt-in.

geans-pin: lguohan,

The infra is widely used in Broadcom SONIC for all the Broadcom SONIC platforms including TD2/TD3/TH2/TH3. BRCM SONIC team have also created the common config in sonic-buildimage
for maintaining each silicon common config. See the PR (https://github.com/Azure/sonic-buildimage/pull/5818). So, the infra is qualified. The option for each platform is that if there is the same property in ODM config.bcm, the common config infra will not overwrite it and will still use the ODM config.bcm property. So, we still reserve the option for ODM to maintain their platform specific property setting.  What we do is just merged common config to it, but not overwrite ODM config.bcm existing setting. Hope this info can resolve what your concern. 

Geans

lguohan: no, not fully addressed my concern. since the common config.bcm has some new attributes introduced, we do not know if those has some un-wanted effects or not. I think we need to give each platform owner the option to do a full test and then opt-in.

syncd/scripts/syncd_init_common.sh
```diff
@@ -87,7 +87,56 @@ function set_start_type()
...
 {
-    if [ -f "/etc/sai.d/sai.profile" ]; then
+    PLATFORM_COMMON_DIR=/usr/share/sonic/device/x86_64-broadcom_common
+
+    cp -f $HWSKU_DIR/*.config.bcm /tmp
+    cp -f /etc/sai.d/sai.profile /tmp
+    CONFIG_BCM=$(find /tmp -name '*.bcm')
+    PLT_CONFIG_BCM=$(find $HWSKU_DIR -name '*.bcm')
+    SAI_PROFILE=$(find /tmp -name 'sai.profile')
+    sed -i 's+/usr/share/sonic/hwsku+/tmp+g' $SAI_PROFILE
```
lguohan, line 97: line 92-97 can be moved to build time.

syncd/scripts/syncd_init_common.sh
geans-pin: Actually, we have used the infra in the BRCM SONIC PDE to work with each ODM and run the related qualification testing. Not sure if you want to create an option in this infra to enable or disable this feature for each ODM. Maybe we can think about to add some config in makefile to enable/disable the feature.

geans-pin: lguohan,

I have added an option for this. By default the feature is disabled, and ODM can enable this feature by touch the common_config_support file in the $PLATFORM_DIR. Let me know if this works for you

Geans

Geans

geans-pin:
lguohan,

I have added an option for this in new commit. By default the common config feature is disabled, and ODM can enable this feature by touch the common_config_support file in the $PLATFORM_DIR. Let me know if this works for you

Geans

geans-pin:
> /Azurepipelines run

Not sure why see this ERR ?

main::test_brcm_full_to_full_no_bridge: Fresh start
Killing syncd
Flushing redis
Starting syncd
Replay BCM56850/full.rec
Replay BCM56850/full_second_no_bridge.rec
main::test_brcm_full_to_full_no_bridge_restart: Fresh start
Killing syncd
Flushing redis
Starting syncd
Replay BCM56850/full.rec
  what():  :- performNotifySyncd: response status SAI_STATUS_SUCCESS is different than syncd status SAI_STATUS_FAILURE
player BCM56850/full.rec: exitcode: 134:
FAIL: BCM56850.pl
Using scripts dir 'MLNX2700'
main::test_mlnx_nhg_member: Fresh start

geans-pin:
Resubmit and close this one

lguohan:
@gechiang  to review this one as well.

gechiang (approved):

PR #695, opened 2020-11-06
[syncd][sairedis] Change pub/sub model to push/pull in zmq notification

kcudnik:
There is issue with publisher/subscriber, which I use in notification delivery in zmq channel in sairedis. Issue is that first message/notification I never delivered from publisher to subscriber, and seems like this issue is still not solved ☹ Here are details: https://github.com/zeromq/libzmq/issues/2267


kcudnik:
retest vs please

qiluo-msft (approved):

kcudnik:
retest vs please

PR #698, opened 2020-11-10
[vs] SAI support for VOQ switches

vganesan-nokia:
Core Port-Index Map support for VOQ system ports. This is part of SAI support of VOQ VS testing

### DIFF
--- a/vslib/src/CorePortIndexMap.cpp
+++ b/vslib/src/CorePortIndexMap.cpp
@@ -6,6 +6,8 @@
 
 using namespace saivs;
 
+#define VS_IF_PREFIX   "eth"
+
 CorePortIndexMap::CorePortIndexMap(
         _In_ uint32_t switchIndex):
     m_switchIndex(switchIndex)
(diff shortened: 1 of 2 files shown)


vslib/src/CorePortIndexMap.cpp
```diff
@@ -0,0 +1,183 @@
...
+std::string CorePortIndexMap::getInterfaceFromCorePortIndex(
+        _In_ const std::vector<uint32_t>& corePortIndex) const
+{
+    SWSS_LOG_ENTER();
+
+    auto it = m_corePortIndexToIfName.find(corePortIndex);
+
+    if (it == m_corePortIndexToIfName.end())
+    {
+        SWSS_LOG_WARN("Core port index (%d, %d) not found on index %u", corePortIndex[0], corePortIndex[1], m_switchIndex);
```
kcudnik, line 177: how you know if corePortIndex is at least 2 indexes ? use at() to get value it will throw when not enough items in vector

kcudnik (changes requested):

vslib/src/CorePortIndexMap.cpp
vganesan-nokia: Fixed. Thanks.

kcudnik:
retest vs please

kcudnik (approved):

PR #701, opened 2020-11-10
[vs] SAI support for VOQ switches - Switch State Initialization

vganesan-nokia:
SAI emulation support added for VOQ system ports. The changes include switch create initialization for VOQ switch objects. APIs for VOQ switch objects initialization, System Port configuration and System Port list initialization are provided.

### DIFF
--- a/vslib/src/SwitchStateBase.cpp
+++ b/vslib/src/SwitchStateBase.cpp
@@ -2523,7 +2523,7 @@ sai_status_t SwitchStateBase::initialize_voq_switch_objects(
 sai_status_t SwitchStateBase::create_system_ports(
        _In_ int32_t voq_switch_id,
        _In_ uint32_t sys_port_count,
-       _In_ sai_system_port_config_t *sys_port_cfg_list)
+       _In_ const sai_system_port_config_t *sys_port_cfg_list)
 {
     SWSS_LOG_ENTER();
 
--- a/vslib/src/SwitchStateBase.cpp
+++ b/vslib/src/SwitchStateBase.cpp
@@ -1786,7 +1786,7 @@ sai_status_t SwitchStateBase::refresh_read_only(
 
             case SAI_SWITCH_ATTR_NUMBER_OF_SYSTEM_PORTS:
             case SAI_SWITCH_ATTR_SYSTEM_PORT_LIST:
-                return SAI_STATUS_SUCCESS;
+                return refresh_system_port_list(meta);
         }
     }
 
@@ -2515,8 +2515,6 @@ sai_status_t SwitchStateBase::initialize_voq_switch_objects(
 
     CHECK_STATUS(set_system_port_list());
 
-    CHECK_STATUS(set_voq_switch_attributes(voq_switch_id, voq_max_cores));
-
     return SAI_STATUS_SUCCESS;
 }
 
@@ -2593,28 +2591,16 @@ sai_status_t SwitchStateBase::set_system_port_list()
     return set(SAI_OBJECT_TYPE_SWITCH, m_switch_id, &attr);
 }
 
-sai_status_t SwitchStateBase::set_voq_switch_attributes(
-        _In_ int32_t voq_switch_id,
-        _In_ uint32_t voq_max_cores)
+sai_status_t SwitchStateBase::refresh_system_port_list(
+        _In_ const sai_attr_metadata_t *meta)
 {
     SWSS_LOG_ENTER();
 
-    SWSS_LOG_INFO("set voq switch attributes");
+    // Currently, system ports info are not changing and also since currently the local
+    // port oid for local system ports is retrieved only during portsorch initialization,
+    // no referesh is done
 
-    sai_attribute_t attr;
+    // In the future, when dynamic system port config (add/delete) is implemented, re-visit TODO
 
-    attr.id = SAI_SWITCH_ATTR_TYPE;
-    attr.value.s32 = SAI_SWITCH_TYPE_VOQ;
-
-    CHECK_STATUS(set(SAI_OBJECT_TYPE_SWITCH, m_switch_id, &attr));
-
-    attr.id = SAI_SWITCH_ATTR_SWITCH_ID;
-    attr.value.u32 = voq_switch_id;
-
-    CHECK_STATUS(set(SAI_OBJECT_TYPE_SWITCH, m_switch_id, &attr));
-
-    attr.id = SAI_SWITCH_ATTR_MAX_SYSTEM_CORES;
-    attr.value.u32 = voq_max_cores;
-
-    return set(SAI_OBJECT_TYPE_SWITCH, m_switch_id, &attr);
+    return SAI_STATUS_SUCCESS;
 }
--- a/vslib/src/SwitchStateBase.cpp
+++ b/vslib/src/SwitchStateBase.cpp
@@ -1783,6 +1783,10 @@ sai_status_t SwitchStateBase::refresh_read_only(
             case SAI_SWITCH_ATTR_AVAILABLE_DNAT_ENTRY:
             case SAI_SWITCH_ATTR_AVAILABLE_DOUBLE_NAT_ENTRY:
                 return SAI_STATUS_SUCCESS;
+
+            case SAI_SWITCH_ATTR_NUMBER_OF_SYSTEM_PORTS:
+            case SAI_SWITCH_ATTR_SYSTEM_PORT_LIST:
+                return SAI_STATUS_SUCCESS;
         }
     }
 
@@ -1807,6 +1811,17 @@ sai_status_t SwitchStateBase::refresh_read_only(
                  */
 
             case SAI_PORT_ATTR_OPER_STATUS:
+                return SAI_STATUS_SUCCESS;
+        }
+    }
+
+    if (meta->objecttype == SAI_OBJECT_TYPE_SYSTEM_PORT)
+    {
+        switch (meta->attrid)
+        {
+            case SAI_SYSTEM_PORT_ATTR_TYPE:
+            case SAI_SYSTEM_PORT_ATTR_CONFIG_INFO:
+            case SAI_SYSTEM_PORT_ATTR_PORT:
                 return SAI_STATUS_SUCCESS;
         }
     }
@@ -2423,4 +2438,177 @@ void SwitchStateBase::debugSetStats(
     {
         m_countersMap[key][kvp.first] = kvp.second;
     }
+}
+
+sai_status_t SwitchStateBase::initialize_voq_switch_objects(
+                    _In_ uint32_t attr_count,
+                    _In_ const sai_attribute_t *attr_list)
+{
+    SWSS_LOG_ENTER();
+
+    bool voq_switch = false;
+    int32_t voq_switch_id = -1;
+    uint32_t voq_max_cores = 0;
+    uint32_t sys_port_count = 0;
+    sai_system_port_config_t *sys_port_cfg_list = NULL;
+
+    for (uint32_t i = 0; i < attr_count; i++)
+    {
+        switch (attr_list[i].id)
+        {
+            case SAI_SWITCH_ATTR_TYPE:
+                if (attr_list[i].value.u32 != SAI_SWITCH_TYPE_VOQ)
+                {
+                    //Switch is not being set as VOQ type.
+                    return SAI_STATUS_SUCCESS;
+                }
+                else
+                {
+                    voq_switch = true;
+                }
+                break;
+
+            case SAI_SWITCH_ATTR_SWITCH_ID:
+                voq_switch_id = (int32_t) attr_list[i].value.u32;
+                if (voq_switch_id < 0)
+                {
+                    SWSS_LOG_ERROR("Invalid VOQ switch id %d", voq_switch_id);
+                    return (int32_t)SAI_STATUS_INVALID_ATTR_VALUE_0 + (int32_t)i;
+                }
+                break;
+
+            case SAI_SWITCH_ATTR_MAX_SYSTEM_CORES:
+                voq_max_cores = attr_list[i].value.u32;
+                if (voq_max_cores < 1)
+                {
+                    SWSS_LOG_ERROR("Invalid VOQ max system cores %d", voq_max_cores);
+                    return (int32_t)SAI_STATUS_INVALID_ATTR_VALUE_0 + (int32_t)i;
+                }
+                break;
+
+            case SAI_SWITCH_ATTR_SYSTEM_PORT_CONFIG_LIST:
+                sys_port_count = attr_list[i].value.sysportconfiglist.count;
+                sys_port_cfg_list = attr_list[i].value.sysportconfiglist.list;
+                if (sys_port_count < 1 || !sys_port_cfg_list)
+                {
+                    SWSS_LOG_ERROR("Invalid voq system port config info! sys port count %d, sys port list %p", sys_port_count, sys_port_cfg_list);
+                    return (int32_t)SAI_STATUS_INVALID_ATTR_VALUE_0 + (int32_t)i;
+                }
+                break;
+
+            default:
+                //Ignore other attributes. They are processed elsewhere.
+                break;
+        }
+    }
+
+    if (!voq_switch)
+    {
+        //No switch type attribute in the attribute list.
+        return SAI_STATUS_SUCCESS;
+    }
+
+    CHECK_STATUS(create_system_ports(voq_switch_id, sys_port_count, sys_port_cfg_list));
+
+    CHECK_STATUS(set_system_port_list());
+
+    CHECK_STATUS(set_voq_switch_attributes(voq_switch_id, voq_max_cores));
+
+    SWSS_LOG_NOTICE("Initialized VOQ switch.");
+
+    return SAI_STATUS_SUCCESS;
+}
+
+sai_status_t SwitchStateBase::create_system_ports(
+       _In_ int32_t voq_switch_id,
+       _In_ uint32_t sys_port_count,
+       _In_ sai_system_port_config_t *sys_port_cfg_list)
+{
+    SWSS_LOG_ENTER();
+
+    SWSS_LOG_INFO("create system ports");
+
+    m_system_port_list.clear();
+
+    for (uint32_t i = 0; i < sys_port_count; i++)
+    {
+        SWSS_LOG_DEBUG("create system port index %u", i);
+
+        sai_object_id_t system_port_id;
+
+        CHECK_STATUS(create(SAI_OBJECT_TYPE_SYSTEM_PORT, &system_port_id, m_switch_id, 0, NULL));
+
+        m_system_port_list.push_back(system_port_id);
+
+        sai_attribute_t attr;
+
+        attr.id = SAI_SYSTEM_PORT_ATTR_CONFIG_INFO;
+        attr.value.sysportconfig = sys_port_cfg_list[i];
+
+        CHECK_STATUS(set(SAI_OBJECT_TYPE_SYSTEM_PORT, system_port_id, &attr));
+
+        attr.id = SAI_SYSTEM_PORT_ATTR_TYPE;
+        attr.value.s32 = SAI_SYSTEM_PORT_TYPE_REMOTE;
+
+        if ((int32_t)sys_port_cfg_list[i].attached_switch_id == voq_switch_id)
+        {
+            attr.value.s32 = SAI_SYSTEM_PORT_TYPE_LOCAL;
+
+            //Need to set local port oid attribute TODO
+        }
+
+        CHECK_STATUS(set(SAI_OBJECT_TYPE_SYSTEM_PORT, system_port_id, &attr));
+    }
+
+    return SAI_STATUS_SUCCESS;
+}
+
+sai_status_t SwitchStateBase::set_system_port_list()
+{
+    SWSS_LOG_ENTER();
+
+    SWSS_LOG_INFO("set system port list");
+
+    // NOTE: this is static, but will be refreshed on read only get
+
+    sai_attribute_t attr;
+
+    uint32_t sys_port_count = (uint32_t)m_system_port_list.size();
+
+    attr.id = SAI_SWITCH_ATTR_SYSTEM_PORT_LIST;
+    attr.value.objlist.count = sys_port_count;
+    attr.value.objlist.list = m_system_port_list.data();
+
+    CHECK_STATUS(set(SAI_OBJECT_TYPE_SWITCH, m_switch_id, &attr));
+
+    attr.id = SAI_SWITCH_ATTR_NUMBER_OF_SYSTEM_PORTS;
+    attr.value.u32 = sys_port_count;
+
+    return set(SAI_OBJECT_TYPE_SWITCH, m_switch_id, &attr);
+}
+
+sai_status_t SwitchStateBase::set_voq_switch_attributes(
+        _In_ int32_t voq_switch_id,
+        _In_ uint32_t voq_max_cores)
+{
+    SWSS_LOG_ENTER();
+
+    SWSS_LOG_INFO("set voq switch attributes");
+
+    sai_attribute_t attr;
+
+    attr.id = SAI_SWITCH_ATTR_TYPE;
+    attr.value.s32 = SAI_SWITCH_TYPE_VOQ;
+
+    CHECK_STATUS(set(SAI_OBJECT_TYPE_SWITCH, m_switch_id, &attr));
+
+    attr.id = SAI_SWITCH_ATTR_SWITCH_ID;
+    attr.value.u32 = voq_switch_id;
+
+    CHECK_STATUS(set(SAI_OBJECT_TYPE_SWITCH, m_switch_id, &attr));
+
+    attr.id = SAI_SWITCH_ATTR_MAX_SYSTEM_CORES;
+    attr.value.u32 = voq_max_cores;
+
+    return set(SAI_OBJECT_TYPE_SWITCH, m_switch_id, &attr);
 }
--- a/vslib/src/SwitchStateBase.cpp
+++ b/vslib/src/SwitchStateBase.cpp
@@ -2596,11 +2596,9 @@ sai_status_t SwitchStateBase::refresh_system_port_list(
 {
     SWSS_LOG_ENTER();
 
-    // Currently, system ports info are not changing and also since currently the local
-    // port oid for local system ports is retrieved only during portsorch initialization,
-    // no referesh is done
+    // Currently, system ports info are not changing. So no referesh is done
 
-    // In the future, when dynamic system port config (add/delete) is implemented, re-visit TODO
+    // TODO In the future, when dynamic system port config (add/delete) is implemented, re-visit
 
     return SAI_STATUS_SUCCESS;
 }
--- a/vslib/src/SwitchStateBase.cpp
+++ b/vslib/src/SwitchStateBase.cpp
@@ -1822,9 +1822,6 @@ sai_status_t SwitchStateBase::refresh_read_only(
             case SAI_SYSTEM_PORT_ATTR_TYPE:
             case SAI_SYSTEM_PORT_ATTR_PORT:
                 return SAI_STATUS_SUCCESS;
-
-            case SAI_SYSTEM_PORT_ATTR_CONFIG_INFO:
-                return refresh_system_port_config_info(meta, object_id);
         }
     }
 
@@ -2532,6 +2529,11 @@ sai_status_t SwitchStateBase::create_system_ports(
 
     SWSS_LOG_INFO("create system ports");
 
+    if(!sys_port_cfg_list)
+    {
+        return SAI_STATUS_FAILURE;
+    }
+
     m_system_port_list.clear();
 
     for (uint32_t i = 0; i < sys_port_count; i++)
@@ -2615,17 +2617,4 @@ sai_status_t SwitchStateBase::set_voq_switch_attributes(
     attr.value.u32 = voq_max_cores;
 
     return set(SAI_OBJECT_TYPE_SWITCH, m_switch_id, &attr);
-}
-
-sai_status_t SwitchStateBase::refresh_system_port_config_info(
-        _In_ const sai_attr_metadata_t *meta,
-        _In_ sai_object_id_t system_port_oid)
-{
-    SWSS_LOG_ENTER();
-
-    // Eventhough, this is not a read-only attribute, currently the system ports are configured
-    // only during creation. No dynamic configuration changes after creation. Threfore currently
-    // no referesh is needed
-
-    return SAI_STATUS_SUCCESS;
 }
--- a/vslib/src/SwitchStateBase.cpp
+++ b/vslib/src/SwitchStateBase.cpp
@@ -1820,9 +1820,11 @@ sai_status_t SwitchStateBase::refresh_read_only(
         switch (meta->attrid)
         {
             case SAI_SYSTEM_PORT_ATTR_TYPE:
-            case SAI_SYSTEM_PORT_ATTR_CONFIG_INFO:
             case SAI_SYSTEM_PORT_ATTR_PORT:
                 return SAI_STATUS_SUCCESS;
+
+            case SAI_SYSTEM_PORT_ATTR_CONFIG_INFO:
+                return refresh_system_port_config_info(meta, object_id);
         }
     }
 
@@ -2457,6 +2459,7 @@ sai_status_t SwitchStateBase::initialize_voq_switch_objects(
         switch (attr_list[i].id)
         {
             case SAI_SWITCH_ATTR_TYPE:
+
                 if (attr_list[i].value.u32 != SAI_SWITCH_TYPE_VOQ)
                 {
                     //Switch is not being set as VOQ type.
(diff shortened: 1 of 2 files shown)


vslib/src/SwitchStateBase.cpp
```diff
@@ -1783,6 +1783,10 @@ sai_status_t SwitchStateBase::refresh_read_only(
             case SAI_SWITCH_ATTR_AVAILABLE_DNAT_ENTRY:
             case SAI_SWITCH_ATTR_AVAILABLE_DOUBLE_NAT_ENTRY:
                 return SAI_STATUS_SUCCESS;
+
+            case SAI_SWITCH_ATTR_NUMBER_OF_SYSTEM_PORTS:
+            case SAI_SWITCH_ATTR_SYSTEM_PORT_LIST:
+                return SAI_STATUS_SUCCESS;
```
kcudnik, line 1789: this is wrong, i dont see actual list to br refreshed here or ports

> This is needed. PortsOrch reads this read-only attributes (SWITCH object). These are all fixed one time static configuration. They are not changed by SAI. No refreshing is needed. But returning failure will return failure for GET. So we always return SUCCESS.

i see create_system_port api in SAI/inc/saisystemport.h so there is a possibility to create new system ports in sysstem.

if this configuration is static per switch creation, add refresh function here like refresh_system_port_list() put a comment explaining there why it's static and why its not chaning, and then return success

vslib/src/SwitchStateBase.cpp
```diff
@@ -2424,3 +2439,176 @@ void SwitchStateBase::debugSetStats(
...
+    uint32_t voq_max_cores = 0;
+    uint32_t sys_port_count = 0;
+    sai_system_port_config_t *sys_port_cfg_list = NULL;
+
+    for (uint32_t i = 0; i < attr_count; i++)
+    {
+        switch (attr_list[i].id)
+        {
+            case SAI_SWITCH_ATTR_TYPE:
+                if (attr_list[i].value.u32 != SAI_SWITCH_TYPE_VOQ)
```
kcudnik, line 2460: empty line before each if in every file

vslib/src/SwitchStateBase.cpp
```diff
@@ -2424,3 +2439,176 @@ void SwitchStateBase::debugSetStats(
...
+                if (voq_switch_id < 0)
+                {
+                    SWSS_LOG_ERROR("Invalid VOQ switch id %d", voq_switch_id);
+                    return (int32_t)SAI_STATUS_INVALID_ATTR_VALUE_0 + (int32_t)i;
+                }
+                break;
+
+            case SAI_SWITCH_ATTR_MAX_SYSTEM_CORES:
+                voq_max_cores = attr_list[i].value.u32;
+                if (voq_max_cores < 1)
```
kcudnik, line 2482: empty line before each if in every file

vslib/src/SwitchStateBase.cpp
```diff
@@ -2424,3 +2439,176 @@ void SwitchStateBase::debugSetStats(
...
+                {
+                    SWSS_LOG_ERROR("Invalid VOQ max system cores %d", voq_max_cores);
+                    return (int32_t)SAI_STATUS_INVALID_ATTR_VALUE_0 + (int32_t)i;
+                }
+                break;
+
+            case SAI_SWITCH_ATTR_SYSTEM_PORT_CONFIG_LIST:
+                sys_port_count = attr_list[i].value.sysportconfiglist.count;
+                sys_port_cfg_list = attr_list[i].value.sysportconfiglist.list;
+                if (sys_port_count < 1 || !sys_port_cfg_list)
```
kcudnik, line 2492: empty line before each if in every file

vslib/src/SwitchStateBase.cpp
```diff
@@ -2424,3 +2439,176 @@ void SwitchStateBase::debugSetStats(
...
+                {
+                    SWSS_LOG_ERROR("Invalid voq system port config info! sys port count %d, sys port list %p", sys_port_count, sys_port_cfg_list);
+                    return (int32_t)SAI_STATUS_INVALID_ATTR_VALUE_0 + (int32_t)i;
+                }
+                break;
+
+            default:
+                //Ignore other attributes. They are processed elsewhere.
```
kcudnik, line 2500: always add space after //

vslib/src/SwitchStateBase.cpp
```diff
@@ -2424,3 +2439,176 @@ void SwitchStateBase::debugSetStats(
...
+                break;
+        }
+    }
+
+    if (!voq_switch)
+    {
+        //No switch type attribute in the attribute list.
```
kcudnik, line 2507: always add space after //

vslib/src/SwitchStateBase.cpp
```diff
@@ -2424,3 +2439,176 @@ void SwitchStateBase::debugSetStats(
...
+        return SAI_STATUS_SUCCESS;
+    }
+
+    CHECK_STATUS(create_system_ports(voq_switch_id, sys_port_count, sys_port_cfg_list));
+
+    CHECK_STATUS(set_system_port_list());
+
+    CHECK_STATUS(set_voq_switch_attributes(voq_switch_id, voq_max_cores));
+
+    SWSS_LOG_NOTICE("Initialized VOQ switch.");
```
kcudnik, line 2517: dot not needed at the end

vslib/src/SwitchStateBase.cpp
```diff
@@ -2424,3 +2439,176 @@ void SwitchStateBase::debugSetStats(
...
+        CHECK_STATUS(set(SAI_OBJECT_TYPE_SYSTEM_PORT, system_port_id, &attr));
+
+        attr.id = SAI_SYSTEM_PORT_ATTR_TYPE;
+        attr.value.s32 = SAI_SYSTEM_PORT_TYPE_REMOTE;
+
+        if ((int32_t)sys_port_cfg_list[i].attached_switch_id == voq_switch_id)
+        {
+            attr.value.s32 = SAI_SYSTEM_PORT_TYPE_LOCAL;
+
+            //Need to set local port oid attribute TODO
```
kcudnik, line 2557: always add space after //

vslib/src/SwitchStateBase.cpp
```diff
@@ -1811,6 +1815,17 @@ sai_status_t SwitchStateBase::refresh_read_only(
         }
     }
 
+    if (meta->objecttype == SAI_OBJECT_TYPE_SYSTEM_PORT)
+    {
+        switch (meta->attrid)
+        {
+            case SAI_SYSTEM_PORT_ATTR_TYPE:
+            case SAI_SYSTEM_PORT_ATTR_CONFIG_INFO:
```
kcudnik, line 1823: please remove SAI_SYSTEM_PORT_ATTR_CONFIG_INFO it's not READ_ONLY attribute

vslib/src/SwitchStateBase.cpp
```diff
@@ -1811,6 +1815,17 @@ sai_status_t SwitchStateBase::refresh_read_only(
...
+            case SAI_SYSTEM_PORT_ATTR_PORT:
+                return SAI_STATUS_SUCCESS;
```
kcudnik, line 1825: could SAI_SYSTEM_PORT_ATTR_PORT and SAI_SYSTEM_PORT_ATTR_TYPE ? do they need refresh ?

vslib/src/SwitchStateBase.cpp
```diff
@@ -2424,3 +2439,176 @@ void SwitchStateBase::debugSetStats(
         m_countersMap[key][kvp.first] = kvp.second;
     }
 }
+
+sai_status_t SwitchStateBase::initialize_voq_switch_objects(
```
kcudnik, line 2443: seems like this is main initialization function for voq, but it's not executed anywhere in SwitchStateBae (if it's generic for both Brcm/Mlnx switches)
should be called in: SwitchStateBase::initialize_default_objects()

kcudnik: you also want to take a look at warm_boot_initialize_objects

kcudnik (changes requested):

vslib/src/SwitchStateBase.cpp
vganesan-nokia: No. These are read-only attributes of switch object. Please refer to SAI/inc/saiswitch.h. However I took this feedback to apply on SAI_SYSTEM_PORT_ATTR_CONFIG_INFO attribute below. Thanks

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Fixed

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Fixed

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Fixed

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Fixed

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Removed.

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Fixed in this PR. But this line is removed in other PR #703

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Yes. This is not a read-only attribute. I added a function for refresh system port config info and return success. And also gave comment for doing so.

vslib/src/SwitchStateBase.cpp
vganesan-nokia: No. These are read-only attributes and they are not modified after initialization time. No refresh is needed for these

vslib/src/SwitchStateBase.cpp
vganesan-nokia: For comment 1: The VOQ switch objects initialized here are not default objects. They need specific configuration that should be supplied by orchagent during switch creation for all types of VoQ switches. The configurations are in the attribute list. Right now this is called from VirtualSwitchSaiInterface class from the place where init_swith() is called. This is the bottom most place where attribute list is visible. This is the reason why this API is invoked from here. If it is still preferred to get this API invoked from SwitchStateBase form initialize_default_object(), I can modify the init_switch() and initialize_default_objects() carry forward the attr_count and attr_list.

For comment 2: I had a look at the warm_boot_initialize_objects(). There is refreshing of m_port_list(). An equivalent for VoQ system will be refreshing m_system_port_list. As of now, there is nothing in the SAI or hardware or SDK that changes the system port info once the system is initialized. It is required that all system ports of the VoQ system must be pre-configured during bootup, irrespective of whether cold boot or warmboot. So at least for now, no need to do anything for warm boot as far as VOQ switch objects are concerned.

vslib/src/SwitchStateBase.cpp
kcudnik: read_only attributes are considered that they can change value during switch runtime, for example port_number, where ports can be removed or added during switch runtime, and those are always recalculated

vslib/src/SwitchStateBase.cpp
```diff
@@ -1820,9 +1820,11 @@ sai_status_t SwitchStateBase::refresh_read_only(
         switch (meta->attrid)
         {
             case SAI_SYSTEM_PORT_ATTR_TYPE:
-            case SAI_SYSTEM_PORT_ATTR_CONFIG_INFO:
             case SAI_SYSTEM_PORT_ATTR_PORT:
                 return SAI_STATUS_SUCCESS;
+
+            case SAI_SYSTEM_PORT_ATTR_CONFIG_INFO:
+                return refresh_system_port_config_info(meta, object_id);
```
kcudnik, line 1827: SAI_SYSTEM_PORT_ATTR_CONFIG_INFO is create_only attribute, it should not be refreshed

vslib/src/SwitchStateBase.cpp
```diff
@@ -2612,3 +2616,16 @@ sai_status_t SwitchStateBase::set_voq_switch_attributes(
...
     return set(SAI_OBJECT_TYPE_SWITCH, m_switch_id, &attr);
 }
+
+sai_status_t SwitchStateBase::refresh_system_port_config_info(
+        _In_ const sai_attr_metadata_t *meta,
+        _In_ sai_object_id_t system_port_oid)
+{
+    SWSS_LOG_ENTER();
+
+    // Eventhough, this is not a read-only attribute, currently the system ports are configured
```
kcudnik, line 2626: SAI_SYSTEM_PORT_ATTR_CONFIG_INFO is create_only attribute, it should not be refreshed

vslib/src/SwitchStateBase.cpp
kcudnik: if attribute is NOT read_only, it should not be in this function

vganesan-nokia: Yes. I'll remove SAI_SYSTEM_PORT_ATTR_CONFIG_INFO attribute and also remove the refresh function. Thanks

vslib/src/SwitchStateBase.cpp
kcudnik: Ad 1: If you need some extra action to be performed when creating switch, move this initialization to special case in function: sai_status_t SwitchStateBase::create, there are already some custom functions like port or hostif  that require some extra steps, and that would be the way to go, although, since switch object is special,  there could be some other issues with this, but please try to make it like port/hostif

Ad 2: ok

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Got it. Thanks for clarification. For a any given system port configuration (that is done only at the creation time), this is fixed depending on <switch_id, core index and core port index> - part of system port configuration. Since this is fixed through out the life of the system, we do not need refreshing of these attributes. However if the local port oid changes during run time, which unlikely, these need refresh

vslib/src/SwitchStateBase.cpp
```diff
@@ -2424,3 +2441,191 @@ void SwitchStateBase::debugSetStats(
...
+
+    CHECK_STATUS(set_voq_switch_attributes(voq_switch_id, voq_max_cores));
+
+    return SAI_STATUS_SUCCESS;
+}
+
+sai_status_t SwitchStateBase::create_system_ports(
+       _In_ int32_t voq_switch_id,
+       _In_ uint32_t sys_port_count,
+       _In_ sai_system_port_config_t *sys_port_cfg_list)
```
kcudnik, line 2529: can be const

vslib/src/SwitchStateBase.cpp
kcudnik: local port oid is which object type ? SAI_OBJECT_TYPE_PORT ? or SAI_OBJECT_TYPE_SYSTEM_PORT?
since PORT can be removed/added during runtime when doing port split. if your voq information don't change during lifetime of switch object existence, regardless of any other operation, then you don't need to do any refreshing here

vslib/src/SwitchStateBase.cpp
```diff
@@ -2424,3 +2441,191 @@ void SwitchStateBase::debugSetStats(
...
+        CHECK_STATUS(set(SAI_OBJECT_TYPE_SYSTEM_PORT, system_port_id, &attr));
+
+        attr.id = SAI_SYSTEM_PORT_ATTR_TYPE;
+        attr.value.s32 = SAI_SYSTEM_PORT_TYPE_REMOTE;
+
+        if ((int32_t)sys_port_cfg_list[i].attached_switch_id == voq_switch_id)
+        {
+            attr.value.s32 = SAI_SYSTEM_PORT_TYPE_LOCAL;
+
+            // Need to set local port oid attribute TODO
```
kcudnik, line 2561: what local port ?

vslib/src/SwitchStateBase.cpp
kcudnik: so far initialize_default_objects did not need attr_count/attr_list since creation of internal objects didnt depend on any attributes, in virtual switch each attribute is just stored in map, so there are no actual actions needed. in your case this is different, since you create objects system ports inside switch create based on switch attributes.  This maybe require to pass attr list and count to init_switch and pass to constructiors of SwitchStateBase
is system port object related anyhow to SAI_OBJECT_TYPE_PORT ?

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Local port oid is SAI_OBJECT_TYPE_PORT. I read it from m_port_list to associate it with system port corresponding to the local ports (Refer PR #703). The system port config info tuple mentioned above will not change. If m_port_list is not changed during runt time, no need to do refresh here. But I see there is refresh for m_port_list in refresh_ports(). Does this mean previously assigned oid of the existing port will change? I am wondering, should I refresh system port config to associate system port with changed local port oid, if at all changed.

vslib/src/SwitchStateBase.cpp
vganesan-nokia: System port config list passed at the time system initialization is the system port config of all ports in entire voq system. If any given system port config corresponds to local port (identified by switch_id), the attribute SAI_SYSTEM_PORT_ATTR_TYPE should be SAI_SYSTEM_PORT_TYPE_LOCAL when read. Also, if the type is local, the corresponding local port id should be available in SAI_SYSTEM_PORT_ATTR_PORT. To get local port oid from m_port_list corresponding to the system port, the core and core port index values are used. This processing is in PR #703 (vslib/src/SwitchStateBase.cpp: line number 2557 - 2594)

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Yes. I'll remove this function

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Yes. I'll remove this

vslib/src/SwitchStateBase.cpp
kcudnik: object type port list can change during runtime, ports can be removed and added, i dont know details about voq and system ports, but seems to me that they are somehow related/dependent and it will be tricky to make this change and refresh operations

vslib/src/SwitchStateBase.cpp
vganesan-nokia: o.k. I'll fix for making this function cost. I'll check for null of sys_port_cfg_list. However it is unlikely that this will happen since these attributes are mandatory for switch type VOQ, Either there will be all attributes or this function will not be called. If there is this attribute, the null check happens in the calller of this API.

vslib/src/SwitchStateBase.cpp
kcudnik: from your description seems like some system ports may correspond to local port and some not is that correct ? what is the purpose of such configuration ? Is there ad design spec somewhere about system ports ?

vganesan-nokia: Please refer to the VOQ HLD: https://github.com/Azure/SONiC/blob/master/doc/voq/voq_hld.md
If the system has only one asic, the number of system ports will be same as number of local ports. We have just system ports configured for the local ports only. We still need the mapping between local port and system port as needed by the portsorch to work (PR: https://github.com/Azure/sonic-swss/pull/1431) . If we have multi asic/multi line card system, System port config will have list of system ports all asics from all line cards. This mapping is need by PortsOrch to differentiate between local ports and system Ports initialization in portsorch.

kcudnik: i will dive to this document and see what i can understand. for now it seems to me like voq switch is another switch inside current switch, and it will be really hard to track all dependencies between those 2 internal switches and monitor all read only variables to be refreshed in correct way.
**_also currently when switch is created only 3 other objects are created internally: queues, ingress_priority_gorups and scheduler_groups. and all those 3 objects are considered dependencies of each port. and there are special task taken inside sairedis and syncd to process those dependencies especially on port add and port remove.
this voq system is adding another dependency which is system_port, and when this will be in use, another special handling  for this needs to be added in syncd/sairedis, this will be essential in warm boot scenario to not lost any objects,
also i dont know what will/should happen to voq system ports that have related local ports in case when local port will be removed  _**

kcudnik: also, currently sonic is using only 1 asic, but its written in way to support multiple asic/switches, but that path is not tested in any way yet, so i would not be surprised if there are many bugs related to multi asic sonic at current stage, not even mentioning support for multi asic voq

vganesan-nokia: As per current design,  for the PR: Azure/sonic-swss#1431 system port action for port removal does not exist. For now, to get 1431 through, I'll associate local port to system ports using info from current m_port_list. When the remove is handled in system port, we would consider an appropriate refresh model for system ports.

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Yes. I Agree. Without any processing existing for system port when local port is removed, it is unclear about having refreshing, especially since reading this attribute is one time during initialization of system ports. So for now, I'll not provide any refreshing for SAI_SYSTEM_PORT_ATTR_PORT. We'll re-visit this in the future.

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Yes. There is system port config representing every physical port in the system.

vganesan-nokia: > **there can be issue here, since initialize_default_objects is called inside VirtualSwitchSaiInterface::init_switch and this is called inside VirtualSwitchSaiInterface::create, which after init_switch calls ss->create(..) which creates switch, and this actually works, since create() will just populate internal string values of the switch, and it can't call initialize_voq inside this create function, let leave this for now and let's use that initialize after init_switch() as you had so far**

o.k. I'll leave it like how it is now. I mention this as response to comment in PR #702. Thanks.

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Fixed

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Fixed

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Fixed

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Fixed as mentioned

kcudnik: i don't see any place where sys_port_cfg_list is modified in this function

vslib/src/SwitchStateBase.cpp
kcudnik: ok, just keep in mind that in case of warm boot scenario some wired things can happen, like duplicated port/system_ports etc and it can be hard to identify where this actually happens

vganesan-nokia: o.k. Thanks for the feedback.

vslib/src/SwitchStateBase.cpp
kcudnik: yes, system port remove/create dont exits, but local port remove/create exists.
Also is this VOQ switch supported on Broadcom or Mellanox chip ? or this is on some new asic/silocon ?

vslib/src/SwitchStateBase.cpp
vganesan-nokia: I thought the comment was to consider making the function constant. I'll make sys_port_cfg_list constant so that its contents can not be changed.

vslib/src/SwitchStateBase.cpp
vganesan-nokia: As far as I know, currently VOQ switch support is available for Broadcom chip (DNX/Jerico). No idea about Mellanox and other chipsets.

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Fixed for *sys_port_cfg_list

vslib/src/SwitchStateBase.cpp
```diff
@@ -2424,3 +2438,183 @@ void SwitchStateBase::debugSetStats(
...
+
+    CHECK_STATUS(set(SAI_OBJECT_TYPE_SWITCH, m_switch_id, &attr));
+
+    attr.id = SAI_SWITCH_ATTR_NUMBER_OF_SYSTEM_PORTS;
+    attr.value.u32 = sys_port_count;
+
+    return set(SAI_OBJECT_TYPE_SWITCH, m_switch_id, &attr);
+}
+
+sai_status_t SwitchStateBase::set_voq_switch_attributes(
```
kcudnik, line 2596: as for this function, set of those 2 attributes is not required
https://github.com/opencomputeproject/SAI/blob/master/inc/saiswitch.h#L2175 - you already check that in initialize_voq_objects
also SAI_SWITCH_ATTR_TYPE is checked to be equal to VOQ in initialize objects
https://github.com/opencomputeproject/SAI/blob/master/inc/saiswitch.h#L2184 also checked in initialize objects

those 3 attributes you SET here, they are already passed on attr_list when calling switch_create, so there is not need for this entire function at all

vslib/src/SwitchStateBase.cpp
```diff
@@ -2424,3 +2438,183 @@ void SwitchStateBase::debugSetStats(
...
+            // Need to set local port oid attribute TODO
+        }
+
+        CHECK_STATUS(set(SAI_OBJECT_TYPE_SYSTEM_PORT, system_port_id, &attr));
+    }
+
+    return SAI_STATUS_SUCCESS;
+}
+
+sai_status_t SwitchStateBase::set_system_port_list()
```
kcudnik, line 2572: SAI_SWITCH_ATTR_SYSTEM_PORT_LIST and SAI_SWITCH_ATTR_NUMBER_OF_SYSTEM_PORTS are read only members, so this should be done as refresh function. even if system ports currently don't changed, its better to put this in proper location then later or wondering where it should be
i would propose to keep this is same scheme as local ports, just add refresh function for those 2 attributes similar to refresh_port_list, you can return there just success, and add comment that currently system ports are not changing so no refresh function is done and put TODO there (even better add log notice there "that list was not refreshed, fixme" or you can actually refresh those 2 read only values

kcudnik: i already commented this on 1789

kcudnik (changes requested):

vslib/src/SwitchStateBase.cpp
vganesan-nokia: I want values of these three attributes or at least swith_id and max_cores in the objectHash for retrieval later (by GET from orchagent, even though not doing now, we'll need it in the future) and to emulate the SAI as close as possible to real SAI. If I do not "set" in this function, will they be in objectHash? Would you please let me know where they are set?

vslib/src/SwitchStateBase.cpp
minionatwork: > what local port ? what is relation between system_port and port ?

SystemPort is applicable in VOQ based chassis system. When we say "system_port" it either port in local asic or port in remote asic on the same chassis. Every local port(Front panel port) will system port defined in VOQ chassis system. 

This doesnt apply any system that aren't VOQ.

minionatwork: > yes, system port remove/create dont exits, but local port remove/create exists.
> Also is this VOQ switch supported on Broadcom or Mellanox chip ? or this is on some new asic/silocon ?

VOQ switch is supported by broadcom currently in SONiC. I dont think/know, Mellanox has VOQ based chip. 

This work is for new chipset which are based on VOQ. (first one to onboard to SONiC is broadcom DNX chipset).

vslib/src/SwitchStateBase.cpp
vganesan-nokia: I'll fix as proposed.

vslib/src/SwitchStateBase.cpp
kcudnik: they will be posted into objectHash after init_switch, and orchagent will be able to get them each api call is under mutex so no issue here,
objectHash is set in VirtualSwitchSaiInterface::create(
675     return ss->create(object_type, serializedObjectId, switchId, attr_count, attr_list);

vganesan-nokia: o.k. Thanks. I'll fix as proposed

vganesan-nokia: Fixed

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Fixed.

vganesan-nokia:
retest vs please

vslib/src/SwitchStateBase.cpp
```diff
@@ -2593,28 +2591,16 @@ sai_status_t SwitchStateBase::set_system_port_list()
...
-    attr.id = SAI_SWITCH_ATTR_TYPE;
-    attr.value.s32 = SAI_SWITCH_TYPE_VOQ;
-
-    CHECK_STATUS(set(SAI_OBJECT_TYPE_SWITCH, m_switch_id, &attr));
-
-    attr.id = SAI_SWITCH_ATTR_SWITCH_ID;
-    attr.value.u32 = voq_switch_id;
+    // Currently, system ports info are not changing and also since currently the local
+    // port oid for local system ports is retrieved only during portsorch initialization,
+    // no referesh is done
```
kcudnik, line 2601: > local port oid for local system ports is retrieved only during portsorch initialization, no referesh is done

this is actually not true:
https://github.com/Azure/sonic-swss/blob/master/orchagent/portsorch.cpp#L1759
https://github.com/Azure/sonic-swss/blob/master/orchagent/portsorch.cpp#L1778

ports are remove and added inside portsorch, and this is happening after switch is created, so let it may happen that after switch_create you have 32 local ports, and then orchagent will remove some ports and add some ports, and you can end up in total 48 local ports (since for example 100G port was splitted to 4x25G ports) then probably VOQ configuration is not valid any more at this point

vslib/src/SwitchStateBase.cpp
```diff
@@ -2593,28 +2591,16 @@ sai_status_t SwitchStateBase::set_system_port_list()
...
 
-    CHECK_STATUS(set(SAI_OBJECT_TYPE_SWITCH, m_switch_id, &attr));
+    // In the future, when dynamic system port config (add/delete) is implemented, re-visit TODO
```
kcudnik, line 2603: this is ok, but please put TODO string in front: "//  TODO" its is done in all files all over the project, keep consistent

vslib/src/SwitchStateBase.cpp
vganesan-nokia: I meant this with respect to the upcoming PR https://github.com/Azure/sonic-swss/pull/1431. While adding System Ports (already pre-configured as part of switch create) in PortOrch constructor, to map the system ports corresponding to the local ports, the "local port oid" attribute of the system port info is retrieved. This is the only place we will be in need of this until dynamic system port add/removal is handled. 
If this is not clear, I'll remove this statement.

vslib/src/SwitchStateBase.cpp
vganesan-nokia: o.k. I'll fix

vslib/src/SwitchStateBase.cpp
kcudnik: i have only a little bit knowledge on orchagent, but im better on sairedis/syncd/vs part, but from recordings what i saw, that ports are dynamically splitted in orchagent, and i remember that since i added unisttests for this and support to VS for that scenario. I also read https://github.com/Azure/SONiC/blob/master/doc/voq/voq_hld.md, but its mostly high level description, where on SAI level is very little description, and bottom page just says dynamic part is future work

I'm pointing this dynamic part here since i have been in similar scenarios previously, and it was on my head to find issues why X is not working, because Y was not included in dynamic breakout scenario, and this VOQ switch system seems to be to be even more complicated in this part when it will come to create/remove ports.

I think also question will be here for this feature be supported on sonic or you just adding support to sairedis/syncd/vs to use in internal (nokia?) projects? I already know some features that other vendors added, but they will not be used publicly in sonic

vganesan-nokia: Yes. Dynamic system port handling is future work. Not sure when it will be taken. I agree with you that when the dynamic system ports add/del is taken, it is going to be bit complex. That is one of the reasons why it was proposed to start with simple static pre-configuration and do dynamic system ports add/del later. This VoQ System Ports is part of SONiC on Chassis System development.

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Fixed

kcudnik (approved):

PR #702, opened 2020-11-10
[vs] Initialization of VOQ switch objects

vganesan-nokia:
VOQ switch objects are initialized during swich create. This VOQ objects
are initialized after regular switch objects are initialized. This PR depends on https://github.com/Azure/sonic-sairedis/pull/701
for switch state initialization

### DIFF
--- a/vslib/inc/SwitchBCM81724.h
+++ b/vslib/inc/SwitchBCM81724.h
@@ -97,8 +97,14 @@ namespace saivs
 
         protected:
 
-            virtual sai_status_t refresh_read_only( _In_ const sai_attr_metadata_t *meta, _In_ sai_object_id_t object_id) override;
+            virtual sai_status_t refresh_read_only(
+                    _In_ const sai_attr_metadata_t *meta,
+                    _In_ sai_object_id_t object_id) override;
+
             virtual sai_status_t set_switch_default_attributes();
-            virtual sai_status_t initialize_default_objects( _In_ uint32_t attr_count, _In_ const sai_attribute_t *attr_list) override;
+
+            virtual sai_status_t initialize_default_objects(
+                    _In_ uint32_t attr_count,
+                    _In_ const sai_attribute_t *attr_list) override;
     };
 }
--- a/vslib/inc/SwitchBCM81724.h
+++ b/vslib/inc/SwitchBCM81724.h
@@ -99,6 +99,6 @@ namespace saivs
 
             virtual sai_status_t refresh_read_only( _In_ const sai_attr_metadata_t *meta, _In_ sai_object_id_t object_id) override;
             virtual sai_status_t set_switch_default_attributes();
-            virtual sai_status_t initialize_default_objects() override;
+            virtual sai_status_t initialize_default_objects( _In_ uint32_t attr_count, _In_ const sai_attribute_t *attr_list) override;
     };
 }
--- a/vslib/src/VirtualSwitchSaiInterface.cpp
+++ b/vslib/src/VirtualSwitchSaiInterface.cpp
@@ -540,7 +540,9 @@ std::shared_ptr<SwitchStateBase> VirtualSwitchSaiInterface::init_switch(
         _In_ sai_object_id_t switch_id,
         _In_ std::shared_ptr<SwitchConfig> config,
         _In_ std::shared_ptr<WarmBootState> warmBootState,
-        _In_ std::weak_ptr<saimeta::Meta> meta)
+        _In_ std::weak_ptr<saimeta::Meta> meta,
+        _In_ uint32_t attr_count,
+        _In_ const sai_attribute_t *attr_list)
 {
     SWSS_LOG_ENTER();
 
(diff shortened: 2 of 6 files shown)


vslib/src/VirtualSwitchSaiInterface.cpp
```diff
@@ -659,6 +659,12 @@ sai_status_t VirtualSwitchSaiInterface::create(
             return SAI_STATUS_FAILURE;
         }
 
+        // Initialize switch for VOQ attributes
+        if ((ss->initialize_voq_switch_objects(attr_count, attr_list)) != SAI_STATUS_SUCCESS)
```
kcudnik, line 663: this is in the wrong place, it should be inside SwitchStateBase.cpp class in initialize objects

kcudnik (changes requested):

vslib/src/VirtualSwitchSaiInterface.cpp
vganesan-nokia: As discussed in PR #701, this is unchanged.

kcudnik: agree,  but there are places init_default_objects inside _init_switch and warm_boot_initialize_objects,
 current implementation you want to execute only for default objects, and this place here is actually executing it for both, which will cause to duplicate system port existence in virtual switch for warm boot, 
so i would propose to add attr_count/list to init_switch and move this initialization in line 604

kcudnik: also when you will want to support warm boot scenario for this, then you will need to add warm boot case in init_switch also

vganesan-nokia: For comment 1: o.k. I fix proposed. 
For comment 2: o.k. when we support warm boot (only when system ports become dynamic config. Not sure when), this place will be used.

vganesan-nokia: Fixed. Moved initialize_voq_switch_objects() to init_switch() as proposed.

vslib/src/VirtualSwitchSaiInterface.cpp
```diff
@@ -602,6 +604,15 @@ std::shared_ptr<SwitchStateBase> VirtualSwitchSaiInterface::init_switch(
             SWSS_LOG_THROW("unable to init switch %s", sai_serialize_status(status).c_str());
         }
 
+        // Initialize switch for VOQ attributes
+
+        status = ss->initialize_voq_switch_objects(attr_count, attr_list);
```
kcudnik, line 609: ok, this looks better, but i would like you to make 1 more change, so i think it would be actually better to propagate those count/list to parameters initialize_default_objects() function, and call voq initialize inside initialize_devault_objects, this way we will have initialization part in 1 single function, what do you think?

vganesan-nokia: No issue in doing what you are proposing. However, here the voq objects are not defaults. They are configured attributes. Not sure if it will be clearer to have non default object initialization in function whose name indicates default object initialization

vganesan-nokia: I also see a comment "TODO" to move the calling "initialize_default_objects()" to constructor. If we modify to propagate attr_count/attr_list to this function, we may not be able move this to constructor. Is it still o.k, if I do the modification as proposed?

kcudnik: that move to constructor cant be moved to constructor since that is virtual function, it could be moved, but it would need to be called explicitly in each constructor, and not in SwitchStateBase

vganesan-nokia: o.k. I'll modify as proposed, then.

kcudnik: thanks, this is actually interesting that some of the internal objects depends on the input attributes to create_switch function, and not actual hardware profile, this is wired to me

kcudnik:
retest vs please

kcudnik:
lgtm is stuck, so you will need to do force push or add another commit to trigger lgtm rerun

vslib/src/VirtualSwitchSaiInterface.cpp
vganesan-nokia: Modified as proposed.

vslib/inc/SwitchBCM81724.h
```diff
@@ -99,6 +99,6 @@ namespace saivs
 
             virtual sai_status_t refresh_read_only( _In_ const sai_attr_metadata_t *meta, _In_ sai_object_id_t object_id) override;
             virtual sai_status_t set_switch_default_attributes();
-            virtual sai_status_t initialize_default_objects() override;
+            virtual sai_status_t initialize_default_objects( _In_ uint32_t attr_count, _In_ const sai_attribute_t *attr_list) override;
```
kcudnik, line 102: one param per line, you can also fix the one in line 100

kcudnik (changes requested):

vslib/inc/SwitchBCM81724.h
vganesan-nokia: o.k. I'll fix both.

vganesan-nokia: Fixed

kcudnik (approved):

PR #709, opened 2020-11-16
[meta] Use custom hash in SaiObjectCollection

kcudnik:
Currently unordered_map with std::string key is used, but since sai_serialize_object_meta_key is expensive, it make sense to keep original meta_key as key and write custom hash function. For 100k routes, metadata speeds up 3x when using custom hash.

### DIFF
--- a/meta/Meta.cpp
+++ b/meta/Meta.cpp
@@ -2897,11 +2897,8 @@ void Meta::clean_after_switch_remove(
         }
     }
 
-    for (auto& key: m_saiObjectCollection.getAllKeys())
+    for (auto& mk: m_saiObjectCollection.getAllKeys())
     {
-        sai_object_meta_key_t mk;
-        sai_deserialize_object_meta_key(key, mk);
-
         // we guarantee that switch_id is first in the key structure so we can
         // use that as object_id as well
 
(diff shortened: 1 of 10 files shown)


kcudnik:
retest vs please

meta/Meta.cpp
```diff
@@ -5424,10 +5411,9 @@ sai_status_t Meta::meta_generic_validation_get(
             // (this will not respect create_only with default)
             if (get_object_previous_attr(meta_key, md) == NULL)
             {
-                std::string key = sai_serialize_object_meta_key(meta_key);
-
                 // XXX produces too much noise
```
qiluo-msft, line 5414: >// XXX produces too much noise [](start = 16, length = 30)

This is an empty block. Just reverse the if-statement?

qiluo-msft:
>     if (!SAI_HAS_FLAG_READ_ONLY(md.flags) && md.allowedobjecttypeslength) // md.isoidattribute)

Another empty block. Just remove?

---
Refers to: meta/Meta.cpp:6331 in eb01334. [](commit_id = eb01334936323ed418f92e728c4deb211a8bc67e, deletion_comment = False)

qiluo-msft:
>     if (!SAI_HAS_FLAG_READ_ONLY(md.flags) && md.allowedobjecttypeslength) // md.isoidattribute)

Another empty block. Just remove?  

---
Refers to: meta/Meta.cpp:6763 in eb01334. [](commit_id = eb01334936323ed418f92e728c4deb211a8bc67e, deletion_comment = False)

meta/Meta.cpp
kcudnik: i want to keep this for future possible debug investigation purpose, compiler will optimize this anyway to single block

kcudnik:
for all comments: i want to keep this for future possible debug investigation purpose, compiler will optimize this anyway to single block, currently those are not used since we don't have such complicated SAI internals query so far, but at the beginning i already put the effort to figure out where there will be good hook points for debug when something will go wrong, and it will be much harder to go back and find this in some previous commits when times comes

kcudnik:
retest vs please

qiluo-msft (approved):

PR #711, opened 2020-11-17
Add support for default zmq synchronous mode flag

kcudnik:


lib/src/RedisRemoteSaiInterface.cpp
```diff
@@ -373,9 +374,46 @@ sai_status_t RedisRemoteSaiInterface::setRedisExtensionAttribute(
...
+                // so this command will replace it with zmq channel
+
+                m_communicationChannel = std::make_shared<ZeroMQChannel>(
+                        m_contextConfig->m_zmqEndpoint,
+                        m_contextConfig->m_zmqNtfEndpoint,
+                        std::bind(&RedisRemoteSaiInterface::handleNotification, this, _1, _2, _3));
+
+                SWSS_LOG_NOTICE("zmq enabled, forcing sync mode");
+
+                m_syncMode = true;
```
qiluo-msft, line 395: >m_syncMode [](start = 16, length = 10)

Does it make more sense to return SAI_STATUS_NOT_SUPPORTED if m_zmqSyncMode==True and m_syncMode==False?
This way the caller will know there is a logic error.

syncd/CommandLineOptionsParser.cpp
```diff
@@ -164,6 +169,8 @@ void CommandLineOptionsParser::printUsage()
     std::cout << "        Enable consisteny check DB vs ASIC after comparison logic" << std::endl;
     std::cout << "    -s --syncMode" << std::endl;
     std::cout << "        Enable synchronous mode" << std::endl;
+    std::cout << "    -z --zmqSyncMode" << std::endl;
```
qiluo-msft, line 172: >zmqSyncMode [](start = 27, length = 11)

How about:
-s --syncMode
-z --zmqMode
Throw errors if there are unsupported combination.

syncd/Syncd.cpp
```diff
@@ -63,6 +63,13 @@ Syncd::Syncd(
         SWSS_LOG_THROW("no context config defined at global context %u", m_commandLineOptions->m_globalContext);
     }
 
+    if (m_commandLineOptions->m_enableZmqSyncMode)
+    {
+        SWSS_LOG_NOTICE("zmq sync mode enabled via cmd line");
+
+        m_contextConfig->m_zmqEnable = true;
```
qiluo-msft, line 70: >m_zmqEnable [](start = 25, length = 11)

Do you mean `m_zmqSyncMode` ? It is confusing if `sync` is not in the name.

qiluo-msft (commented):
As comments

lib/src/RedisRemoteSaiInterface.cpp
kcudnik: actually zmq syncd mode forces sync mode internally, so i don't think there is error here

syncd/CommandLineOptionsParser.cpp
kcudnik: you can specify -s -z, zmq forces -s internelly

syncd/Syncd.cpp
kcudnik: there is no m_zmqSyncMode in that class, m_zmqEnable was introduced previously in ContextConfig class without intention that it will be added as separate flag "syncd mode", since zmq is always synchronous, agree that those names could be confusing here, any proposal ?

kcudnik: as your other comment should i drop "sync" from names ? and what about SAI_REDIS_SWITCH_ATTR_ZMQ_SYNC_MODE ?? actually "sync mode" should be referred as "redisdb sync mode" or "redis sync mode"

syncd/CommandLineOptionsParser.cpp
shi-su: `-s -z` and `-z` seem to behave identically. I am thinking if it is more clear to do something like a) `-s -z`: log a notice that it is doing zmq sync mode; b) `-z`: log a warning that there is no `-s` given but we are forcing sync mode? Not completely sure if having an overlapped configuration between zmq sync mode and redis sync mode is good. Maybe we can make `-s` and `-z` mutually exclusive?

kcudnik: so there is common part in syncd, when synchronous mode is enabled, common parts are executed, and this does not depend on whether is redis syncd mode or zmq sync mode, just need to be enabled in case of any synchronization
originally i was planning do to sync mode as a enum, so in this case it could be like none/redis/zmq (maybe in future boost), but user would need to specify another flag with that  enum string to be parsed internally to actual enum value. this way we would have only 1 possible way to enable single mode. but this would break backward compatibility with flag "-s" that we have right now

since at the syncd design stage there was no plan to add multiple modes since redis was acting as pipeline, there was no issue there. so currently i decided that when zmq sync mode will be set (-z), it will just override "-s" parameter everywhere (in OA/saiplayer) (and syncd synmode is enabled by -m param) so there will be no issue with possibly specifying 2 of them

but if you want, i can convert that param to actual enum value for potential future expansion. Also note that zmq mode requires some extra parameters, which at this point are just set to some hard coded values, and actual values could be changed via context_config.json

kcudnik: so im actually thinking about 2 solusions here
* we have "-s" flag, that enables/disables synchronous mode (since there can be only 2 modes async/sync) and then we have one enum communication_mode (also 1 redis atribute) which specifies what mode is is (redis_async, redis_sync, zmq_sync) and this mode could be specified by -z param like "-s -z zmq_sync"
* we have "-s" flag same as above, and 2 enums async_mode_type (redis), and sync_mode_type (redis,zmq, possibly boost later) then default values could be specified for both via 2 extra params from command lines, for example "-s -z zmq" or "-Z redis", one of the params dont need to be implemented now since we have only 1 async mode for now 

both solutions have pros and cons, for first one, we only have 1 enum, but we need to deal with incompatible -s param (if we want to still keep "-s" param) and do error when "-s -Z redis" is specified, for second we keep -s as selector for sync/async mode, and we can specify "-s Z redis -z zmq" and it will be fine,  downside is that we need 2 enums to handle

any thoughts on that ?

shi-su: I got a little confused between `-z` and `-Z`. Do they mean communication mode for sync_mode and async_mode, respectively? If so, why is `-s -Z redis` a problem for the first option (`-Z` should not be valid anyway)? Shouldn't `-z zmq_sync` without `-s` be an error?

shi-su: Personally speaking, I would like one communication mode better since the redundant `-Z redis` in `-s -Z redis -z zmq` looks a little confusing to me. But either way looks OK to me.

kcudnik: addressed, added communication mode enum as single param

kcudnik:
retest vs please

kcudnik:
retest vs please

kcudnik:
retest vs please

kcudnik:
retest vs please

shi-su (commented):
Do you know if there are any other places that use the sync mode option? I know it is used in `syncd_init_common.sh` start syncd in sync mode. If you know any other places that use it, probably we need to note it somewhere so that it would be easier to update all of them to communication mode.

syncd/CommandLineOptionsParser.cpp
```diff
@@ -164,6 +172,8 @@ void CommandLineOptionsParser::printUsage()
     std::cout << "        Enable consisteny check DB vs ASIC after comparison logic" << std::endl;
     std::cout << "    -s --syncMode" << std::endl;
     std::cout << "        Enable synchronous mode" << std::endl;
```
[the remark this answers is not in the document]
kcudnik: i already added this in log warn message, do you want me to update it here as well?

saiplayer/CommandLineOptionsParser.cpp
```diff
@@ -124,11 +137,14 @@ void CommandLineOptionsParser::printUsage()
     std::cout << "        Sleep after success reply, to notice any switch notifications" << std::endl << std::endl;
     std::cout << "    -m --syncMode:" << std::endl;
     std::cout << "        Enable synchronous mode" << std::endl << std::endl;
```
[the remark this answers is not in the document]
kcudnik: i already added this in log warn message, do you want me to update it here as well?

syncd/CommandLineOptionsParser.cpp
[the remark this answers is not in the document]
shi-su: I thought the warn message could only be seen in the log after running the command. When someone uses this command, he/she might read this help message. So I thought adding a message here might be useful to prevent people from using this option in the first place?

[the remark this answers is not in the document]
kcudnik: updated

saiplayer/CommandLineOptionsParser.cpp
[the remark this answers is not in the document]
kcudnik: updated

kcudnik:
retest vs please

kcudnik:
retest vs please

kcudnik:
retest vs please

kcudnik:
retest vs please

kcudnik:
retest vs please

kcudnik:
> Do you know if there are any other places that use the sync mode option? I know it is used in `syncd_init_common.sh` start syncd in sync mode. If you know any other places that use it, probably we need to note it somewhere so that it would be easier to update all of them to communication mode.

i grepped via sonic build image scripts, but i couldn't find anything relevant, since like in "syncd_init_common.sh" syncd process name and cmd args are splitted to different lines in script, or they could be combined, there is also similar issue with swss/OA where also takes -s param and i would like to get rid of it, i already prepared PR (https://github.com/Azure/sonic-swss/pull/1505) but this PR needs to be checked in first

qiluo-msft (approved):
Please also check with other reviewers.

shi-su (approved):

PR #719, opened 2020-11-23
[vslib]Add MACsec forward and filters to HostInterfaceInfo

Pterosaur:
Add MACsec filter before VLAN Tag and FDB event processing.

Refer: MACsec in VLAN-aware Bridges
image

### DIFF
--- a/vslib/inc/TrafficForwarder.h
+++ b/vslib/inc/TrafficForwarder.h
@@ -5,6 +5,9 @@
 #include <stddef.h>
 #include <sys/socket.h>
 
+static constexpr size_t ETH_FRAME_BUFFER_SIZE = 0x4000;
+static constexpr size_t CONTROL_MESSAGE_BUFFER_SIZE = 0x1000;
+
 namespace saivs
 {
     class TrafficForwarder
@@ -15,15 +18,15 @@ namespace saivs
     protected:
         TrafficForwarder() = default;
 
-        void add_vlan_tag(
+        static void addVlanTag(
             _Inout_ unsigned char *buffer,
             _Inout_ size_t &length,
-            _Inout_ struct msghdr &msg) const;
+            _Inout_ struct msghdr &msg);
 
-        bool send_to(
+        bool sendTo(
             _In_ int fd,
             _In_ const unsigned char *buffer,
-            _In_ size_t &length) const;
+            _In_ size_t length) const;
 
     };
 }
--- a/vslib/inc/TrafficForwarder.h
+++ b/vslib/inc/TrafficForwarder.h
@@ -5,11 +5,11 @@
 #include <stddef.h>
 #include <sys/socket.h>
 
-static constexpr size_t ETH_FRAME_BUFFER_SIZE = 0x4000;
-static constexpr size_t CONTROL_MESSAGE_BUFFER_SIZE = 0x1000;
-
 namespace saivs
 {
+    static constexpr size_t ETH_FRAME_BUFFER_SIZE = 0x4000;
+    static constexpr size_t CONTROL_MESSAGE_BUFFER_SIZE = 0x1000;
+
     class TrafficForwarder
     {
     public:
--- a/vslib/inc/TrafficForwarder.h
+++ b/vslib/inc/TrafficForwarder.h
@@ -23,7 +23,7 @@ namespace saivs
             _Inout_ size_t &length,
             _Inout_ struct msghdr &msg);
 
-        bool sendTo(
+        virtual bool sendTo(
             _In_ int fd,
             _In_ const unsigned char *buffer,
             _In_ size_t length) const;
--- a/vslib/inc/TrafficForwarder.h
+++ b/vslib/inc/TrafficForwarder.h
@@ -1,0 +1,29 @@
+#pragma once
+
+#include "swss/sal.h"
+
+#include <stddef.h>
+#include <sys/socket.h>
+
+namespace saivs
+{
+    class TrafficForwarder
+    {
+    public:
+        virtual ~TrafficForwarder() = default;
+
+    protected:
+        TrafficForwarder() = default;
+
+        void add_vlan_tag(
+            _Inout_ unsigned char *buffer,
+            _Inout_ size_t &length,
+            _Inout_ struct msghdr &msg) const;
+
+        bool send_to(
+            _In_ int fd,
+            _In_ const unsigned char *buffer,
+            _In_ size_t &length) const;
+
+    };
+}
--- a/vslib/src/HostInterfaceInfo.cpp
+++ b/vslib/src/HostInterfaceInfo.cpp
@@ -90,6 +90,40 @@ void HostInterfaceInfo::async_process_packet_for_fdb_event(
     m_eventQueue->enqueue(std::make_shared<Event>(EventType::EVENT_TYPE_PACKET, payload));
 }
 
+bool HostInterfaceInfo::installEth2TapFilter(
+        _In_ int priority,
+        _In_ std::shared_ptr<TrafficFilter> filter)
+{
+    SWSS_LOG_ENTER();
+
+    return m_e2tFilters.installFilter(priority, filter);
+}
+
+bool HostInterfaceInfo::uninstallEth2TapFilter(
+        _In_ std::shared_ptr<TrafficFilter> filter)
+{
+    SWSS_LOG_ENTER();
+
+    return m_e2tFilters.uninstallFilter(filter);
+}
+
+bool HostInterfaceInfo::installTap2EthFilter(
+        _In_ int priority,
+        _In_ std::shared_ptr<TrafficFilter> filter)
+{
+    SWSS_LOG_ENTER();
+
+    return m_t2eFilters.installFilter(priority, filter);
+}
+
+bool HostInterfaceInfo::uninstallTap2EthFilter(
+        _In_ std::shared_ptr<TrafficFilter> filter)
+{
+    SWSS_LOG_ENTER();
+
+    return m_t2eFilters.uninstallFilter(filter);
+}
+
 #define ETH_FRAME_BUFFER_SIZE (0x4000)
 #define CONTROL_MESSAGE_BUFFER_SIZE (0x1000)
 #define IEEE_8021Q_ETHER_TYPE (0x8100)
@@ -159,6 +193,20 @@ void HostInterfaceInfo::veth2tap_fun()
         {
             SWSS_LOG_ERROR("invalid ethernet frame length: %zu", msg.msg_controllen);
             continue;
+        }
+
+        size_t length = static_cast<size_t>(size);
+        auto ret = m_e2tFilters.execute(buffer, length);
+        size = static_cast<ssize_t>(length);
+
+        if (ret == TrafficFilter::TERMINATE)
+        {
+            continue;
+        }
+        else if (ret == TrafficFilter::ERROR)
+        {
+            // Error log should be recorded in filter
+            return;
         }
 
         struct cmsghdr *cmsg;
@@ -266,6 +314,20 @@ void HostInterfaceInfo::tap2veth_fun()
             }
 
             continue;
+        }
+
+        size_t length = static_cast<size_t>(size);
+        auto ret = m_t2eFilters.execute(buffer, length);
+        size = static_cast<ssize_t>(length);
+
+        if (ret == TrafficFilter::TERMINATE)
+        {
+            continue;
+        }
+        else if (ret == TrafficFilter::ERROR)
+        {
+            // Error log should be recorded in filter
+            return;
         }
 
         if (write(m_packet_socket, buffer, (int)size) < 0)
--- a/vslib/src/HostInterfaceInfo.cpp
+++ b/vslib/src/HostInterfaceInfo.cpp
@@ -124,12 +124,6 @@ bool HostInterfaceInfo::uninstallTap2EthFilter(
     return m_t2eFilters.uninstallFilter(filter);
 }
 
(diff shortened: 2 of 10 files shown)


vslib/src/MACsecForwarder.cpp
```diff
@@ -112,6 +118,25 @@ void MACsecForwarder::forward()
 
     while (m_runThread)
```
kcudnik, line 119: why you need separate thread in macsec ? there is already thread thats deals with packet processing, why not tap into that ?

vslib/src/MACsecForwarder.cpp
```diff
@@ -151,6 +176,55 @@ void MACsecForwarder::forward()
...
+        }
+
+        if (m_info == nullptr)
+        {
+            SWSS_LOG_ERROR("The HostInterfaceInfo on the MACsec port %s is empty", m_macsecInterfaceName.c_str());
+
+            break;
+        }
+
+        m_info->async_process_packet_for_fdb_event(buffer, size);
```
kcudnik, line 227: This seems really bad that you moved this here , also line from 188 to 218 are i think direct copy from HostInterface.cpp file, basically you are double the work, i think we already discussed this, that packet arrives at:
 HostInterfaceInfo::veth2tap_fun() then it's received and processed, then you should execute all the filters, and decide whether fdb event should be processed or not.

This entire logic here that you added seems redundant to me, and MACsecForwarder should have no idea aboug HostInterfaceInfo class, it should return data whether to process it later or not.

kcudnik (changes requested):

vslib/src/MACsecForwarder.cpp
kcudnik: we already have 2 threads per port that transfer the data, and you are having some another thread for the path that is already processing data, this seems a waist of resources

Pterosaur: I remember you asked this question before, sorry that I'm not explain it clear.
I think we need three threads in current design. The previous two are for forwarding data between tap and veth, and the new one is for forwarding the plaintext data from macsec to tap.

Meanwhile, all encrypted data on the thread 2 will be dropped by the macsec ingress filter. Because linux kernel will help us to forward the encrypted data.
All plaintext data, except EAPOL traffic, from tap interface will be captured by the MACsec egress filter that forwards to the macsec device for encryption.
Who will help us to forward the plaintext data from macsec device? I believe it should be thread 3.

image

vslib/src/MACsecForwarder.cpp
Pterosaur: My understanding is that the fdb processing and vlan tag adding needs the plaintext data instead of encrypted data. So they cannot be in that thread, veth2tap.

vslib/src/MACsecForwarder.cpp
kcudnik: so mu understanding here is now like this: we have regular traffic, and encrypted traffic coming to the same interface then based on something (dont know what it is) you can distinguish which one is which, and for encrypted traffic you are sending this traffic to separate created socket that will decrypt this traffic inside kernel and then you can read that socket (or other one?) to get planetext traffic, process it, block or forward, maybe generate fdb notification, and then forward back encrypted traffic (re-encrypted if some filters are applied that modify pakcket)  to the existing tap device?

if this is the case, both encrypted and plain text traffic come to the single tap interface, and right now this traffic is processing 1 by 1 packet in order that they arrived, with your async solution there is possibility that arrived packets will be reordered, and this is probably whats not happening in real hardware, but thats my guess. Anyway in this case i think you should block for processing in the pipeline.

if my understanding here is wrong, please setup a call meeting with me, we can talk and share on teams. I'm in CET timezone and available from 11am to 10pm everyday

kcudnik (changes requested):

vslib/src/HostInterfaceInfo.cpp
```diff
@@ -268,6 +316,20 @@ void HostInterfaceInfo::tap2veth_fun()
             continue;
         }
 
+        size_t length = static_cast<size_t>(size);
+        auto ret = m_t2eFilters.execute(buffer, length);
```
Pterosaur, line 320: add comment: Plaintext data

vslib/src/HostInterfaceInfo.cpp
```diff
@@ -161,6 +195,20 @@ void HostInterfaceInfo::veth2tap_fun()
             continue;
         }
 
+        size_t length = static_cast<size_t>(size);
```
Pterosaur, line 198: add comment: Encrypted data

vslib/src/MACsecForwarder.cpp
```diff
@@ -151,6 +176,55 @@ void MACsecForwarder::forward()
...
+        else if (size < (ssize_t)sizeof(ethhdr))
+        {
+            SWSS_LOG_ERROR("invalid ethernet frame length: %zu", msg.msg_controllen);
+
+            continue;
+        }
+
+        struct cmsghdr *cmsg;
+
+        for (cmsg = CMSG_FIRSTHDR(&msg); cmsg != NULL; cmsg = CMSG_NXTHDR(&msg, cmsg))
```
Pterosaur, line 188: Extract these code to a function.

vslib/src/MACsecForwarder.cpp
```diff
@@ -151,6 +176,55 @@ void MACsecForwarder::forward()
...
+        if (m_info == nullptr)
+        {
+            SWSS_LOG_ERROR("The HostInterfaceInfo on the MACsec port %s is empty", m_macsecInterfaceName.c_str());
+
+            break;
+        }
+
+        m_info->async_process_packet_for_fdb_event(buffer, size);
 
         if (write(m_tapfd, buffer, static_cast<int>(size)) < 0)
```
kcudnik, line 229: this also seems like a  HostInterfaceInfo::veth2tap_fun end of the function, so this is also could be moved to a common function?

vslib/src/MACsecManager.cpp
```diff
@@ -579,7 +579,7 @@ bool MACsecManager::add_macsec_forwarder(
 
     auto &manager = itr->second;
 
-    manager.m_forwarder = std::make_shared<MACsecForwarder>(macsecInterface, manager.m_info->m_tapfd);
+    manager.m_forwarder = std::make_shared<MACsecForwarder>(macsecInterface, manager.m_info->m_tapfd, manager.m_info);
```
kcudnik, line 582: manager.m_info->m_tapfd in not necessary, remove; you are already passing manager.m_info inside, so you can internally use that m_tapfd, and also m_tapfd should be private, did you change visibility of that item ?

kcudnik (changes requested):

vslib/src/MACsecForwarder.cpp
kcudnik: basically this MACsecForwarder::forward() is exactly the same as HostInterfaceInfo::veth2tap_fun, but not having this filter option, and just uses different FD, but other processes are the same, i think this could be somehow unified, but maybe not this PR

vslib/src/MACsecForwarder.cpp
```diff
@@ -151,6 +176,55 @@ void MACsecForwarder::forward()
...
+                memcpy(pvlan, &tpid, sizeof(uint16_t));
+                memcpy(pvlan + sizeof(uint16_t), &tci, sizeof(uint16_t));
+
+                size += VLAN_TAG_SIZE;
+
+                break;
+            }
+        }
+
+        if (m_info == nullptr)
```
kcudnik, line 220: i think this should be checked on the constructor and throw if empty

vslib/src/MACsecForwarder.cpp
Pterosaur: Agree. I add a new super class `TrafficForwarder`. Maybe we can leverage it to refactor this part in the another PR.

vslib/src/MACsecForwarder.cpp
Pterosaur: Done

vslib/src/MACsecManager.cpp
Pterosaur: Done

vslib/src/MACsecForwarder.cpp
Pterosaur: Done

vslib/src/TrafficForwarder.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+            break;
+        }
+    }
+}
+
+bool TrafficForwarder::send_to(
+    _In_ int fd,
+    _In_ const unsigned char *buffer,
+    _In_ size_t &length) const
```
kcudnik, line 73: length is not modified here no need for &

vslib/src/TrafficForwarder.cpp
```diff
@@ -0,0 +1,100 @@
...
+    {
+        /*
+        * We filter out EIO because of this patch:
+        * https://github.com/torvalds/linux/commit/1bd4978a88ac2589f3105f599b1d404a312fb7f6
+        */
+
+        if (errno != ENETDOWN && errno != EIO)
+        {
+            SWSS_LOG_ERROR("failed to write to device fd %d, errno(%d): %s",
+                    fd, errno, strerror(errno));
```
kcudnik, line 87: 1 param per line or join everything to 1 line

vslib/inc/TrafficForwarder.h
```diff
@@ -0,0 +1,29 @@
...
+{
+    class TrafficForwarder
+    {
+    public:
+        virtual ~TrafficForwarder() = default;
+
+    protected:
+        TrafficForwarder() = default;
+
+        void add_vlan_tag(
```
kcudnik, line 18: seems to me likt those methods can be static, also use camelcase method naming

vslib/src/TrafficForwarder.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include <net/if_arp.h>
+#include <unistd.h>
+#include <net/ethernet.h>
+#include <arpa/inet.h>
+#include <linux/if_packet.h>
+#include <linux/if_ether.h>
+
+using namespace saivs;
+
+#define ETH_FRAME_BUFFER_SIZE (0x4000)
```
kcudnik, line 22: is this redefined in many places ?

vslib/src/TrafficForwarder.cpp
```diff
@@ -0,0 +1,100 @@
...
+#define CONTROL_MESSAGE_BUFFER_SIZE (0x1000)
+#define IEEE_8021Q_ETHER_TYPE (0x8100)
+#define MAC_ADDRESS_SIZE (6)
+#define VLAN_TAG_SIZE (4)
+
+void TrafficForwarder::add_vlan_tag(
+    _Inout_ unsigned char *buffer,
+    _Inout_ size_t &length,
```
kcudnik, line 30: so there is a potential problem here, which also exists in original code, so buffer is allocated with ETH_FRAME_BUFFER_SIZE, but length here is specifying valid data in the buffer, which probably all the time is less then buffer allocated size, but since buffer is extended here, and vlan tag is inserted, then in case when length will be exact == ETH_FRAME_BUFFER_SIZE, then write will occur outside buffer bounds, which can cause issues. maybe we could add check for that ? and throw before buffer is modified ?

Pterosaur: Done, please check it.

vslib/src/TrafficForwarder.cpp
Pterosaur: Done, I move it to the header. Please check it.

vslib/inc/TrafficForwarder.h
Pterosaur: Done.

vslib/src/TrafficForwarder.cpp
Pterosaur: Done

vslib/src/TrafficForwarder.cpp
Pterosaur: Done

vslib/inc/TrafficForwarder.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
+
+#include "swss/sal.h"
+
+#include <stddef.h>
+#include <sys/socket.h>
+
+static constexpr size_t ETH_FRAME_BUFFER_SIZE = 0x4000;
+static constexpr size_t CONTROL_MESSAGE_BUFFER_SIZE = 0x1000;
```
kcudnik, line 9: please move both inside saivs namespace

vslib/inc/TrafficForwarder.h
```diff
@@ -0,0 +1,32 @@
...
+
+    protected:
+        TrafficForwarder() = default;
+
+        static void addVlanTag(
+            _Inout_ unsigned char *buffer,
+            _Inout_ size_t &length,
+            _Inout_ struct msghdr &msg);
+
+        bool sendTo(
```
kcudnik, line 26: this method also can be static since there are no members, sorry i didnt mentioned to make all methods static

Pterosaur: Yes, I thought this design you mentioned.
But I hope `TrafficForwarder` could be an abstract interface which means you can did some customized `sendTo` flow by overriding it. How about to make it becomes a virtual function?

kcudnik: ok, make this virtual

Pterosaur: Done

kcudnik:
retest vs please

kcudnik (approved):

PR #721, opened 2020-11-29
Code clean

kcudnik:


syncd/AsicView.cpp
```diff
@@ -829,7 +829,7 @@ void AsicView::asicRemoveObject(
                 break;
 
             case SAI_OBJECT_TYPE_NAT_ENTRY:
-                m_soNatEntries[currentObj->m_str_object_id] = currentObj;
+                m_soNatEntries.erase(currentObj->m_str_object_id);
```
jleveque, line 832: Was this a bug?

kcudnik: yes :)

jleveque: Good find!

jleveque (approved):
LGTM.

PR #722, opened 2020-12-01
[vslib]: Add MACsec state to state base

Pterosaur:

### DIFF
--- a/vslib/inc/SwitchStateBase.h
+++ b/vslib/inc/SwitchStateBase.h
@@ -253,12 +253,6 @@ namespace saivs
                     _In_ const std::string &serializedObjectId,
                     _In_ const sai_attribute_t* attr);
 
-            virtual sai_status_t get_internal(
-                    _In_ sai_object_type_t objectType,
-                    _In_ const std::string &serializedObjectId,
-                    _In_ uint32_t attr_count,
-                    _Out_ sai_attribute_t* attr_list);
-
         private:
 
             sai_object_type_t objectTypeQuery(
--- a/vslib/inc/SwitchStateBase.h
+++ b/vslib/inc/SwitchStateBase.h
@@ -468,6 +468,12 @@ namespace saivs
                     _In_ sai_object_id_t entry_id,
                     _In_ const sai_attribute_t* attr);
 
+            sai_status_t createMACsecPort(
+                    _In_ sai_object_id_t macsec_sa_id,
+                    _In_ sai_object_id_t switch_id,
+                    _In_ uint32_t attr_count,
+                    _In_ const sai_attribute_t *attr_list);
+
             sai_status_t createMACsecSA(
                     _In_ sai_object_id_t macsec_sa_id,
                     _In_ sai_object_id_t switch_id,
@@ -497,9 +503,8 @@ namespace saivs
                     _In_ sai_object_id_t macsec_flow_id,
                     _Out_ sai_object_id_t &line_port_id);
 
-            sai_status_t findHostInterfaceInfoByPort(
-                    _In_ sai_object_id_t &line_port_id,
-                    _Out_ std::shared_ptr<HostInterfaceInfo> &info);
+            std::shared_ptr<HostInterfaceInfo> findHostInterfaceInfoByPort(
+                    _In_ sai_object_id_t line_port_id);
 
             sai_status_t loadMACsecAttrFromMACsecPort(
                     _In_ sai_object_id_t object_id,
--- a/vslib/inc/SwitchStateBase.h
+++ b/vslib/inc/SwitchStateBase.h
@@ -7,6 +7,7 @@
 #include "SwitchConfig.h"
 #include "RealObjectIdManager.h"
 #include "EventPayloadNetLinkMsg.h"
+#include "MACsecManager.h"
 
 #include <set>
 #include <unordered_set>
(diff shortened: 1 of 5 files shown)


vslib/inc/SwitchStateBase.h
```diff
@@ -249,6 +250,12 @@ namespace saivs
                     _In_ const std::string &serializedObjectId,
                     _In_ const sai_attribute_t* attr);
 
+            virtual sai_status_t get_internal(
```
kcudnik, line 253: please remove, no need for get internal function

vslib/src/SwitchStateBase.cpp
```diff
@@ -629,6 +685,33 @@ sai_status_t SwitchStateBase::get(
...
+    if (objectType == SAI_OBJECT_TYPE_MACSEC)
+    {
+        return getMACsecAttr(
+            serializedObjectId,
+            attr_count,
+            attr_list);
+    }
+
+    return get_internal(objectType, serializedObjectId, attr_count, attr_list);
+}
```
kcudnik, line 713: please remove get_inernal function, there are no special action needed for GET, those attributes SAI_MACSEC_ATTR_SCI_IN_INGRESS_MACSEC_ACL you are referring, are READ_ONLY and should be refressed in "status = refresh_read_only(meta, oid);" function internals

kcudnik: SAI_MACSEC_SA_ATTR_XPN should be set during CREATE, since it's create and set attribute, so get will work fine for this attribute, same for SAI_MACSEC_SA_ATTR_MINIMUM_XPN, those attributes have default values, but they are not set, when object is created, probably you should set them on custom create function

kcudnik: those 2 functions getMACsecSAAttr getMACsecAttr can be removed

kcudnik (changes requested):

kcudnik:
can you check  LGTM warnings ?

vslib/src/MACsecManager.cpp
```diff
@@ -818,7 +818,7 @@ size_t MACsecManager::get_macsec_sa_count(
 
     size_t sa_count = 0;
 
-    for (macsec_an_t an = 0; an <= MAX_MACSEC_SA_NUMBER; an++)
+    for (macsec_an_t an = 0; an <= MAX_MACSEC_SA_NUMBER; an++) // lgtm [cpp/constant-comparison]
```
kcudnik, line 821: can't figure out why this is alert ?

vslib/src/SwitchStateBase.cpp
```diff
@@ -433,6 +466,22 @@ sai_status_t SwitchStateBase::setPort(
...
 
+sai_status_t SwitchStateBase::setAclEntry(
+        _In_ sai_object_id_t entry_id,
+        _In_ const sai_attribute_t* attr)
+{
+    SWSS_LOG_ENTER();
+
+    if (attr && attr->id == SAI_ACL_ENTRY_ATTR_ACTION_MACSEC_FLOW)
+    {
+        return setAclEntryMACsecFlowActive(entry_id, attr);
```
kcudnik, line 477: is this calling set_internal internally ? every custom api must call set_internal

vslib/src/SwitchStateBaseMACsec.cpp
```diff
@@ -748,40 +748,14 @@ sai_status_t SwitchStateBase::loadMACsecAttrsFromACLEntry(
...
-
 sai_status_t SwitchStateBase::getMACsecSAAttr(
         _In_ const std::string &serializedObjectId,
         _In_ uint32_t attrCount,
         _Out_ sai_attribute_t *attrList)
 {
     SWSS_LOG_ENTER();
 
-    CHECK_STATUS(get_internal(SAI_OBJECT_TYPE_MACSEC_SA, serializedObjectId, attrCount, attrList));
+    CHECK_STATUS(get(SAI_OBJECT_TYPE_MACSEC_SA, serializedObjectId, attrCount, attrList));
```
kcudnik, line 758: this and below  loop seems wrong to me, get() api will already populate all attributeList, and iy seems like you in 765 doing again some population? if those attributes must be refreshed, then they need to be done in refresh functions,
basically there should be no custom GET api implementations

kcudnik (changes requested):

vslib/src/MACsecManager.cpp
Pterosaur: Yes, I cannot understand why it's an alert. Maybe it's a false positive, I guess.

vslib/src/SwitchStateBase.cpp
Pterosaur: Yes, the `set_internal` will be called in this function.

vslib/src/SwitchStateBaseMACsec.cpp
Pterosaur: Actually, this function will never be used currently. We haven't had a finally conclusion about how to implement fetching the packet number of MACsec SA. Guohan's suggestion is we can skip this feature temporarily.
But I will remove this function if it's not acceptable to you.

kcudnik: i think you can leave it if you want, ideally we would not like to have dead code in source, but if you gonna leave it, add TODO comment, with statement that i made, that get should have not other custom attribute changes

Pterosaur: I removed it and backup this version in my private repo. Please check it.

kcudnik (approved):

lguohan:
what is this new alerts?

Pterosaur:
> what is this new alerts?

I didn't find any alert in the LGTM link. I'm not sure whether it's a false report of LGTM.

PR #723, opened 2020-12-01
[meta] Use memcpy instead of cast to prevent strict-aliasing error

kcudnik:


kcudnik:
hope will fix https://github.com/Azure/sonic-buildimage/pull/6064 

```
18:28:15  MetaKeyHasher.cpp: In function 'std::size_t sai_get_hash(const sai_fdb_entry_t&)':
18:28:15  MetaKeyHasher.cpp:168:49: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
18:28:15       return *(const uint32_t*)(&fe.mac_address[2]);
18:28:15                                                   ^
```

kcudnik:
retest vs please

meta/MetaKeyHasher.cpp
```diff
@@ -160,12 +160,20 @@ static inline std::size_t sai_get_hash(
     return ne.ip_address.addr_family;
 }
 
+static_assert(sizeof(uint32_t) == 4, "uint32_t expected to be 4 bytes");
```
qiluo-msft, line 163: >static_assert [](start = 0, length = 13)

This assertion seems not needed. Do you anticipate some failure case?

qiluo-msft (approved):

meta/MetaKeyHasher.cpp
kcudnik: just a surprise for some wired platforms :D

PR #725, opened 2020-12-03
[sairedis]vs SAI support for voq neighbor

vganesan-nokia:
Signed-off-by: vedganes <vedavinayagam.ganesan@nokia.com>

Changes include SAI support for voq neighbor. VOQ neighbor requires an
additional attribute to set the encap index. For local neighbors the
encap index is assigned by SAI, if not imposed. For remote neighbors,
the encap index is required to be supplied by the upper layer. The
changes validate these conditions for remote neighbors.


vslib/src/SwitchStateBase.cpp
```diff
@@ -2662,6 +2668,111 @@ sai_status_t SwitchStateBase::refresh_system_port_list(
...
+    }
+
+    if (is_system_neigh && is_local)
+    {
+        SWSS_LOG_ERROR("VOQ neigh info mismatch for %s. RIF is remote. attr is_local is true", serializedObjectId.c_str());
+
+        return SAI_STATUS_FAILURE;
+    }
+
+    if(!is_local && encap_index == 0)
```
kcudnik, line 2749: space after if, and remove empty line 2751

vslib/src/SwitchStateBase.cpp
```diff
@@ -2662,6 +2668,111 @@ sai_status_t SwitchStateBase::refresh_system_port_list(
...
+
+    CHECK_STATUS(create_internal(SAI_OBJECT_TYPE_NEIGHBOR_ENTRY, serializedObjectId, switch_id, attr_count, attr_list));
+
+    if (encap_index == 0)
+    {
+        // Encap index not provided. Assign encap index.
+        // The only requirement for the encap index is it must be locally
+        // unique in asic. Lower 32 bits of the ip address is used as encap index
+
+        encap_index = nbr_entry.ip_address.addr.ip4;
```
kcudnik, line 2765: according to:

```
119     /**
120      * @brief Encapsulation Index
121      *
122      * Defines the neighbor's encapsulation index
123      *
124      * @type sai_uint32_t
125      * @flags CREATE_AND_SET
126      * @default 0
127      */
128     SAI_NEIGHBOR_ENTRY_ATTR_ENCAP_INDEX,
```

if encap index is not provided default value is zero and not ipv4

kcudnik (changes requested):

vslib/src/SwitchStateBase.cpp
vganesan-nokia: When we create **local** neighbors, we either set this to 0 or do not send this attribute at all. Imposing encap index is optional for local  neighbors.  In SAI, when we see encap index = 0 and if the neighbor is local SAI knows that it has to assign the encap index by itself.  This is what I do in here. If it was remote neighbor, encap index = 0 is an error, because for remote neighbor a valid encap index must be imposed. This is what is done in bcm SAI code, which I also emulate here. 

Please also note that using last 32 bits of ip address as the encap index is vs test thing. I just wanted to have a locally unique number. We can use any mechanism to get a unique number. I just chose IP address since under normal conditions, it is expected that host ip address attached to any given asic must be unique. In actual SAI, encap index will be a unique number starting from a base number representing the FEC (used in egress encapsulation, depending on platfrom).

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Thanks. I'll fix them

vslib/src/SwitchStateBase.cpp
kcudnik: issue i have here is that this attribute is set internally, by headers definition this value is 0, so possibly when we create this neighbor, this value will be not zero, but current sonic code will expect zero, if attribute is not provided, seems like this default value should be set to "inernal" in SAI headers ?

kcudnik: if user will pass this attribute would it be set to user value or also assigned internally ?

vganesan-nokia: The encap index is not set by user but assigned internally by SAI. The normal work flow is: When a neighbor is learnt, it is sent to SAI without encap index (since it is local). For VOQ switch, SAI assigns a unique encap index to the neighbor internally. The upper layer (orchagent), retrieves this encap index after neighbor is created and syncs the neighbor info including the retrieved encap index with other asics (via chassis app db in supervisor card). In other asics, this neighbor is remote neighbor. For VOQ switches, all asics should have  equivalent neighbor record for every neighbor from other asics to have proper forwarding hosts from different asics. So when an asic creates neighbor record for remote neighbors, it uses the same exact encap index synced and the encap index attribute is sent along with attributes for **is local** flag = false and **impose index** flag = true.

vganesan-nokia: > issue i have here is that this attribute is set internally, by headers definition this value is 0, so possibly when we create this neighbor, this value will be not zero, but current sonic code will expect zero, if attribute is not provided, seems like this default value should be set to "inernal" in SAI headers ?

Sonic will not expect 0 since it sends 0 (or does not send encap index attribute) only for local neighbors to let the SAI assign encap index internally. Actually sonic expects a valid non-zero encap index from SAI after creating neighbor. Please note that as per bcm SAI, encap index =  0 is invalid. When SAI sees encap index = 0, if all other conditions are met, it assigns a valid non-zero encap index internally.

kcudnik: is this is assigned by internal SAI and does not change, then this attribute must be marked by READ_ONLY and not create_and_set

vganesan-nokia: This is still create_and_set for remote neighbors.

kcudnik: then we have some conflict here, @lguohan can you take a look here ? seems like we have some conditional flags depending on that attribute

kcudnik: what attribute determines that neighbor is local or remote ?, in case of that this attribute should be conditional for remote neighbors, and for local neighbors another RO attribute should be introduced. By design SAI attributes that are create_and_set or create_only can't change on it's own during runtime

vganesan-nokia: No attribute, but the RIF id in the neighbor entry key indicates whether the neighbor is local or remote.

Please note that, SAI does not prevent upper layer from setting encap index for local neighbors. This is the reason why the encap index attribute is not read_only. If upper layer gives a valid non-zero encap index for local neighbors, SAI accepts it. Then, it is upto the upper layer to make sure that chassis-wide unique encap index is provided for local neighbors. 

The logic is: For remote neighbor a valid encap index must be supplied. For local neighbors, it is optional, i.e., if a valid encap index is supplied, it is used, if not SAI internally assigns one.

vslib/src/SwitchStateBase.cpp
vganesan-nokia: Fixed

vslib/src/SwitchStateBase.cpp
kcudnik: then this seems to me to be default internal rather than 0

vganesan-nokia: Actually, there is an attribute "is local" that is set to false for remote neighbors. We can use this as condition for the encap index attribute.

vganesan-nokia: > 
> 
> No attribute, but the RIF id in the neighbor entry key indicates whether the neighbor is local or remote.
> 

Actually, there is an attribute "is local" that is set to false whenever remote neighbors send encap index

anshuv-mfst:
@eswaranb can you please review/approve this PR.

vslib/src/SwitchStateBase.cpp
vganesan-nokia: This condition check is done line #2834

vganesan-nokia: Sorry. I gave the line #2834 after I rebased to resolve conflict. The correct line number (from commit 1) is #2732.

rlhui: but when it's at line 2759, it could still mean SAI_NEIGHBOR_ENTRY_ATTR_ENCAP_IMPOSE_INDEX is true, and doing the encap_index assignment when  (encap_index == 0) is not the intended usage of the SAI API.

vganesan-nokia: Yes.  impose encap index == true and encap_index == 0 is invalid combination.  We'll return error (based on  check in line #2732) even before hitting #2759.

vganesan-nokia: Changed to check for impose encap index flag instead of ecap index value == 0 to decide on allocating unique encap index. This is to align with SAI API intended usage as suggested.

rlhui (approved):

kcudnik (approved):

vslib/src/SwitchStateBase.cpp
kcudnik: need fixing: https://github.com/opencomputeproject/SAI/issues/1165

PR #726, opened 2020-12-03
[syncd] Process all cases fdb flush notification

kcudnik:


syncd/syncd_notifications.cpp
```diff
@@ -176,13 +173,34 @@ void redisPutFdbEntryToAsicView(
...
+
+                    if (bp && *bp == strBridgePortId)
+                    {
+                        SWSS_LOG_DEBUG("remove fdb entry %s for SAI_FDB_EVENT_FLUSHED",fdbkey.c_str());
+                        g_redisClient->del(fdbkey);
+                    }
+                }
+                else
+                {
+                    SWSS_LOG_ERROR("found unknown type fdb entry, key %s", fdbkey.c_str());
```
prsunny, line 198: With FLUSH_TYPE ALL,  should we also get flush notification for static?

```
    /** Flush static and dynamic FDB entries */
    SAI_FDB_FLUSH_ENTRY_TYPE_ALL,
```

kcudnik: yes

kcudnik: actually

kcudnik: no

kcudnik: this is notification processing, and notification don't get type type, that's why if vendor wants to flush all static and non static, then its required for vendor to send 2 flush notifications, one, when type is set to dynamic, and second for static

kcudnik: since sai_fdb_event_notification_data_t contains sai_fdb_event_t event_type, and not sai_fdb_flush_entry_type_t

kcudnik: fixed

prsunny (approved):

PR #727, opened 2020-12-03
Dynamic Port breakout fix the crash, port down event processing after…

vishnushetty:
Fix the syncd crash during Dynamic Port Breakout (DPB).  The syncd is processing in-transit port events after port has been removed. Add check to avoid it. Please refer the inline comments for enhancement required in this area.

Nov 23 03:26:21.607291 sonic ERR syncd#syncd: :- allocateNewObjectId: object type of switch oid:0x0 is SAI_OBJECT_TYPE_NULL, should be SWITCH

### DIFF
--- a/syncd/NotificationProcessor.cpp
+++ b/syncd/NotificationProcessor.cpp
@@ -387,6 +387,7 @@ void NotificationProcessor::process_on_port_state_change(
     for (uint32_t i = 0; i < count; i++)
     {
         sai_port_oper_status_notification_t *oper_stat = &data[i];
+        sai_object_id_t rid = oper_stat->port_id;
 
         /*
          * We are using switch_rid as null, since port should be already
@@ -397,7 +398,16 @@ void NotificationProcessor::process_on_port_state_change(
          * switch vid.
          */
 
-        oper_stat->port_id = m_translator->translateRidToVid(oper_stat->port_id, SAI_NULL_OBJECT_ID);
+        oper_stat->port_id = m_translator->translateRidToVid(rid, SAI_NULL_OBJECT_ID);
+
+        if(oper_stat->port_id == SAI_NULL_OBJECT_ID)
+        {
+            /* TODO: Port might have already removed. Dont send the notification higher.
+               Delete the port from the list.
+             */
+            SWSS_LOG_NOTICE("Port RID %s notification delayed", sai_serialize_object_id(rid).c_str());
+            continue;
+        }
     }
 
     std::string s = sai_serialize_port_oper_status_ntf(count, data);
--- a/syncd/NotificationProcessor.cpp
+++ b/syncd/NotificationProcessor.cpp
@@ -387,6 +387,7 @@ void NotificationProcessor::process_on_port_state_change(
     for (uint32_t i = 0; i < count; i++)
     {
         sai_port_oper_status_notification_t *oper_stat = &data[i];
+        sai_object_id_t rid = oper_stat->port_id;
 
         /*
          * We are using switch_rid as null, since port should be already
@@ -398,14 +399,11 @@ void NotificationProcessor::process_on_port_state_change(
          */
 
         SWSS_LOG_DEBUG("Port RID %s state change notification", 
-                sai_serialize_object_id(oper_stat->port_id).c_str());
+                sai_serialize_object_id(rid).c_str());
 
-        if (false == m_translator->tryTranslateRidToVid(oper_stat->port_id, oper_stat->port_id))
+        if (false == m_translator->tryTranslateRidToVid(rid, oper_stat->port_id))
         {
-            oper_stat->port_id = SAI_NULL_OBJECT_ID;
-
-            SWSS_LOG_WARN("Port RID transalted to null VID!!!");
-
+            SWSS_LOG_WARN("Port RID %s transalted to null VID!!!", sai_serialize_object_id(rid).c_str());
         }
 
         /*
--- a/syncd/NotificationProcessor.cpp
+++ b/syncd/NotificationProcessor.cpp
@@ -387,7 +387,6 @@ void NotificationProcessor::process_on_port_state_change(
     for (uint32_t i = 0; i < count; i++)
     {
         sai_port_oper_status_notification_t *oper_stat = &data[i];
-        sai_object_id_t rid = oper_stat->port_id;
 
         /*
          * We are using switch_rid as null, since port should be already
@@ -398,16 +397,18 @@ void NotificationProcessor::process_on_port_state_change(
          * switch vid.
          */
 
-        oper_stat->port_id = m_translator->translateRidToVid(rid, SAI_NULL_OBJECT_ID);
+        SWSS_LOG_DEBUG("Port RID %s state change notification", 
+                sai_serialize_object_id(oper_stat->port_id).c_str());
 
-        if(oper_stat->port_id == SAI_NULL_OBJECT_ID)
-        {
-            /* TODO: Port might have already removed. Dont send the notification higher.
-               Delete the port from the list.
-             */
-            SWSS_LOG_NOTICE("Port RID %s notification delayed", sai_serialize_object_id(rid).c_str());
-            continue;
-        }
+        oper_stat->port_id = m_translator->tryTranslateRidToVid(oper_stat->port_id, SAI_NULL_OBJECT_ID);
+
+        /*
+         * port may be in process of removal. OA may recieve notification for VID either
+         * SAI_NULL_OBJECT_ID or non exist at time of processing 
+         */
+
+        SWSS_LOG_DEBUG("Port VID %s state change notification", 
+                sai_serialize_object_id(oper_stat->port_id).c_str());
     }
 
     std::string s = sai_serialize_port_oper_status_ntf(count, data);
--- a/syncd/NotificationProcessor.cpp
+++ b/syncd/NotificationProcessor.cpp
@@ -400,14 +400,16 @@ void NotificationProcessor::process_on_port_state_change(
         SWSS_LOG_DEBUG("Port RID %s state change notification", 
                 sai_serialize_object_id(oper_stat->port_id).c_str());
 
-        oper_stat->port_id = m_translator->tryTranslateRidToVid(oper_stat->port_id, SAI_NULL_OBJECT_ID);
-
+        if (false == m_translator->tryTranslateRidToVid(oper_stat->port_id, oper_stat->port_id))
+        {
+            oper_stat->port_id = SAI_NULL_OBJECT_ID;
+        }
         /*
          * port may be in process of removal. OA may recieve notification for VID either
          * SAI_NULL_OBJECT_ID or non exist at time of processing 
          */
 
-        SWSS_LOG_DEBUG("Port VID %s state change notification", 
+        SWSS_LOG_INFO("Port VID %s state change notification", 
                 sai_serialize_object_id(oper_stat->port_id).c_str());
     }
 
--- a/syncd/NotificationProcessor.cpp
+++ b/syncd/NotificationProcessor.cpp
@@ -398,7 +398,7 @@ void NotificationProcessor::process_on_port_state_change(
          * switch vid.
          */
 
-        SWSS_LOG_DEBUG("Port RID %s state change notification", 
+        SWSS_LOG_INFO("Port RID %s state change notification", 
                 sai_serialize_object_id(rid).c_str());
 
         if (false == m_translator->tryTranslateRidToVid(rid, oper_stat->port_id))
--- a/syncd/NotificationProcessor.cpp
+++ b/syncd/NotificationProcessor.cpp
@@ -403,9 +403,13 @@ void NotificationProcessor::process_on_port_state_change(
         if (false == m_translator->tryTranslateRidToVid(oper_stat->port_id, oper_stat->port_id))
         {
             oper_stat->port_id = SAI_NULL_OBJECT_ID;
+
+            SWSS_LOG_WARN("Port RID transalted to null VID!!!");
+
         }
+
         /*
-         * port may be in process of removal. OA may recieve notification for VID either
+         * Port may be in process of removal. OA may recieve notification for VID either
          * SAI_NULL_OBJECT_ID or non exist at time of processing 
          */
 
--- a/syncd/VirtualOidTranslator.cpp
+++ b/syncd/VirtualOidTranslator.cpp
@@ -31,15 +31,28 @@ bool VirtualOidTranslator::tryTranslateRidToVid(
     std::lock_guard<std::mutex> lock(m_mutex);
 
     if (rid == SAI_NULL_OBJECT_ID)
-        return false;
+    {
+        SWSS_LOG_DEBUG("translated RID null to VID null");
 
-    if (m_rid2vid.find(rid) == m_rid2vid.end())
-        return false;
+        vid = SAI_NULL_OBJECT_ID;
+        return true;
+    }
+
+    auto it = m_rid2vid.find(vid);
+
+    if (it != m_rid2vid.end())
+    {
+        vid = it->second;
+        return true;
+    }
 
     vid = m_client->getVidForRid(rid);
 
-    SWSS_LOG_DEBUG("translated RID %s to VID %s",
-            sai_serialize_object_id(rid).c_str(), sai_serialize_object_id(vid).c_str());
+    if(vid == SAI_NULL_OBJECT_ID)
+    {
+        SWSS_LOG_DEBUG("translated RID %s to VID null", sai_serialize_object_id(rid).c_str());
+        return false;
+    }
 
     return true;
 }
--- a/syncd/VirtualOidTranslator.cpp
+++ b/syncd/VirtualOidTranslator.cpp
@@ -22,26 +22,26 @@ VirtualOidTranslator::VirtualOidTranslator(
     // empty
 }
 
-sai_object_id_t VirtualOidTranslator::tryTranslateRidToVid(
+bool VirtualOidTranslator::tryTranslateRidToVid(
         _In_ sai_object_id_t rid,
-        _In_ sai_object_id_t switchVid)
+        _Out_ sai_object_id_t &vid)
 {
     SWSS_LOG_ENTER();
 
-    /*
-     * NOTE: switch_vid here is Virtual ID of switch for which we need
-     *  get VID for given RID. No check now! 
-     */
+    std::lock_guard<std::mutex> lock(m_mutex);
 
-    if(false == checkRidExists(rid))
-    {
-        SWSS_LOG_NOTICE("translated RID %s to VID null", sai_serialize_object_id(rid).c_str());
-        return SAI_NULL_OBJECT_ID;
-    }
+    if (rid == SAI_NULL_OBJECT_ID)
+        return false;
 
-    auto vid = m_client->getVidForRid(rid);
+    if (m_rid2vid.find(rid) == m_rid2vid.end())
+        return false;
 
-    return vid;
+    vid = m_client->getVidForRid(rid);
+
+    SWSS_LOG_DEBUG("translated RID %s to VID %s",
+            sai_serialize_object_id(rid).c_str(), sai_serialize_object_id(vid).c_str());
+
+    return true;
 }
 
 sai_object_id_t VirtualOidTranslator::translateRidToVid(
--- a/syncd/VirtualOidTranslator.cpp
+++ b/syncd/VirtualOidTranslator.cpp
@@ -48,7 +48,7 @@ bool VirtualOidTranslator::tryTranslateRidToVid(
 
     vid = m_client->getVidForRid(rid);
 
-    if(vid == SAI_NULL_OBJECT_ID)
+    if (vid == SAI_NULL_OBJECT_ID)
     {
         SWSS_LOG_DEBUG("translated RID %s to VID null", sai_serialize_object_id(rid).c_str());
         return false;
--- a/syncd/VirtualOidTranslator.cpp
+++ b/syncd/VirtualOidTranslator.cpp
@@ -83,6 +83,16 @@ sai_object_id_t VirtualOidTranslator::translateRidToVid(
         SWSS_LOG_THROW("RID 0x%" PRIx64 " is switch object, but not in local or redis db, bug!", rid);
     }
 
+    if((object_type == SAI_OBJECT_TYPE_PORT) && (switchVid == SAI_NULL_OBJECT_ID))
+    {
+        /* Port might have been removed. Dont create new object from 
+           notification handler. switchVid == NULL is right check? TODO
+         */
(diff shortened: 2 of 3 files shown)


syncd/NotificationProcessor.cpp
```diff
@@ -387,6 +387,7 @@ void NotificationProcessor::process_on_port_state_change(
     for (uint32_t i = 0; i < count; i++)
     {
         sai_port_oper_status_notification_t *oper_stat = &data[i];
+        sai_object_id_t rid = oper_stat->port_id;
```
kcudnik, line 390: this is not needed, just use port_id directly

syncd/NotificationProcessor.cpp
```diff
@@ -397,7 +398,16 @@ void NotificationProcessor::process_on_port_state_change(
          * switch vid.
          */
 
-        oper_stat->port_id = m_translator->translateRidToVid(oper_stat->port_id, SAI_NULL_OBJECT_ID);
+        oper_stat->port_id = m_translator->translateRidToVid(rid, SAI_NULL_OBJECT_ID);
+
+        if(oper_stat->port_id == SAI_NULL_OBJECT_ID)
```
kcudnik, line 403: if port_id is NULL then it will return NULL, if port_id is removed but it's not NULL, then translateRidToVid will  throw, since it will try to create new VID for null switch

what is actual scenario you are getting this error ? since if port was removed there should be no notification about this port any more

syncd/VirtualOidTranslator.cpp
```diff
@@ -83,6 +83,16 @@ sai_object_id_t VirtualOidTranslator::translateRidToVid(
         SWSS_LOG_THROW("RID 0x%" PRIx64 " is switch object, but not in local or redis db, bug!", rid);
     }
 
+    if((object_type == SAI_OBJECT_TYPE_PORT) && (switchVid == SAI_NULL_OBJECT_ID))
```
kcudnik, line 86: please remove this from here, notice that you are introducing bug here, since this translation function will refer to all ports if switch is null

syncd/NotificationProcessor.cpp
```diff
@@ -397,7 +398,16 @@ void NotificationProcessor::process_on_port_state_change(
...
+        {
+            /* TODO: Port might have already removed. Dont send the notification higher.
```
kcudnik, line 405: if  you use multiline comment, follow style like other multiline comments

kcudnik (changes requested):

kcudnik:
What is actual scenario you are hitting ?

syncd/NotificationProcessor.cpp
vishnushetty: port removal and port events (up/down) race condition (processed in separate thread). How on-fly (in transit) message being treated? SAI would have reported port event which got removed just after event has been raised. This is race condition. Need to drop this event, should not process.

syncd/NotificationProcessor.cpp
vishnushetty: ok.

syncd/NotificationProcessor.cpp
vishnushetty: ok

syncd/VirtualOidTranslator.cpp
vishnushetty: Shouldnt allocateNewObjectId() be skipped from notification?

kcudnik: no, since when syncd is doing nothing it can still generate some notification with internal object, we try to make sure that will not happen when discovering new objects

kcudnik: maybe this is a good point, to not allocate new object, i would need to think about this

syncd/NotificationProcessor.cpp
kcudnik: all notifications are synchronized with mutex that is uses to execute all api's, but yes notification arrives from vendor SAI internals in asynchronous way, so if  there is a waiting notification in the sai queue, to be posted, when another api is executed (like remove that port) then this notification will be processed under mutex, after api finished, so port will not be existing, and we will get error here.

so far i think we only hit this scenario once, in some wired case, but in general this is unavoidable, since notifications from SAI are designed to arrive in asynchronous way.

my proposal her would to to add method "tryTranslateRidToVid" which would return null when objects is not present, and log warning message with original RID value, and not send notification, or send notification with null port id just to OA awarness, which could be ignored then, since there is nothing that that could be done with that (better not send, and just log warn)

how about that ?

vishnushetty: tryTranslateRidToVid - would be good option from notification (or overload same function with try option) and send NULL object to OA. This gives an indication something being missed and can correlate (useful for debugging). Here is the trace which was seen. 

Nov 23 03:26:21.447281 sonic NOTICE swss#portsyncd: :- onMsg: Delete Ethernet24(ok) from state db
Nov 23 03:26:21.475279 sonic INFO syncd#syncd: [0] SAI_API_PORT:_brcm_sai_link_event_cb:432 Port 49 link down event cause: ADMIN_DOWN
Nov 23 03:26:21.475947 sonic NOTICE syncd#syncd: :- collectPortRelatedObjects: obtained 39 port oid:0x100000031 related RIDs
Nov 23 03:26:21.475947 sonic INFO syncd#syncd: [0] SAI_API_PORT:brcm_sai_remove_port:4531 REMOVE_PORT port_id:4294967345, bcm_port:49, phy:113
Nov 23 03:26:21.555213 sonic NOTICE swss#portsyncd: :- onMsg: nlmsg type:16 key:Ethernet16 admin:1 oper:0 addr:b8:6a:97:fd:d0:fa ifindex:148 master:0
Nov 23 03:26:21.555500 sonic NOTICE swss#portsyncd: :- onMsg: Publish Ethernet16(ok) to state db
Nov 23 03:26:21.555534 sonic NOTICE swss#portsyncd: :- onMsg: nlmsg type:16 key:Ethernet17 admin:1 oper:0 addr:b8:6a:97:fd:d0:fa ifindex:149 master:0
Nov 23 03:26:21.555843 sonic NOTICE swss#portsyncd: :- onMsg: Publish Ethernet17(ok) to state db
Nov 23 03:26:21.555843 sonic NOTICE swss#portsyncd: :- onMsg: nlmsg type:16 key:Ethernet18 admin:1 oper:0 addr:b8:6a:97:fd:d0:fa ifindex:150 master:0
Nov 23 03:26:21.556160 sonic NOTICE swss#portsyncd: :- onMsg: Publish Ethernet18(ok) to state db
Nov 23 03:26:21.606997 sonic NOTICE syncd#syncd: :- postPortRemove: removed 4 lanes from redis lane map for port RID oid:0x100000031
Nov 23 03:26:21.606997 sonic NOTICE syncd#syncd: :- postPortRemove: post port remove actions succeeded
Nov 23 03:26:21.607291 sonic ERR syncd#syncd: :- allocateNewObjectId: object type of switch oid:0x0 is SAI_OBJECT_TYPE_NULL, should be SWITCH

kcudnik: will  you be willing to add tryTranslateRidToVid ?

vishnushetty: ok. let me add and push for review again. Thanks.

syncd/VirtualOidTranslator.cpp
```diff
@@ -22,6 +22,28 @@ VirtualOidTranslator::VirtualOidTranslator(
...
+        _In_ sai_object_id_t switchVid)
+{
+    SWSS_LOG_ENTER();
+
+    /*
+     * NOTE: switch_vid here is Virtual ID of switch for which we need
+     *  get VID for given RID. No check now! 
+     */
+
+    if(false == checkRidExists(rid))
```
kcudnik, line 36: this entire method must be under mutex, since there can be context switch between checkRidExists and getVidForRid
also, trytranslate should return bool if success, and it assumes that will not create new vid if rid exists, so switchVid is not necessary param here, you are not even using it

syncd/NotificationProcessor.cpp
```diff
@@ -398,16 +397,18 @@ void NotificationProcessor::process_on_port_state_change(
          * switch vid.
          */
 
-        oper_stat->port_id = m_translator->translateRidToVid(rid, SAI_NULL_OBJECT_ID);
+        SWSS_LOG_DEBUG("Port RID %s state change notification",
```
kcudnik, line 400: make this to info

syncd/VirtualOidTranslator.cpp
kcudnik: space after if

syncd/VirtualOidTranslator.cpp
```diff
@@ -22,6 +22,28 @@ VirtualOidTranslator::VirtualOidTranslator(
...
+    {
+        SWSS_LOG_NOTICE("translated RID %s to VID null", sai_serialize_object_id(rid).c_str());
```
kcudnik, line 38: make this debug

kcudnik (changes requested):

syncd/VirtualOidTranslator.cpp
vishnushetty: ok. I removed the lock since checkRidExists() already has it and  getVidForRid is DB call.  I can add it.
Will remove switchVid  and introduce  out parameter for vid.
Will add space after if.

Please confirm.

syncd/NotificationProcessor.cpp
vishnushetty: ok.

syncd/VirtualOidTranslator.cpp
vishnushetty: ok

syncd/VirtualOidTranslator.cpp
vishnushetty: need lock to get the right data what we checked for.

kcudnik: checkRidExists uses mutex internally, so you cant use this method directly when you will add mutex here, it would be better to use checkRidExists contents here

syncd/NotificationProcessor.cpp
kcudnik: make to info ?

syncd/NotificationProcessor.cpp
```diff
@@ -397,7 +397,20 @@ void NotificationProcessor::process_on_port_state_change(
...
-        oper_stat->port_id = m_translator->translateRidToVid(oper_stat->port_id, SAI_NULL_OBJECT_ID);
+        SWSS_LOG_DEBUG("Port RID %s state change notification", 
+                sai_serialize_object_id(oper_stat->port_id).c_str());
+
+        if (false == m_translator->tryTranslateRidToVid(oper_stat->port_id, oper_stat->port_id))
+        {
+            oper_stat->port_id = SAI_NULL_OBJECT_ID;
+        }
+        /*
+         * port may be in process of removal. OA may recieve notification for VID either
```
kcudnik, line 408: start with Capital letter, and add empty line before startin comment between 406 and 407

syncd/NotificationProcessor.cpp
```diff
@@ -397,7 +397,20 @@ void NotificationProcessor::process_on_port_state_change(
          * switch vid.
          */
 
-        oper_stat->port_id = m_translator->translateRidToVid(oper_stat->port_id, SAI_NULL_OBJECT_ID);
+        SWSS_LOG_DEBUG("Port RID %s state change notification", 
+                sai_serialize_object_id(oper_stat->port_id).c_str());
+
+        if (false == m_translator->tryTranslateRidToVid(oper_stat->port_id, oper_stat->port_id))
+        {
+            oper_stat->port_id = SAI_NULL_OBJECT_ID;
```
kcudnik, line 405: does it make sense to send notification with null port ?
log warning message here, that RID was not found and you skip this notification, and just add "continue"

syncd/VirtualOidTranslator.cpp
```diff
@@ -22,6 +22,28 @@ VirtualOidTranslator::VirtualOidTranslator(
...
+bool VirtualOidTranslator::tryTranslateRidToVid(
+        _In_ sai_object_id_t rid,
+        _Out_ sai_object_id_t &vid)
+{
+    SWSS_LOG_ENTER();
+
+    std::lock_guard<std::mutex> lock(m_mutex);
+
+    if (rid == SAI_NULL_OBJECT_ID)
+        return false;
```
kcudnik, line 34: NULL value should always be translated to NULL, and this is perfect scenario, event checkRidExists return true in that case

syncd/VirtualOidTranslator.cpp
```diff
@@ -22,6 +22,28 @@ VirtualOidTranslator::VirtualOidTranslator(
...
+
+    if (m_rid2vid.find(rid) == m_rid2vid.end())
```
kcudnik, line 36: condition is wrong here, if rid is not found in local map, it can still be in m_clinet, so reverse this, if rid is found, return true, and if rid is not found, then check m_client

syncd/VirtualOidTranslator.cpp
kcudnik: please remove this

syncd/VirtualOidTranslator.cpp
```diff
@@ -22,6 +22,28 @@ VirtualOidTranslator::VirtualOidTranslator(
...
+        return false;
+
+    vid = m_client->getVidForRid(rid);
+
+    SWSS_LOG_DEBUG("translated RID %s to VID %s",
+            sai_serialize_object_id(rid).c_str(), sai_serialize_object_id(vid).c_str());
```
kcudnik, line 42: one param per line

kcudnik (changes requested):

syncd/VirtualOidTranslator.h
```diff
@@ -42,6 +42,15 @@ namespace syncd
                     _In_ sai_object_id_t rid,
                     _In_ sai_object_id_t switchVid);
 
+            /*
+             * This method will try get VID for given RID.
+             * returns false if vid doesn't exists.
```
kcudnik, line 47: return false if RID doesn't exists

syncd/NotificationProcessor.cpp
vishnushetty: Ok.

syncd/NotificationProcessor.cpp
vishnushetty: Adding WARN log is fine. If not sending  NULL object then  "oper_stat->port_id" holds invalid RID otherwise need to adjust  "data" and "count" variable data (delete NULL entries). Need to build fresh entries. Will do it that in next check-in, let report NULL object for now. Agree?

syncd/VirtualOidTranslator.cpp
vishnushetty: ok

syncd/NotificationProcessor.cpp
kcudnik: ok

syncd/VirtualOidTranslator.cpp
vishnushetty: if (rid == SAI_NULL_OBJECT_ID)
    {
        SWSS_LOG_DEBUG("translated RID null to VID null");

        vid = SAI_NULL_OBJECT_ID;
        return true;
    }

    auto it = m_rid2vid.find(vid);

    if (it != m_rid2vid.end())
    {
        vid = it->second;
        return true;
    }

    vid = m_client->getVidForRid(rid);

    if(vid == SAI_NULL_OBJECT_ID)
    {
        SWSS_LOG_DEBUG("translated RID %s to VID null", sai_serialize_object_id(rid).c_str());
        return false;
    }

    return true;

I s this correction fine?

syncd/VirtualOidTranslator.cpp
vishnushetty: ok

syncd/VirtualOidTranslator.h
vishnushetty: ok

syncd/VirtualOidTranslator.cpp
kcudnik: looks good, also you write the code so you should test this scenario :P

vishnushetty: UT log:

root@sonic:~# show log process_on_port_state_change
Dec  7 02:40:39.168385 sonic DEBUG syncd#syncd: :> process_on_port_state_change: enter
Dec  7 02:40:39.168409 sonic DEBUG syncd#syncd: :- process_on_port_state_change: port notification count: 1
Dec  7 02:40:39.168442 sonic DEBUG syncd#syncd: :- process_on_port_state_change: Port RID oid:0x10000006b state change notification
Dec  7 02:40:39.168626 sonic INFO syncd#syncd: :- process_on_port_state_change: Port VID oid:0x1000000000034 state change notification
Dec  7 02:40:39.169020 sonic DEBUG syncd#syncd: :< process_on_port_state_change: exit
root@sonic:~# redis-cli -n 1 hgetall RIDTOVID > rid2vid.txt
root@sonic:~# grep -n 0x10000006b rid2vid.txt 
1081:oid:0x10000006b
root@sonic:~# grep -n 0x1000000000034 rid2vid.txt 
1082:oid:0x1000000000034
root@sonic:~#

syncd/VirtualOidTranslator.h
```diff
@@ -44,8 +44,9 @@ namespace syncd
 
             /*
              * This method will try get VID for given RID.
-             * returns false if vid doesn't exists.
-             * returns true if exists and vid.
+             * returns true if input RID is null object and out VID is null object
+             * returns true if able to find VID and out VID object.
+             * returns false if not able to find VID and out VID is null object.
```
kcudnik, line 49: if able to find RID
if not able to find RID

syncd/VirtualOidTranslator.cpp
```diff
@@ -31,15 +31,28 @@ bool VirtualOidTranslator::tryTranslateRidToVid(
...
+    {
+        vid = it->second;
+        return true;
+    }
 
     vid = m_client->getVidForRid(rid);
 
-    SWSS_LOG_DEBUG("translated RID %s to VID %s",
-            sai_serialize_object_id(rid).c_str(), sai_serialize_object_id(vid).c_str());
+    if(vid == SAI_NULL_OBJECT_ID)
```
kcudnik, line 51: space after if

syncd/NotificationProcessor.cpp
```diff
@@ -403,9 +403,13 @@ void NotificationProcessor::process_on_port_state_change(
         if (false == m_translator->tryTranslateRidToVid(oper_stat->port_id, oper_stat->port_id))
         {
             oper_stat->port_id = SAI_NULL_OBJECT_ID;
+
+            SWSS_LOG_WARN("Port RID transalted to null VID!!!");
+
```
kcudnik, line 408: unnecessary empty space

syncd/NotificationProcessor.cpp
```diff
@@ -403,9 +403,13 @@ void NotificationProcessor::process_on_port_state_change(
         if (false == m_translator->tryTranslateRidToVid(oper_stat->port_id, oper_stat->port_id))
         {
             oper_stat->port_id = SAI_NULL_OBJECT_ID;
+
+            SWSS_LOG_WARN("Port RID transalted to null VID!!!");
```
kcudnik, line 407: please also log exact RID value, it cannot be oper_stat->port_id, since that is overrided, you need to introduce new variable

syncd/VirtualOidTranslator.cpp
```diff
@@ -83,6 +118,16 @@ sai_object_id_t VirtualOidTranslator::translateRidToVid(
...
+    if((object_type == SAI_OBJECT_TYPE_PORT) && (switchVid == SAI_NULL_OBJECT_ID))
+    {
+        /* Port might have been removed. Dont create new object from 
+           notification handler. switchVid == NULL is right check? TODO
+         */
+        SWSS_LOG_NOTICE("Port RID %s VID %s ignore allocation (from notify)!", 
+                sai_serialize_object_id(rid).c_str(), sai_serialize_object_id(vid).c_str());
+        return SAI_NULL_OBJECT_ID;
+    }
+
```
kcudnik, line 130: Remove this

syncd/VirtualOidTranslator.h
kcudnik: also if you start new sentence after dot "." start word with Capital letter

kcudnik (changes requested):

syncd/VirtualOidTranslator.cpp
kcudnik: please remove this check, you dont need it any more

syncd/VirtualOidTranslator.cpp
vishnushetty: forgot to do. Now removed it. Thanks

syncd/NotificationProcessor.cpp
```diff
@@ -397,7 +398,21 @@ void NotificationProcessor::process_on_port_state_change(
          * switch vid.
          */
 
-        oper_stat->port_id = m_translator->translateRidToVid(oper_stat->port_id, SAI_NULL_OBJECT_ID);
+        SWSS_LOG_INFO("Port RID %s state change notification", 
+                sai_serialize_object_id(rid).c_str());
+
+        if (false == m_translator->tryTranslateRidToVid(rid, oper_stat->port_id))
```
kcudnik, line 404: Just FYI, this is not entirely correct behavior, since after switch create we could receive some notifications with port id's and they maybe not found yet. we actually prevent this from happening by doing "discovery" of all possible objects on the switch after create switch will happen. More likely this scenario will happen when actual port will be removed, similar scenario has been already observed when a fdb flush operation is performed, and then bridge port is removed, same issue happens since fdb notification check will not find existing bridge port.

Anyway i have solution which im working on. to have a set of removed objects id's to see if object was removed.

kcudnik (approved):

kcudnik:
retest vs please

syncd/NotificationProcessor.cpp
vishnushetty: Thanks, good to know.  The discovery will resolve this, but need to enable notification on the switch after discovery?

kcudnik: they are already enabled

vishnushetty:
VS not related to this. 

cmd = 'ip link add zen_morse-sr type veth peer name eth1'

    def ensure_system(cmd):
        rc, output = subprocess.getstatusoutput(cmd)
        if rc:
>           raise RuntimeError(f"Failed to run command: {cmd}. rc={rc}. output: {output}")
E           RuntimeError: Failed to run command: ip link add zen_morse-sr type veth peer name eth1. rc=2. output: RTNETLINK answers: File exists

conftest.py:36: RuntimeError

kcudnik:
retest vs please

PR #729, opened 2020-12-08
[syncd] Translate removed RIDs in fdb notification

kcudnik:
So it’s basically race condition between removing bridge port and fdb flush notification. After bridge port is removed, fdb flush notification arrives, and tries to translate removed port, and fails. This PR fixes that.

Also this PR also removes all fdb entries from redis db right after fdb flush success, and not in fdb notification.

### DIFF
--- a/syncd/syncd.cpp
+++ b/syncd/syncd.cpp
@@ -328,6 +328,8 @@ sai_object_id_t redis_create_virtual_object_id(
 std::unordered_map<sai_object_id_t, sai_object_id_t> local_rid_to_vid;
 std::unordered_map<sai_object_id_t, sai_object_id_t> local_vid_to_rid;
 
+std::unordered_map<sai_object_id_t, sai_object_id_t> g_removed_rid_to_vid;
+
 void save_rid_and_vid_to_local(
         _In_ sai_object_id_t rid,
         _In_ sai_object_id_t vid)
@@ -346,6 +348,8 @@ void remove_rid_and_vid_from_local(
 
     local_rid_to_vid.erase(rid);
     local_vid_to_rid.erase(vid);
+
+    g_removed_rid_to_vid[rid] = vid;
 }
 
 /*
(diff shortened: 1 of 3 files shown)


syncd/syncd_notifications.cpp
```diff
@@ -66,199 +66,212 @@ sai_fdb_entry_type_t getFdbEntryType(
...
+            fdb->attr,
+            false);
+
+    sai_object_type_t objectType = SAI_OBJECT_TYPE_FDB_ENTRY;
+
+    std::string strObjectType = sai_serialize_object_type(objectType);
+
+    std::string strFdbEntry = sai_serialize_fdb_entry(fdb->fdb_entry);
+
+    std::string key = ASIC_STATE_TABLE + (":" + strObjectType + ":" + strFdbEntry);
```
qiluo-msft, line 255: >ASIC_STATE_TABLE [](start = 22, length = 16)

Seem you add `()` to convert `char *` into `string`. More readable way is explicitly convert first one into string.

syncd/syncd.cpp
```diff
@@ -346,6 +348,8 @@ void remove_rid_and_vid_from_local(
 
     local_rid_to_vid.erase(rid);
     local_vid_to_rid.erase(vid);
+
+    g_removed_rid_to_vid[rid] = vid;
```
qiluo-msft, line 352: >g_removed_rid_to_vid [](start = 4, length = 20)

I have concern about the long run. If the switch is running health for long, `g_removed_rid_to_vid` will accumulate all the history `rid`. There is no way to release the memory.

kcudnik: yes there is - apply view will clear that, until restart, yes, it can accumulate, but this is temporary solution, to move things going on 201911 branch, with fdb notification for removed bridge port. Im also working on fix on master. 
Do you have any proposal for making this better? Note that those oids will be checked in fdb notification, and we don't know how long it takes to process notifications, and how many notifications there can be to process.
please refer to in depth problem description on "Re: PRs for FDB fix issue" internal thread

syncd/syncd_notifications.cpp
kcudnik: not the scope of this PR, i extracted part of this function to separate method, but somehow git diff messed things around since those 2 functions are too close to each other in code

kcudnik:
ping

lguohan (approved):

rlhui:
@abdosi - fyi

PR #732, opened 2020-12-09
CPU queues on vslib

PrabhuSreenivasan:
Signed-off-by: Prabhu Sreenivasan <prabhu.sreenivasan@broadcom.com>

Add Queues to CPU port on vslib

### DIFF
--- a/vslib/inc/SwitchBCM56850.h
+++ b/vslib/inc/SwitchBCM56850.h
@@ -24,6 +24,9 @@ namespace saivs
 
         protected:
 
+            virtual sai_status_t create_cpu_qos_queues(
+                    _In_ sai_object_id_t port_id) override;
+
             virtual sai_status_t create_qos_queues_per_port(
                     _In_ sai_object_id_t port_id) override;
 
--- a/vslib/src/SwitchBCM56850.cpp
+++ b/vslib/src/SwitchBCM56850.cpp
@@ -78,6 +78,56 @@ sai_status_t SwitchBCM56850::create_qos_queues_per_port(
     return SAI_STATUS_SUCCESS;
 }
 
+sai_status_t SwitchBCM56850::create_cpu_qos_queues(
+        _In_ sai_object_id_t port_id)
+{
+    SWSS_LOG_ENTER();
+
+    sai_attribute_t attr;
+
+    // CPU queues are of type multicast queues
+    const uint32_t port_qos_queues_count = 32;
+
+    std::vector<sai_object_id_t> queues;
+
+    for (uint32_t i = 0; i < port_qos_queues_count; ++i)
+    {
+        sai_object_id_t queue_id;
+
+        CHECK_STATUS(create(SAI_OBJECT_TYPE_QUEUE, &queue_id, m_switch_id, 0, NULL));
+
+        queues.push_back(queue_id);
+
+        attr.id = SAI_QUEUE_ATTR_TYPE;
+        attr.value.s32 = SAI_QUEUE_TYPE_MULTICAST;
+
+        CHECK_STATUS(set(SAI_OBJECT_TYPE_QUEUE, queue_id, &attr));
+
+        attr.id = SAI_QUEUE_ATTR_INDEX;
+        attr.value.u8 = (uint8_t)i;
+
+        CHECK_STATUS(set(SAI_OBJECT_TYPE_QUEUE, queue_id, &attr));
+
+        attr.id = SAI_QUEUE_ATTR_PORT;
+        attr.value.oid = port_id;
+
+        CHECK_STATUS(set(SAI_OBJECT_TYPE_QUEUE, queue_id, &attr));
+    }
+
+    attr.id = SAI_PORT_ATTR_QOS_NUMBER_OF_QUEUES;
+    attr.value.u32 = port_qos_queues_count;
+
+    CHECK_STATUS(set(SAI_OBJECT_TYPE_PORT, port_id, &attr));
+
+    attr.id = SAI_PORT_ATTR_QOS_QUEUE_LIST;
+    attr.value.objlist.count = port_qos_queues_count;
+    attr.value.objlist.list = queues.data();
+
+    CHECK_STATUS(set(SAI_OBJECT_TYPE_PORT, port_id, &attr));
+
+    return SAI_STATUS_SUCCESS;
+}
+
 sai_status_t SwitchBCM56850::create_qos_queues()
 {
     SWSS_LOG_ENTER();
@@ -90,6 +140,8 @@ sai_status_t SwitchBCM56850::create_qos_queues()
     {
         CHECK_STATUS(create_qos_queues_per_port(port_id));
     }
+
+    CHECK_STATUS(create_cpu_qos_queues(m_cpu_port_id));
 
     return SAI_STATUS_SUCCESS;
 }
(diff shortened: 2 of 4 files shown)


vslib/src/SwitchBCM56850.cpp
```diff
@@ -91,6 +91,8 @@ sai_status_t SwitchBCM56850::create_qos_queues()
         CHECK_STATUS(create_qos_queues_per_port(port_id));
     }
 
+    CHECK_STATUS(create_qos_queues_per_port(m_cpu_port_id));
```
kcudnik, line 94: if i remember BCM56850 asic didn't have qos queues on cpu port, at least that was broadcom implementation, why this is changing right now ? this can lead to some wired stuff, also are you sure that those cpu queues are exactly the same as on port queue ?

kcudnik (changes requested):

vslib/src/SwitchBCM56850.cpp
PrabhuSreenivasan: Not aware of such restrictions. Most of the recent Broadcom devices support 48 CPU queues.  copp.json.j2 does reference multiple CPU for different traffic.

kcudnik: Reason im asking, is that when VS switch is created, internally it should reflect SAI objects that are in real hardware, it should be 1 to 1 match, and last time (long time ago, sai 0.9.4 ?) when i did discovery objects on cpu port for brcm, there were no queues associated with it. Maybe brcm didn't implemented them at that time or something else.
You can actually use "saidiscovery" tool in syncd package to get all objects on that specific asic to what is it right now.

PrabhuSreenivasan: When I tried the "/usr/bin/saidiscovery" command, I got the below error error on my image Maybe I need to try it on a community image.
```
root@sonic:/# /usr/bin/saidiscovery
/var/sonicbld/workspace/Build/broadcom/broadcom_sai/20-sai-build-brcm-4.2.1.3/output/x86-xgs5-deb80/saimeta/meta/saimetadata.c:79753 sai_metadata_apis_query: :- failed to query api SAI_API_BMTOR: SAI_STATUS_INVALID_PARAMETER (-5)
/var/sonicbld/workspace/Build/broadcom/broadcom_sai/20-sai-build-brcm-4.2.1.3/output/x86-xgs5-deb80/saimeta/meta/saimetadata.c:79873 sai_metadata_apis_query: :- failed to query api SAI_API_MACSEC: SAI_STATUS_NOT_IMPLEMENTED (-15)

```

PrabhuSreenivasan: I checked on a real hardware(BCM56850) and its showing 32 CPU queue.

```
root@sonic:/home/admin# show platform summary
Platform: x86_64-accton_as5712_54x-r0
HwSKU: Accton-AS5712-54X
ASIC: broadcom
ASIC Count: 1
root@sonic:/home/admin# show queue watermark CPU
COUNTERS_BUFFER_POOL_NAME_MAP is empty!
Egress shared pool occupancy per CPU queue
  Queue    Bytes
-------  -------
  CPU:0        0
  CPU:1        0
  CPU:2        0
  CPU:3        0
  CPU:4        0
  CPU:5        0
  CPU:6        0
  CPU:7        0
  CPU:8        0
  CPU:9        0
 CPU:10        0
 CPU:11        0
 CPU:12        0
 CPU:13        0
 CPU:14        0
 CPU:15        0
 CPU:16        0
 CPU:17        0
 CPU:18        0
 CPU:19        0
 CPU:20        0
 CPU:21        0
 CPU:22        0
 CPU:23        0
 CPU:24        0
 CPU:25        0
 CPU:26        0
 CPU:27        0
 CPU:28        0
 CPU:29        0
 CPU:30        0
 CPU:31        0
```

kcudnik: those errors from discovery can be ignored, you will also probably need -p parameter to pass profile, and actual errors are logged into syslog

those cpu queues, are interesting, since i dont know if we actually support for them, maybe in swss (i need to check) but anyway, ifthere are 32 of them, then that function which you added is creating only 20 queues

kcudnik: i checked swss, and they get queue list from port list, and they explicitly add cpu port to that port list
but it seems like cpu have more queues than normal ports

kcudnik: and it would actually matter how they are arranged, unicast/multicast or if they are are of different type, and whether that type is implemented in internal SAI
this could be checked by /usr/bin/saidiscoviery -p profile.ini -f -D
but i notice that it need to be fixed for mandatory attributes, i will post pr to fix this

kcudnik: Updated here: https://github.com/Azure/sonic-sairedis/pull/736 if you want to cherry pick that and check

kcudnik: Already merged, you can use saidiscovery tool to see what those queues actually look like

kcudnik: if you could run "/usr/bin/saidiscoviery -p profile.ini -f -D" and share output i could guide you how to proceed from there

PrabhuSreenivasan: Do we have any Jenkins PR image with this change? 
Got the below error when I ran it after pulling the changes to my branch,  Will the output of saidump help?

```
Feb 19 21:27:28.159701 sonic NOTICE syncd#saidiscovery: :- main: creating switch                                                          
Feb 19 21:27:28.159733 sonic ERR syncd#saidiscovery: :- create: virtual sai_status_t syncd::VendorSai::create(sai_object_type_t, sai_object_id_t*, sai_object_id_t, uint32_t, const sai_attribute_t*): api not initialized                                                          
Feb 19 21:27:28.159733 sonic ERR syncd#saidiscovery: :- create: create status: SAI_STATUS_FAILURE                                         
Feb 19 21:27:28.159775 sonic ERR syncd#saidiscovery: :- main: failed to create a switch: SAI_STATUS_FAILURE 
```

PrabhuSreenivasan: On BRCM switches all CPU queues are multicast. BCM56850 has 32 multicast queues on CPU port whereas newer devices have 48 multicast queues. I can write a new function similar to create_qos_queues_per_port() to populate 32 queues on BCM56850.

kcudnik: at which sairedis commit did you built sairedis/saidiscovery ? at lread at this one https://github.com/Azure/sonic-sairedis/commit/94fc99f899638f788824107916438d8087d2597c ?
this is intersting since initialize is called here: https://github.com/Azure/sonic-sairedis/blob/master/saidiscovery/saidiscovery.cpp#L555
and create switch after that:
https://github.com/Azure/sonic-sairedis/blob/master/saidiscovery/saidiscovery.cpp#L581

so initialize is returning success, since it's passed to create function, interesting

vslib/inc/SwitchBCM56850.h
```diff
@@ -24,6 +24,9 @@ namespace saivs
 
         protected:
 
+            sai_status_t create_cpu_qos_queues(
```
kcudnik, line 27: since this could be the case also in other platforms maybe this could be virtual functions override and not implemented function in base class SwitchStateBase like create_qos_queues_per_port?/

kcudnik (approved):

PrabhuSreenivasan:
Can you please merge this PR to help https://github.com/Azure/sonic-swss/pull/1544 pass its test.

PR #734, opened 2020-12-09
[syncd] Translate removed RIDs in fdb notification

kcudnik:


kcudnik:
ping

lguohan (approved):

inspurSDN:
hi kcudnik, 
    Hope to know why added the m_removedVidToRid for fdb. In this situation, the fdb LEARNED notification will be sent up even the bridge port id had been removed which should be discard before. And meta will add reference for the vlan related, this can make this vlan remove failed and crash.
    Although maybe this can still happen if this translation happened before the bridge port id removed, but can you add some comments for the benefit of translate it even the bridge port id had been removed? Or what is your thoughts about it?

kcudnik:
not sure if i follow correctly, probably this situation can happen, since fdb event is asynchronous, and we dont know at which point event will fire, even if bridge port was removed, do you suggest we add a comment here?

inspurSDN:
In my opinion, if bridge port id is removed, we should not learn this mac, because of the asynchronous, if mac learned translate before bridge port id, it seems only meta can refuse it. but if added m_removedVidToRid, then more mac learned can be sent up. So this may happen easier.
I think you added m_removedVidToRid should have something to deal with, but no comment here, so hope you add some comments here to tell us what is your thoughts. for example, if I want to disable it, what should we do to avoid some bad things happened.
thanks.

kcudnik:
i dont remember correctly now, but if bridge port is removed, then translation will fail on learn mac in fdb event, function check_fdb_event_notification_data will check if RID for bridge port id exists in existing database, and will not translate that

inspurSDN:
I think in check_fdb_event_notification_data, you have modified it to if (!m_translator->checkRidExists(attr.value.oid, **true**)), so if the bridge port id have removed, if may return true and translate it as if it exists still. Have I misunderstand it?

kcudnik:
i remember adding  this for a reason, but cant remember what was the reason exactly

PR #736, opened 2020-12-10
[saidiscovery] Update saidiscovery to use VendorSai object and metadata

kcudnik:
Replace static check_mandatory_attributes with Meta object, which automatically checks for mandatory objects on create switch. Previous function was incorrectly demanding mandatory objects which are on condition.

### DIFF
--- a/meta/Meta.cpp
+++ b/meta/Meta.cpp
@@ -55,14 +55,14 @@ sai_status_t Meta::initialize(
 {
     SWSS_LOG_ENTER();
 
-    return SAI_STATUS_SUCCESS;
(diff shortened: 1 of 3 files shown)


kcudnik:
retest vs please

lguohan (dismissed):

lguohan:
can you check the lgtm failure?

kcudnik:
> can you check the lgtm failure?

relates to Winline warning and recent changes to swss-common
im disabling it since it's not critical https://github.com/Azure/sonic-sairedis/pull/741

PR #737, opened 2020-12-10
Add fabric port test to vslib.

ngoc-do:
In this PR, I'm adding test for fabric ports. 

Testing asic supported in this PR is VOQ because fabric ports are enabled on a VOQ asic so that it can connect to other VOQ asics or fabric asics.

This PR is to add support for testing changes in https://github.com/Azure/sonic-swss/pull/1459. 

Together with this PR is https://github.com/Azure/sonic-buildimage/pull/6185 that defines lane mapping for fabric part in Force10-S6000.

Signed-off-by: ngocdo <ngocdo@arista.com>


vslib/src/SwitchStateBase.cpp
```diff
@@ -1793,6 +1800,10 @@ sai_status_t SwitchStateBase::refresh_read_only(
             case SAI_SWITCH_ATTR_NUMBER_OF_SYSTEM_PORTS:
             case SAI_SWITCH_ATTR_SYSTEM_PORT_LIST:
                 return refresh_system_port_list(meta);
+
+            case SAI_SWITCH_ATTR_NUMBER_OF_FABRIC_PORTS:
+            case SAI_SWITCH_ATTR_FABRIC_PORT_LIST:
+                return SAI_STATUS_SUCCESS;
```
kcudnik, line 1806: is this happening on each brcm chip ?

vslib/src/SwitchStateBase.cpp
```diff
@@ -1690,6 +1693,10 @@ sai_status_t SwitchStateBase::refresh_port_list(
         if (port_id == cpu_port_id)
             continue;
 
+        // don't put fabric ports on the list
+        if (fabric_port_set.find(port_id) != fabric_port_set.end())
```
kcudnik, line 1697: just not iterate it here?

ngoc-do: Yes, it is to speed up the search.

vslib/src/SwitchStateBase.cpp
ngoc-do: This should be on both NPU and fabric asics.

Can I have a context why you asked this? Maybe I missed something?

kcudnik: in number of fabric port can change on runtime, then thosevalues must be refreshed in similar way as other ports

kcudnik:
Are there any queues IPGs or scheduler groups related with those ports ?

vslib/src/SwitchStateBase.cpp
ngoc-do: Different from front panel ports which could be disabled/enabled in runtime, currently fabric ports don't. All fabric ports are brought up at chip init, and then they are always there. But I could have `refresh_fabric_port_list` in case in future we want to enable/disable fabric ports, but of course currently `refresh_fabric_port_list` does nothing and just returns `SAI_STATUS_SUCCESS`. Does it make sense to you?

ngoc-do:
> Are there any queues IPGs or scheduler groups related with those ports ?

Hmm which change in this commit your comment refers to?

ngoc-do:
> its not about this change, its about port properties, take a look inside source that for each port when it's created, another SG's, queues and IPGs are created for each port, but not in your case
> 
> take a look at methods:
> create_ingress_priority_groups
> create_qos_queues
> create_qos_queues_per_port
> 
> in each of specific classes

Hello Kamil, thanks for pointing these functions out -- very useful information.

Fabric ports are different from regular front panel ports. They don't have properties like INGRESS_PRIORITY_GROUP,  INGRESS_PRIORITY_GROUP_LIST, QOS_NUMBER_OF_SCHEDULER_GROUPS, QOS_SCHEDULER_GROUP_LIST, and etc...

-------------------------
            case SAI_PORT_ATTR_NUMBER_OF_INGRESS_PRIORITY_GROUPS:
                if (SAI_PORT_TYPE_FABRIC == port_type)
                {
                    rv = SAI_STATUS_ATTR_NOT_SUPPORTED_0+i;
                    break;
                }

            case SAI_PORT_ATTR_INGRESS_PRIORITY_GROUP_LIST:
            {
                if (SAI_PORT_TYPE_FABRIC == port_type)
                {
                    rv = SAI_STATUS_ATTR_NOT_SUPPORTED_0+i;
                    break;
                }

-------------------------
            case SAI_PORT_ATTR_QOS_NUMBER_OF_SCHEDULER_GROUPS:
                if (SAI_PORT_TYPE_FABRIC == port_type)
                {
                    rv = SAI_STATUS_ATTR_NOT_SUPPORTED_0+i;
                    break;
                }

            case SAI_PORT_ATTR_QOS_SCHEDULER_GROUP_LIST:
            {
                if (SAI_PORT_TYPE_FABRIC == port_type)
                {
                    rv = SAI_STATUS_ATTR_NOT_SUPPORTED_0+i;
                    break;
                }

QOS_NUMBER_OF_QUEUES is available for fabric ports, but currently not supported in network forwarding asics. I can't test on real hardware, and so queue counters for fabric ports are not collected yet.

            case SAI_PORT_ATTR_QOS_NUMBER_OF_QUEUES:
                if (SAI_PORT_TYPE_FABRIC == port_type)
                {
                    {
                        /* FIXME: Add fabric queue support in JR2 if applicable */
                        rv = SAI_STATUS_ATTR_NOT_SUPPORTED_0+i;
                    }
                }

Once Broadcom enables this feature, I will update swss, test on real hardware, as well as update vstest lib for this property.

kcudnik:
so if SAI_PORT_ATTR_QOS_NUMBER_OF_QUEUES will be supported, then also qos queues will also be supported later probably, so i think for now we can go with this approach

ngoc-do:
retest this please

ngoc-do:
Hello Kamil, do you have any advices for vs test failures? The first time I saw 4 failures, which looked weird to me (very unlikely to related to my changes). I retested, and now there were 299 failures. Those were due to another issue that seemed to system setup:

```
def __init__(self, *args):
> this = _swsscommon.new_Table(*args)
E RuntimeError: Sonic database config file doesn't exist at /var/run/redis/sonic-db/database_config.json
```
What should I do next? retest one more time?

kcudnik:
yea, this is some new error, not related to your changes

ngoc-do:
> yea, this is some new error, not related to your changes

Cool, so would you approve this PR to merge?

ngoc-do:
Hello, I'm trying to follow up what the next step I should do to merge it up?

kcudnik:
retest vs please

ngoc-do:
Thank you. It passed all. Definitely inside things got fixed.

ngoc-do:
Could we merge it now?

kcudnik (approved):

PR #738, opened 2020-12-11
[meta] Enable strict cast-align warning

kcudnik:
Required on ARM architecture where strict alignment is required.
Also fixes cast to std::size_t which size is 4 bytes


jleveque (approved):

meta/MetaKeyHasher.cpp
```diff
@@ -132,7 +132,9 @@ static inline std::size_t sai_get_hash(
 
     if (re.destination.addr_family == SAI_IP_ADDR_FAMILY_IPV6)
     {
-        const uint32_t* ip6 = (const uint32_t*)re.destination.addr.ip6;
+        // cast is not good enough for arm (cast align)
+        uint32_t ip6[4];
+        memcpy(ip6, re.destination.addr.ip6, sizeof(ip6));
```
qiluo-msft, line 137: >addr [](start = 35, length = 4)

I see `sai_ip_addr_t` is already 4 bytes aligned because `ip4` is `uint32_t`. So the `memcpy` is preventable.

kcudnik: somehow on arm architecture it still gives this warning, also i checked assembly on O2, and memcpy is optimized, and it's not there, just direct ip access

PR #747, opened 2020-12-15
[logfile]: Add handling of Sairedis rec filename 

arlakshm:
The PR implements the handling of the sairedis rec filename passed to the orchagent.
Coresponding swss pr : https://github.com/Azure/sonic-swss/pull/1546
This change is mainly to be used for multi asic devices where each asic.


kcudnik:
this will also needs to be handled by sonic syslog rotate, since how it's hardcoded to sairedis.rec i think

arlakshm:
> this will also needs to be handled by sonic syslog rotate, since how it's hardcoded to sairedis.rec i think

Yes, I will create a PR in sonic-buildimage for logrotate

lguohan:
@kcudnik  to approve

lib/inc/sairedis.h
```diff
@@ -194,4 +194,15 @@ typedef enum _sai_redis_switch_attr_t
...
     SAI_REDIS_SWITCH_ATTR_CONTEXT,
 
+    /**
+     * @brief Recording log filename.
+     *
+     * Default valus is sairedis.rec
+     *
+     * @type sai_s8_list_t
+     * @flags CREATE_AND_SET
+     * @default empty
```
kcudnik, line 204: you could make this @default "sairedis.rec"

lib/src/Recorder.cpp
```diff
@@ -97,6 +97,58 @@ bool Recorder::setRecordingOutputDirectory(
...
 }
 
+bool Recorder::setRecordingFilename(
+    _In_ const sai_attribute_t &attr)
+{
+    SWSS_LOG_ENTER();
+
+    if (attr.value.s8list.count == 0)
+    {
+        m_recordingFileName = "sairedis.rec";
```
kcudnik, line 107: this name is also used in constructor, can you make a define on top of file and reuse define here ?

kcudnik (changes requested):

lib/src/Recorder.cpp
arlakshm: address the commit in the latest commit

lib/inc/sairedis.h
arlakshm: address the commit in the latest commit

kcudnik (approved):

PR #751, opened 2020-12-16
[syncd][tests] Add syncd deprecated attribute value test

kcudnik:


tests/BCM56850.pl
```diff
@@ -594,7 +594,31 @@ sub test_bulk_object
...
+
+    my $key = `redis-cli -n 1 "keys" "*"|grep SAI_OBJECT_TYPE_NEXT_HOP_GROUP`;
+
+    chomp $key;
+
+    `redis-cli -n 1 hset "$key" "SAI_NEXT_HOP_GROUP_ATTR_TYPE" "SAI_NEXT_HOP_GROUP_TYPE_ECMP"`;
+
+    start_syncd_warm;
+
+    play "non_depreacated.rec", 0;
```
shi-su, line 615: Just wanted to confirm if 0 here means expecting 0 operations?

tests/BCM56850.pl
```diff
@@ -594,7 +594,31 @@ sub test_bulk_object
...
+
+    exit;
```
shi-su, line 617: I suspect this will exit the whole script and skip everything afterward.

kcudnik: leftover from debugging :P

tests/BCM56850.pl
kcudnik: yes, this is not the best design, i will have to redesign those tests some day

shi-su (approved):

PR #752, opened 2020-12-17
[vs] Add workaround for clean up macsec ports

kcudnik:
@Pterosaur please address this when you will be available

### DIFF
--- a/vslib/src/MACsecManager.cpp
+++ b/vslib/src/MACsecManager.cpp
@@ -845,7 +845,10 @@ void MACsecManager::cleanup_macsec_device() const
 
     if (!exec("/sbin/ip macsec show", macsecInfos))
     {
-        SWSS_LOG_THROW("Cannot show MACsec ports");
+        // this is workaround, there was exception thrown here, and it probably
+        // suggest that, ip command is in place, but don't support "macsec"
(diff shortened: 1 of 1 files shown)


kcudnik:
retest vs please

vslib/src/MACsecManager.cpp
```diff
@@ -845,7 +845,10 @@ void MACsecManager::cleanup_macsec_device() const
 
     if (!exec("/sbin/ip macsec show", macsecInfos))
     {
-        SWSS_LOG_THROW("Cannot show MACsec ports");
+        // this is workaround, there was exception thrown here, and it probably
+        // suggest that, ip command is in place, but don't support "macsec"
+        SWSS_LOG_ERROR("Cannot show MACsec ports");
```
lguohan, line 850: i think this will block log analyzer, better to change to notice level.

kcudnik: what log analyzer ? out internal ? do you propose warning ?

lguohan: the log analyzer in sonic-mgmt repo, it will report error when it does the vsimage check, warning should be fine.

lguohan (approved):

PR #753, opened 2020-12-18
[meta] Fix stat_mode enums to sai_bulk_op_error_mode_t

kcudnik:


daall (approved):
thanks Kamil!

PR #760, opened 2021-01-04
[tests] Enable azure pipeline make check to respect unittests

kcudnik:
* add check target for python module Makefile
* fix all aspell errors
* extract syslog in azure pipeline as artifact for debugging
* enable make check to be respected in azure pipeline


lguohan:
can you modify the azure-pipeline.yml to remove || true so that we will honor the test results in the build pipeline?

kcudnik:
> can you modify the azure-pipeline.yml to remove || true so that we will honor the test results in the build pipeline?

sure, but also this change:
```
04:02:04  RedisClient.cpp:675:29: error: could not convert 'swss::DBConnector::hgetall(const string&)((* & key))' from 'std::map<std::__cxx11::basic_string<char>, std::__cxx11::basic_string<char> >' to 'std::unordered_map<std::__cxx11::basic_string<char>, std::__cxx11::basic_string<char> >'
04:02:04       return m_dbAsic->hgetall(key);
04:02:04              ~~~~~~~~~~~~~~~~~^~~~~
```

will require Qi change https://github.com/Azure/sonic-sairedis/pull/759

kcudnik:
retest vs please

kcudnik:
```
tests: tests.cpp:784: void test_watchdog_timer_clock_rollback(): Assertion `settimeofday(&currentTime, NULL) == 0' failed.
/bin/bash: line 5: 13731 Aborted                 (core dumped) ${dir}$tst
```
this test requires root privilege's to be performed, i think this was never executed, so this test can be disabled

syncd/tests.cpp
```diff
@@ -767,6 +767,12 @@ void test_watchdog_timer_clock_rollback()
 {
     SWSS_LOG_ENTER();
 
+    if (getuid() != 0)
+    {
+        // this test requires root for set time
```
lguohan, line 772: print log here?

azure-pipelines.yml
```diff
@@ -68,7 +68,7 @@ jobs:
       cp ../*.deb .
     displayName: "Compile sonic sairedis"
   - script: |
-      make check || true
+      make check
```
lguohan, line 71: i think you can add sudo here if root is required to run the test.

syncd/tests.cpp
kcudnik: done

azure-pipelines.yml
kcudnik: done

kcudnik:
```Generating script.
Script contents:
sudo make check

tests: tests.cpp:790: void test_watchdog_timer_clock_rollback(): Assertion `settimeofday(&currentTime, NULL) == 0' failed.
/bin/bash: line 5: 13728 Aborted                 (core dumped) ${dir}$tst
```
seems like sudo doesent have permissions to change time, i dont know whether this is issue or sudo needs to be propagated

azure-pipelines.yml
kcudnik: reverted back since don't work

lguohan:
are you able to repro it locally?

kcudnik:
locally all my tests are passing, how i can setup and run locally Azure.sonic-siredis contaner ?

azure-pipelines.yml
```diff
@@ -200,8 +240,28 @@ jobs:
...
     displayName: "Run sonic sairedis unit tests"
   - publish: $(System.DefaultWorkingDirectory)/
     artifact: sonic-sairedis.armhf
     displayName: "Archive sonic sairedis debian packages"
+  - script: |
+      pwd
+      sudo chmod a+r /var/log/syslog*
+      tar -cf syslog-all.tar /var/log/syslog*
```
lguohan, line 251: zip it?

lguohan (approved):

lguohan:
can you make the pr title accurate to reflect the content?

mybe fix issues in make check unit tests and enable make check in ci.

kcudnik:
enabling tests requires to fix all aspell errors and they are fixed here, i can change name of the PR

kcudnik:
arm64 and armhf are failing unittests run, since they are so slow when compiling/execution, that they hit syncd/sairedis timeout on communitation channel which is 1 minute :(

lguohan:
i think we only run the unittest on amd64. i cannot solve the emulation slowness on arm64 and armhf.

kcudnik:
should i remove arm64 and armhf from yaml file ?

lguohan:
for testing yes keep the build

kcudnik:
Ok, i will merge that, and then you can disable arm64 and armhf

lguohan:
please update the pr to remove the test first.

kcudnik:
which test? this one https://github.com/Azure/sonic-sairedis/pull/760#issuecomment-765367993?? you said to keep those

lguohan:
i meant, yes, remove the test part, keep the build part.

kcudnik:
retest vs please

kcudnik:
im not sure why VS is failing, the packages is not even built

lguohan:
retest vs please

lguohan:
retest this please

kcudnik:
retest vs please

kcudnik:
success, please sign off

azure-pipelines.yml
```diff
@@ -134,11 +158,32 @@ jobs:
...
   - script: |
-      make check || true
-    displayName: "Run sonic sairedis unit tests"
+      sudo cp azsyslog.conf /etc/rsyslog.conf
+      sudo service rsyslog restart
+    displayName: "Update rsyslog.conf"
   - publish: $(System.DefaultWorkingDirectory)/
     artifact: sonic-sairedis.arm64
     displayName: "Archive sonic sairedis debian packages"
+  - script: |
```
lguohan, line 167: do you still needs this if not running the test?

kcudnik: this script is exposing syslog as artefact, so if it could be helpful if build on azure pipeline will fail, since build is always performed, just tests for arm64 and armhf was removed, tests from amd64 still are active

lguohan (approved):

PR #761, opened 2021-01-04
[syncd] Fix bulk multi attrs for same key db update

kcudnik:
attributes will be set for the same key, then when we want to, push them to redis database, we need to combine all attributes, to a single vector of attributes

issue was related to bulk set operation, when multiple attributes were set for the same key, in this case bug was causing that only last attribute was actually written to database and previous ones were skipped


yxieca (dismissed):

syncd/Syncd.cpp
```diff
@@ -2074,7 +2074,19 @@ void Syncd::syncUpdateRedisBulkQuadEvent(
 
         keys.push_back(key);
 
-        multiHash[key] = strAttributes.at(idx);
+        if (api == SAI_COMMON_API_BULK_SET)
+        {
+            // in case of bulk set operation, it can happen that multiple
+            // attributes will be set for the same key, then wehen we want to
```
lguohan, line 2080: wehen -> when

lguohan:
wonder if you add more description on the commit?

kcudnik:
done

shi-su (approved):

kcudnik:
retest vs please

kcudnik:
@lguohan should all vs tests pass? in previous PR you mentioned that 4 tests are fine ?

lguohan:
yes, if you see 4 tests failing, then it means this pr is ok. the four failed test cases are known issue.

yxieca (approved):

lguohan:
retest this please

PR #764, opened 2021-01-10
[syncd] Comparison logic add support to LABEL attribute with higher priority

kcudnik:


lguohan:
do you support the comparison logic between without label to having the label?

kcudnik:
yes, this is supported, but since there is no label to match in current view, matching will be based on lag_member and port or if lag is empty then match by number for same attributes

lguohan:
can you addd unit test for this pr, then we should be good.

kcudnik:
i added test case where we don't have label, and then we do, for 2 lags, 2 operations are expected to SET label from new viw

lguohan (approved):

PR #766, opened 2021-01-12
[ci]: use sonicbld pool

lguohan:

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -64,8 +64,7 @@ jobs:
     submodules: true
   - script: |
       ./autogen.sh
-      fakeroot dpkg-buildpackage -b -us -uc -Tbinary-syncd-vs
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -64,8 +64,7 @@ jobs:
     submodules: true
   - script: |
       ./autogen.sh
-      fakeroot dpkg-buildpackage -b -us -uc -Tbinary-syncd-vs
-      cp ../*.deb .
+      fakeroot dpkg-buildpackage -b -us -uc -Tbinary-syncd-vs -j$(nproc) && cp ../*.deb .
```
kcudnik, line 67: just of out curiosity why this was merged to 1 line?

lguohan: this is to prorogate the error code to the script

PR #774, opened 2021-01-22
[sairedis] Add get response timeout knob

kcudnik:
Knob needed by mlnx when doing firmware update on some platforms, which can exceed default 1 min timeout.
Need to be set by orchagent.

### DIFF
--- a/lib/inc/sairedis.h
+++ b/lib/inc/sairedis.h
@@ -205,4 +205,15 @@ typedef enum _sai_redis_switch_attr_t
      */
     SAI_REDIS_SWITCH_ATTR_RECORDING_FILENAME,
 
+    /**
+     * @brief Get operation response timeout in milliseconds.
+     *
+     * Also used for every synchronous API call.
+     *
+     * @type sai_uint64_t
+     * @flgs CREATE_AND_SET
+     * @default 60000
+     */
+    SAI_REDIS_SWITCH_ATTR_GET_RESPONSE_TIMEOUT_MS,
+
 } sai_redis_switch_attr_t;
--- a/lib/inc/sairedis.h
+++ b/lib/inc/sairedis.h
@@ -13,6 +13,11 @@ extern "C" {
  */
 #define SAI_REDIS_KEY_CONTEXT_CONFIG              "SAI_REDIS_CONTEXT_CONFIG"
 
+/**
+ * @brief Default synchronous operation response timeout in milliseconds.
+ */
+#define SAI_REDIS_DEFAULT_SYNC_OPERATION_RESPONSE_TIMEOUT (60*1000)
+
 typedef enum _sai_redis_notify_syncd_t
 {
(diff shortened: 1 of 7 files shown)


lguohan:
can this be set on-the-fly?

kcudnik:
yes it can

kcudnik:
retest vs please

lib/inc/sairedis.h
```diff
@@ -205,4 +205,15 @@ typedef enum _sai_redis_switch_attr_t
...
+    /**
+     * @brief Get operation response timeout in milliseconds.
+     *
+     * Also used for every synchronous API call.
+     *
+     * @type sai_uint64_t
+     * @flgs CREATE_AND_SET
+     * @default 60000
+     */
+    SAI_REDIS_SWITCH_ATTR_GET_RESPONSE_TIMEOUT_MS,
```
liorghub, line 217: I believe it should be SET_RESPONSE_TIMEOUT_MS

kcudnik: mainly this timeout was designed for GET api in any quad api, but with adding synchronous mode it's used for all of them

lib/src/Channel.cpp
```diff
@@ -4,9 +4,12 @@
 
 using namespace sairedis;
 
+#define REDIS_ASIC_STATE_COMMAND_GETRESPONSE_TIMEOUT_MS (60*1000)
```
liorghub, line 7: I would like to use this define in sonic-swss/orchagent/saihelper.cpp
Can you please move this define to .h file that can be included from both Channel.cpp and saihelper.cpp?
If there is such...

kcudnik: why would you need this define on saihelper?

liorghub: I am using the new API to change the timeout and later on I want to change it back to default, this is where I want to use your define.
In the PR bellow I used new define SAI_REDIS_GET_RESPONSE_TIMEOUT_MS_DEFAULT but I would like to replace it to REDIS_ASIC_STATE_COMMAND_GETRESPONSE_TIMEOUT_MS if it is possible.
[(https://github.com/liorghub/sonic-swss/pull/3)]

kcudnik: updated

kcudnik: i was actually thinking, that this define maybe should not be exposed as define here, but you should do SAI api GET api, to obtain this attribute default value

liorghub: Generally, if we have set API we can have get API, basic stuff.
However, there is nothing wrong with exposing the default value in header file and for my needs it will do the work just fine. I usually vote for the option with less lines of code.

kcudnik: ok i leave it as is

liat-grozovik:
@liorghub and @nazariig  can you please approve so we can move forward and merge?

liorghub (approved):
LGTM

PR #777, opened 2021-01-25
[syncd armhf] Fix syncd crash when running community test suites

dflynn-Nokia:
This commit fixes a syncd crash seen when running the sonic-mgmt comunity test
suites against the Marvell armhf platform. Analysis of the generated core file
points to improper format specifiers used when writing SWSS log entries. This
commit fixes that.

### DIFF
--- a/lib/src/VirtualObjectIdManager.cpp
+++ b/lib/src/VirtualObjectIdManager.cpp
@@ -2,6 +2,7 @@
 
 #include "meta/sai_serialize.h"
 #include "swss/logger.h"
+#include <inttypes.h>
 
 extern "C" {
 #include "saimetadata.h"
@@ -242,7 +243,7 @@ sai_object_id_t VirtualObjectIdManager::allocateNewObjectId(
 
     if (objectIndex > SAI_REDIS_OBJECT_INDEX_MAX)
     {
-        SWSS_LOG_THROW("no more object indexes available, given: 0x%lx but limit is 0x%llx",
+        SWSS_LOG_THROW("no more object indexes available, given: 0x%" PRIu64 " but limit is 0x%" PRIu64 " ",
                 objectIndex,
                 SAI_REDIS_OBJECT_INDEX_MAX);
     }
--- a/lib/src/PerformanceIntervalTimer.cpp
+++ b/lib/src/PerformanceIntervalTimer.cpp
@@ -1,6 +1,7 @@
 #include "PerformanceIntervalTimer.h"
 
 #include "swss/logger.h"
+#include <inttypes.h>
 
 using namespace sairediscommon;
 
@@ -46,7 +47,7 @@ void PerformanceIntervalTimer::inc(
     {
         if (m_enable)
         {
-            SWSS_LOG_NOTICE("%zu (calls %zu) %s op took: %zu ms", m_count, m_calls, m_msg.c_str(), m_total/1000000);
+            SWSS_LOG_NOTICE("%" PRIu64 " (calls %" PRIu64 ") %s op took: %" PRIu64 " ms", m_count, m_calls, m_msg.c_str(), m_total/1000000);
         }
 
         reset();


lib/src/VirtualObjectIdManager.cpp
```diff
@@ -242,7 +243,7 @@ sai_object_id_t VirtualObjectIdManager::allocateNewObjectId(
 
     if (objectIndex > SAI_REDIS_OBJECT_INDEX_MAX)
     {
-        SWSS_LOG_THROW("no more object indexes available, given: 0x%lx but limit is 0x%llx",
+        SWSS_LOG_THROW("no more object indexes available, given: 0x%" PRIu64 " but limit is 0x%" PRIu64 " ",
```
kcudnik, line 246: PRIx64 ?

kcudnik (changes requested):

lib/src/VirtualObjectIdManager.cpp
dflynn-Nokia: My thinking is that PRIu64 is the way to go here given the following

    uint64_t objectIndex = m_oidIndexGenerator->increment();

and ...

    #define SAI_REDIS_OBJECT_INDEX_MAX ( (1ULL << SAI_REDIS_OBJECT_INDEX_BITS_SIZE) - 1 )
    #define SAI_REDIS_OBJECT_INDEX_MASK (SAI_REDIS_OBJECT_INDEX_MAX)

Both cases indicate unsigned.

dflynn-Nokia:
retest vs please

lib/src/VirtualObjectIdManager.cpp
kcudnik: ok

dflynn-Nokia:
It looks like the vs test failed due to running out of disk space on jenkins worker vstest-w7bdd50.
Can someone clean that up?

[2021-01-25T16:06:59.455Z] Also:   java.io.IOException: No space left on device

dflynn-Nokia:
Retest vs please

kcudnik:
@daall hey, can you take a look ?

dflynn-Nokia:
@daall Can you take a look at this please?

lguohan:
@kcudnik, can you check? build passed?

kcudnik:
all builds are now passing, but i didnt change anything

lguohan:
@kcudnik, can you approve?

kcudnik (approved):

kcudnik:
this is interesting, since we compiled armhf before without this changes in azure armhf pipeline, and we didnt get those errors

dflynn-Nokia:
@kcudnik Thanks for moving this one along.
Actually this was a syncd run-time failure rather than a build failure. More specifically syncd, started successfully and appeared to be operational but crashed and core'd when running the community test suites. The syncd backtrace showed a segment violation within libc. I would have thought that the compiler would have caught this but not the case this time.

kcudnik:
but PRI prefixes should warn on compile time not runtime

dflynn-Nokia:
Agreed.
I'm thinking this slipped through the compiler due to a compiler bug or crashed within libc due to a libc bug.
Not sure which is in play here.

kcudnik:
somehow i get locally error like this:
```
In file included from ../meta/sai_serialize.h:19,
                 from ../lib/src/VirtualObjectIdManager.cpp:3:
../lib/src/VirtualObjectIdManager.cpp: In member function 'sai_object_id_t sairedis::VirtualObjectIdManager::allocateNewObjectId(sai_object_type_t, sai_object_id_t)':
/home/sonic/include/swss/logger.h:24:101: error: format '%lu' expects argument of type 'long unsigned int', but argument 6 has type 'long long unsigned int' [-Werror=format=]
   24 | #define SWSS_LOG_THROW(MSG, ...)       swss::Logger::getInstance().wthrow(swss::Logger::SWSS_ERROR, "e:- %s: " MSG, __FUNCTION__, ##__VA_ARGS__)
../lib/src/VirtualObjectIdManager.cpp:246:9: note: in expansion of macro 'SWSS_LOG_THROW'
  246 |         SWSS_LOG_THROW("no more object indexes available, given: 0x%" PRIu64 " but limit is 0x%" PRIu64 " ",
```

kcudnik:
also if you want to use "u" instead "x", remove "0x" from the log line

PR #779, opened 2021-01-26
Remove unregisterMessageHandler from NetMsgRegistrar thread

vaibhavhd:
Fixes: https://github.com/Azure/sonic-buildimage/issues/6466

This is a temporary fix to avoid double-free memory corruption (and core) that is consistently seen in syncd shutdown in KVM switch.
The call made to unregisterMessageHandler as part of NetMsgRegistrar thread causes, double-free memory corruption in one of the executing threads.

### DIFF
--- a/vslib/src/NetMsgRegistrar.cpp
+++ b/vslib/src/NetMsgRegistrar.cpp
@@ -134,9 +134,6 @@ void NetMsgRegistrar::run()
         }
     }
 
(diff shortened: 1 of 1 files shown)


kcudnik:
just confirmed that this fixes issue

kcudnik (approved):

kcudnik:
retest vs please

PR #784, opened 2021-02-01
[build-vm] Don't Install Lib Boost 1.67

tahmed-dev:
Libboost 1.71 is already installed in slave buster. Do not install v1.67 on top of it.

singed-off-by: Tamer Ahmed <tamer.ahmed@microsoft.com>

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -35,8 +35,7 @@ jobs:
         libnl-cli-3-dev \
         swig3.0 \
         libpython2.7-dev \
-        libzmq5 libzmq3-dev \
-        libboost1.71-dev
+        libzmq5 libzmq3-dev
 
       sudo apt-get install -y redis-server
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -36,7 +36,7 @@ jobs:
         swig3.0 \
         libpython2.7-dev \
         libzmq5 libzmq3-dev \
-        libboost-all-dev
+        libboost1.71-dev
```
qiluo-msft, line 39: >libboost1.71-dev [](start = 8, length = 16)

I see it already in slave container: sonic-slave-buster/Dockerfile.j2:324
Will you just remove `libboost-all-dev` ?

tahmed-dev: I did not notice this docker is deriving from slave buster. I was under the impression it is a parallel one. Will remove

qiluo-msft (approved):

kcudnik (approved):

PR #785, opened 2021-02-02
[201911][Flex Counters] Add PFC pause duration counters in microseconds

volodymyrsamotiy:
Signed-off-by: Volodymyr Samotiy <volodymyrs@nvidia.com>

Extended supported counters for "flex counter" logic in order to add support of PFC pause duration counters in microseconds.


abdosi (approved):
same change as 201811

abdosi:
@volodymyrsamotiy I will update submodule

PR #786, opened 2021-02-03
[syncd] Create base class SaiSwitchInterface

kcudnik:
This class will be needed for future asic compare.

### DIFF
--- a/syncd/SaiSwitchInterface.h
+++ b/syncd/SaiSwitchInterface.h
@@ -1,0 +1,123 @@
+#pragma once
+
+extern "C" {
+#include "sai.h"
+#include "saimetadata.h"
+}
+
+#include <set>
+#include <unordered_map>
+#include <map>
+
+namespace syncd
+{
+    class SaiSwitchInterface
+    {
+        private:
+
+            SaiSwitchInterface(const SaiSwitchInterface&);
+            SaiSwitchInterface& operator=(const SaiSwitchInterface&);
+
+        public:
+
+            SaiSwitchInterface(
+                    _In_ sai_object_id_t switchVid,
+                    _In_ sai_object_id_t switchRid);
+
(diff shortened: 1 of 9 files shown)


syncd/SaiSwitchInterface.h
```diff
@@ -0,0 +1,123 @@
...
+                    _In_ sai_object_id_t switchRid);
+
+            virtual ~SaiSwitchInterface() = default;
+
+        public:
+
+            sai_object_id_t getVid() const;
+            sai_object_id_t getRid() const;
+
+        public:
```
qiluo-msft, line 34: >public: [](start = 8, length = 7)

Wondering why you need consequent `public:`

syncd/SaiSwitchInterface.h
```diff
@@ -0,0 +1,123 @@
...
+
+            virtual std::unordered_map<sai_object_id_t, sai_object_id_t> getVidToRidMap() const = 0;
+
+            virtual std::unordered_map<sai_object_id_t, sai_object_id_t> getRidToVidMap() const = 0;
+
+            virtual bool isDiscoveredRid(
+                    _In_ sai_object_id_t rid) const  = 0;
```
qiluo-msft, line 41: >   [](start = 51, length = 2)

Use one blank?

qiluo-msft (commented):
As comments

syncd/SaiSwitchInterface.h
kcudnik: done

syncd/SaiSwitchInterface.h
kcudnik: i use this as separator for similar blocks, here block of non virtual public functions , and next block of public virtual functions

qiluo-msft (approved):

PR #788, opened 2021-02-03
[ci]: use publish pipeline artifacts to publish syslog

lguohan:
avoid to use publish build artifacts and pipeline artifacts
at the same time.


kcudnik (approved):
why to change it like this ? current approach works and publishes correctly

lguohan:
mainly downstream project could not consume the artficats.

PR #789, opened 2021-02-04
syncd: Fix profile iterator bug

abanu-ms:
As I didn't see SAI specifying that the user of the service table should reset the iterator before iterating over the values, I think this would come as a surprise to most vendors implementing SAI. Because the iterator was first stored before adding any values to the map, it would be pointing to the profile's .end(), thus if a SAI implementation would try iterating over the values,  it'd get none until it'd reset the iterator.


kcudnik:
good catch, probably so SAI was iterating all values, just directly query for specific values

abanu-ms:
@kcudnik Could you approve the PR? (thought you might forgot about that - sorry if that's not the case)

kcudnik (approved):

PR #791, opened 2021-02-08
[saiasiccmp] Add saiasiccmp tool to compare 2 asic views

kcudnik:
Will be used to compare 2 independent ASIC view dumps for unittest.
File format must by provided by command:
```
$ ./redisdl.py -d 1 -y
```
from https://github.com/p/redis-dump-load repo


kcudnik:
vs- no space left on device

saiasiccmp/test.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+./saiasiccmp dump1.json dump2.json
+
+if [ $? != 0 ]; then
+    echo "ERROR: expected dumps do be equal"
+    exit 1
+fi
```
lguohan, line 8: can we add a negative test as well?

kcudnik: done

lguohan (approved):

PR #796, opened 2021-02-14
[vs] Add special warm boot logic to populate default attributes

kcudnik:
When doing warm boot from older SAI (201811) to master some attributes
maybe not implemented in previous version, but they are expected to be
present on new version. Special logic is required to populate those
default attributes after warm boot.

### DIFF
--- a/vslib/src/SwitchMLNX2700.cpp
+++ b/vslib/src/SwitchMLNX2700.cpp
@@ -48,6 +48,8 @@ sai_status_t SwitchMLNX2700::create_qos_queues_per_port(
         attr[1].id = SAI_QUEUE_ATTR_PORT;
         attr[1].value.oid = port_id;
 
+        // TODO add type
+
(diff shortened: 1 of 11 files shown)


vslib/src/SwitchMLNX2700.cpp
```diff
@@ -48,6 +48,8 @@ sai_status_t SwitchMLNX2700::create_qos_queues_per_port(
         attr[1].id = SAI_QUEUE_ATTR_PORT;
         attr[1].value.oid = port_id;
 
+        // TODO add type
```
shi-su, line 51: Just want to confirm if I understood this correctly. Does this mean `SAI_QUEUE_ATTR_TYPE`?

kcudnik: yes

shi-su (approved):

PR #797, opened 2021-02-16
Update FlexCounter.cpp, use m_pollInterval in MUTEX lock

ryan44guo:
m_pollInterval is used out of the MUTEX lock，when set to a smaller value, it may result of wait for a very long time.


kcudnik:
can you explain case in the description? also value supplied to wait_for is the same, so what the actual change is made here?

kcudnik (approved):

PR #798, opened 2021-02-16
Update FlexCounter.cpp, use m_pollInterval in MUTEX lock

ryan44guo:
m_pollInterval is used out of the MUTEX lock，when set to a smaller value, it may result of wait for a very long time.

### DIFF
--- a/syncd/syncd_flex_counter.cpp
+++ b/syncd/syncd_flex_counter.cpp
@@ -1720,12 +1720,12 @@ void FlexCounter::flexCounterThread(void)
         uint32_t delay = static_cast<uint32_t>(
                 std::chrono::duration_cast<std::chrono::milliseconds>(finish - start).count());
         uint32_t correction = delay % m_pollInterval;
-
+        correction = m_pollInterval - correction;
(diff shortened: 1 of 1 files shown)


kcudnik (changes requested):
can you explain case in the description? also value supplied to wait_for is the same, so what the actual change is made here?

ryan44guo:
I have encountered one case. the m_pollInterval is 1000 at first, and the delay is 98870, so in the old code, the correction is 870. you know, the lock is unlocked, and the action setting m_pollInterval to 100 which waiting for this lock is executed. so when the wait_for function executed, the value used is 100-870=0xfffffcfe, and this thread is blocking here.

kcudnik:
im not sure if i get it, the value still will be 0xfffffcfe before mutex or after mutex anyway

ryan44guo:
after my modification, the use of m_pollInterval is used in the lock, it can not be modified before unlock. cause the set action is in the lock too.

kcudnik (commented):
got it

kcudnik (approved):

PR #800, opened 2021-02-26
Check if libsaibcm is installed and install it dynamically on syncd docker start

judyjoseph:
This change is to support the support for DNX libsaibcm modules to be auto installed in docker based on the platform type. 
It checks first to see if libsaibcm package is installed and proceeds if not installed.
Ref PR : https://github.com/Azure/sonic-buildimage/pull/6882

With the approach in this PR, there will be an additional ~4 secs which syncd docker takes to install the saibcm DEB package. Since swss/syncd/teamd docker startup happens in parallel, this should not have a bigger impact ( not adding up the time it takes for control plane to be up ) in case of warm boot.

### DIFF
--- a/syncd/scripts/syncd_init_common.sh
+++ b/syncd/scripts/syncd_init_common.sh
@@ -87,6 +87,26 @@ function set_start_type()
 
 config_syncd_bcm()
 {
+    # Check if the libsaibcm package is installed.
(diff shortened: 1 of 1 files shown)


judyjoseph:
Closing this as Azure/sonic-buildimage#6882 is closed.

PR #804, opened 2021-03-04
[vslib] Allow host interface object ID to be "VLAN" type

volodymyrsamotiy:
Signed-off-by: Volodymyr Samotiy <volodymyrs@nvidia.com>

Allow host interface object ID to be "VLAN" type.
Needed for VS test for the following changes: https://github.com/Azure/sonic-swss/pull/1645

### DIFF
--- a/vslib/src/SwitchStateBaseHostif.cpp
+++ b/vslib/src/SwitchStateBaseHostif.cpp
@@ -560,9 +560,15 @@ sai_status_t SwitchStateBase::vs_create_hostif_tap_interface(
 
     sai_object_type_t ot = objectTypeQuery(obj_id);
 
-    if ((ot != SAI_OBJECT_TYPE_PORT) && (ot != SAI_OBJECT_TYPE_VLAN))
+    if (ot == SAI_OBJECT_TYPE_VLAN)
     {
-        SWSS_LOG_ERROR("SAI_HOSTIF_ATTR_OBJ_ID=%s expected to be PORT or VLAN but is: %s",
+        SWSS_LOG_DEBUG("Skipping tap creation for hostif with object type VLAN");
+        return SAI_STATUS_SUCCESS;
+    }
+
+    if (ot != SAI_OBJECT_TYPE_PORT)
+    {
+        SWSS_LOG_ERROR("SAI_HOSTIF_ATTR_OBJ_ID=%s expected to be PORT but is: %s",
                 sai_serialize_object_id(obj_id).c_str(),
                 sai_serialize_object_type(ot).c_str());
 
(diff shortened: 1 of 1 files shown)


prsunny (dismissed):

vslib/src/SwitchStateBaseHostif.cpp
```diff
@@ -560,9 +560,9 @@ sai_status_t SwitchStateBase::vs_create_hostif_tap_interface(
 
     sai_object_type_t ot = objectTypeQuery(obj_id);
 
-    if (ot != SAI_OBJECT_TYPE_PORT)
+    if ((ot != SAI_OBJECT_TYPE_PORT) && (ot != SAI_OBJECT_TYPE_VLAN))
```
kcudnik, line 563: i think this is not so easy, since inside there is logic, that actually depends that this host object id to be port, when packet arrives that port is used to decode interface name

kcudnik (changes requested):

vslib/src/SwitchStateBaseHostif.cpp
kcudnik: curently all traffic forwarding and "netdev" device is only PORT based, there is no logic that would support vlan here, it's not that simple just to allow that object type

volodymyrsamotiy: For now this is just partial request for virtual switch in order to make the test working in this PR https://github.com/Azure/sonic-swss/pull/1645.
We are not sending traffic in VS test only configuration is verified.
Can we have this only configuration support merged for the mentioned test?
And in the future if it will be needed VS traffic support can be added.

kcudnik: if there is no support for it implemented now, why the test should pass then ?

volodymyrsamotiy: Because VS test just covers configuration part for now, without any traffic.

kcudnik: then im confused here, port objects is covered, there is no implementation for vlan, so test is passing for port then ?

volodymyrsamotiy: For VLAN there is implementation in orchagent, added by this PR https://github.com/Azure/sonic-swss/pull/1645.
This PR adds possibility to configure hostif attribute for the VLAN.
And for that changes also VS test was added which just configures VLAN with hostif and checks if it was added to ASIC DB.
That is why changes in current PR are needed in order to allow such config so VS test can pass.

kcudnik: but this PR does not provide support for VLAN implementation, just object type check, internally code is executed as if this would be PORT not vlan

volodymyrsamotiy:
Updated PR with the new changes.
Currently it is under verification and I will update once finished but appreciate if you can review and provide your feedback menwhile.

kcudnik (approved):

PR #805, opened 2021-03-11
[python] Extend python module to support all SAI objects

kcudnik:


lguohan:
is this needed?

kcudnik:
This is first of python module that i started to implementing, where everything was a string, and each object was a dictionary. There is already 2nd module which is generated using SWIG, and after playing with it i decided that version will easier to maintain. So this module will not be needed any more. 

Just wondering if we could check that in and then remove it from repo, to have a footprint in commits, if we would like whenever to get back to this, or to use technology for submodule used in this solution. We would make sure that is still in sairedis repo itself instead of grepping through branches of github.

PR #809, opened 2021-03-19
Introduce gbsyncdmgrd; Deprecate gbsyncd_startup.py

jleveque:
Introduce `gbsyncdmgrd`, which will be responsible for spawning and watching all gearbox syncd processes.

`gbsyncdmgrd` is intended to supplant `gbsyncd_startup.py`. But until all references have been updated, we keep both, and I have added a deprecation warning to `gbsyncd_startup.py`.

### DIFF
--- a/syncd/scripts/gbsyncd_startup.py
+++ b/syncd/scripts/gbsyncd_startup.py
@@ -1,6 +1,12 @@
 #!/usr/bin/env python3
 
 '''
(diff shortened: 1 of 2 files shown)


jleveque:
@sydlogan: Please review.

lguohan (dismissed):

lguohan (approved):

PR #821, opened 2021-04-13
[syncd] Bring back TimerWatchdog

kcudnik:


syncd/Syncd.cpp
```diff
@@ -4295,6 +4296,14 @@ bool Syncd::isVeryFirstRun()
     return firstRun;
 }
 
+static void timerWatchdogCallback(
+        _In_ int64_t span)
+{
+    SWSS_LOG_ENTER();
+
+    SWSS_LOG_ERROR("main loop execution exceeded %ld ms", span);
```
shi-su, line 4304: This is in millisecond (ms) or microsecond (us)

kcudnik: microseconds :P

kcudnik: fixed

shi-su (approved):

PR #830, opened 2021-05-21
[VS] Add support for context and multiple switches

kcudnik:
So far virtual switch supported only 1 switch to be created and only default hwinfo which was empty string. This change allows to reuse context_config.json to allow multiple switches to be created in virtual switch and each can have different hwinfo. This scenario can be useful when running VS test with multiple swss containers running.


SuvarnaMeenakshi (approved):
Tested with https://github.com/Azure/sonic-sairedis/pull/832 changes on multi-asic VS switch to make sure that swss/syncd dockers are coming and interfaces are getting created.
Tested on single asic - comes up with all interfaces.

PR #831, opened 2021-05-22
Add the option to provide the package name in dh build

judyjoseph:
Ref: https://github.com/Azure/sonic-buildimage/pull/7598

**NOTE** : Will be closing this PR as it is not needed.


kcudnik (approved):

judyjoseph:
Discussed with Kamil, this change won't be needed as syncd directly don't need SAIbcm. It just needs the SAI headers which is there in saibcm-dev. Closing this.

PR #838, opened 2021-06-08
[sairedis] Add support for client server architecture

kcudnik:
With this feature libsairedis will have ability to act as a client and server, for example as client (pbhorch) can connect to existing server (OA) and call SAI api in server scope. Client connects over zmq channel and can be running in separate docker, connecting over ipc pipe or tcp.

TODO support on client:
- notifications
- non quad api (fdb flush, stats, bulk)
- context and multiple switches
- client/server zmq channel configuration file
- client sai api recording


akokhan:
Will this affect existing approach how libsairedis is used? Thanks.

kcudnik:
> Will this affect existing approach how libsairedis is used? Thanks.

no, there is no impact on existing sairedis, everything is backward compatible

PR #843, opened 2021-06-17
[recorder] Fix incorrect attribute enum value capability query

stepanblyschak:
Recorder was assuming that enum value capability query is executed for an
attribute that has a value type of a s32 list.

When querying SAI_DEBUG_COUNTER_ATTR_TYPE using
sai_query_attribute_enum_values_capability a warning is printed in
syslog: "enum value 4 not found in enum sai_debug_counter_type".

This happens because SAI_DEBUG_COUNTER_ATTR_TYPE attrvaluetype is int32
(enum value) and sai_s32_list_t structure was casted to int32. Since we
initialize sai_s32_list with .count = 4 (the number of values in
sai_debug_counter_type_t enum) value 4 is printed in the syslog.

Signed-off-by: Stepan Blyschak <stepanb@nvidia.com>


kcudnik:
int32 is exact same as s32 type, what's the problem here? if the problem i that count is treated as first value, then point it to list.list

stepanblyschak:
@kcudnik 

This serialization works fine, because SAI_DEBUG_COUNTER_ATTR_IN_DROP_REASON_LIST has attrvaluetype of SAI_ATTR_VALUE_TYPE_INT32_LIST -> ```sai_serialize_enum_list(attr.value.s32list, meta.enummetadata, countOnly);``` will be called.:
```
attr.id = SAI_DEBUG_COUNTER_ATTR_IN_DROP_REASON_LIST;
attr.value.s32list = *enumValuesCapability;
auto values = SaiAttributeList::serialize_attr_list(objectType, 1, &attr, true);
```

And below code will not work and throw a warning in log, attrvaluetype for SAI_DEBUG_COUNTER_ATTR_TYPE is SAI_ATTR_VALUE_TYPE_INT32 -> ```sai_serialize_enum(attr.value.s32, meta.enummetadata);``` will be called :

```
attr.id = SAI_DEBUG_COUNTER_ATTR_TYPE;
attr.value.s32list = *enumValuesCapability;
auto values = SaiAttributeList::serialize_attr_list(objectType, 1, &attr, true);
```

Overall this looks strange because enum value capability serialization is handled as an attribute serialization but enum value capability is not an attribute.

[KC] if the problem i that count is treated as first value, then point it to list.list
Didn't get this part.

kcudnik:
please provide unittest that will test both test cases

stepanblyschak:
@kcudnik I have thought about it, I wanted to use existing infra (sonic-sairedis/tests), but I found that even existing tests BCM56850/query_attr_enum_values_capability.rec are not checked because 'q' and 'Q' types of operations are skipped - https://github.com/Azure/sonic-sairedis/blob/master/saiplayer/SaiPlayer.cpp#L1980

kcudnik:
could you then add explicit test in lib/src/tests.cpp for Recorder.cpp class for this api?

lib/src/Recorder.cpp
```diff
@@ -1041,14 +1043,18 @@ void Recorder::recordQueryAattributeEnumValuesCapability(
...
+        return;
+    }
 
-    sai_attribute_t attr;
+    auto key = sai_serialize_object_type(SAI_OBJECT_TYPE_SWITCH) + ":" + sai_serialize_object_id(switchId);
 
-    attr.id = attrId;
-    attr.value.s32list = *enumValuesCapability;
+    auto str_attr_id = sai_serialize_attr_id(*meta);
+    auto str_enum_list = sai_serialize_enum_list(*enumValuesCapability, meta->enummetadata, true); // we only need to serialize count
```
kcudnik, line 1055: also, does this problem applies to RedisRemoteInterfaceSai.cpp ? and same on Syncd.cpp on desterilize and need to be corrected in both of them?

kcudnik (changes requested):

lib/src/Recorder.cpp
stepanblyschak: No, in RedisRemoteInterfaceSai/Syncd there is different serialization/deserialization implemented. All works well, it just the recorder that fails to record when querying SAI_DEBUG_COUNTER_ATTR_TYPE

kcudnik:
```
 *** ./../lib/src/tests.cpp : method is missing SWSS_LOG_ENTER():
 std::vector<std::string> parseFirstRecordedAPI()
{
```

kcudnik (approved):

liat-grozovik:
@kcudnik could you please merge?
@stepanblyschak  should it be taken for other active branches such as 201911 and 202012? can it be cleanly cherry picked?

kcudnik:
something is broken here in tests, can you force push to this ? or add some dumy commit/squast, to retrigger build?

stepanblyschak:
@liat-grozovik Yes, cherry-pick into 202012 is clean.
201911 I don't think issues exists there, it is a pretty different code base.

liat-grozovik:
@qiluo-msft or @kcudnik have no permission to run tests on this repo. Can you please retrigger?

kcudnik:
something is wrong with this tests, they get canceled, not sure whats happening on this pipeline

qiluo-msft:
It always timeout after 180 min. So something is hanging at "Install dependencies" step.

PR #854, opened 2021-07-04
[syncd] bulk OID remove requires RID

qbdwlr:
Syncd::processBulkOidRemove() is incorrectly passing objectVids.data() to m_vendorSai->bulkRemove() when it should be passing objectRids.data(). 
Only RID should be passed to m_vendorSai calls. 
Issue can be seen when SAI bulk APIs are enabled in syncd and the following sequence of iproute2 commands are executed:
"ip route add \<ip-prefix> nexthop via inet \<ip-gateway-1> dev \<port-1> nexthop via inet \<ip-gateway-2> dev \<port-2>"
"ip route del \<ip-prefix> nexthop via inet \<ip-gateway-1> dev \<port-1> nexthop via inet \<ip-gateway-2> dev \<port-2>"

### DIFF
--- a/syncd/Syncd.cpp
+++ b/syncd/Syncd.cpp
@@ -1495,7 +1495,7 @@ sai_status_t Syncd::processBulkOidRemove(
     status = m_vendorSai->bulkRemove(
                                 objectType,
                                 (uint32_t)object_count,
-                                objectVids.data(),
+                                objectRids.data(),
                                 mode,
                                 statuses.data());
 


kcudnik:
thanks for catching this bug !

this is interesting, since we have test case that is testing this scenario in tests/BCM56850.pl test_bulk_object, there is function in file /tests/BCM56850/bulk_object.rec
```
2017-06-14.01:55:56.555975|R|SAI_OBJECT_TYPE_NEXT_HOP_GROUP||oid:0x5000000000005||oid:0x5000000000006||oid:0x5000000000007
```
and that should fail and syncd should crash in this case, interesting that this is not happening, ahh, i think i know what's happening, it's falling back to 1 by 1 instead of executing bulk, here is fallback line: https://github.com/Azure/sonic-sairedis/blob/master/syncd/Syncd.cpp#L1579

Ok, i run the tests, and it actually not executes this, since "-l" param is missing in tests, so actual bulk code is not executed at all, but even then if it's executed, then if it fails it fallbacks, and it succeed

kcudnik (approved):

prsunny:
@kcudnik , looks like this needs to be cherry-picked to 202012.. can you please confirm?

kcudnik:
yes

PR #858, opened 2021-07-09
[syncd][bcm] Start syncd by passing context configuration file

mlok-nokia:
In multi-asic BCM switch, there are multiple swss and syncd dockers. Orchagent process in each swss is started with a different hwinfo(asic_id). This PR is to ensure that context configuration is passed in syncd which has the hwinfo information with which each orchagent is started.
This change is similar to done for VS target

### DIFF
--- a/syncd/scripts/syncd_init_common.sh
+++ b/syncd/scripts/syncd_init_common.sh
@@ -163,7 +163,7 @@ config_syncd_bcm()
     fi
     
     if [ -f "$HWSKU_DIR/context_config.json" ]; then 
-        CMD_ARGS+=" -x $HWSKU_DIR/context_config.json"
+        CMD_ARGS+=" -x $HWSKU_DIR/context_config.json -g 0"
     fi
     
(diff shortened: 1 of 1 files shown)


kcudnik (dismissed):

syncd/scripts/syncd_init_common.sh
```diff
@@ -161,7 +161,11 @@ config_syncd_bcm()
       fi
 
     fi
-
+    
+    if [ -f "$HWSKU_DIR/context_config.json" ]; then 
+        CMD_ARGS+=" -x $HWSKU_DIR/context_config.json"
```
kcudnik, line 166: when context config is passed, and there are multiple contexts defined, a "-g" parameter is needed to specify which context should be used, by default is "0" but if you have many syncds then it maybe different number

mlok-nokia: Added "-g" option with id 0 for now.

kcudnik (approved):

mlok-nokia:
@kcudnik @judyjoseph @minionatwork  Since this PR has been approval, please help to merge it to the master stream.

PR #864, opened 2021-07-16
[syncd] Clean up log messages and add performance timer for bulk route create

kcudnik:


syncd/Syncd.cpp
```diff
@@ -917,6 +917,10 @@ sai_status_t Syncd::processBulkCreateEntry(
                 entries[it].vr_id = m_translator->translateVidToRid(entries[it].vr_id);
             }
 
+            static PerformanceIntervalTimer timer("Syncd::processBulkCreateEntry(route_entry) CREATE");
```
qiluo-msft, line 920: >static [](http://example.com/codeflow?start=12&length=6)

Better to use class static member. There could be multiple Syncd instances, right?

kcudnik: per process, no, i will fix that on the road, so far here are many performance counters instantiated like here

lguohan (approved):

PR #875, opened 2021-07-30
[Draft]: Add MACsec filter state guard

Pterosaur:


vslib/inc/MACsecFilterStateGuard.h
```diff
@@ -0,0 +1,20 @@
...
+    public:
+        MACsecFilterStateGuard(
+            _Inout_ MACsecFilter::MACsecFilterState &guarded_state,
+            _In_ MACsecFilter::MACsecFilterState target_state);
+
+        ~MACsecFilterStateGuard();
+
+    private:
+        MACsecFilter::MACsecFilterState m_old_state;
+        MACsecFilter::MACsecFilterState &m_guarded_state;
```
kcudnik, line 18: why this is a reference ?

vslib/inc/MACsecFilterStateGuard.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include "MACsecFilter.h"
+
+namespace saivs
+{
+    class MACsecFilterStateGuard
+    {
+    public:
+        MACsecFilterStateGuard(
```
kcudnik, line 10: add empty line after public

vslib/inc/MACsecFilter.h
```diff
@@ -10,6 +10,12 @@ namespace saivs
         public TrafficFilter
     {
     public:
+        enum MACsecFilterState
```
kcudnik, line 13: add empty line after public

kcudnik: use typedef for defining enums ,for example similar to this: vslib/inc/Event.h

vslib/inc/MACsecFilter.h
```diff
@@ -10,6 +10,12 @@ namespace saivs
...
+        {
+            IDLE,
```
kcudnik, line 15: enum names should be prefixed as entire enum name like here: vslib/inc/Event.h

vslib/src/MACsecFilter.cpp
```diff
@@ -30,6 +32,13 @@ void MACsecFilter::enable_macsec_device(
     SWSS_LOG_ENTER();
 
     m_macsecDeviceEnable = enable;
+    // The function, MACsecFilter::execute(), may be running in another thread,
```
kcudnik, line 35: add empty line before comment

vslib/src/MACsecFilter.cpp
```diff
@@ -30,6 +32,13 @@ void MACsecFilter::enable_macsec_device(
...
+    // the macsec device enable state cannot be changed in the busy state.
+    // Because if the macsec device was deleted in the busy state,
+    // the error value of function, MACsecFilter::forward, will be returned
+    // to the caller of MACsecFilter::execute()
+    // and the caller thread will exit due to the error return value.
+    while(get_state() != MACsecFilter::MACsecFilterState::IDLE);
```
kcudnik, line 41: dont put ; after while statment in the same line, please make this proper { } and insert comment // loop etc

kcudnik (changes requested):

vslib/inc/MACsecFilter.h
```diff
@@ -10,6 +10,15 @@ namespace saivs
...
     {
     public:
+
+        typedef enum _MACsecFilterState
+        {
+            MACSEC_FILTER_STATE_IDLE,
+
+            MACSEC_FILTER_STATE_BUSY,
+
+        } _MACsecFilterState;
```
kcudnik, line 20: this one without underscore

vslib/inc/MACsecFilterStateGuard.h
```diff
@@ -0,0 +1,22 @@
...
+#include "MACsecFilter.h"
+
+namespace saivs
+{
+    class MACsecFilterStateGuard
+    {
+    public:
+
+        MACsecFilterStateGuard(
+            _Inout_ volatile MACsecFilter::MACsecFilterState &guarded_state,
```
kcudnik, line 12: seems like a overkill here, did you hit scenario that this was optimized in wrong way by compiler that you need volatile here?

vslib/src/MACsecFilterStateGuard.cpp
```diff
@@ -0,0 +1,23 @@
+#include "MACsecFilterStateGuard.h"
+
+#include <swss/logger.h>
+
+using namespace saivs;
+
+MACsecFilterStateGuard::MACsecFilterStateGuard(
+    _Inout_ volatile MACsecFilter::MACsecFilterState &guarded_state,
```
kcudnik, line 8: seems like a overkill here, did you hit scenario that this was optimized in wrong way by compiler that you need volatile here?

vslib/inc/MACsecFilterStateGuard.h
Pterosaur: Agree, it looks like overkill. I will remove the qualifier.

vslib/inc/MACsecFilterStateGuard.h
Pterosaur: I want to use the RAII mechanism to protect the state at
https://github.com/Azure/sonic-sairedis/blob/cac30e9f4705e88973929a5155a9e76ba23af815/vslib/src/MACsecFilter.cpp#L69
So it's a reference.

vslib/src/MACsecFilter.cpp
```diff
@@ -30,6 +32,17 @@ void MACsecFilter::enable_macsec_device(
...
 
     m_macsecDeviceEnable = enable;
+
+    // The function, execute(), may be running in another thread,
+    // the macsec device enable state cannot be changed in the busy state.
+    // Because if the macsec device was deleted in the busy state,
+    // the error value of function, forward, will be returned
+    // to the caller of execute()
+    // and the caller thread will exit due to the error return value.
+    while(get_state() == MACsecFilter::MACsecFilterState::MACSEC_FILTER_STATE_BUSY)
```
kcudnik, line 42: this is just empty loop since get_state return just state, so it's like a spin lock using 100% cpu, how long this while will potentially spin?

Pterosaur: It will only be locked during the below block once, if you don't like a pure spin lock using 100% cpu, I can add a tiny sleep about 1 ms.
https://github.com/Azure/sonic-sairedis/blob/276eb3506f3b93ce41436134448a241e9aa5187d/vslib/src/MACsecEgressFilter.cpp#L19-L50

kcudnik (approved):

vslib/MACsecFilterStateGuard.cpp
```diff
@@ -0,0 +1,23 @@
+#include "MACsecFilterStateGuard.h"
+
+#include <swss/logger.h>
+
+using namespace saivs;
+
+MACsecFilterStateGuard::MACsecFilterStateGuard(
```
kcudnik, line 7: could you also add unittests for this file ? in similar way like are for other?

unittest/vslib/TestMACsecFilterStateGuard.cpp
```diff
@@ -0,0 +1,18 @@
+#include "MACsecFilterStateGuard.h"
+
+#include <gtest/gtest.h>
+
+using namespace saivs;
+
```
kcudnik, line 6: you are missing update Makefile.am for tests

Pterosaur: Done

kcudnik (approved):

kcudnik:
could you resolve conflicts?

PR #877, opened 2021-08-02
Install sonic-buildimage libnl3 for AZP

qbdwlr:
What I did
AZP changes to install libnl3 from sonic-buildimage

Why I did it
Prevent sonic-sairedis AZP build failures in docker-sonic-vs.
fpmsyncd in sonic-swss now relies on modified version of libnl3 in sonic-buildimage with additional support for MPLS.
Any AZP build of docker-sonic-vs must now install libnl3 from sonic-buildimage build artifact rather than Debian released version of libnl3.

How I verified it
Unit-tests in sonic-swss/tests/test_mpls.py
System tests in sonic-mgmt

Details if related
Refer to PR:  https://github.com/Azure/sonic-swss/pull/1871

### DIFF
--- a/.azure-pipelines/build-swss-template.yml
+++ b/.azure-pipelines/build-swss-template.yml
@@ -50,10 +50,6 @@ jobs:
       sudo apt-get install -y libzmq5 libzmq3-dev
       sudo apt-get install -qq -y \
           libhiredis-dev \
-          libnl-3-dev \
-          libnl-genl-3-dev \
-          libnl-route-3-dev \
-          libnl-nf-3-dev \
           swig3.0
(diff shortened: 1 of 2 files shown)


kcudnik:
why you making this changes? please add description

kcudnik (approved):

qbdwlr:
@kcudnik @liat-grozovik Is it really the intention to build sonic-sairedis from 202106 with sonic-swss from master?  I would think this could lead to issues with incompatibility eventually.  Perhaps build-swss-template.yml in 202106 should be updated to use sonic-swss and sonic-buildimage artifacts from the appropriate branch?

PR #878, opened 2021-08-03
Fix flex counters logic of converting poll interval to seconds from MS

noaOrMlnx:
According to https://github.com/Azure/SONiC/blob/ec6d35dd2c28491bfade19cfee990fe612f1e5e9/doc/rates-and-utilization/Rates_and_utilization_HLD.md, counterpoll command gives polling interval in milliseconds. 
So when converting them to seconds to be supplied to lua script this should be divide by 1000. However, syncd multiplies it by 1000.

Changed the multiplication to none, and did the converting in lua script - https://github.com/Azure/sonic-swss/pull/1855
Fixed issue - https://github.com/Azure/sonic-buildimage/issues/8392

### DIFF
--- a/syncd/FlexCounter.cpp
+++ b/syncd/FlexCounter.cpp
@@ -933,7 +933,7 @@ void FlexCounter::addCounterPlugin(
             for (auto& sha: shaStrings)
             {
                 addRifCounterPlugin(sha);
-            } 
+            }
         }
         else if (field == BUFFER_POOL_PLUGIN_FIELD)
         {
@@ -1597,7 +1597,7 @@ void FlexCounter::runPlugins(
     {
         std::to_string(counters_db.getDbId()),
         COUNTERS_TABLE,
-        std::to_string(m_pollInterval * 1000)
+        std::to_string(m_pollInterval)
     };
 
     std::vector<std::string> portList;
--- a/lib/inc/Recorder.h
+++ b/lib/inc/Recorder.h
@@ -157,6 +157,9 @@ namespace sairedis
                     _In_ sai_status_t status,
                     _In_ const std::vector<swss::FieldValueTuple>& arguments);
 
+            void recordGenericResponse(
+                    _In_ sai_status_t status);
+
         public: // create ENTRY
 
             SAI_REDIS_RECORDER_DECLARE_RECORD_CREATE(fdb_entry);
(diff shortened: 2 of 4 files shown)


syncd/FlexCounter.cpp
```diff
@@ -1597,7 +1597,7 @@ void FlexCounter::runPlugins(
     {
         std::to_string(counters_db.getDbId()),
         COUNTERS_TABLE,
-        std::to_string(m_pollInterval * 1000)
+        std::to_string(m_pollInterval / 1000)
```
kcudnik, line 1600: m_pollInterval is set via orchagent with values in milliseconds (MS), currently this value was extended to microseconds, and now you changing this to seconds, so if m_pollInterval will be equal to let say 60ms or 100ms or 900ms, then your value in db will be set to zero (0) because of integer division, is that what you want ?

kcudnik (changes requested):

liat-grozovik:
@kcudnik could you please help review?

kcudnik:
> @kcudnik could you please help review?

i already posted my concerns in comment above, and still this is not addressed nor responded how this will be handled

syncd/FlexCounter.cpp
noaOrMlnx: Done - removed the multiplication and did the required multiplication in the lua script - https://github.com/Azure/sonic-swss/pull/1855

noaOrMlnx:
@kcudnik Hi Kamil,
the request was addressed and the code has been changed. 
appreciate if you'll take a look.
Thanks

kcudnik:
i still don't see consistency here, what value is set in OA, what pooliterval value is stored in db and what value  valie of pollinterval i passed to lua script

dgsudharsan:
> i still don't see consistency here, what value is set in OA, what pooliterval value is stored in db and what value valie of pollinterval i passed to lua script

Hi Kamil. The value set in CLI is in milliseconds which is again set as milliseconds in orchagent. This value is again passed to lua script in milliseconds. So end to end we are using milliseconds. The lua scripts are updated to treat the input arg in milliseconds which is here https://github.com/Azure/sonic-swss/pull/1855
Your initial concern was correct and we avoided converting milliseconds to seconds (since lua script expected seconds in arg). Now we have changed end to end and we do not convert anywhere and lua script rate calculation logic is updated considering poll interval as milliseconds.

kcudnik:
thanks for deep answer on that, no i get the flow

kcudnik (approved):

liat-grozovik:
@noaOrMlnx could you please review the failures?

noaOrMlnx:
@kcudnik could you please re-approve?

kcudnik (dismissed):

qiluo-msft:
The PR check "license/cla" seems have some issue. Could you commit an empty commit, to retrigger the check.
I checked https://cla.opensource.microsoft.com/Azure/sonic-sairedis?pullRequest=878, and it is not signed.

noaOrMlnx:
@qiluo-msft Done

qiluo-msft (approved):

qiluo-msft:
@kcudnik Could you review again, and merge after all checkers pass?

kcudnik (approved):

PR #881, opened 2021-08-03
Add code coverage to configure.ac and Makefile.am

kcudnik:


.azure-pipelines/build-template.yml
```diff
@@ -66,7 +66,9 @@ jobs:
         libnl-cli-3-dev \
         swig3.0 \
         libpython2.7-dev \
-        libzmq5 libzmq3-dev
+        libzmq5 \
+        libzmq3-dev \
+        autoconf-archive
```
qiluo-msft, line 71: >autoconf-archive [](http://example.com/codeflow?start=8&length=16)

As my experiment, this package is available in slave docker.

PR #885, opened 2021-08-07
Enhancing VS support for running platform specific tests in VXLAN EVPN scenario

dgsudharsan:
For VxLAN EVPN VS testing where certain vendors support only P2MP tunnels, we needed to have a mechanism in the vs to identify the platform on which the test is running and thereby making returning the desired SAI capability.
When fake_platform is mellanox, the attribute SAI_TUNNEL_ATTR_PEER_MODE value capability will not include SAI_TUNNEL_PEER_MODE_P2P to simulate the tests written for mellanox based platforms.


vslib/VirtualSwitchSaiInterface.cpp
```diff
@@ -831,6 +834,40 @@ sai_status_t VirtualSwitchSaiInterface::queryAattributeEnumValuesCapability(
...
+
+        enum_values_capability->count = 1;
+        enum_values_capability->list[0] = SAI_TUNNEL_PEER_MODE_P2MP;
+        if (platform != "mellanox")
+        {
+            enum_values_capability->count++;
+            enum_values_capability->list[1] = SAI_TUNNEL_PEER_MODE_P2P;
+        }
+        return SAI_STATUS_SUCCESS;
+    }
```
kcudnik, line 870: if you need some special capability for mellanox, then it should be implemented in mlnx switch, not in generic claas

vslib/VirtualSwitchSaiInterface.cpp
```diff
@@ -786,6 +786,9 @@ sai_status_t VirtualSwitchSaiInterface::queryAattributeEnumValuesCapability(
 {
     SWSS_LOG_ENTER();
 
+    char *env_platform = std::getenv("fake_platform");
+    std::string platform = (env_platform != NULL) ? std::string(env_platform): "";
```
kcudnik, line 790: this is really ugly, this should depend on vs switch

kcudnik: take a look at SwitchMLNX2700.cpp

vslib/VirtualSwitchSaiInterface.cpp
kcudnik: take a look at SwitchMLNX2700.cpp

kcudnik (changes requested):
take a look at SwitchMLNX2700.cpp

PR #886, opened 2021-08-07
Added Flex Counters support for tunnel counters

dgsudharsan:
Added Flex counter support for  tunnel counters and rates implementation


syncd/FlexCounter.cpp
```diff
@@ -114,6 +114,14 @@ FlexCounter::MACsecSAAttrIds::MACsecSAAttrIds(
     // empty intentionally
 }
 
+FlexCounter::TunnelCounterIds::TunnelCounterIds(
+        _In_ sai_object_id_t tunnel,
+        _In_ const std::vector<sai_tunnel_stat_t> &tunnelIds):
+    tunnelId(tunnel), tunnelCounterIds(tunnelIds)
```
kcudnik, line 120: put each variable in separate line, also name members with m_ prefix

syncd/FlexCounter.cpp
```diff
@@ -114,6 +114,14 @@ FlexCounter::MACsecSAAttrIds::MACsecSAAttrIds(
...
+{
+    SWSS_LOG_ENTER();
+}
```
kcudnik, line 123: add // empty
to notify that this is empty intentionally

kcudnik (changes requested):

syncd/FlexCounter.cpp
```diff
@@ -114,6 +114,14 @@ FlexCounter::MACsecSAAttrIds::MACsecSAAttrIds(
     // empty intentionally
 }
 
+FlexCounter::TunnelCounterIds::TunnelCounterIds(
+        _In_ sai_object_id_t tunnel,
```
kcudnik, line 118: tunnel vid or rid?

syncd/FlexCounter.cpp
dgsudharsan: Done

syncd/FlexCounter.cpp
dgsudharsan: Done

syncd/FlexCounter.cpp
dgsudharsan: tunnelRid. I modified variable name to reflect it.

kcudnik:
seems like build its failing because it requires your changes from swss-common

dgsudharsan:
@kcudnik The pipeline is passing now. Could you please help merging the PR?

kcudnik (approved):

PR #887, opened 2021-08-09
[meta] Simplify compilation for saimetadata.c

kcudnik:
All those flags are just warnings, and were already checked in SAI repo during PR validation.


kcudnik:
error not related to PR

PR #888, opened 2021-08-09
Move sairedis to c++14 standard

kcudnik:


kcudnik:
error not related to PR

PR #889, opened 2021-08-09
[tests] Add vssyncd it's own main.cpp

kcudnik:
Main motivation is to avoid linking the same files across different directories to simplify future coverage data collection.


kcudnik:
error not related to PR

PR #890, opened 2021-08-09
[pyext] Add built sources list to makefiles

kcudnik:
Note which source files are auto generated for python modules


kcudnik:
error not related to PR

PR #894, opened 2021-08-13
Fix cast align warning

Pterosaur:
Fix warning: cast from 'unsigned char*' to 'tpacket_auxdata*' increases required alignment of target type [-Wcast-align]


vslib/src/TrafficForwarder.cpp
```diff
@@ -37,7 +37,7 @@ void TrafficForwarder::addVlanTag(
         if (cmsg->cmsg_level != SOL_PACKET || cmsg->cmsg_type != PACKET_AUXDATA)
             continue;
 
-        struct tpacket_auxdata* aux = (struct tpacket_auxdata*)CMSG_DATA(cmsg);
+        struct tpacket_auxdata* aux = (struct tpacket_auxdata*)(void *)CMSG_DATA(cmsg);
```
kcudnik, line 40: seems like this is not the proper solution here, issue is about cast align, which void* probably will not solve, but i didnt tested that on gcc10, i had similar issue with armhf which is different platform and have different memory requirements 
which i fixed this way: https://github.com/Azure/sonic-sairedis/pull/738

Pterosaur: OK, I will follow your solution.
Actually, this solution comes from the source code of OVS (https://mail.openvswitch.org/pipermail/ovs-dev/2014-January/279545.html)

kcudnik: Hey, i tested this 
on x64 gcc10: https://godbolt.org/z/rv4fKxTMb (no errors in any situation)
but on arm32: https://godbolt.org/z/T73h7v5jd you get original warning, but void seems to fix it, you can get back to that void* if you want

Pterosaur: Sure, I reset it.

kcudnik (approved):

PR #901, opened 2021-08-21
[tests] [vslib] Add bunch of unittests for vslib

kcudnik:
Add a lot of unit test for vslib part, fix small errors in vslib for not null checking, bring code coverage over 60%


kcudnik:
arm build succeeded, but publish artefacts already exists

PR #904, opened 2021-08-24
[tests] Adding tests to meta lib and also moving legacy tests to gtest evn

kcudnik:


kcudnik:
Errors on arm64 because syslog artifacts were already published

kcudnik:
replaced by https://github.com/Azure/sonic-sairedis/pull/913

PR #910, opened 2021-08-27
[syncd] Add stack dumper for better crash reports

abanu-ms:
* Register stack dumper for syncd crashes

Signed-off-by: Alexandru Banu <v-albanu@microsoft.com>


abanu-ms:
> stack is already available in code dump with all the nice features of gdb

I did this improvement a while back. I'll do some re-testing without this fix in order to remember what exactly it was fixing.

abanu-ms:
Okay, so the purpose of adding `-rdynamic` flag was to demangle the symbols in a core dump when the crash is generated by the SAI library. If a crash happens in the SAI library, say an assert, the core dump will decode only the symbols of the SAI library, but not the ones from syncd. See the screenshots below:

Without `-rdyanmic` flag:
MicrosoftTeams-image

With the `-rdynamic` flag:
MicrosoftTeams-image (1)

Actually, I'd rather add that flag for all builds, not just the debug ones as I think this would be useful in a production environment as well.

As for the stack dumper, that was in order to get the stack dumped in syslog as well in order for someone to have easy access to the crash's stack instead of going through the work necessary of opening a core. If you don't want/need that kind of stuff, I'm happy to remove that part and keep just the `-rdynamic` flag change.

kcudnik:
first stack with ?? seem to me like missing dbg package for resolving symbols 
if you want just add permanently -rdynamic flag to syncd, but do not add any custom stack printing

abanu-ms:
@kcudnik can you re-review and approve if happy? Thanks!

kcudnik (approved):

abanu-ms:
Can someone also merge this PR? It doesn't look like I have the access to do so.

PR #930, opened 2021-09-15
[sairedis] Fail get/clear api on objects created in init view mode

kcudnik:
Add also nice syslog message why api failed. This will prevent going to syncd and fail at translateVidToRid, it will now return failure on sairedis orchagent level.


lguohan (approved):

kcudnik:
seems like  this PR is braking 28 tests, but i can't figure why since logs are not published from failed tests :/

kcudnik:
this PR needs fix, since syncd on first init view goes directly to appl view, but OA thinks it's actually init view. probably fix for that would be to move this change to syncd side

PR #931, opened 2021-09-17
[vslib]: Skip to create tap interface if the KVM interface isn't enabled

Pterosaur:
This PR https://github.com/Azure/sonic-sairedis/pull/923/files fixed the return value of `ifup`, but it introduces a new bug. If the SONiC KVM didn't enable full 32 ports.
Here is the qemu xml example that only enables two interfaces.
image

The SWSS will exit with an unexpected error as follow
image
image

So, I add a new function `ifexists` to check whether the target interface of KVM has been enabled, I will skip to create the tap interface if the target interface wasn't enabled.

### DIFF
--- a/vslib/SwitchStateBase.h
+++ b/vslib/SwitchStateBase.h
@@ -455,6 +455,8 @@ namespace saivs
                     _In_ const char*name,
                     _In_ int mtu);
 
+            bool ifexists(const char *dev);
+
             int ifup(
                     _In_ const char *dev,
                     _In_ sai_object_id_t port_id,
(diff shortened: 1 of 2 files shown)


lguohan:
i do not feel this is the right approach. sai should return failure, if the host interface cannot be created. A better approach is to make sure swss won't create such host interface.

please check example in sonic-buildimage as below.

```
platform/vs/docker-sonic-vs/start.sh:# filter available front panel ports in lanemap.ini
platform/vs/docker-sonic-vs/start.sh:[ -f lanemap.ini.orig ] || cp lanemap.ini lanemap.ini.orig
platform/vs/docker-sonic-vs/start.sh:    grep ^$p: lanemap.ini.orig
platform/vs/docker-sonic-vs/start.sh:done > lanemap.ini
```

PR #934, opened 2021-09-20
[ci]: Support code diff coverage

xumia:
Support code diff coverage
1. Support to add the pull request checks for Code diff coverage in GitHub.
2. Support to display the code diff coverage html report in Azure Pipelines.

### DIFF
--- a/.azure-pipelines/build-template.yml
+++ b/.azure-pipelines/build-template.yml
@@ -43,8 +43,8 @@ jobs:
   timeoutInMinutes: ${{ parameters.timeout }}
   variables:
     DIFF_COVER_CHECK_THRESHOLD: 0
-    ${{ if ne(parameters.run_unit_test, false) }}:
-      DIFF_COVER_DISABLE: 'false'
+    ${{ if eq(parameters.run_unit_test, true) }}:
+      DIFF_COVER_ENABLE: 'true'
 
   pool:
     ${{ if ne(parameters.pool, 'default') }}:
--- a/.azure-pipelines/build-template.yml
+++ b/.azure-pipelines/build-template.yml
@@ -41,6 +41,10 @@ jobs:
 - job:
   displayName: ${{ parameters.arch }}
   timeoutInMinutes: ${{ parameters.timeout }}
+  variables:
+    DIFF_COVER_CHECK_THRESHOLD: 0
+    ${{ if ne(parameters.run_unit_test, false) }}:
+      DIFF_COVER_DISABLE: 'false'
 
   pool:
     ${{ if ne(parameters.pool, 'default') }}:


.azure-pipelines/build-template.yml
```diff
@@ -41,6 +41,10 @@ jobs:
 - job:
   displayName: ${{ parameters.arch }}
   timeoutInMinutes: ${{ parameters.timeout }}
+  variables:
+    DIFF_COVER_CHECK_THRESHOLD: 0
```
qiluo-msft, line 45: Do you have some document about the two variables? What is the value range, and how do we choose the value?

kcudnik:
how this entire feature should work ? is already framework for support that build in azpipeline and sairedis azpipeline configuration?

.azure-pipelines/build-template.yml
xumia: @qiluo-msft , add the doc https://github.com/Azure/sonic-build-web/tree/main/code-diff-coverage

xumia:
@kcudnik , I added an azure devops extension to support it, see https://github.com/Azure/sonic-build-web/tree/main/code-diff-coverage

.azure-pipelines/build-template.yml
```diff
@@ -41,6 +41,10 @@ jobs:
...
+    ${{ if ne(parameters.run_unit_test, false) }}:
+      DIFF_COVER_DISABLE: 'false'
```
qiluo-msft, line 47: >DIFF_COVER_DISABLE [](http://example.com/codeflow?start=6&length=18)

Negative name is harder to read than postiive one. Suggest `DIFF_COVER_ENABLED` with default value `true`. Not sure how difficult for you to change code and redeploy.

qiluo-msft (approved):

.azure-pipelines/build-template.yml
xumia: Maybe some day we can change the default value to false, so no one need to set the value. The feature is in preview status, we do not want to have any impact, so disable it by default.

qiluo-msft: I am not asking to change the semantics of default value, only the name. The negative default value is actually same as your intention.

kcudnik (approved):

.azure-pipelines/build-template.yml
xumia: Fixed, change to use DIFF_COVER_ENABLE=true.

kcudnik (approved):

qiluo-msft (approved):

xumia:
@kcudnik , can we merge the code, there is one failure check, looks like it is not relative to the change.

kcudnik:
your diff coverage plugin broken full coverage stats file by file:
https://dev.azure.com/mssonic/build/_build/results?buildId=39429&view=results

clicking you can't see any more how much covered is each added file , please revert your change or make it possible to view both results

PR #938, opened 2021-09-22
[vslib]: Remove invalid lane when create ports

Pterosaur:
To make the virtual SAI supports partial ports enabled.

Here is the qemu xml example that only enables two interfaces.
image

The SWSS will exit with an unexpected error as follow
image
image

I check the availability of eth-x ports at creating port.

### DIFF
--- a/vslib/SwitchStateBase.cpp
+++ b/vslib/SwitchStateBase.cpp
@@ -1123,43 +1123,11 @@ sai_status_t SwitchStateBase::create_ports()
 
     auto lanesVector = map->getLaneVector();
 
-    if (m_useTapDevice)
+    if (m_switchConfig->m_useTapDevice)
     {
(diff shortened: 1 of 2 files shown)


vslib/SwitchStateBase.cpp
```diff
@@ -1120,7 +1121,46 @@ sai_status_t SwitchStateBase::create_ports()
         return SAI_STATUS_FAILURE;
     }
 
-    auto& lanesVector = map->getLaneVector();
+    auto lanesVector = map->getLaneVector();
+
+    if (m_useTapDevice)
+    {
+        SWSS_LOG_DEBUG("Check available lane");
```
kcudnik, line 1128: could you make this code as separate function?

kcudnik (approved):

PR #939, opened 2021-09-22
[gearbox] startup gbsyncd in sync mode in vSonic

jimmyzhai:
Need to startup gbsyncd in sync mode, since orchagent is using sync mode.

### DIFF
--- a/syncd/scripts/gbsyncdmgrd
+++ b/syncd/scripts/gbsyncdmgrd
@@ -33,7 +33,7 @@ def physyncd_spawn(gearbox_config):
     proc_list = []
 
     for i, phy in enumerate(gearbox_config['phys'], 1):
-        cmd = '/usr/bin/syncd -p /etc/sai.d/pai.profile -x /usr/share/sonic/hwsku/context_config.json -g {}"'.format(i)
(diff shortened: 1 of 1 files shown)


syncd/scripts/gbsyncdmgrd
```diff
@@ -33,7 +33,7 @@ def physyncd_spawn(gearbox_config):
     proc_list = []
 
     for i, phy in enumerate(gearbox_config['phys'], 1):
-        cmd = '/usr/bin/syncd -p /etc/sai.d/pai.profile -x /usr/share/sonic/hwsku/context_config.json -g {}"'.format(i)
+        cmd = '/usr/bin/syncd -s -p /etc/sai.d/pai.profile -x /usr/share/sonic/hwsku/context_config.json -g {}'.format(i)
```
akokhan, line 36: I think `-s` is deprecated and `-z` should be used instead.
https://github.com/Azure/sonic-sairedis/blob/40e9c745aa3cfb75dbff91973d2b6dff1e3b61fb/syncd/CommandLineOptionsParser.cpp#L39

jimmyzhai: I see. The option `-s` is equative to `-z redis_sync`, which is still used in syncd and orchagent command line parameter.

lguohan (approved):

PR #954, opened 2021-10-11
[FlexCounter] Add trap flow counter support

Junchao-Mellanox:
**What I did**

Add trap flow counter support. See HLD: https://github.com/Azure/SONiC/pull/858

**Why I did it**

Flow counters are usually used for debugging, troubleshooting and performance enhancement processes. Host interface trap counter can get number of received traps per Trap ID.

**How I verified it**

Manual test/VS test/sonic mgmt test

**Details if related**

### DIFF
--- a/meta/SaiSerialize.cpp
+++ b/meta/SaiSerialize.cpp
@@ -1007,6 +1007,14 @@ std::string sai_serialize_tunnel_stat(
     return sai_serialize_enum(counter, &sai_metadata_enum_sai_tunnel_stat_t);
 }
 
+std::string sai_serialize_counter_stat(
+        _In_ const sai_counter_stat_t counter)
+{
+    SWSS_LOG_ENTER();
+
+    return sai_serialize_enum(counter, &sai_metadata_enum_sai_counter_stat_t);
+}
+
 std::string sai_serialize_queue_attr(
         _In_ const sai_queue_attr_t attr)
 {
--- a/meta/sai_serialize.h
+++ b/meta/sai_serialize.h
@@ -122,6 +122,9 @@ std::string sai_serialize_buffer_pool_stat(
 std::string sai_serialize_tunnel_stat(
         _In_ const sai_tunnel_stat_t counter);
 
+std::string sai_serialize_counter_stat(
+        _In_ const sai_counter_stat_t counter);
+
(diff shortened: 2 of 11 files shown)


kcudnik (approved):

kcudnik:
can  you try to add empty commit just to trigger build, i restarted it from github, but no clue why it's timing out

kcudnik:
you need to have diff code coverage passed:
Coverage: 0%
Threshold: 50%

Junchao-Mellanox:
Hi @kcudnik , I don't see any unit test for FlexCounter, could you tell where can I find the test case?

kcudnik:
you would need to start create them, not all components have started to have unittests yet, you could follow schema in unittests/meta in unittests/syncd directory

kcudnik:
please resolve conflicts after merging ACL counters

Junchao-Mellanox:
Hi @kcudnik , the failure "Cancelled after 180m" is not caused by my change. I see that the step "Install dependencies" takes almost 3hrs.

kcudnik:
why you need mockable interface at all? and not use "-fno-access-control" to have access to private methods

Junchao-Mellanox:
> why you need mockable interface at all? and not use "-fno-access-control" to have access to private methods and also please dont put code in header files

Hi @kcudnik , 

1. Why we need mockable interface? I don't understand here. Say we have SaiInterface::func, we want to mock it in test case A and test case B, and we want it behave differently in A and B. How can we achieve this without mockable interface?
2. I will move the code from header to cpp file.

kcudnik:
in my understanding Mock class should be only wrapper to expose methods/variables for which we don't have have access since they are ether protected or private, and testing them directly would be much simpler. Behavior should be the same since mock should be created for specific class ?, it could be for base interface also i guess

Junchao-Mellanox:
Maybe "-fno-access-control" can do the job. But MockableInterface provide following capability:
1. Mock any SaiInterface members in any test case with different implementation. E.g. the function getStatExt can fill different counter value if need.
2. It is easy to mock negative case. E.g, the function create can return SAI_INVALID_PARAMTER if need.
3. The implementation of the mocked function can be change dynamically without affecting any other test case.

kcudnik (approved):

PR #955, opened 2021-10-19
Updating SAI to v1.9.1

vmittal-msft:
Updating sonic-sairedis to point to SAI header 1.9.1.


kcudnik:
seems like there is some issue in swss building pipeline:
```
cp: cannot stat 'sonic-swss/tests/log': No such file or directory
##[error]Bash exited with code '1'.
```

PR #973, opened 2021-11-23
[lgtm] Add gmock libs to lgtm

kcudnik:


gechiang (approved):

qiluo-msft:
This is to fix lgtm checker, so bypass the failed checkers.

qiluo-msft (approved):

PR #984, opened 2021-12-07
Fix check_object_default_state on objlist. 

baxia-lan:
Return error only when non SAI_NULL_OBJECT_ID present in objlist.

### DIFF
--- a/vslib/SwitchStateBase.cpp
+++ b/vslib/SwitchStateBase.cpp
@@ -2721,16 +2721,23 @@ bool SwitchStateBase::check_object_default_state(
         {
             if (objlist.size())
             {
-                SWSS_LOG_ERROR("expected empty list on %s on %s, contents:",
-                        meta->attridname,
-                        sai_serialize_object_id(object_id).c_str());
-
+                bool all_null_oid = true;
                 for (auto oid: objlist)
                 {
-                    SWSS_LOG_ERROR(" - oid: %s", sai_serialize_object_id(oid).c_str());
+                    if (oid != SAI_NULL_OBJECT_ID)
+                    {
+                        all_null_oid = false;
+                        SWSS_LOG_ERROR(" - oid: %s", sai_serialize_object_id(oid).c_str());
+                    }
                 }
 
-                return false;
+                if (!all_null_oid)
+                {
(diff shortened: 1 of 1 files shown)


kcudnik:
i don't remember where there would be use case for a object_list to contain null object id's, could you explain more what exact issue/scenario are you addressing here ?

baxia-lan:
> i don't remember where there would be use case for a object_list to contain null object id's, could you explain more what exact issue/scenario are you addressing here ?

Updated issue https://github.com/Azure/sonic-sairedis/issues/983#issue-1073850741 describing the scenario.
When deleting a port object, even though a port attribute with objlist type has all SAI_NULL_OBJECT_ID, due to code [here](https://github.com/Azure/sonic-sairedis/blob/master/vslib/SwitchStateBase.cpp#L2664-L2677), the length of the `std::vector<sai_object_id_t> objlist` is not 0.

kcudnik:
> Updated issue [#983 (comment)](https://github.com/Azure/sonic-sairedis/issues/983#issue-1073850741) describing the scenario.
> When deleting a port object, even though a port attribute with objlist type has all SAI_NULL_OBJECT_ID, due to code [here](https://github.com/Azure/sonic-sairedis/blob/master/vslib/SwitchStateBase.cpp#L2664-L2677), the length of the `std::vector<sai_object_id_t> objlist` is not 0.

not sure if i follow this exactly, port attribute with objlits should not have null objects list (not null) it should be empty (in SAI vendor), code you pointes is using MAX list size since after that there is a GET api called, and objlist object should be resized to ad list.count received from SAI (which is not, which is a bug here which you are tryin to address? :P)

kcudnik:
it should be more like this:
```
objlist.resize(attr.value.objlist.count); // this should be correct count returned
 if (objlist.size())
...
```

baxia-lan:
> it should be more like this:
> 
> ```
> objlist.resize(attr.value.objlist.count); // this should be correct count returned
>  if (objlist.size())
> ...
> ```

Done

kcudnik (approved):

baxia-lan:
Hi @kcudnik  I don't have write access to the repo, can you please help me merge the PR? Thanks!

kcudnik:
please add some unittests to cover at least 50% of the code, we have policy now to enforce that

PR #985, opened 2021-12-14
[syncd][vslib] Keep new warm boot discovered SERDES objects

kcudnik:
when updating 3.5 to 4.3 brcm SAI, new objects are discovered - PORT_SERDES, and each one is assigned to existing port, by default, since not present in temporary view CL will remove them, but we want to keep them for user to explicitly make operations if required


vslib/SwitchStateBase.cpp
```diff
@@ -2635,6 +2681,9 @@ bool SwitchStateBase::check_object_default_state(
         if (!meta->isoidattribute)
             continue;
 
+        if (meta->ismandatoryoncreate && meta->iscreateonly)
```
kcudnik, line 2684: could be explicit SAI_OBJECT_TYPE_PORT_SERDES and SAI_PORT_SERDES_ATTR_PORT

qiluo-msft:

kcudnik:
here is fixed PR for 202012 branch: https://github.com/Azure/sonic-sairedis/pull/1011

PR #987, opened 2021-12-20
SAI NAT aging notification

arvbb:
The sairedis changes related to the SAI here: [https://github.com/opencomputeproject/SAI/pull/1365](https://github.com/opencomputeproject/SAI/pull/1365)


arlakshm:
@AkhileshSamineni , can you review this PR

kperumalbfn:
@arlakshm @AkhileshSamineni could you please review the PR.

arlakshm:
Can you please add Unit tests

kcudnik:
please also satisfy code coverage

kcudnik (approved):

arvbb:
> Can you please add Unit tests

Hi @arlakshm, added,.. thanks

meta/Meta.cpp
```diff
@@ -7594,6 +7594,52 @@ void Meta::meta_sai_on_fdb_event(
...
+        _In_ const sai_nat_event_notification_data_t& data)
+{
+    SWSS_LOG_ENTER();
+
+    const sai_object_meta_key_t meta_key_nat = { .objecttype = SAI_OBJECT_TYPE_NAT_ENTRY, .objectkey = { .key = { .nat_entry = data.nat_entry } } };
+
+    switch (data.event_type)
+    {
+        case SAI_NAT_EVENT_AGED:
+
```
kcudnik, line 7607: should nat entry be removed here when aged event was received ? similar like in fdb entry aged ?

arvbb: Hi Kamil, thanks for reviewing.
Yes, nat entry gets removed when aged event is processed in SWSS. Natorch triggers deletion of the kernel conntrack entry for the cleanup. This is similar to the existing case where platforms that do not have NAT aging will rely on hitbit timer for cleanup.

meta/SaiSerialize.cpp
```diff
@@ -1940,6 +1940,28 @@ static json sai_serialize_json_fdb_event_notification_data(
...
+}
+
+static json sai_serialize_json_nat_event_notification_data(
+        _In_ const sai_nat_event_notification_data_t& nat_event)
+{
+    SWSS_LOG_ENTER();
+
+    json j;
+
+    j["nat_event"] = sai_serialize_nat_event(nat_event.event_type);
```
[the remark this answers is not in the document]
arvbb: I was following similar to fdb to be consistent with naming.
j["fdb_event"] = sai_serialize_fdb_event(fdb_event.event_type);

If this is not a big deal, is it okay to keep the same to avoid the testing again.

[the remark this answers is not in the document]
kcudnik: ok

meta/Meta.cpp
kcudnik: yes, but i mean should it be removed here from Meta database, since if it will be removed automatically then meta database will have inconsistent state with actual device

kcudnik: from your pictures seems like nat entry will be deleted after event explicitly by natorch, so no code will be required here, but i would prefer to put at least comment in this part of code why there is no explicit delete

this behavior then is different than fdb, so SAI is just generating NAT aged event but entry is still there preserved ? then this would need to be explicitly added in sai documentation repo so all vendors would have the same view on this

arvbb: Added comments

meta/SaiSerialize.cpp
```diff
@@ -3906,6 +3961,37 @@ void sai_deserialize_fdb_event_ntf(
     *fdb_event = data;
 }
 
+static void sai_deserialize_json_nat_event_notification_data(
+        _In_ const json& j,
+        _Out_ sai_nat_event_notification_data_t& nat)
+{
+    SWSS_LOG_ENTER();
+
+    sai_deserialize_nat_event(j["nat_event"], nat.event_type);
```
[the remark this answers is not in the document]
arvbb: Similar to fdb, resolving.

kcudnik (approved):

PR #989, opened 2021-12-27
Added Support for enum query capability of Nexthop Group Type.

abdosi:
What/Why I did:
Added Support for enum query capability of Nexthop Group Type.
This is needed for DVS testing of Ordered ECMP Feature.
Design Doc: https://github.com/Azure/SONiC/pull/896

Signed-off-by: Abhishek Dosi <abdosi@microsoft.com>

### DIFF
--- a/unittest/vslib/TestSwitchBCM56850.cpp
+++ b/unittest/vslib/TestSwitchBCM56850.cpp
@@ -418,3 +418,64 @@ TEST(SwitchBCM56850, test_vlan_flood_capability)
 
     EXPECT_EQ(flood_types_found, 3);
 }
+
+TEST(SwitchBCM56850, test_nexthop_group_type_enum_values_capability)
+{
+    auto sc = std::make_shared<SwitchConfig>(0, "");
+    auto signal = std::make_shared<Signal>();
+    auto eventQueue = std::make_shared<EventQueue>(signal);
+
+    sc->m_saiSwitchType = SAI_SWITCH_TYPE_NPU;
+    sc->m_switchType = SAI_VS_SWITCH_TYPE_BCM56850;
+    sc->m_bootType = SAI_VS_BOOT_TYPE_COLD;
+    sc->m_useTapDevice = false;
+    sc->m_laneMap = LaneMap::getDefaultLaneMap(0);
+    sc->m_eventQueue = eventQueue;
+
+    auto scc = std::make_shared<SwitchConfigContainer>();
+
+    scc->insert(sc);
+
+    SwitchBCM56850 sw(
+            0x2100000000,
+            std::make_shared<RealObjectIdManager>(0, scc),
+            sc);
+
+    sai_s32_list_t enum_val_cap;
+    int32_t list[5];
+    memset(list, 0, sizeof(list));
+
+    enum_val_cap.count = 4;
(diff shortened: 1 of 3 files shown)


abdosi:
@shi-su i see vnet orch test case 7/8/9/11 are failing in the PR job. Can you please check ?

cc @prsunny

shi-su:
> @shi-su i see vnet orch test case 7/8/9/11 are failing in the PR job. Can you please check ? I ran job again and it is still failing. Looks to be consistent failure.
> 
> cc @prsunny

Ack checking

shi-su:
> @shi-su i see vnet orch test case 7/8/9/11 are failing in the PR job. Can you please check ? I ran job again and it is still failing. Looks to be consistent failure.
> 
> cc @prsunny

This is a very weird failure. It seems that the same nexthop group gets recreated when a new route is pointing to it. This is an unexpected behavior since the nexthop group should be reused. I did not find any clue why the same nexthop group could get created multiple times. Interestingly, this failure only seems to happen in sairedis pipeline vs test while it is not observed in swss pipeline. On the other hand, I do not think there is anything in sairedis that may affect the behavior. Still investigating.

shi-su:
I tried to overwrite swss deb package with the one built from swss pipeline. The failure does not happen anymore after doing so. This seems to indicate that the swss package built from swss and sairedis pipelines are somewhat different. Any idea about what might cause such a difference?

abdosi:
> > @shi-su i see vnet orch test case 7/8/9/11 are failing in the PR job. Can you please check ? I ran job again and it is still failing. Looks to be consistent failure.
> > cc @prsunny
> 
> This is a very weird failure. It seems that the same nexthop group gets recreated when a new route is pointing to it. This is an unexpected behavior since the nexthop group should be reused. I did not find any clue why the same nexthop group could get created multiple times. Interestingly, this failure only seems to happen in sairedis pipeline vs test while it is not observed in swss pipeline. On the other hand, I do not think there is anything in sairedis that may affect the behavior. Still investigating.

@shi-su may be it is not binary based of your code . Can we put breakpoint and see the code flow when it is failure.

shi-su:
> 

Actually, passes and failures are both pretty consistent from my observation.

shi-su:
> > 
> 
> Actually, passes and failures are both pretty consistent from my observation.

The test cases constantly fail with the deb package from sairedis pipeline and they constantly pass with the package from swss pipeline..

shi-su:
I suspect the problem is because the variable `weight` here https://github.com/Azure/sonic-swss/blob/master/orchagent/nexthopkey.h#L22 is not initialized for nexthops in vnetorch. Created this PR https://github.com/Azure/sonic-swss/pull/2096 and let's check if it resolves the problem.

shi-su (approved):

shi-su (approved):

kcudnik (approved):

abdosi:
> I suspect the problem is because the variable `weight` here https://github.com/Azure/sonic-swss/blob/master/orchagent/nexthopkey.h#L22 is not initialized for nexthops in vnetorch. Created this PR [Azure/sonic-swss#2096](https://github.com/Azure/sonic-swss/pull/2096) and let's check if it resolves the problem.

@shi-su thanks the change worked..

PR #994, opened 2022-01-19
Add to save sai objects which are created during onPostPortCreate

gord1306:
The ports which is without QUEUE related configurations still created related
SCHEDULER_GROUPs in the onPostPortCreate() during the syncd startup in some
platform - e.g. barefoot montara. But these created objects did NOT be recorded
in the `COLDVIDS` and `m_warmBootDiscoveredVids` set. While swss is perform
warmstart, the `compareViews()` would be called before apply the view, and it
would detect the differences between the current view and temp view. Then syncd
would remove these schedulers groups in the finalize.

This patch adds to record the objects which are created during onPostPortCreate
into the `m_warmBootDiscoveredVids` set.

###  Verified by 
1. Applied sonic-mgmt t0 minigraph
2. SONiC CLI command: `config warm_restart enable swss`
3. System CLI command: `service swss restart`
Verify the syslog is there any WARNING/ERR in `checkInternalObjects`

### DIFF
--- a/syncd/SaiSwitch.cpp
+++ b/syncd/SaiSwitch.cpp
@@ -1030,6 +1030,20 @@ void SaiSwitch::onPostPortCreate(
          */
 
         redisSetDummyAsicStateForRealObjectId(rid);
+
+	/*
+         * The ports which is without QUEUE related configurations still created
+         * related SCHEDULER_GROUPs in the onPostPortCreate() during the syncd
+         * startup in some platform - e.g. montara. But these created objects
+	 * did NOT be recorded in the `COLDVIDS` and `m_warmBootDiscoveredVids`
+         * set. While swss is perform warmstart, the `compareViews()` would be
+         * called before apply the view, and it would detect the differences
+         * between the current view and temp view. Then syncd would remove these
+         * schedulers groups in the finalize. Therefore, add to record these
+         * object into the m_warmBootDiscoveredVids
+         */
+        sai_object_id_t vid = m_translator->translateRidToVid(rid, m_switch_vid);
+        m_warmBootDiscoveredVids.insert(vid);
     }
 
     redisUpdatePortLaneMap(port_rid);


syncd/SaiSwitch.cpp
```diff
@@ -1030,6 +1030,20 @@ void SaiSwitch::onPostPortCreate(
...
+         * startup in some platform - e.g. montara. But these created objects
+	 * did NOT be recorded in the `COLDVIDS` and `m_warmBootDiscoveredVids`
+         * set. While swss is perform warmstart, the `compareViews()` would be
+         * called before apply the view, and it would detect the differences
+         * between the current view and temp view. Then syncd would remove these
+         * schedulers groups in the finalize. Therefore, add to record these
+         * object into the m_warmBootDiscoveredVids
+         */
+        sai_object_id_t vid = m_translator->translateRidToVid(rid, m_switch_vid);
+        m_warmBootDiscoveredVids.insert(vid);
```
kcudnik, line 1046: not sure why this is needed here, why m_warmBootDiscoveredVids is used at all, since onPostPortCreate is called AFTER cold boot or warm boot was performed already, so all discovered objects were already discovered during "create switch" at that phase. i dont quite understand description here, also you don't know whether onPostPortCreate which is called here is after cold boot or after warm boot, but yet  you always add vids to m_warmBootDiscoveredVids

this translation is not needed here, if you are missing some vids, then it means problem is somewhere else

kcudnik (changes requested):
please see comments

syncd/SaiSwitch.cpp
gord1306: Thanks for reply. The case that I encountered was to perform warmstart swss only, not syncd. And the onPostPortCreate() would not be called again. And it causes some vids miss that which are created during previous onPostPortCreate(), because there is no records for this vids.

PR #995, opened 2022-01-21
[pipeline] Download swss common artifact in a separated directory

jimmyzhai:
* Put the downloaded swss common artifact in a separated directory while build sairedis target. 
* The downloaded swss common artifact is in a sub-directory of sairedis artifact, then it avoids overriding
   the previous swss common packages.

### DIFF
--- a/.azure-pipelines/build-template.yml
+++ b/.azure-pipelines/build-template.yml
@@ -58,6 +58,9 @@ jobs:
     image: sonicdev-microsoft.azurecr.io:443/${{ parameters.sonic_slave }}:latest
 
   steps:
+  - checkout: self
+    clean: true
+    submodules: true
   - script: |
       sudo apt-get install -qq -y \
         qtbase5-dev \
@@ -106,8 +109,6 @@ jobs:
       sudo dpkg -i libswsscommon-dev_1.0.0_${{ parameters.arch }}.deb
     workingDirectory: $(Pipeline.Workspace)
     displayName: "Install sonic swss Common"
-  - checkout: self
-    submodules: true
   - script: |
       set -ex
       ./autogen.sh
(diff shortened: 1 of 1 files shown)


.azure-pipelines/build-template.yml
```diff
@@ -98,8 +98,10 @@ jobs:
       artifact: ${{ parameters.swss_common_artifact_name }}
       runVersion: 'latestFromBranch'
       runBranch: 'refs/heads/master'
+      path: '$(Build.SourcesDirectory)/${{ parameters.swss_common_artifact_name }}'
     displayName: "Download sonic swss common deb packages"
   - script: |
+      cd $(Build.SourcesDirectory)/${{ parameters.swss_common_artifact_name }}
       sudo dpkg -i libswsscommon_1.0.0_${{ parameters.arch }}.deb
       sudo dpkg -i libswsscommon-dev_1.0.0_${{ parameters.arch }}.deb
     workingDirectory: $(Pipeline.Workspace)
```
xumia, line 107: How about changing the workDirectroy to $(Build.SourcesDirectory)/${{ parameters.swss_common_artifact_name }}?

xumia (dismissed):

qiluo-msft:
> The sonic-sairedis artifact will not include the downloaded swss common artifact, otherwise its using
may have issue of overriding the previous swss common packages.

Then in the artifacts view, we have no way to check libswsscommon packages. This will possible limit our observation and the troubleshooting capability.

.azure-pipelines/build-template.yml
jimmyzhai: I move the step of sairedis checkout to the first.

xumia (approved):

PR #1011, opened 2022-02-24
[syncd][vslib] Keep new warm boot discovered SERDES objects (#985)

kcudnik:


kcudnik:
test failures not related to PR

PR #1012, opened 2022-03-01
[202012] For LGTM, fix up Doxyfile to make it work for new versions of Doxygen

saiarcot895:
LGTM is now using Ubuntu 20.04 for compilations. As part of this, newer versions of Doxygen use a different syntax for representing newlines in aliases defined in Doxyfile. Try to fix this from the LGTM config.


saiarcot895:
LGTM build with this commit is passing, but the base commit (the current 202012 branch) is failing, which makes sense since the fix isn't there.

kcudnik (changes requested):
this is not correct fix, correct fix is to update SAI submodule with this change: https://github.com/opencomputeproject/SAI/pull/1424

saiarcot895:
I was looking into that, but for some reason, I thought there were multiple commits that would come in if i did a submodule update to bring in that fix. I think I was looking at the v1.9 branch instead of the v1.7 branch.

I've raised a PR at opencomputeproject/SAI#1414 to backport the fix to v1.7.

kcudnik (approved):

PR #1016, opened 2022-03-08
Enable SAI_SWITCH_ATTR_UNINIT_DATA_PLANE_ON_REMOVAL attribute

tbgowda:
Enable SAI_SWITCH_ATTR_UNINIT_DATA_PLANE_ON_REMOVAL attribute for all platforms based on capability

- Remove Mellanox specific checks
- Check whether the attribute is supported
- Call SAI API if it is supported, else ignore

Cherry-picking https://github.com/Azure/sonic-sairedis/pull/975

### DIFF
--- a/syncd/Syncd.cpp
+++ b/syncd/Syncd.cpp
@@ -4181,15 +4181,36 @@ sai_status_t Syncd::setUninitDataPlaneOnRemovalOnAllSwitches()
 
         auto strRid = sai_serialize_object_id(rid);
 
-        auto status = m_vendorSai->set(SAI_OBJECT_TYPE_SWITCH, rid, &attr);
+        sai_attr_capability_t attr_capability = {};
 
-        if (status != SAI_STATUS_SUCCESS)
+        sai_status_t queryStatus;
+
+        queryStatus = sai_query_attribute_capability(rid,
+                                                     SAI_OBJECT_TYPE_SWITCH,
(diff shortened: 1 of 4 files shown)


tbgowda:
This is a double commit of https://github.com/Azure/sonic-sairedis/pull/975.

a. it is tested on cisco platforms
b. don't have access to other platforms

syncd/Syncd.cpp
```diff
@@ -4181,15 +4181,36 @@ sai_status_t Syncd::setUninitDataPlaneOnRemovalOnAllSwitches()
 
         auto strRid = sai_serialize_object_id(rid);
 
-        auto status = m_vendorSai->set(SAI_OBJECT_TYPE_SWITCH, rid, &attr);
+        sai_attr_capability_t attr_capability = {};
 
-        if (status != SAI_STATUS_SUCCESS)
+        sai_status_t queryStatus;
+
+        queryStatus = sai_query_attribute_capability(rid,
```
kcudnik, line 4188: don't use SAI api directly, use m_vendorSai

tbgowda: Apologies for the delay. I have addressed this comment. Please take a look.

tbgowda:
Thank you! will do. Please help merge the change so that it can be released into sonic-buildimage.

PR #1022, opened 2022-03-22
[syncd] Fix comparison logic for new transferred objects

kcudnik:
For created objects like buffer pool or buffer profile
those are not on the switch, but they were treated as such
and transferred to temporary view during comparison logic
causing to duplicate objects on temporary view


kcudnik:
Ignore code coverage, since i fixed some log messages in unused class

tjchadaga (approved):

PR #1028, opened 2022-04-06
[counter] Support gearbox and macsec counters

jimmyzhai:
1/ Collect gearbox port counters
2/ Run macsec sa plugin

### DIFF
--- a/syncd/FlexCounter.cpp
+++ b/syncd/FlexCounter.cpp
@@ -3505,6 +3505,8 @@ void FlexCounter::notifyPoll()
 
 bool FlexCounter::isOnGearbox()
 {
+    SWSS_LOG_ENTER();
+
     return m_dbCounters == "GB_COUNTERS_DB";
 }
 
@@ -3517,6 +3519,8 @@ bool FlexCounter::isGearboxEnabled()
 
 bool FlexCounter::isPortPhyCounter(sai_port_stat_t portCounterId)
 {
+    SWSS_LOG_ENTER();
+
     return portCounterId == SAI_PORT_STAT_IF_IN_ERRORS ||
         portCounterId == SAI_PORT_STAT_IF_OUT_ERRORS ||
         portCounterId == SAI_PORT_STAT_IF_IN_DISCARDS ||
@@ -3534,6 +3538,8 @@ void FlexCounter::mapGearboxPort(
         _In_ const std::string &name,
         _In_ sai_object_id_t vid)
 {
+    SWSS_LOG_ENTER();
+
     if (!isOnGearbox())
     {
         return;
@@ -3579,6 +3585,8 @@ void FlexCounter::mapGearboxPort(
 void FlexCounter::unmapGearboxPort(
         _In_ sai_object_id_t vid)
 {
+    SWSS_LOG_ENTER();
+
     if (!isOnGearbox())
     {
         return;
--- a/syncd/FlexCounter.cpp
+++ b/syncd/FlexCounter.cpp
@@ -1111,6 +1111,7 @@ void FlexCounter::checkPluginRegistered(
 
     if (
             m_portPlugins.find(sha) != m_portPlugins.end() ||
+            m_macsecSAPlugins.find(sha) != m_macsecSAPlugins.end() ||
             m_rifPlugins.find(sha) != m_rifPlugins.end() ||
             m_queuePlugins.find(sha) != m_queuePlugins.end() ||
             m_priorityGroupPlugins.find(sha) != m_priorityGroupPlugins.end() ||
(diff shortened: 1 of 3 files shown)


syncd/FlexCounter.h
```diff
@@ -13,9 +13,14 @@ extern "C" {
 #include <condition_variable>
 #include <unordered_map>
 #include <memory>
+#include <utility>
 
 namespace syncd
 {
+    typedef std::pair<sai_object_id_t, sai_object_id_t> gearboxPortPair;
+#define vidSystemSide std::get<0>
+#define vidLineSide   std::get<1>
```
kcudnik, line 22: move this out of namespace, defines are not scoped by namespaces

syncd/FlexCounter.cpp
```diff
@@ -3433,3 +3502,120 @@ void FlexCounter::notifyPoll()
...
+        portCounterId == SAI_PORT_STAT_IF_OUT_ERRORS ||
+        portCounterId == SAI_PORT_STAT_IF_IN_DISCARDS ||
+        portCounterId == SAI_PORT_STAT_IF_OUT_DISCARDS ||
+        portCounterId == SAI_PORT_STAT_ETHER_RX_OVERSIZE_PKTS ||
+        portCounterId == SAI_PORT_STAT_ETHER_TX_OVERSIZE_PKTS ||
+        portCounterId == SAI_PORT_STAT_ETHER_STATS_UNDERSIZE_PKTS ||
+        portCounterId == SAI_PORT_STAT_ETHER_STATS_JABBERS ||
+        portCounterId == SAI_PORT_STAT_ETHER_STATS_FRAGMENTS ||
+        portCounterId == SAI_PORT_STAT_IF_IN_FEC_NOT_CORRECTABLE_FRAMES ||
+        portCounterId == SAI_PORT_STAT_IF_IN_FEC_SYMBOL_ERRORS;
```
kcudnik, line 3534: this does not seem generic :/

syncd/FlexCounter.cpp
```diff
@@ -3433,3 +3502,120 @@ void FlexCounter::notifyPoll()
     m_readyToPoll = true;
     m_pollCond.notify_all();
 }
+
+bool FlexCounter::isOnGearbox()
+{
+    SWSS_LOG_ENTER();
+
+    return m_dbCounters == "GB_COUNTERS_DB";
```
kcudnik, line 3510: this is not the best solution, since that db name could be changed to something else using config.json, this db name is default when no config is actually specified, maybe you could find a different way to figure out whether this is gearbox ? btw. why do you actually need to check this ?

kcudnik (changes requested):

syncd/FlexCounter.h
jimmyzhai: Move it to FlexCounter.cpp

syncd/FlexCounter.cpp
jimmyzhai: Fix it with a unordered_set

syncd/FlexCounter.cpp
jimmyzhai: I have not found an easier and better solution. For default db name, that must not be a case for gbsyncd/gearbox, which always utilizes global context guid 1 in `/usr/share/sonic/hwsku/context_config.json`. Like `COUNTERS_DB` for asic, the name GB_COUNTERS_DB is generally common for gearbox.

Only on gbsyncd, it needs to run a specific plugin gb_port.lua for dealing with gearbox port counter. This plugin needs a list of tuples (portName, systemSidePortVid, lineSidePortVid) as KEYS.

kcudnik: not sure what you meant by always utilizing guid 1, since that also don't seems to generic solution, using context_config.json, gearbox could be put anywhere, and it this will be some case in the future, this will cause some weird issues hard to find if some of those values will be hard coded :(

syncd/FlexCounter.cpp
```diff
@@ -3433,3 +3519,110 @@ void FlexCounter::notifyPoll()
...
+}
+
+bool FlexCounter::isGearboxEnabled()
+{
+    SWSS_LOG_ENTER();
+
+    return access("/usr/share/sonic/hwsku/gearbox_config.json", F_OK) != -1;
```
kcudnik, line 3534: this is also not good to hardcode config path :( please don't use hardcoded config location to figure out whether this is gearbox or not

isOnGearbox and isGearboxEnabled basically are the same? why there could not be only 1 function for this? do you need some special action if gearbox is enabled but non on gearbox? seems like just putting that config would have some impact even if syncd will not run in gearbox

i have a proposal here: can you have 1 flag maybe as commandline option or an extra option in context_config.json that specifies this is gearbox ? and then syncd would be able to figure this out based on that parameter rather hardcoding db names or config location

kcudnik (changes requested):
please read comments

syncd/FlexCounter.cpp
jimmyzhai: Have a new design https://github.com/Azure/sonic-swss-common/pull/622. In application (counter reader) side, it handles the logic of counter combining of asic port and gearbox port. Then we do not need to change anything in syncd/gbsyncd, which only simply collects counters from asic/phy chip.

kcudnik: ok, then after fixing this in common repo, please close this pr

jimmyzhai: Sure, will do it.

jimmyzhai:
Close it. New design https://github.com/Azure/sonic-swss-common/pull/622 is merged.

PR #1029, opened 2022-04-06
Use priority attribute matching acl counter using acl entry

kcudnik:
In case where multiple exact the sam ACL counter exists, and they are assigned to different ACL entries, it can happen that when CL is searching for correct ACL counter for given ACL entry, it uses any ACL entry attribute to find correct ACL counter, and it this case here it happen that at random first attribute was  SAI_ACL_TABLE_ATTR_FIELD_ETHER_TYPE, which has the same value on all ACL entries defined. This caused to all ACL counters look the same for CL, and actually ACL counter was selected at random. This could lead to potential 2 situations that caused unnecessary acis operations:
* wrong selected ACL counters was causing SET operation on some ACL entries to correct the state of ACL counter, which in transition state was leading to have same ACL counter briefly present on more than 1 ACL entry, which maybe not supported by vendor hardware, in this case BRCM, and SET operation failed causing entire apply view CL to fail and crash syncd
* similar scenario to above, but in some cases multiple ACL entries could match when selecting best match on ACL entry, which causing to remove one of the ACL entries and create new one, and on others also SET ACL counter attribute

In this PR first attribute was selected SAI_ACL_ENTRY_ATTR_PRIORITY, which probably will be unique for all ACL entries, since we want to control exact order or ACL on the devie.

This PR addresses both those issue and no more asic operations will be preformed.
NOTE:  this is just a patch fix for this particular scenario, ideally we want to use all ACL entry non oid attributes to act as a hash, when selecting acl counter, and put that logic inside pre match logic to pre match acl counters


vaibhavhd (approved):

qiluo-msft:

kcudnik:
> This commit could not be cleanly cherry-picked to 202012. Please submit another PR.

202012 PR is here: https://github.com/Azure/sonic-sairedis/pull/1033

PR #1032, opened 2022-04-08
Advance submodule SAI for 202012 branch

bingwang-ms:
Changes include

```
c460675 [SAI]Fix build in v1.7 and remove unnecessary change (#1468)
bfaaa0e Tunnel Scoped TC MAP and Remarking (#1452)
```

Signed-off-by: bingwang <bingwang@microsoft.com>

### DIFF
--- a/lib/inc/sai_redis.h
+++ b/lib/inc/sai_redis.h
@@ -424,3 +424,15 @@ PRIVATE extern std::shared_ptr<sairedis::Sai>   redis_sai;
     _X(ROUTE_ENTRY,route_entry);            \
     _X(NAT_ENTRY,nat_entry);                \
 
+// BULK get/set DECLARE
+
+#define REDIS_BULK_GET_SET(OT,ot)   \
+    REDIS_BULK_GET(OT,ot);          \
+    REDIS_BULK_SET(OT,ot);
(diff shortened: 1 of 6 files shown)


kcudnik:
seems like that branch on SAI is incorrect, failed to build metadata

richardyu-ms:
> seems like that branch on SAI is incorrect, failed to build metadata

PR https://github.com/opencomputeproject/SAI/pull/1468, for fix

bingwang-ms:
Update PR to include change in https://github.com/opencomputeproject/SAI/pull/1468

kcudnik:
something wrong with swss common download, could you add empty dummy commit here to trigger full pipeline build?

also, this change of SAI module is adding new SAI apis, and they will need to be handled in the code of virtual switch and sairedis apis to correctly compile

bingwang-ms:
> something wrong with swss common download, could you add empty dummy commit here to trigger full pipeline build?
> 
> also, this change of SAI module is adding new SAI apis, and they will need to be handled in the code of virtual switch and sairedis apis to correctly compile

Updated. Please take a look again. Thanks

kcudnik (approved):

PR #1040, opened 2022-05-11
[ci] Paralize azure pipeline

liushilongbuaa:
1. Setup pipeline without manual effort when checkout new release branch.
2. Use correct branch when downloading artifacts or checkout relative repos.
3. Clear downloaded artifacts to avoid using outdated dependencies.
4. Use commonlib pipeline to download libnl3 and libyang instead of vs image build, to increase success rate.
5. Add weekly build to keep artifacts remaining.

### DIFF
--- a/.azure-pipelines/build-docker-sonic-vs-template.yml
+++ b/.azure-pipelines/build-docker-sonic-vs-template.yml
@@ -58,7 +58,7 @@ jobs:
     inputs:
       source: specific
       project: build
-      pipeline: sonic-buildimage.vs
+      pipeline: Azure.sonic-buildimage.official.vs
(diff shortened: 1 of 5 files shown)


kcudnik:
could you provide extensive description in this PR what it's exactly for ?

liushilongbuaa:
> could you provide extensive description in this PR what it's exactly for ?

Updated description. From this [page](https://dev.azure.com/mssonic/build/_build?definitionId=12&_a=summary&repositoryFilter=10&branchFilter=7408%2C7408%2C7408), we can see 202111 branch azp is failing. While the same PR for 202111 branch( https://github.com/Azure/sonic-sairedis/pull/1043) will fix this issue.

PR #1048, opened 2022-05-26
[syncd] Remove error message when tryTranslateVidToRid fails

### LINKED ISSUE #1047: [LAG] received extra error log "unable to get RID for VID" even when code flow handle this exception successfully

kcudnik:
Fixes #1047

### DIFF
--- a/syncd/VirtualOidTranslator.cpp
+++ b/syncd/VirtualOidTranslator.cpp
@@ -338,16 +338,45 @@ bool VirtualOidTranslator::tryTranslateVidToRid(
 {
     SWSS_LOG_ENTER();
 
-    try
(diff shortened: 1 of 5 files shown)


lguohan (dismissed):

stephenxs:
Hi @kcudnik 
I'm just curious why we still need to retrieve RID from VID when a flex counter is removed as the RID is not really used in the removing flow and we can run into a racing condition that RID is not available anymore.
Just for logging the debug information to notify the user of the fact that we are facing a racing condition or other purposes?
thanks

stephenxs:
hi @kcudnik can we cherry-pick it to 202205?

PR #1049, opened 2022-05-27
[configure.ac] SAI version check fix for gcc 10.2 (bullseye)

stepanblyschak:
The command to build a check program was:

```
gcc -lsai -I./SAI/inc -I./SAI/experimental -I./SAI/meta conftest.cpp -o conftest
```

This, however, does not work on gcc 10.2 where the linker wants first
the executable and then the shared library:

```
gcc -I./SAI/inc -I./SAI/experimental -I./SAI/meta conftest.cpp -lsai -o conftest
```

However CXX_FLAGS comes first before the source file, so added
--no-as-needed as a fix for this issue to make it link against libsai in
any order passed to the linker.

Signed-off-by: Stepan Blyschak <stepanb@nvidia.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -185,7 +185,7 @@ AC_SUBST(SAIINC, "-I\$(top_srcdir)/SAI/inc -I\$(top_srcdir)/SAI/experimental -I\
 AM_COND_IF([SYNCD], [
 AM_COND_IF([SAIVS], [], [
 SAVED_FLAGS="$CXXFLAGS"
-CXXFLAGS="-lsai -I$srcdir/SAI/inc -I$srcdir/SAI/experimental -I$srcdir/SAI/meta"
+CXXFLAGS="-Xlinker --no-as-needed -lsai -I$srcdir/SAI/inc -I$srcdir/SAI/experimental -I$srcdir/SAI/meta"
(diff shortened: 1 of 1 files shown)


kcudnik:
> This, however, does not work on gcc 10.2 where the linker wants first the executable and then the shared library:

what is the error here? and did you mean source file before shared library ?

stepanblyschak:
@kcudnik Yes, and the error is undefined reference to sai_query_api_version

kcudnik:
why not pou -lsai to the end of flags, instead of adding extra linker flags?

stepanblyschak:
@kcudnic From the configure.ac:
CXXFLAGS="-lsai -I$srcdir/SAI/inc -I$srcdir/SAI/experimental -I$srcdir/SAI/meta"

No matter where I put -lsai, the CXXFLAGS is placed before conftest.cpp when compile command is generated.
I also tried to put -lsai in LDFLAGS but it is also put before conftest.cpp.

I found a flag -Xlinker=--no-as-needed to keep previous behaviour and it was the simplest solution I found

kcudnik (approved):

kcudnik (approved):

PR #1052, opened 2022-05-30
[counter] Fix port flex counter 

jimmyzhai:
Fix https://github.com/Azure/sonic-buildimage/issues/10850.

A fact is there might be different port types on asic, then different port stats capabilities. Instead of using a cached supported port counter ID list for all ports, it gets supported port counter list per port.

### DIFF
--- a/syncd/FlexCounter.cpp
+++ b/syncd/FlexCounter.cpp
@@ -2451,6 +2451,10 @@ sai_status_t FlexCounter::querySupportedPortCounters(
         SAI_OBJECT_TYPE_PORT,
         &stats_capability);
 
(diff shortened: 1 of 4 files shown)


prgeor (approved):

syncd/FlexCounter.cpp
```diff
@@ -2437,7 +2431,7 @@ void FlexCounter::endFlexCounterThread(void)
 }
 
 sai_status_t FlexCounter::querySupportedPortCounters(
-        _In_ sai_object_id_t portRid)
+        _In_ sai_object_id_t portRid, _Out_ PortCountersSet &supportedPortCounters)
```
kcudnik, line 2434: one parameter per line

kcudnik (approved):

PR #1058, opened 2022-06-06
[sairedis] Perform log rotate on request

kcudnik:
Previously log rotate was performed after request and on first log line
that was recorded which caused delay and issues with syslog logrotate
when sending HUP signal, actual log rotate was not performed and handle
to a sairedis.rec was sill open preventing logrotate from happening.

Because of this change, perform log rotate CAN'T be done in signal handler, it needs to be postponed to main thread.

### DIFF
--- a/lib/Recorder.cpp
+++ b/lib/Recorder.cpp
@@ -179,27 +179,20 @@ void Recorder::recordLine(
     {
         m_ofstream << getTimestamp() << "|" << line << std::endl;
     }
-
-    if (m_performLogRotate)
-    {
-        m_performLogRotate = false;
-
-        recordingFileReopen();
-
-        /* double check since reopen could fail */
-
-        if (m_ofstream.is_open())
-        {
-            m_ofstream << getTimestamp() << "|" << "#|logrotate on: " << m_recordingFile << std::endl;
-        }
-    }
 }
 
 void Recorder::requestLogRotate()
 {
     SWSS_LOG_ENTER();
 
-    m_performLogRotate = true;
+    recordingFileReopen();
+
+    /* double check since reopen could fail */
+
+    if (m_ofstream.is_open())
+    {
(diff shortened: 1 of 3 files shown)


lguohan:
@kcudnik , where pr do we need? i am not quite clear.

kcudnik:
i explained this in detail in internal thread on 100% /var/log utilization, that when perform logrotate is requested in sairedis, it's not done immediately, but after next api or notification will be called, which can take some time and it will hold handle to sairedis rec preventing logrotating

lguohan (approved):

qiluo-msft:

kcudnik:
@qiluo-msft PR for 202012 is here : https://github.com/Azure/sonic-sairedis/pull/1083

stephenxs:
Looks like this PR introduces an error message
`ERR swss#orchagent: :- recordingFileReopen: failed to open recording file : No such file or directory`
The function `recordingFileReopen` is called each time log rotating but was not called before.
I suspect the `m_recordingFile` is not initialized correctly.
Can you please check it? @kcudnik

kcudnik:
recording file is initialized correctly here:
https://github.com/Azure/sonic-sairedis/blob/master/lib/Recorder.cpp#L29

but not sure why this error would pop up

kcudnik:
ahh i see the issue, will try to address, seems like recordingFileReopen is called before startRecording
but this should be not major issue, recording still will work after startRecording called
i will make proper PR

stephenxs:
> ahh i see the issue, will try to address, seems like recordingFileReopen is called before startRecording but this should be not major issue, recording still will work after startRecording called i will make proper PR

Thank you.
yes, only an error message. not functional.

kcudnik:
Added fix here: https://github.com/Azure/sonic-sairedis/pull/1087

PR #1068, opened 2022-06-17
[pipeline] Set CAP_SYS_TIME for settimeofday ok in syncd test

jimmyzhai:
The fix https://github.com/Azure/sonic-sairedis/pull/1067 is not enough. If docker user is non-root, set capability CAP_SYS_TIME for settimeofday success in syncd test, then `test_watchdog_timer_clock_rollback` can be run.

### DIFF
--- a/syncd/Makefile.am
+++ b/syncd/Makefile.am
@@ -86,6 +86,7 @@ syncd_request_shutdown_LDADD = libSyncdRequestShutdown.a $(top_srcdir)/lib/libSa
 
 tests_SOURCES = tests.cpp
 tests_CXXFLAGS = $(DBGFLAGS) $(AM_CXXFLAGS) $(CXXFLAGS_COMMON)
+tests_LDFLAGS = -Wl,-rpath,$(top_srcdir)/lib/.libs -Wl,-rpath,$(top_srcdir)/meta/.libs
 tests_LDADD = libSyncd.a -lhiredis -lswsscommon -lpthread -L$(top_srcdir)/lib/.libs -lsairedis \
 			  -L$(top_srcdir)/meta/.libs -lsaimetadata -lsaimeta -lzmq $(CODE_COVERAGE_LIBS)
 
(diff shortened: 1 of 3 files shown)


syncd/Makefile.am
```diff
@@ -86,6 +86,7 @@ syncd_request_shutdown_LDADD = libSyncdRequestShutdown.a $(top_srcdir)/lib/libSa
 
 tests_SOURCES = tests.cpp
 tests_CXXFLAGS = $(DBGFLAGS) $(AM_CXXFLAGS) $(CXXFLAGS_COMMON)
+tests_LDFLAGS = -Wl,-rpath,$(top_srcdir)/lib/.libs -Wl,-rpath,$(top_srcdir)/meta/.libs
```
kcudnik, line 89: why this is required here ? those directories are already passed in below LDADD

kcudnik (approved):

syncd/Makefile.am
jimmyzhai: Running `syncd/tests` depends on env LD_LIBRARY_PATH:
```
+ LD_LIBRARY_PATH=/sonic/src/sonic-sairedis/lib/.libs:/sonic/src/sonic-sairedis/meta/.libs
+ export LD_LIBRARY_PATH
...
+ exec /sonic/src/sonic-sairedis/syncd/.libs/tests
```

After setting linux capability, for security reason, LD_LIBRARY_PATH is ignored, referring to **man ld.so**:

> **Secure-execution mode**
    For security reasons, if the dynamic linker determines that a binary should be run in secure-execution mode,  the  effects  of  some  environment variables  are  voided  or  modified, and furthermore those environment variables are stripped from the environment, so that the program does not even see the definitions...
A binary is executed in secure-execution mode ...
> \* A process with a non-root user ID executed a binary that conferred capabilities to the process.

The have to specify `-rpath` option for runtime linker to locate shared library libsairedis, etc.

PR #1071, opened 2022-06-27
Add support of mdio access using sai switch api and unix socket IPC s…

jiahua-wang:
…erver

Signed-off-by: Jiahua Wang <jiahua.wang@broadcom.com>


syncd/GlobalSwitchId.cpp
```diff
@@ -15,6 +15,10 @@ sai_object_id_t gSwitchId = SAI_NULL_OBJECT_ID;
 
 #endif
 
+#ifdef MDIO_ACCESS_USE_NPU
+sai_object_id_t mdioSwitchId = SAI_NULL_OBJECT_ID;
+#endif
+
```
kcudnik, line 21: why you suddenly need this instead of gSwitchId? also i strongly advise to not use global objects, this file is added only for thrift support

syncd/Syncd.cpp
```diff
@@ -4472,6 +4472,11 @@ static void timerWatchdogCallback(
     SWSS_LOG_ERROR("main loop execution exceeded %ld ms", span/1000);
 }
 
+#ifdef MDIO_ACCESS_USE_NPU
+extern int startIpcMdioProcessingThread(sai_switch_api_t *sai_switch_api);
+extern void stopIpcMdioProcessingThread(void);
+#endif
+
```
kcudnik, line 4479: move this to separate file
also why those 2 functions needs to be explicitly called instead on internals SAI when calling create_switch ?

syncd/syncd_mdio_ipc.cpp
```diff
@@ -0,0 +1,440 @@
+#include <stdio.h>
```
kcudnik, line 1: if  you want to keep this file, make this to a proper class, and then instantiate object and call explicitly functions of that object

syncd/syncd_mdio_ipc.cpp
```diff
@@ -0,0 +1,440 @@
...
+
+    if (argc < 3)
+    {
+        return SAI_STATUS_INVALID_PARAMETER;
+    }
+
+    mdio_addr = (uint32_t)strtoul(argv[1], NULL, 0);
+    reg_addr = (uint32_t)strtoul(argv[2], NULL, 0);
+
+    if (mdioSwitchId == SAI_NULL_OBJECT_ID)
```
kcudnik, line 110: this mdioSwitchId  needs to be passed to this class that you will create in the constructor, then remove global definition

syncd/GlobalSwitchId.cpp
```diff
@@ -33,4 +37,10 @@ void GlobalSwitchId::setSwitchId(
             sai_serialize_object_id(gSwitchId).c_str());
 #endif
 
+#ifdef MDIO_ACCESS_USE_NPU
+    if (mdioSwitchId == SAI_NULL_OBJECT_ID)
+    {
+        mdioSwitchId = switchRid;
+    }
+#endif
```
kcudnik, line 45: this mdioSwitchId  needs to be passed to this class that you will create in the constructor, then remove global definition

syncd/Syncd.cpp
```diff
@@ -4472,6 +4472,11 @@ static void timerWatchdogCallback(
     SWSS_LOG_ERROR("main loop execution exceeded %ld ms", span/1000);
 }
 
+#ifdef MDIO_ACCESS_USE_NPU
+extern int startIpcMdioProcessingThread(sai_switch_api_t *sai_switch_api);
```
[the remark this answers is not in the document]
kcudnik: those functions needs to be object methods

syncd/syncd_mdio_ipc.cpp
```diff
@@ -0,0 +1,440 @@
...
+                    break;
+                ++argc;
+            }
+
+#if 0
+            printf("CMD: ");
+            for (j = 0; j < argc; ++j)
+                printf("[%s],", argv[j]);
+            printf("\n");
+#endif
```
kcudnik, line 369: remove debug

syncd/syncd_mdio_ipc.cpp
```diff
@@ -0,0 +1,440 @@
...
+}
+
+void stopIpcMdioProcessingThread(void)
+{
+    int *err = NULL;
+
+    taskAlive = 0;
+    pthread_join(taskId, (void **)&err);
+    SWSS_LOG_NOTICE("IPC task thread is stopped\n");
+    return;
```
kcudnik, line 422: unnecessary return

syncd/syncd_mdio_ipc.cpp
```diff
@@ -0,0 +1,440 @@
...
+}
+
+static syncd_mdio_ipc_cmd_t ipc_cmds[] =
+{
+    { "mdio", syncd_ipc_cmd_mdio },
+    { "mdio-cl22", syncd_ipc_cmd_mdio_cl22 },
+};
+
+static void *syncd_ipc_task_main(void *ctx)
+{
```
kcudnik, line 189: all functions are missing SWSS_LOG_ENTER()

syncd/syncd_mdio_ipc.cpp
```diff
@@ -0,0 +1,440 @@
...
+    if (argc > 3)
+    {
+        val = (uint32_t)strtoul(argv[3], NULL, 0);
+        ret = sai_switch_api->switch_mdio_write(mdioSwitchId, mdio_addr, reg_addr, 1, &val);
+        sprintf(resp, "%d\n", ret);
+    }
+    else
+    {
+        ret = sai_switch_api->switch_mdio_read(mdioSwitchId, mdio_addr, reg_addr, 1, &val);
+        sprintf(resp, "%d 0x%x\n", ret, val);
```
kcudnik, line 125: output to a unknown buffer size out the function scope, use std::string to produce response or log errors here instead of passing it through the stack

kcudnik (changes requested):

syncd/Syncd.cpp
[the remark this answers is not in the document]
kcudnik: you can pass std::shared_ptr<VendorSai> as api

syncd/syncd_mdio_ipc.cpp
```diff
@@ -0,0 +1,440 @@
...
+}
+
+/*
+ * mdio <port_oid> reg:     Read from the PHY register
+ * mdio <port_oid> reg val: Write to the PHY register
+ */
+static sai_status_t syncd_ipc_cmd_mdio_common(char *resp, int argc, char *argv[], sai_switch_api_t *sai_switch_api)
+{
+    int ret = 0;
+    uint32_t mdio_addr = 0, reg_addr = 0, val = 0;
```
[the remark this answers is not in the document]
jiahua-wang: I am rewriting the code in c++ class. I will try to address most of the comments in new code.

jiahua-wang:
This PR is replaced by #1080

PR #1074, opened 2022-07-04
[fastboot] fastboot enhancement: Use warm-boot infrastructure for fast-boot

arfeigin:
Rename WarmRestartTable to AdvancedRestartTable since it is used for both reboot scenarios.
Ignore warm logic if syncd starts in fast-boot in addition to fastfast-boot.

This PR is dependent on https://github.com/Azure/sonic-swss-common/pull/644

Signed-off-by: Aryeh Feigin [afeigin@nvidia.com](mailto:afeigin@nvidia.com)


liat-grozovik:
@arfeigin please review failures.
it will be great if you can rename the PR name and add the following prefix '[fastboot] fastboot enhancement: '

arfeigin:
> @arfeigin please review failures. it will be great if you can rename the PR name and add the following prefix '[fastboot] fastboot enhancement: '

It fails due to dependency - it relies on schema change. (https://github.com/Azure/sonic-swss-common/pull/644)

syncd/Syncd.cpp
```diff
@@ -203,9 +203,9 @@ void Syncd::performStartupLogic()
 {
     SWSS_LOG_ENTER();
 
-    // ignore warm logic here if syncd starts in Mellanox fastfast boot mode
+    // ignore warm logic here if syncd starts in fastboot or in Mellanox fastfast boot mode
 
-    if (m_isWarmStart && (m_commandLineOptions->m_startType != SAI_START_TYPE_FASTFAST_BOOT))
+    if (m_isAdvancedStart && m_commandLineOptions->m_startType == SAI_START_TYPE_WARM_BOOT)
```
kcudnik, line 208: changing fastboot to warm boot may have some issues here, or is this the way it should be ?

kcudnik: rest looks good, please resolve issues, or because of the type of this change it cant befixed? is this breaking change and not backward compatible?

PR #1076, opened 2022-07-08
[vslib] Add VLAN stacking on libvs

chaoskao:
libvs support VLAN stacking operation


kcudnik:
please fix conflicts and compilation issues

PR #1079, opened 2022-07-14
Do not fetch rid during removing counter

stephenxs:
It's possible an object was removed before the corresponding counter is removed even if orchagent removed the counter first and then the object itself because they are handled in different tables in syncd.
In such a case, an error message will be reported in tryTranslateVidToRid.
To avoid the error, the function is called for SET only.

Signed-off-by: Stephen Sun <stephens@nvidia.com>

### DIFF
--- a/syncd/Syncd.cpp
+++ b/syncd/Syncd.cpp
@@ -2122,7 +2122,12 @@ void Syncd::processFlexCounterEvent( // TODO must be moved to go via ASIC channe
 
     sai_object_id_t rid;
 
-    if (!m_translator->tryTranslateVidToRid(vid, rid))
+    // It's possible the object was removed before the corresponding counter is removed
+    // even if orchagent removed counter first and then the object itself because
+    // they are handled in different tables in syncd
+    // In such case, error message will be reported in tryTranslateVidToRid
+    // To avoid the error, the function is called for SET only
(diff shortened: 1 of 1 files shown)


stephenxs:
Working on the coverage issue.
VS test failure isn't related to the PR.

syncd/Syncd.cpp
```diff
@@ -2122,7 +2122,12 @@ void Syncd::processFlexCounterEvent( // TODO must be moved to go via ASIC channe
 
     sai_object_id_t rid;
 
-    if (!m_translator->tryTranslateVidToRid(vid, rid))
+    // It's possible the object was removed before the corresponding counter is removed
+    // even if orchagent removed counter first and then the object itself because
+    // they are handled in different tables in syncd
+    // In such case, error message will be reported in tryTranslateVidToRid
+    // To avoid the error, the function is called for SET only
+    if (op == SET_COMMAND && !m_translator->tryTranslateVidToRid(vid, rid))
```
kcudnik, line 2130: comment is correct, but please remove this operation, this issue was fixed here: https://github.com/Azure/sonic-sairedis/pull/1048/files but not merged yet, if you want you can leave comment why trytranslate is used

stephenxs: Thanks. Then I'll close this PR and let's discuss there.

PR #1082, opened 2022-07-19
Support SAI_PORT_ATTR_PORT_SERDES_ID on vs gearbox

byu343:
This change is to support SAI_PORT_ATTR_PORT_SERDES_ID on the SAI for vs gearbox

### DIFF
--- a/vslib/SwitchBCM81724.cpp
+++ b/vslib/SwitchBCM81724.cpp
@@ -58,6 +58,11 @@ sai_status_t SwitchBCM81724::create_port_dependencies(
 
     CHECK_STATUS(set(SAI_OBJECT_TYPE_PORT, port_id, &attr));
 
+    attr.id = SAI_PORT_ATTR_PORT_SERDES_ID;
+    attr.value.oid = SAI_NULL_OBJECT_ID;
+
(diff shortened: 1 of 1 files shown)


vslib/SwitchBCM81724.cpp
```diff
@@ -273,6 +278,7 @@ sai_status_t SwitchBCM81724::refresh_read_only(
                  */
 
             case SAI_PORT_ATTR_OPER_STATUS:
+            case SAI_PORT_ATTR_PORT_SERDES_ID:
```
kcudnik, line 281: SAI_PORT_ATTR_PORT_SERDES_ID is readonly attribute, and normally it should be recalculated each time is queried but if it's null, then is ok to jus return here

kcudnik:
please address errors and issues

vslib/SwitchBCM81724.cpp
byu343: @kcudnik Do I need to fix something here? Other than here, it looks like all checks haven been pass.

kcudnik: you are fine

PR #1100, opened 2022-08-01
[fastboot] fastboot enhancement: Use warm-boot infrastructure for fast-boot

arfeigin:
This PR should be merged together with sonic-utilities PR (https://github.com/sonic-net/sonic-utilities/pull/2286) and sonic-buildimage PR (https://github.com/sonic-net/sonic-buildimage/pull/11594).

Fast-reboot is utilizing warm-reboot infrastructure to improve its performance, but it should ignore warm-boot logic when syncd starts in fast-boot.
As well it shouldn't use temporary view between init and apply.

### DIFF
--- a/syncd/Syncd.cpp
+++ b/syncd/Syncd.cpp
@@ -204,9 +204,9 @@ void Syncd::performStartupLogic()
 {
     SWSS_LOG_ENTER();
 
-    // ignore warm logic here if syncd starts in Mellanox fastfast boot mode
+    // ignore warm logic here if syncd starts in fastboot or Mellanox fastfast boot mode
 
-    if (m_isWarmStart && (m_commandLineOptions->m_startType != SAI_START_TYPE_FASTFAST_BOOT))
+    if (m_isWarmStart && m_commandLineOptions->m_startType != SAI_START_TYPE_FASTFAST_BOOT && m_commandLineOptions->m_startType != SAI_START_TYPE_FAST_BOOT)
     {
         SWSS_LOG_WARN("override command line startType=%s via SAI_START_TYPE_WARM_BOOT",
                 CommandLineOptions::startTypeToString(m_commandLineOptions->m_startType).c_str());
--- a/syncd/Syncd.cpp
+++ b/syncd/Syncd.cpp
@@ -204,7 +204,7 @@ void Syncd::performStartupLogic()
 {
     SWSS_LOG_ENTER();
 
-    // ignore warm logic here if syncd starts in fastboot or Mellanox fastfast boot mode
+    // ignore warm logic here if syncd starts in fast-boot or Mellanox fastfast boot mode
 
     if (m_isWarmStart && m_commandLineOptions->m_startType != SAI_START_TYPE_FASTFAST_BOOT && m_commandLineOptions->m_startType != SAI_START_TYPE_FAST_BOOT)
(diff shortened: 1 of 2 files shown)


kcudnik:
please address build issues

kcudnik (approved):

syncd/Syncd.cpp
```diff
@@ -204,9 +204,9 @@ void Syncd::performStartupLogic()
 {
     SWSS_LOG_ENTER();
 
-    // ignore warm logic here if syncd starts in Mellanox fastfast boot mode
+    // ignore warm logic here if syncd starts in fast-boot or Mellanox fastfast boot mode
 
-    if (m_isWarmStart && (m_commandLineOptions->m_startType != SAI_START_TYPE_FASTFAST_BOOT))
+    if (m_isWarmStart && m_commandLineOptions->m_startType != SAI_START_TYPE_FASTFAST_BOOT && m_commandLineOptions->m_startType != SAI_START_TYPE_FAST_BOOT)
```
vaibhavhd, line 209: It may be better if this condition checks for positive match for `SAI_START_TYPE_WARM_BOOT` instead of a negative match for fastfast and fast reboot.

vaibhavhd (approved):
Minor suggestion added. LGTM.

vaibhavhd:
Is this also needed for 202205?

vaibhavhd:
Will this break the logic to fall back to old way of fast-reboot?
We still need old fast-reboot logic to work in case of vendor-os to SONiC conversion scenario. In that scenario there won't be any db to reconcile.

kcudnik:
> Will this break the logic to fall back to old way of fast-reboot? We still need old fast-reboot logic to work in case of vendor-os to SONiC conversion scenario. In that scenario there won't be any db to reconcile.

yea, this could be potential issue, good catch

syncd/Syncd.cpp
arfeigin: In this case the assignment in line 214: "m_commandLineOptions->m_startType = SAI_START_TYPE_WARM_BOOT;" is redundant, not?
Should I change to the positive match and remove the redundancy?

arfeigin:
@vaibhavhd @kcudnik Thanks for your comment, the scenario of vendor-os to SONiC conversion was addressed in the HLD (https://github.com/shlomibitton/SONiC/blob/436cb6a4280cc0214d45e3e41c6d843eaec8fadb/doc/fast-reboot/Fast-reboot_Flow_Improvements_HLD.md#32-in-service-restart-and-upgrade-from-a-vendor-nos-to-sonic-nos). I will add a test according to Shlomi's description.

syncd/Syncd.cpp
arfeigin: What do you think?

arfeigin:
This PR should be merged first together with the sonic-utilities PR (https://github.com/sonic-net/sonic-utilities/pull/2286) before merging the sonic-buildimage PR (https://github.com/sonic-net/sonic-buildimage/pull/11594).

Automatic functional test case covering the scenario of vendor NOS to SONiC NOS reboot is in work. The test will simulate such scenario by dropping the dump files and perform fast-reboot without it.

liat-grozovik:
@kcudnik could you please help to merge?

PR #1103, opened 2022-08-09
[vslib] Add VLAN stacking on libvs

chaoskao:
libvs support VLAN stacking operation


kcudnik:
please fix build issues, probably first SAI pointer needs to be updated

chaoskao:
Yes, when this PR (https://github.com/opencomputeproject/SAI/pull/1400) merged, I will update SAI pointer

PR #1105, opened 2022-08-17
Fix issue: bulk counter feature cannot compile on platforms having no sai_bulk_object_get_stats/sai_bulk_object_clear_stats

Junchao-Mellanox:
Why I did this?

bulk counter feature uses two new SAI API sai_bulk_object_get_stats/sai_bulk_object_clear_stats. Some vendors have no such API implemented in their SAI code. Need a way to make compilation pass.

How I fix this?

1. Add AC_CHECK_FUNCS for sai_bulk_object_get_stats/sai_bulk_object_clear_stats which generates macro HAVE_SAI_BULK_OBJECT_CLEAR_STATS/HAVE_SAI_BULK_OBJECT_GET_STATS in config.h.
2. Use ifdef to avoid compilation failure.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -207,6 +207,8 @@ AC_MSG_ERROR("SAI headers API version and library version mismatch")])])
 CXXFLAGS="$SAVED_FLAGS"
 ])])
 
+AC_CHECK_FUNCS(sai_bulk_object_clear_stats sai_bulk_object_get_stats)
+
 AC_OUTPUT(Makefile
           meta/Makefile
           lib/Makefile
--- a/syncd/VendorSai.cpp
+++ b/syncd/VendorSai.cpp
@@ -661,6 +661,7 @@ sai_status_t VendorSai::bulkGetStats(
     SWSS_LOG_ENTER();
     VENDOR_CHECK_API_INITIALIZED();
 
+#ifdef HAVE_SAI_BULK_OBJECT_GET_STATS
     return sai_bulk_object_get_stats(
             switchId,
             object_type,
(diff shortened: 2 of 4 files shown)


jimmyzhai (dismissed):

Junchao-Mellanox:
Hi @kcudnik , can we ignore the coverage checker? I don't think we can cover that.

kcudnik:
no we can't merge without coverage fail, since it will show up on monthly report and i will get into trouble based on that, this test must pass

kcudnik:
why you need to add ifdef for those stats ?

kcudnik:
> Some vendors do not have `sai_bulk_object_get_stats` and `sai_bulk_object_clear_stats`, it would cause compile issue. So, I have to use ifdef to make those vendor pass the compilation. Could you suggest a way to cover the #ifdef?

but why this is vendor specific? you are adding this in virutal switch not for specific vendor are you?
if you want to test specific fidef, then extra tests would need to be added for example in unitests/vslib/ and new executable would need to be added where you specify extra defines to c compiler

kcudnik (approved):

liat-grozovik:
@Junchao-Mellanox could you please check the LGTM failure?

Junchao-Mellanox:
> @Junchao-Mellanox could you please check the LGTM failure?

Not an issue introduced by this PR, I have sent email to sonic admin add cc'ed you.

kcudnik:
its probably missing library for swss-common on lgtm container build docker needs to be updated

kcudnik:
i merged my fix, so if you want to retrigger you would need to submit some dummy code change since re-run from is not working somehow

Junchao-Mellanox:
Thanks, will re-trigger

kcudnik (approved):

Issue #1106, opened 2022-08-19
Zmq function failed with EINTR errno when log rotation happened


mint570:
In orchagent, the [SIGHUP handler](https://github.com/sonic-net/sonic-swss/blob/9eb44220819acca2fb8d4740443aee53f1f77061/orchagent/main.cpp#L325) is used for log rotation purpose. When signaling happens, zmq functions might fail with EINTR errno.
We should do retry in [ZeroMQChannel.cpp](https://github.com/sonic-net/sonic-sairedis/blob/ffc410969a907836c4d66165ab4738b9a54a6e65/lib/ZeroMQChannel.cpp) if zmq functions fail with EINTR.

kcudnik:
when did you catched this ?

mint570:
> when did you catched this ?

We recently enabled zmq in our device and found this issue.

PR #1109, opened 2022-08-22
Add retry on zmq functions if fail with EINTR.

mint570:
Fixed https://github.com/sonic-net/sonic-sairedis/issues/1106

### DIFF
--- a/lib/ZeroMQChannel.cpp
+++ b/lib/ZeroMQChannel.cpp
@@ -247,31 +247,7 @@ void ZeroMQChannel::del(
 
     std::vector<swss::FieldValueTuple> values;
 
-    swss::FieldValueTuple opdata(key, command);
-
-    values.insert(values.begin(), opdata);
-
-    std::string msg = swss::JSon::buildJson(values);
-
-    SWSS_LOG_DEBUG("sending: %s", msg.c_str());
-
-    for (int i = 0; true ; ++i)
-    {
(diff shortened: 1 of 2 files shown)


kcudnik:
please fix build errors

mint570:
Build was fine but the vs test failed.
I didn't know that the sairedis repo also run the vs test.

The vs test failure is most likely https://github.com/sonic-net/sonic-swss/issues/2330.
Recently we submitted https://github.com/sonic-net/sonic-swss/pull/2237, which added a few more pytests. Without the fix, vs test will almost guarantee fail.

Also, about code coverage, I don't think zmq is enabled in vs test. So after we fix the vs test, this PR might still fail in code coverage.

mint570:
Correction.
https://github.com/sonic-net/sonic-swss/issues/2330 shouldn't be the issue for sairedis because coverage is not enabled here.
Will re-run the test and see.
I don't think zmq is enabled in vs docker. So this change should have zero impact on the vs test.

kcudnik:
you dont need entire VS for testing, you can use unittests and test only zmqchannel class, check in sairedis/unitests/ directory

mint570:
@kcudnik 
This PR is ready. Please take a look.

kcudnik (approved):

mint570:
@kcudnik can this PR be merged now?

PR #1110, opened 2022-08-22
[sairedis] Replace memset functions

maipbui:
Signed-off-by: maipbui <maibui@microsoft.com>

Why I did it
`memset()` is an insecure function that can cause buffer overflow.

How I did it
Replace `memset()` by `memset_s()`


lib/Switch.cpp
```diff
@@ -43,7 +42,11 @@ void Switch::clearNotificationsPointers()
 {
     SWSS_LOG_ENTER();
 
+#ifdef __STDC_LIB_EXT1__
+    memset_s(&m_switchNotifications, sizeof(m_switchNotifications), 0, sizeof(m_switchNotifications));
```
qiluo-msft, line 46: >memset_s [](http://example.com/codeflow?start=4&length=8)

Let's find a generic solution for the pattern:
```
memset_s(&obj, sizeof(obj), 0, sizeof(obj));
memset_s(buf, sizeof(buf), 0, sizeof(buf)); // buf is array, not pointer.
```

kcudnik:
can we use memset_s in all places instead doing ifdef ?

maipbui:
> can we use memset_s in all places instead doing ifdef ?

I don't think so, seems like gcc/g++ doesn't support memset_s

qiluo-msft (approved):

qiluo-msft:
@kcudnik Could you review again?

kcudnik (approved):

kcudnik:
please satisfy code coverage: https://dev.azure.com/mssonic/build/_build/results?buildId=147452&view=codecoverage-tab

kcudnik:
i added some unittest to cover old code https://github.com/sonic-net/sonic-sairedis/pull/1133

PR #1111, opened 2022-08-23
Run 20 vs tests at a time.

mint570:
In vs test, run 20 tests at a time instead of running all the tests together.
This may help in fixing the test issue.

### DIFF
--- a/.azure-pipelines/test-docker-sonic-vs-template.yml
+++ b/.azure-pipelines/test-docker-sonic-vs-template.yml
@@ -64,7 +64,22 @@ jobs:
       docker ps
       ip netns list
       pushd sonic-swss/tests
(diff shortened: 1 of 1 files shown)


kcudnik (approved):

kcudnik:
> In vs test, run 20 tests at a time instead of running all the tests together.
> This may help in fixing the test issue.

what test issue? can you be more specific here?

mint570:
If we run all the tests together, the last few tests will fail. I don't know the exact reason. It might due to out of memory.
The last few tests are always p4rt tests. We saw this failure in https://github.com/sonic-net/sonic-sairedis/pull/1109.
We recently submitted new development and added more vs tests. So this failure might happen for all PRs.

kcudnik:
ok

mint570:
Can we merge this PR now?

PR #1112, opened 2022-08-23
Add Xsight specific syncd start options

deran1980:
Signed-off-by: Roman Zhurakivsky <rzhurak@larch-networks.com>

**What I did**
Added xsight platform specific options for syncd common

**How I did it**

Modify syncd_init_common.sh script. added a function to support xsight based platforms. 

**How to verify it**
Verify on Accton platform (based on Xsight ASIC)

### DIFF
--- a/syncd/scripts/syncd_init_common.sh
+++ b/syncd/scripts/syncd_init_common.sh
@@ -287,6 +287,55 @@ config_syncd_innovium()
     mkdir -p $II_ROOT
 }
 
+config_syncd_xsight()
+{
+    SYS_MODE="asic"
+    CFG_FILE="/etc/sonic/xlink.cfg"
+    LABEL_REVISION_FILE="/etc/sonic/hw_revision"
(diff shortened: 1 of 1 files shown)


kcudnik (approved):

deran1980:
Hi @kcudnik  - is it going to be merged or we need to wait for sairedis master branch to pass all pipelines tests?

deran1980:
Guess not... 
I can try helping review the failures but it was not introduced as part of this PR.

deran1980:
Hi @kcudnik - I ran the jobs several times and it seems very unstable. 
it always fails as you can see, but every time it fails on different test cases. 
is there a general issue with the stability of the tests or this is a real issue in current master tests that needs to be fixed?
the reason I am asking is because I can see from time to time that there are PRs which are able to pass all tests.

kcudnik:
yes, those test could be very unstable, rest is good so i approve and you can merge

deran1980:
Hi @kcudnik thanks!
But I don't think I have permission to perform merge so I will appreciate if you can merge it. 
Thanks.

PR #1116, opened 2022-08-29
Make changes to building and packaging sairedis

saiarcot895:
This PR makes the following changes:

1. Use Debian build profiles instead of custom build targets to build
different configurations of sairedis. Build profiles were designed for
this purpose. This also makes the debian/rules file a bit cleaner.
2. Rely on the debug packages being automatically created, instead of us
explicitly specifying it in debian/control.
3. Add actual support for excluding Python 2 binding build.
4. Make sure the compile flags used for building Python 2 and Python 3
are actually correct.


debian/libsaimetadata-dev.install
```diff
@@ -1,3 +1,5 @@
 meta/sai*.h usr/include/sai
 meta/Sai*.h usr/include/sai
 SAI/meta/sai*.h usr/include/sai
+usr/lib/*/libsaimetadata.so
+usr/lib/*/libsaimeta.so
```
kcudnik, line 5: should those be in non-dev package?

kcudnik (approved):

debian/libsaimetadata-dev.install
saiarcot895: No, these files are only needed at linking time, not during runtime. At linking time, the SONAME of these libraries (which, I believe, are `libsaimetadata.so.0.0.0` and `libsaimeta.so.0.0.0` for these two libraries) gets recorded in the binary/library, and that file instead gets loaded.

liuh-80:
@saiarcot895 , sonic-swss-common validation break because debian/rules remove 'binary-syncd-vs' in this PR:

https://github.com/sonic-net/sonic-swss-common/blob/master/.azure-pipelines/build-sairedis-template.yml
      DEB_BUILD_OPTIONS=nocheck fakeroot dpkg-buildpackage -b -us -uc -Tbinary-syncd-vs -j$(nproc)


Build error:
 https://dev.azure.com/mssonic/build/_build/results?buildId=165515&view=logs&j=61b7eb0d-78fa-5af5-6df0-b6f82d3891a8&t=f575074c-f5f7-5116-3e48-4a11a237643a

dpkg-buildpackage: info: host architecture amd64
 debian/rules binary-syncd-vs
dh binary-syncd-vs
dh: Unknown sequence binary-syncd-vs (choose from: binary binary-arch binary-indep build build-arch build-indep clean install install-arch install-indep)
make: *** [debian/rules:51: binary-syncd-vs] Error 255
dpkg-buildpackage: error: debian/rules binary-syncd-vs subprocess returned exit status 2

PR #1119, opened 2022-09-05
[lgtm] Add uuid library

kcudnik:
Needed to build swss-common in lgtm container


Pterosaur:
I think this PR should also be merged into 202205. Should we add this label?

kcudnik:
yes, this needs to be ported to this branch

PR #1127, opened 2022-09-16
Manage LANES mapping on VOQ system

abohanyang:
This issue is related to the dynamic port breakout mode support on multi-ASIC platforms. When the port breakout mode changes, SWSS removes ports from LANES mapping and adds new ports to LANES mapping. This logic works properly on the single-chip platform (NPU switch type) but it does not work properly on the multi-ASIC platform (VOQ switch type) because LANES mapping is not created on the multi-ASIC platform.

To fix this issue, my proposal is to create LANES mapping on the multi-ASIC platform.  Though the fix is straightforward, the comment in the existing codebase is a bit concerning to me. I am not sure if creating LANES mapping on the multi-ASIC platform could affect backward compatibility. @arlakshm

### DIFF
--- a/syncd/RedisClient.cpp
+++ b/syncd/RedisClient.cpp
@@ -48,11 +48,20 @@ std::string RedisClient::getRedisLanesKey(
      * NOTE: To support multiple switches LANES needs to be made per switch.
      *
      * return std::string(LANES) + ":" + sai_serialize_object_id(m_switch_vid);
+     *
+     * Only switch with index 0 and global context 0 will have key "LANES" for
+     * backward compatibility. We could convert that during runtime at first
+     * time.
      */
 
     auto index = VidManager::getSwitchIndex(switchVid);
 
     auto context = VidManager::getGlobalContext(switchVid);
+
+    if (index == 0 && context == 0)
+    {
+        return std::string(LANES);
+    }
 
     return (LANES ":") + sai_serialize_object_id(switchVid);
 }
--- a/syncd/RedisClient.cpp
+++ b/syncd/RedisClient.cpp
@@ -48,20 +48,11 @@ std::string RedisClient::getRedisLanesKey(
      * NOTE: To support multiple switches LANES needs to be made per switch.
      *
      * return std::string(LANES) + ":" + sai_serialize_object_id(m_switch_vid);
-     *
-     * Only switch with index 0 and global context 0 will have key "LANES" for
-     * backward compatibility. We could convert that during runtime at first
-     * time.
      */
 
     auto index = VidManager::getSwitchIndex(switchVid);
 
     auto context = VidManager::getGlobalContext(switchVid);
-
-    if (index == 0 && context == 0)
-    {
-        return std::string(LANES);
-    }
 
     return (LANES ":") + sai_serialize_object_id(switchVid);
 }
--- a/syncd/SaiSwitch.cpp
+++ b/syncd/SaiSwitch.cpp
@@ -62,7 +62,10 @@ SaiSwitch::SaiSwitch(
 
     helperInternalOids();
 
-    helperCheckLaneMap();
+    if (getSwitchType() == SAI_SWITCH_TYPE_NPU || getSwitchType() == SAI_SWITCH_TYPE_VOQ)
+    {
+        helperCheckLaneMap();
+    }
 
     helperLoadColdVids();
 
--- a/syncd/SaiSwitch.cpp
+++ b/syncd/SaiSwitch.cpp
@@ -62,10 +62,7 @@ SaiSwitch::SaiSwitch(
 
     helperInternalOids();
 
-    if (getSwitchType() == SAI_SWITCH_TYPE_NPU)
-    {
-        helperCheckLaneMap();
-    }
+    helperCheckLaneMap();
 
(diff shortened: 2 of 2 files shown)


syncd/RedisClient.cpp
```diff
@@ -48,21 +48,12 @@ std::string RedisClient::getRedisLanesKey(
...
 
     auto index = VidManager::getSwitchIndex(switchVid);
 
     auto context = VidManager::getGlobalContext(switchVid);
 
-    if (index == 0 && context == 0)
-    {
-        return std::string(LANES);
-    }
-
```
kcudnik, line 65: removing this will break compatibility with existing platforms and this solution already provides suppor for multiple asic

kcudnik (changes requested):

syncd/RedisClient.cpp
kcudnik: why remove this will solve antything

abohanyang: Hi Kamil - Thank you for taking a look. I have a Linecard with two ASICs. I think I need to change this piece of code because LANES mapping is needed for both syncd0 and syncd1. Without this change, the LANES mapping is created for only one chip and the port change on the second chip will fail because LANES mapping does not exist.

kcudnik: if you have second syncd, then you need to have separate asic database for it, you cant use the same asic db for both, this is not supported, not only on lanes, but on many other components in entire syncd code

kcudnik: i dont understand your problem, since you have 2 different namespaces for syncds there is no conflict there, and lanes mappings should be created correctly, if index and context is not zero, then just oid will be added, but for backward compatibility for index zero and context zero just LANES will be name

if you are saying that lanes are not created then the entry in DB is missong or what do you have in mind here?

abohanyang: I think I understand your point now. the change in getRedisLanesKey might not be needed. I revert them in the latest commit.

abohanyang:
Hi @kcudnik , Could you take another look at the latest commit and see if it looks good to you? The code analysis fails but I don't think it's related to my change.

kcudnik (dismissed):

arlakshm (dismissed):

rlhui:
@shyam77git, @anamehra - would you like to review this as well?  thanks

anamehra:
Hi @abohanyang   How is this change tested? Could you please add any details on the test and logs?

syncd/SaiSwitch.cpp
```diff
@@ -62,10 +62,7 @@ SaiSwitch::SaiSwitch(
 
     helperInternalOids();
 
-    if (getSwitchType() == SAI_SWITCH_TYPE_NPU)
-    {
-        helperCheckLaneMap();
-    }
+    helperCheckLaneMap();
```
rlhui, line 65: why did the original check of SAI_SWITCH_TYPE_NP should be removed, how it prevented lane mapping for multi-asic?

shyam77git:
@abohanyang 
Don't see mention of External-PHY/ gbsyncd in this.
Did this change consider boards/LCs with External-PHY (gbsyncd vertical)? be it single-ASIC or multi-ASIC?
If not, how to assess impact of this change to boards/LCs having External-PHY?

abohanyang:
> @abohanyang Don't see mention of External-PHY/ gbsyncd in this. Did this change consider boards/LCs with External-PHY (gbsyncd vertical)? be it single-ASIC or multi-ASIC? If not, how to assess impact of this change to boards/LCs having External-PHY?

Actually, we just found a new issue last week when we test this change against a system with BCM54182 PHY (using broncos SAI). I submit the new change and you can see that helperCheckLaneMap is called only for NPU and VOQ. This change won't have any impact on SAI_SWITCH_TYPE_PHY.

syncd/SaiSwitch.cpp
abohanyang: to support port breakout, I think both NPU system and VOQ system need to create the lane mapping.

abohanyang: I forget to mention that the modular chassis uses SAI_SWITCH_TYPE_VOQ, not NPU.

arlakshm:
@anamehra @shyam77git , can you please review this PR and signoff if no more comments?

arlakshm:
@anamehra @shyam77git , can you please review this PR again and signoff if no more comments?

abdosi (approved):

abdosi:
can we update PR title to
`Manage LANES mapping on VOQ system`. VOQ can be single/multi-asic.

mlok-nokia:
@rlhui The change looks ok to me.

rlhui:
> can we update PR title to `Manage LANES mapping on VOQ system`. VOQ can be single/multi-asic.

@abohanyang , if you could please just update the title to reflect the exact scope of this, then we can merge this, thanks.

abohanyang:
> > can we update PR title to `Manage LANES mapping on VOQ system`. VOQ can be single/multi-asic.
> 
> @abohanyang , if you could please just update the title to reflect the exact scope of this, then we can merge this, thanks.

Done

rlhui (approved):

PR #1129, opened 2022-09-19
[gbsyncd] Add asic db prefix for channel NOTIFICATIONS

jimmyzhai:
Fix https://github.com/sonic-net/sonic-sairedis/issues/1131.

The communication channel between orchagent and gbsyncd needs different from the channel "NOTIFICATIONS" between orchagent and syncd.


lib/sairediscommon.h
```diff
@@ -67,6 +67,11 @@
  */
 #define REDIS_TABLE_NOTIFICATIONS   "NOTIFICATIONS"
 
+/**
+ * @brief Table which will be used to forward notifications from gbsyncd.
+ */
+#define REDIS_TABLE_GBNOTIFICATIONS "GBNOTIFICATIONS"
```
kcudnik, line 73: why GB needs different communication channel? channels are based on databases, for example ASIC_DB, so if GB will have different database then it will have separate communication channel for notifications and they will not colide

kcudnik (changes requested):

lib/sairediscommon.h
kcudnik: also this could be generalizaed by chaning notifications channel name by prefixing it by db name
for example m_dbASIC + REDIS_TABLE_NOTIFICATIONS, then it would be more generic, and i  think this would be proffered way to make this change

jimmyzhai: The redis channel is global per instance. Do you prefer to the below?

```
#define REDIS_GB_ASIC_NOTIFICATIONS "GB_ASIC_NOTIFICATIONS"
```

"NOTIFICATIONS" has been widely referred as asic channel. I'll not rename it.

kcudnik: i think more preffered way would be to use that asic db prefix

jimmyzhai: Add `REDIS_TABLE_NOTIFICATIONS_PER_DB(dbName)`

kcudnik (approved):

Blueve (approved):

PR #1130, opened 2022-09-19
[syncd] bulk OID remove requires RID

yenlu-keith:
the feature of "SAI bulk OID" on removing requires RID, not VID


prsunny (approved):

kcudnik (approved):

yenlu-keith:
could anyone help me to merge this? I do not have write access right.

PR #1133, opened 2022-09-22
[unittests] Add unittests for SwitchStateBaseHostif

kcudnik:


vslib/SwitchStateBaseHostif.cpp
```diff
@@ -46,7 +47,7 @@ int SwitchStateBase::vs_create_tap_device(
 
     const char *tundev = "/dev/net/tun";
 
-    int fd = open(tundev, O_RDWR);
+    int fd = saimeta::System::open(tundev, O_RDWR);
```
qiluo-msft, line 50: >open [](http://example.com/codeflow?start=30&length=4)

What is the value of adding this function? Improve coverage number?

kcudnik: correct

kcudnik: wrapper on system function is need to simulate function succes on unittests

qiluo-msft: Is the performance a concern? Does debug build help the coverage anyhow?

kcudnik: there is no performance impact since those apis are not on critical paths, coverage is achieved in places required by https://github.com/sonic-net/sonic-sairedis/pull/1110

kcudnik:
ut fail probably not related to the change

PR #1134, opened 2022-09-28
Add mdio IPC client library

jiahua-wang:
Signed-off-by: Jiahua Wang <jiahua.wang@broadcom.com>
This PR is to replace https://github.com/sonic-net/sonic-buildimage/pull/11284

Why I did it
The syncd has a MDIO IPC server to help the access of external PHYs on the NPU MDIO bus. A MDIO IPC client library is required to communicate to the MDIO IPC server. The gbsyncd will pass the function pointers for the function in the client library to the PAI.

How I did it
MDIO IPC client functions are added to connect to the MDIO IPC server socket, to send request and to wait for response IPC messages. The client functions are made into a .so library. The .so library is packaged in the syncd debian package.

How to verify it
The client functions are also available in as a .a library. The unit test code can use the code to test both IPC server and client.


kcudnik:
please add more description about this PR

jiahua-wang:
> please add more description about this PR

Description added

syncd/MdioIpcClient.cpp
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    pthread_mutex_lock(&mutex);
+
+    sprintf(cmd, "mdio 0x%x 0x%x\n", mdio_addr, reg_addr);
+    if (syncd_mdio_ipc_command(cmd, resp) == 0)
+    {
+        *data = (uint32_t)strtoul(strchrnul(resp, ' ') + 1, NULL, 0);
+        rc = SAI_STATUS_SUCCESS;
+    }
```
jimmyzhai, line 144: If syncd_mdio_ipc_command(cmd, resp) < 0, for debugging, should at least add log for error code. Better to translate the error code to SAI status code on `rc`.

kcudnik:
you should add unittests for this code

syncd/MdioIpcClient.cpp
```diff
@@ -0,0 +1,224 @@
...
+#include <fcntl.h>
+#include <sys/socket.h>
+#include <sys/un.h>
+#include <unistd.h>
+#include <time.h>
+#include <pthread.h>
+
+#include "MdioIpcClient.h"
+
+#define SAI_IPC_SOCK_DIR_SYNCD      "/var/run/sswsyncd"
```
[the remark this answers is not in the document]
jiahua-wang: will do that

syncd/MdioIpcClient.cpp
```diff
@@ -0,0 +1,224 @@
...
+        sock = 0;
+    }
+
+    if (strlen(path) == 0)
+    {
+        strcpy(path, SAI_IPC_SOCK_DIR_SYNCD);
+        fd = open(path, O_DIRECTORY);
+        if (fd < 0)
+        {
+            printf("%s: Program is not run on host\n", __func__);
```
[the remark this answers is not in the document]
jiahua-wang: will do that

syncd/MdioIpcClient.cpp
```diff
@@ -0,0 +1,224 @@
...
+#include "MdioIpcClient.h"
+
+#define SAI_IPC_SOCK_DIR_SYNCD      "/var/run/sswsyncd"
+#define SAI_IPC_SOCK_DIR_HOST       "/var/run/docker-syncd"
+#define SAI_IPC_SOCK_FILE           "mdio-ipc"
+#define SAI_IPC_BUFF_SIZE           256
+
+#define CONN_TIMEOUT                25     /* shorter than 30 sec on server side */
+
+static pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
```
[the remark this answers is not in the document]
jiahua-wang: Will change code to use std::mutex

syncd/MdioIpcClient.cpp
```diff
@@ -0,0 +1,224 @@
...
+/* Function to read data from MDIO interface */
+extern "C" sai_status_t mdio_read(uint64_t platform_context, uint32_t mdio_addr, uint32_t reg_addr,
+        uint32_t number_of_registers, uint32_t *data)
+{
+    int rc = SAI_STATUS_FAILURE;
+    char cmd[SAI_IPC_BUFF_SIZE], resp[SAI_IPC_BUFF_SIZE];
+
+    if (number_of_registers > 1)
+    {
+        printf("Multiple register reads are not supported, num_of_registers: %d\n", number_of_registers);
```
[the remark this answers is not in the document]
jiahua-wang: This is an error check for multiple reasons. The caller of the mdio ipc client is the PAI. The mdio read function prototype defined for PAI has already determined the mdio read single register. The mdio ipc server also determines the mdio read/write single register. The IPC message format also determines the mdio access single register.

syncd/MdioIpcClient.cpp
jiahua-wang: Will add error log as the else condition statement.

kcudnik:
please add unittests to satisfy code coverage

jiahua-wang:
> 

I am working on the unit test.

kcudnik (dismissed):

syncd/Makefile.am
```diff
@@ -80,6 +80,10 @@ if SONIC_ASIC_PLATFORM_BROADCOM
 libSyncd_a_CXXFLAGS += -DMDIO_ACCESS_USE_NPU
 endif
 
+if SONIC_ASIC_PLATFORM_VS
```
kcudnik, line 83: you are changind default behaviour by this, since by default vs switch is not considered as npu switch

jiahua-wang: Because some code in MdioIpcServer.cpp is compiled with the condition MDIO_ACCESS_USE_NPU and seems not run in unit test, I tried to add the condition to the vs platform but it did not work. I will revert the change since it does not help the unit test.

syncd/MdioIpcCommon.h
```diff
@@ -0,0 +1,9 @@
+#define SYNCD_IPC_SOCK_SYNCD  "/var/run/sswsyncd"
+#define SYNCD_IPC_SOCK_HOST   "/var/run/docker-syncd"
+#define SYNCD_IPC_SOCK_FILE   "mdio-ipc"
+#define SYNCD_IPC_BUFF_SIZE   256    /* buffer size */
+
+#define MDIO_SERVER_TIMEOUT   30     /* sec, connection timeout */
+#define MDIO_CLIENT_TIMEOUT   25     /* shorter than 30 sec on server side */
+
+#define CONN_MAX              18     /* max. number of connections */
```
kcudnik, line 9: prefix this by MDIO_

jiahua-wang: Will change CONN_MAX to MDIO_CONN_MAX.

syncd/MdioIpcServer.cpp
```diff
@@ -56,6 +50,11 @@ MdioIpcServer::MdioIpcServer(
 
     /* globalContext == 0 for syncd, globalContext > 0 for gbsyncd */
     MdioIpcServer::m_syncdContext = (globalContext == 0);
+#ifdef MDIO_ACCESS_USE_NPU
+    MdioIpcServer::m_accessUseNPU = true;
```
jimmyzhai, line 54: Why not directly `m_accessUseNPU = true` ? It is not defined as a static class member.

Since member `m_accessUseNPU` is added, macro `MDIO_ACCESS_USE_NPU` looks redundant and needs to be removed. Its value can be assigned by adding a parameter like `accessUseNPU` in the constructor.

syncd/MdioIpcClient.cpp
```diff
@@ -0,0 +1,251 @@
...
+    {
+        SWSS_LOG_ERROR("syncd_mdio_ipc_command returns : %d\n", rc);
+    }
+
+    ipcMutex.unlock();
+    return rc;
+}
+
+/* Function to read data using clause 22 from MDIO interface */
+extern "C" sai_status_t mdio_read_cl22(uint64_t platform_context, uint32_t mdio_addr, uint32_t reg_addr,
```
kcudnik, line 189: all those extern C functions should be declared in extern C in header file, and not explicitly here

kcudnik: which they are, so this extern should be redundant here

syncd/MdioIpcServer.cpp
jiahua-wang: MDIO_ACCESS_USE_NPU is only defined in syncd/Malefile.am for "make configure PLATFORM=broadcom". It is only tested on broadcom ASIC. When other platforms have tested the feature, the syncd/Makefile.am can be modified to add the feature. Actually, if m_accessUseNPU = true for vs platform, we see vssyncd test logs error messages.

jiahua-wang:
This PR is replaced by https://github.com/sonic-net/sonic-sairedis/pull/1230

syncd/MdioIpcServer.cpp
jimmyzhai: The value of `m_accessUseNPU`  has indicated feature enablement, which could be TRUE if 

1. platform support, currently only broadcom
2. mock MdioIpcServer instance in unittest

jiahua-wang: Yes, The value of m_accessUseNPU is TRUE if platform support, currently only broadcom. For mock MdioIpcServer instance in unittest, the value of m_accessUseNPU is changed to be true at run time by the function MdioIpcServer::setIpcTestMode().

PR #1143, opened 2022-11-01
Updated logging level for un-supported counters messages

vmittal-msft:
Updating logging level for unsupported counters messages as it keeps on filling syslogs.


arlakshm (approved):

rlhui:
@vmittal-msft , thanks.  do we need same change in master branch?  Thanks.

PR #1144, opened 2022-11-04
[202012][Recorder]: Acquire lock for ofstream changes

theasianpianist:
While the Recorder is writing to the sairedis log file, it's possible for a log rotate to occur at exactly the right time so that file stream used for writing (a `std::ofstream`) is re-opened in the middle of the write operation (since writing and log rotate are handled by separate threads). Since standard library objects are not thread safe, this can cause some pointers used during the write operation to be overwritten, leading to a segmentation fault when the write operation proceeds.

To prevent this from occurring, acquire a lock for the file stream for any methods that change the ofstream (including opening, closing, and writing to it with the `<<` operator).

Signed-off-by: Lawrence Lee <lawlee@microsoft.com>

### DIFF
--- a/lib/src/Recorder.cpp
+++ b/lib/src/Recorder.cpp
@@ -190,10 +190,7 @@ void Recorder::requestLogRotate()
 
     /* double check since reopen could fail */
 
-    if (m_ofstream.is_open())
-    {
-        m_ofstream << getTimestamp() << "|" << "#|logrotate on: " << m_recordingFile << std::endl;
-    }
+    recordLine("#|logrotate on: " + m_recordingFile);
 }
 
 void Recorder::recordingFileReopen()
--- a/lib/src/Recorder.cpp
+++ b/lib/src/Recorder.cpp
@@ -220,18 +220,20 @@ void Recorder::recordingFileReopen()
 
 void Recorder::startRecording()
 {
-    MUTEX();
 
     SWSS_LOG_ENTER();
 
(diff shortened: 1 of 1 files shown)


prsunny:
@theasianpianist , is this a regression from older branches?

theasianpianist:
> @theasianpianist , is this a regression from older branches?

It looks like it might be, I believe #1083 is responsible for introducing this bug on 202012, and #1058 on master. Prior to this, the log rotate operation and writing to the log file were handled by the same thread/function so it was impossible for this to occur

saiarcot895 (approved):

lib/src/Recorder.cpp
```diff
@@ -221,6 +220,8 @@ void Recorder::recordingFileReopen()
 
 void Recorder::startRecording()
 {
+    MUTEX();
```
jimmyzhai, line 223: Could cause a deadlock. Later while calling `recordLine(...)`, it calls `MUTEX()` again. 

From https://en.cppreference.com/w/cpp/thread/mutex, std::mutex offers exclusive, non-recursive ownership semantics.

theasianpianist: Good catch, I've enclosed the lock operation and all the direct ofstream operations in a scoped block to avoid this.

saiarcot895 (approved):

jimmyzhai (approved):

theasianpianist:
202012 PR for #1145

PR #1145, opened 2022-11-04
[Recorder]: Acquire lock for ofstream changes

theasianpianist:
While the Recorder is writing to the sairedis log file, it's possible for a log rotate to occur at exactly the right time so that file stream used for writing (a std::ofstream) is re-opened in the middle of the write operation (since writing and log rotate are handled by separate threads). Since standard library objects are not thread safe, this can cause some pointers used during the write operation to be overwritten, leading to a segmentation fault when the write operation proceeds.

To prevent this from occurring, acquire a lock for the file stream for any methods that change the ofstream (including opening, closing, and writing to it with the << operator). Also use `recordLine` for all writes to the file stream to avoid deadlock.

Signed-off-by: Lawrence Lee [lawlee@microsoft.com](mailto:lawlee@microsoft.com)

### DIFF
--- a/lib/Recorder.cpp
+++ b/lib/Recorder.cpp
@@ -189,10 +189,7 @@ void Recorder::requestLogRotate()
 
     /* double check since reopen could fail */
 
-    if (m_ofstream.is_open())
-    {
-        m_ofstream << getTimestamp() << "|" << "#|logrotate on: " << m_recordingFile << std::endl;
-    }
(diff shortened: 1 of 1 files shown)


saiarcot895 (dismissed):

saiarcot895 (approved):

theasianpianist:
> mutex t here could cause deadlock

which line are you referring to?

PR #1173, opened 2022-12-19
Fix for [EVPN] When MAC moves from remote end point to local, ASIC DB fields are not updated properly for the mac #11503

anilkpan:
Issue:
When a mac moves from remote to local, the local mac is set in ASIC db using hset, which updates common fields, but does not remove fields specific to remote mac.

Fix:
In case of a mac move event, remove the existing FDB entry from ASIC DB first and then set the new one.


anilkpan:
retest this please

dgsudharsan (approved):

yxieca (approved):

PR #1174, opened 2022-12-19
Switch to using stock gcovr 5.2

saiarcot895:
The custom branch we were using previously has since been deleted. That branch appears to have some fix for searching for the source file (for a gcda file) recursively. I don'y know if it's needed or not today, but using the stock gcovr 5.2 (from the official repo) appears to work.


yxieca (approved):

bingwang-ms:
I opened another PR to backport the change into 202205 branch. https://github.com/sonic-net/sonic-sairedis/pull/1190

PR #1194, opened 2023-01-05
Reland 'Make changes to building and packaging sairedis (#1116)'

saiarcot895:
The vstest issues causing random test failures have since been fixed, and were not related to the changes in this PR.

* Make changes to building and packaging sairedis

This commit includes the following changes:
1. Use Debian build profiles instead of custom build targets to build different configurations of sairedis. Build profiles were designed for this purpose. This also makes the debian/rules file a bit cleaner.
2. Rely on the debug packages being automatically created, instead of us explicitly specifying it in debian/control.
3. Add actual support for excluding Python 2 binding during build.
4. Make sure the compile flags used for building Python 2 and Python 3 are actually correct.

* Update pipeline file

* Fix argument for profile

* Exclude libsai package from the list of dependencies

* Allow multiple syslog artifacts during tests

Microsoft ADO: 24405994


saiarcot895:
@qiluo-msft @kcudnik could you review this when you get a chance?

debian/libsaimetadata-dev.install
```diff
@@ -1,3 +1,5 @@
 meta/sai*.h usr/include/sai
 meta/Sai*.h usr/include/sai
 SAI/meta/sai*.h usr/include/sai
+usr/lib/*/libsaimetadata.so
```
kcudnik, line 4: why suddenly dev package contains so library? should that be in normal package? and dev should contain headers only?

saiarcot895: The normal package should just have the so library with the SONAME (i.e. `libsaimetadata.so.0`). The dev package should contain the headers as well as a symlink from the basic library name (`libsaimetadata.so`) to the actual so library (so a symlink from `libsaimetadata.so` to `libsaimetadata.so.0`). This line that's being added is a symlink.

See also https://github.com/sonic-net/sonic-sairedis/pull/1116#discussion_r1003510714

saiarcot895:
@kcudnik Can you review this when you get a chance?

configure.ac
```diff
@@ -111,13 +119,31 @@ if test "x$asan_enabled" = "xtrue"; then
...
+
 AM_CONDITIONAL(ASAN_ENABLED, test x$asan_enabled = xtrue)
+AM_CONDITIONAL(PYTHON2, test x$python2 = xtrue)
 
 AC_PATH_PROGS(SWIG, [swig3.0 swig])
 
 CXXFLAGS_COMMON=""
 CXXFLAGS_COMMON+=" -ansi"
 CXXFLAGS_COMMON+=" -fPIC"
+CXXFLAGS_COMMON+=" -pipe"
```
kcudnik, line 146: why this is added? something wrong with temporary files ?

kcudnik (approved):

configure.ac
saiarcot895: Hmm, I don't remember why I added this. At best, this might give a bit of performance/space benefit since intermediate files will all get transferred in memory rather than needing to be saved to a file first. Using temporary files shouldn't cause any issue, unless disk space on `/tmp` is very low.

PR #1196, opened 2023-01-23
Use new value of STATE_DB FAST_REBOOT entry

arfeigin:
Update syncd_init_common to check if fast-reboot is enabled according to the new value for FAST_REBOOT entry in STATE_DB.

This PR should come along with the following PRs:
https://github.com/sonic-net/sonic-utilities/pull/2621
https://github.com/sonic-net/sonic-buildimage/pull/13484
https://github.com/sonic-net/sonic-swss-common/pull/742
https://github.com/sonic-net/sonic-platform-daemons/pull/335

This set of PRs solves the issue https://github.com/sonic-net/sonic-buildimage/issues/13251

### DIFF
--- a/syncd/scripts/syncd_init_common.sh
+++ b/syncd/scripts/syncd_init_common.sh
@@ -49,7 +49,8 @@ case "$(cat /proc/cmdline)" in
     ;;
   *SONIC_BOOT_TYPE=fast*|*fast-reboot*)
     # check that the key exists
-    if [[ $(sonic-db-cli STATE_DB GET "FAST_REBOOT|system") == "1" ]]; then
+    SYSTEM_FAST_REBOOT=`sonic-db-cli STATE_DB GET "FAST_REBOOT|system"`
+    if [[ ${SYSTEM_FAST_REBOOT} == "enable" ]]; then
        FAST_REBOOT='yes'
     else
        FAST_REBOOT='no'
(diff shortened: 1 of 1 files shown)


stepanblyschak (approved):

syncd/scripts/syncd_init_common.sh
```diff
@@ -49,7 +49,8 @@ case "$(cat /proc/cmdline)" in
     ;;
   *SONIC_BOOT_TYPE=fast*|*fast-reboot*)
     # check that the key exists
-    if [[ $(sonic-db-cli STATE_DB GET "FAST_REBOOT|system") == "1" ]]; then
+    SYSTEM_FAST_REBOOT=`sonic-db-cli STATE_DB GET "FAST_REBOOT|system"`
```
saiarcot895, line 52: Vaibhav and I briefly discussed this change. Since for warm reboot, the name of the table there is `WARM_RESTART_ENABLE_TABLE`, and this key is anyways being changed to use enable/disable instead of 1/not exist, it would be better for consistency purposes to rename the key everywhere to be `FAST_RESTART_ENABLE_TABLE`.

vaibhavhd (approved):
Please wait for @saiarcot895 review.

saiarcot895 (approved):

saiarcot895 (approved):

PR #1198, opened 2023-01-27
[sai_failure_dump]Invoking dump during SAI failure

dgsudharsan:
HLD: https://github.com/sonic-net/SONiC/pull/1212

**What I did**
Added logic to invoke SAI failure dump during any SAI programming failure before invoking abort by orchagent.

**Why I did it**
To collect necessary dumps in problem state in syncd before abort is called and all processes restarts

**How I verified it**
Manual verification. Added UT to cover abort scenario as well.


lib/RedisRemoteSaiInterface.cpp
```diff
@@ -13,6 +13,7 @@
 #include "meta/SaiAttributeList.h"
 #include "meta/PerformanceIntervalTimer.h"
 #include "meta/Globals.h"
+#include "swss/exec.h"
```
prsunny, line 16: why is this included. Please remove if not required

lib/RedisRemoteSaiInterface.cpp
```diff
@@ -276,6 +277,7 @@ sai_status_t RedisRemoteSaiInterface::setRedisExtensionAttribute(
         _In_ const sai_attribute_t *attr)
 {
     SWSS_LOG_ENTER();
+    std::string ret_str;
```
prsunny, line 280: i dont see this being used. Please remove

syncd/Syncd.cpp
```diff
@@ -3273,7 +3290,6 @@ sai_status_t Syncd::processNotifySyncd(
     {
         SWSS_LOG_NOTICE("very first run is TRUE, op = %s", key.c_str());
 
-        sai_status_t status = SAI_STATUS_SUCCESS;
```
prsunny, line 3276: Please revert back the changes not relevant to this PR

syncd/tests.cpp
```diff
@@ -157,6 +176,7 @@ void test_enable_recording()
     ASSERT_SUCCESS("Failed to enable recording");
 }
 
+
```
prsunny, line 179: Extra line

syncd/Syncd.cpp
dgsudharsan: This change is needed because status is being defined above that would lead to compilation issue

prsunny (approved):

kcudnik:
why use extra SAI_REDIS_NOTIFY_SYNCD_INVOKE_DUMP enum instead of actually calling sai_dump api ? that would do exactly the same in more elegant way?

dgsudharsan:
> why use extra SAI_REDIS_NOTIFY_SYNCD_INVOKE_DUMP enum instead of actually calling sai_dump api ? that would do exactly the same in more elegant way?

Hi Kamil. I believe you are referring to saisdkdump which also covers lower layer information. Currently only mellanox platforms collect this information and other vendors may not have implemented it. https://github.com/sonic-net/sonic-utilities/blob/7a604c51671a85470db3d15aaa83b6b39a01531a/scripts/generate_dump#L1075
Other vendors invoke their own proprietary debug dump information as seen in the above generate_dump script. So there needs to be action taken by all vendors to standardize this.

On a note, this feature intends to collect dump immediately after SAI failure before services restart. In order to accommodate all vendors today we took the approach. Once all SAI vendors support the debug dump functionality we can standardize this.

This was also brought up during the HLD discussion and it was decided to take it in the SAI community meeting https://github.com/sonic-net/SONiC/pull/1212#issuecomment-1377827163

PR #1214, opened 2023-03-01
[dash] Add support of DASH API

oleksandrivantsiv:
**What I did**

- Implement DASH API
- Implement DASH API unit tests
- Implement VS switch for Nvidia DPI
- Implement extended unit tests for DASH API

**Why I did it**

Add support for DASH API extension.

**How I verified it**

- Run sairedis unit tests
- Run tests/NVDAMBF2H536C.pl script with extended unit tests for DASH API
- Compile and run `testdash_gtest` test application.

**Details if related**


Yakiv-Huryk (approved):

prsunny:
@rlhui for viz

liat-grozovik:
@marian-pritsak @prsunny I believe this should go to a specific dash branch as long as sonic is not aligned with newer sai version.
what should be the next step here?

kcudnik:
very very very complex change, i looked at it breafly for 30min, and most worring is Meta.cpp and Serialize.cpp, since most complex changed are there, other stuff looks fairly good, but it could be splitted to multiple changes, like 1 dash per PR and also adding new switch type nvdia, also as separate PR, pushing all into one big change, it probably contains some bugs and could potentially break something that is working now

PR #1222, opened 2023-03-09
[ci] Continue on error when running test.

liushilongbuaa:


.azure-pipelines/test-docker-sonic-vs-template.yml
```diff
@@ -95,7 +96,8 @@ jobs:
       fi
       rm -rf $(Build.ArtifactStagingDirectory)/download
     displayName: "Run vs tests"
-    continueOnError: ${{ parameters.asan }}
+    ${{ if ne(variables['Build.Reason'], 'PullRequest') }}:
+      continueOnError: true
```
saiarcot895, line 100: Can `continueOnError: ${{ parameters.asan }}` be preserved in the else case? This is so that if the ASAN test run fails for whatever reason in a PR build, merging is still allowed.

liushilongbuaa: fixed

PR #1216, opened 2023-03-07
Ignore removing switch for mellanox platform due to known limitation

Junchao-Mellanox:
Why I did this?

Due to a limitation in mellanox platform, sonic ignores syncs graceful shutdown in syncd.sh previously. However, ignoring graceful shutdown in syncd.sh also ignores closing other resources such as removing all counters. 
This PR is going to ignore remove_switch for mellanox platform when shutdown type is cold. It make sure other shutdown operations are properly executed in shutdown flow.

How I test this?

Running config reload on many platforms and check result.

PR #1232, opened 2023-04-24
[202211] Revert "Ignore removing switch for mellanox platform due to known limitation (#1216)"

Junchao-Mellanox:
What I did?

This reverts commit 62aa2448328f50815336561e141d076cf03d4087.

Why I did this?

This commit introduces a race condition with would cause syncd crash. The issue happens when:

Syncd.m_sn has been freed
SAI send a port state change notification which will access Syncd.m_sn ===> causes crash


kcudnik (approved):

keboliu:
@StormLiangMS  would you please help merge?

PR #1240, opened 2023-05-16
[syncd] Add pre match logic for acl entry

kcudnik:
this will test scenario where we have 1 acl entry with counter and after
warm boot 2 acl entries and 2 counters the same which could end up
matching wrong acl counter and causing fail "object exist" on broadcom
acl entry set, which will require improve pre match logic for acl entry


syncd/ComparisonLogic.cpp
```diff
@@ -2806,6 +2806,53 @@ void ComparisonLogic::cretePreMatchForLagMembers(
...
+
+        for (auto& tAclEntry: tAclEntries)
+        {
+            if (processed.find(tAclEntry->m_str_object_id) != processed.end())
+                continue;
+
+            auto cPrio = cAclEntry->getSaiAttr(SAI_ACL_ENTRY_ATTR_PRIORITY);
+            auto tPrio = tAclEntry->getSaiAttr(SAI_ACL_ENTRY_ATTR_PRIORITY);
+
+            if (cPrio->getStrAttrValue() != tPrio->getStrAttrValue())
```
vaibhavhd, line 2834: Rookie question: are ACL priorities guaranteed to be always unique? If not, then we will likely run into same issue again.

syncd/ComparisonLogic.cpp
```diff
@@ -3045,7 +3094,8 @@ void ComparisonLogic::applyViewTransition(
     {
         // TODO make generic list of root objects (or meta SAI for this)
 
-        if (obj.second->getObjectType() == SAI_OBJECT_TYPE_ACL_TABLE_GROUP)
+        if (obj.second->getObjectType() == SAI_OBJECT_TYPE_ACL_TABLE_GROUP ||
+            obj.second->getObjectType() == SAI_OBJECT_TYPE_ACL_ENTRY)
```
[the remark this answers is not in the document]
kcudnik: no, since ACL_ENTRY is leaf object

syncd/ComparisonLogic.cpp
kcudnik: no they are not guaranteed in general, but sonic make sure that they are, since we want to control the right order of the acl, so having 2 acls with the same priority not guarantee the order of them

vaibhavhd (dismissed):
Please wait for Ying or Sai's review before merge.

syncd/ComparisonLogic.cpp
```diff
@@ -2929,6 +2976,8 @@ void ComparisonLogic::createPreMatchMap(
     SWSS_LOG_NOTICE("preMatch map size: %zu, tmp oid obj: %zu",
             tmp.m_preMatchMap.size(),
             count);
+
+    cretePreMatchForAclEntries(cur, tmp, processed);
```
saiarcot895, line 2980: Should this be before the log message above, so that the right prematch count gets printed?

kcudnik: yes

kcudnik: fixed

saiarcot895 (approved):

qiluo-msft:

vaibhavhd:
@kcudnik please raise a separate PR for 202012 branch

liuh-80:
Manually cherry-pick PR created: https://github.com/sonic-net/sonic-sairedis/pull/1257

PR #1243, opened 2023-05-18
[syncd] Add the possibility to overwrite the watchdog warning time span via the command line option.

oleksandrivantsiv:
This option is useful on the slower systems where the execution of SAI call might take more than a default time span (30 seconds). The default behavior is preserved if the option is not specified.


kcudnik (dismissed):

kcudnik (dismissed):

unittest/syncd/TestCommandLineOptions.cpp
```diff
@@ -1,9 +1,45 @@
...
+    -l --enableBulk
+        Enable SAI Bulk support
+    -g --globalContext
+        Global context index to load from context config file
+    -x --contextConfig
+        Context configuration file
+    -b --breakConfig
+        Comparison logic 'break before make' configuration file
+    -w --watchdogWarnTimeSpan
+        Watchdog time span to watch for execution
```
kcudnik, line 38: add units in which this timespan is, milliseconds, microsecods etc

kcudnik (approved):

kcudnik (approved):

PR #1244, opened 2023-05-18
[syncd] Extend syncd start script to start SYNCD on nvidia-bluefield.

oleksandrivantsiv:


oleksandrivantsiv:
Depends on https://github.com/sonic-net/sonic-sairedis/pull/1243

kcudnik (approved):

PR #1252, opened 2023-06-16
[warmboot] Add workaround for `INIT_VIEW` failure 

zjswhhh:
**MICROSOFT ADO#: 24107886**

## why change is needed
"SaiGetHardwareInfo" is now returning string "none" instead of empty string (""), which is causing INIT_VIEW failure:
```
May 26 18:57:33.577868 NOTICE syncd#syncd: :- onSwitchCreateInInitViewMode: new switch oid:0x21000000000000 contains hardware info: ''
May 26 18:57:33.577868 ERR syncd#syncd: :- onSwitchCreateInInitViewMode: hardware info mismatch: current 'none' vs new ''
May 26 18:57:33.578017 ERR syncd#syncd: :- run: Runtime error: :- onSwitchCreateInInitViewMode: hardware info mismatch: current 'none' vs new ''
```

Adding this change as a temp workaround before CSP is resolved with BRCM. 

## how I verify 
Test syncd build on lab device, able to pass INIT_VIEW stage. 

sign-off: Jing Zhang zhangjing@microsoft.com

### DIFF
--- a/syncd/Syncd.cpp
+++ b/syncd/Syncd.cpp
@@ -4309,6 +4309,11 @@ void Syncd::onSwitchCreateInInitViewMode(
                 sai_serialize_object_id(switchVid).c_str(),
                 newHw.c_str());
 
+        /*
+         * The line below is added due to a behavior change of SAI call.
+         *
+         * TODO: remove the line when SAI vendor agrees fix on their end.
+        */
         currentHw = currentHw == "none"? "":currentHw;
 
         if (currentHw != newHw)
--- a/syncd/Syncd.cpp
+++ b/syncd/Syncd.cpp
@@ -4309,6 +4309,8 @@ void Syncd::onSwitchCreateInInitViewMode(
                 sai_serialize_object_id(switchVid).c_str(),
                 newHw.c_str());
 
+        currentHw = currentHw == "none"? "":currentHw;
+
         if (currentHw != newHw)
         {
             SWSS_LOG_THROW("hardware info mismatch: current '%s' vs new '%s'", currentHw.c_str(), newHw.c_str());
--- a/syncd/Syncd.cpp
+++ b/syncd/Syncd.cpp
@@ -4313,8 +4313,8 @@ void Syncd::onSwitchCreateInInitViewMode(
          * The line below is added due to a behavior change of SAI call.
          *
          * TODO: remove the line when SAI vendor agrees fix on their end.
-        */
-        currentHw = currentHw == "none"? "":currentHw;
+         */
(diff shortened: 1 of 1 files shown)


syncd/Syncd.cpp
```diff
@@ -4309,6 +4309,8 @@ void Syncd::onSwitchCreateInInitViewMode(
                 sai_serialize_object_id(switchVid).c_str(),
                 newHw.c_str());
 
+        currentHw = currentHw == "none"? "":currentHw;
```
vaibhavhd, line 4312: Please add a comment here explaining why this is needed, and that the behavior change is from SAI call.
Also a TODO to remove this change in the future if SAI vendor agrees to fix this at their end.

vaibhavhd (dismissed):
LGTM, please wait for @kcudnik 's review. Thanks.

syncd/Syncd.cpp
zjswhhh: Done.

saiarcot895 (dismissed):

syncd/Syncd.cpp
```diff
@@ -4309,6 +4309,13 @@ void Syncd::onSwitchCreateInInitViewMode(
                 sai_serialize_object_id(switchVid).c_str(),
                 newHw.c_str());
 
+        /*
+         * The line below is added due to a behavior change of SAI call.
+         *
+         * TODO: remove the line when SAI vendor agrees fix on their end.
+        */
+        currentHw = currentHw == "none"? "":currentHw;
```
yxieca, line 4317: The coding style in this repo seems to require white space before and after ":".

kcudnik: How come suddenly its none? Is it Sai internal vendor?

kcudnik: You miss space on last comment linę before star

vaibhavhd: > How come suddenly its none? Is it Sai internal vendor?

With new SAI code in 202205+ images the return value for GET call on hwinfo is `none`. We have confirmed that older SAI was returning `''`. This will be further  discussed with BRCM SAI team. But, in the meanwhile, we need this workaround to unblock/uncover other issues.

kcudnik (approved):

syncd/Syncd.cpp
zjswhhh: @yxieca @kcudnik Addressed.

vaibhavhd (approved):

yxieca (approved):

PR #1255, opened 2023-06-22
Tunnel Term Table Attributes Tests

mholankar:
Fixed the tests that require SRC_IP attributes for this change https://github.com/opencomputeproject/SAI/pull/1799


mholankar:
@kcudnik these are the tests that I changed.

kcudnik:
we will need to wait with this until we pull new SAI headers

kcudnik:
hey, i'm updating syncd headers to the latest ones, and it requires to add some missing apis, and it wont pass tests because of SAI headers conditions in attributes, so i will let my self to cherrypick this change to my pr, and put everting in one go, so new headers will go, new code updated tests https://github.com/sonic-net/sonic-sairedis/pull/1285

kcudnik:
this is no longer needed when using relaxed flag, i will add extra test case for this scenario in https://github.com/sonic-net/sonic-sairedis/pull/1285

mholankar:
Thank you for taking care of this @kcudnik

PR #1256, opened 2023-06-27
Debug commit to trigger azure ci

vadymhlushko-mlnx:


kcudnik:
please add description

kcudnik:
why you want to make this change ? seems like this is not needed

vadymhlushko-mlnx:
> why you want to make this change ? seems like this is not needed

I just wanted to trigger `sonic-sairedis` `coverity` check for debugging purposes, the changes itself is not relevant

kcudnik:
so is this just dummy PR ?

vadymhlushko-mlnx:
> so is this just dummy PR ?

yes, I will close it

PR #1238, opened 2023-05-09
Fix pipeline issue caused by urllib3 v2

liushilongbuaa:
docker 4.4.1 don't support urllib3 v2.
Update it to latest.

PR #1257, opened 2023-06-28
[syncd] Add pre match logic for acl entry (#1240)

kcudnik:
this will test scenario where we have 1 acl entry with counter and after warm boot 2 acl entries and 2 counters the same which could end up matching wrong acl counter and causing fail "object exist" on broadcom acl entry set, which will require improve pre match logic for acl entry


kcudnik:
seems like build pipeline is broken here

liuh-80:
@xumia @liushilongbuaa , the 202012 branch build pipeline seems break, could you please help check?

saiarcot895:
sonic-net/sonic-swss-common#780 needs to be cherry-picked to 202012 branch (similar to how #1238 was cherry-picked to 202012 branch for this repo).

saiarcot895 (approved):

kcudnik:
> [sonic-net/sonic-swss-common#780](https://github.com/sonic-net/sonic-swss-common/pull/780) needs to be cherry-picked to 202012 branch (similar to how #1238 was cherry-picked to 202012 branch for this repo).

who can fix this ?

saiarcot895:
@kcudnik, I think this can be merged now?

PR #1259, opened 2023-07-01
[vslib]: Temporaily map DPU switch type to NVDA_MBF2H536C

prabhataravind:
NVDA_MBF2H536C HWSKU information is not yet available in sonic-buildimage master and therefore a fake NPU HWKSU is added to sonic-buildimage which is being mapped internally to NVDA_MBF2H536C. This will help to run and pass DASH specific vstests using NPU HWSKU.

Error we are trying to address:

Jun 29 19:34:55.043704 a6a6a10ed28f ERR #syncd: :- parseSwitchType: unknown switch type: 'SAI_VS_SWITCH_TYPE_NPU', expected (SAI_VS_SWITCH_TYPE_BCM81724|SAI_VS_SWITCH_TYPE_BCM56850|SAI_VS_SWITCH_TYPE_BCM56971B0|SAI_VS_SWITCH_TYPE_MLNX2700)
Jun 29 19:34:55.043712 a6a6a10ed28f ERR #syncd: :- Syncd: FATAL: failed to sai_api_initialize: SAI_STATUS_FAILURE
Jun 29 19:34:55.180774 a6a6a10ed28f INFO #supervisord 2023-06-29 19:34:55,180 INFO exited: syncd (terminated by SIGABRT (core dumped); not expected)
Jun 29 19:34:56.182412 a6a6a10ed28f INFO #supervisord 2023-06-29 19:34:56,182 INFO gave up: syncd entered FATAL state, too many start retries too


vslib/SwitchConfig.cpp
```diff
@@ -79,6 +79,15 @@ bool SwitchConfig::parseSwitchType(
     {
         switchType = SAI_VS_SWITCH_TYPE_NVDA_MBF2H536C;
     }
+    else if (st == SAI_VALUE_VS_SWITCH_TYPE_NPU)
```
kcudnik, line 82: nope, this should not be that way, see that NPU seems like generic, look at other implementation that they have specific names, since SAI_VS_SWITCH_TYPE_NVDA_MBF2H536C is already defined, then you you should use SAI_VALUE_VS_SWITCH_TYPE_NVDA_MBF2H536C to se its value

kcudnik (changes requested):

vslib/SwitchConfig.cpp
prabhataravind: @kcudnik, The problem we are faced with is that support for NVDA_MBF2H536C is not yet available in public sonic-buildimage repo. Those are still in the nvidia-sonic buildimage repo. Nvidia is planning to eventually upstream those, but until then we would like to have a way to run vstests without depending on NVDA_MBF2H536C hwsku. Guohan had suggested to create a "generic" NPU hwksu that could be used for overcoming this issue. Please see https://github.com/sonic-net/sonic-buildimage/pull/15604 and https://github.com/sonic-net/sonic-buildimage/pull/15695 which are related sonic-buildimage PRs.

lguohan: this should be SAI_VALUE_VS_SWITCH_TYPE_NPU_2P make it specific. then, we can create such vslib, just copy paste NVDA_MBF2H536C, change to NPU_2P. if we need more specific name, we can say NPU_SIMU_2P, meaning this is to 2 port npu simulator.

lguohan: thought about it, maybe we can just rename to SAI_VALUE_VS_SWITCH_TYPE_NPU_SIMU_2P and it is kind of alias to nvda_mbf2h536c

lguohan (dismissed):
rename to DPU_SIMU_2P

vslib/SwitchConfig.cpp
prabhataravind: Okay, I have changed the name to SAI_VALUE_VS_SWITCH_TYPE_NPU_SIMU_2P.

kcudnik (dismissed):

lguohan (approved):

vslib/SwitchConfig.cpp
```diff
@@ -79,14 +79,25 @@ bool SwitchConfig::parseSwitchType(
     {
         switchType = SAI_VS_SWITCH_TYPE_NVDA_MBF2H536C;
     }
+    else if (st == SAI_VALUE_VS_SWITCH_TYPE_DPU_SIMU_2P)
+    {
+        /*
+         * Temporarily set switchType to SAI_VS_SWITCH_TYPE_NVDA_MBF2H536C
```
kcudnik, line 85: add "TODO" or "FIXME" in the comment for easy spot

kcudnik (dismissed):

PR #1268, opened 2023-07-26
Fix RPC package build failure due to shell syntax issue

saiarcot895:
PR #1194 removed the .ONESHELL flag, causing commands for a target to be executed in separate shell environments. Because of that, multi-line shell code blocks aren't parsed correctly.

Since the `syncd_init_common.sh` file should always exist anyways (it's unconditionally sourced by an init script and a start script), remove the if-check around it, and always run `sed` when building rpc packages.


kcudnik (dismissed):

saiarcot895:
I was hoping to add a pipeline to build the RPC version, but looks like more changes are needed there. I'll try to address that in a separate PR.

kcudnik (approved):

dgsudharsan (approved):

PR #1270, opened 2023-08-02
[syncd] Add pre match logic for acl entry (#1240)

kcudnik:
this will test scenario where we have 1 acl entry with counter and after warm boot 2 acl entries and 2 counters the same which could end up matching wrong acl counter and causing fail "object exist" on broadcom acl entry set, which will require improve pre match logic for acl entry


kcudnik:
this change is not needed, it's already on the branch

PR #1272, opened 2023-08-08
[submodule] Update SAI to latest v1.12 branch

kcudnik:
Also update VendorSai class to use latest generic api from SAI metadata.

Since latest SAI uses version v1.12 on master branch, force require sai_query_api_version to be present at all SAI libs

### DIFF
--- a/syncd/VendorSai.cpp
+++ b/syncd/VendorSai.cpp
@@ -31,8 +31,16 @@ VendorSai::VendorSai()
         .api_query = &sai_api_query,
         .api_uninitialize = &sai_api_uninitialize,
         .bulk_get_attribute = nullptr,
+#ifdef HAVE_SAI_BULK_OBJECT_CLEAR_STATS
+        .bulk_object_clear_stats = &sai_bulk_object_clear_stats,
+#else
         .bulk_object_clear_stats = nullptr,
+#endif
+#ifdef HAVE_SAI_BULK_OBJECT_GET_STATS
+        .bulk_object_get_stats = &sai_bulk_object_get_stats
+#else
         .bulk_object_get_stats = nullptr,
+#endif
         .dbg_generate_dump = nullptr,
         .get_maximum_attribute_count = nullptr,
         .get_object_count = nullptr,
@@ -41,7 +49,7 @@ VendorSai::VendorSai()
         .object_type_get_availability = &sai_object_type_get_availability,
         .object_type_query = &sai_object_type_query,
         .query_api_version = &sai_query_api_version,
-        .query_attribute_capability = &sai_query_attribute_capability ,
+        .query_attribute_capability = &sai_query_attribute_capability,
         .query_attribute_enum_values_capability = &sai_query_attribute_enum_values_capability,
         .query_object_stage = nullptr,
         .query_stats_capability = &sai_query_stats_capability,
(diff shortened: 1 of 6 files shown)


syncd/VendorSai.cpp
```diff
@@ -24,6 +24,40 @@ VendorSai::VendorSai()
     m_apiInitialized = false;
 
     memset(&m_apis, 0, sizeof(m_apis));
+
+    sai_global_apis_t ga =
```
kcudnik, line 28: from now on, we keep sai global api pointers in a metadata generated structs, this will be handy when we will want to load another vendor libsai.so into memory

lguohan (approved):

liushilongbuaa:
@kcudnik ,https://github.com/sonic-net/sonic-buildimage/pull/16168
The submodule update PR always failed on braodcom/barefoot/mellanox build.
Please check.

liushilongbuaa:
https://dev.azure.com/mssonic/build/_build/results?buildId=343938&view=results

kcudnik:
for mellanox and broadcom the same error happens:
```
VendorSai.cpp: In constructor 'syncd::VendorSai::VendorSai()':
VendorSai.cpp:44:10: error: request for member 'dbg_generate_dump' in 'sai_bulk_object_get_stats', which is of non-class type 'sai_status_t(sai_object_id_t, sai_object_type_t, uint32_t, const sai_object_key_t*, uint32_t, const sai_stat_id_t*, sai_stats_mode_t, sai_status_t*, uint64_t*)' {aka 'int(long unsigned int, _sai_object_type_t, unsigned int, const _sai_object_key_t*, unsigned int, const unsigned int*, _sai_stats_mode_t, int*, long unsigned int*)'}
   44 |         .dbg_generate_dump = nullptr,
      |          ^~~~~~~~~~~~~~~~~
```

which build is for barefoot ? 

btw. i found the issue, fixe is here: https://github.com/sonic-net/sonic-sairedis/pull/1278

kcudnik:
already merged, please update sairedis submodule on https://github.com/sonic-net/sonic-buildimage/pull/16168 to help integrate the fix

liushilongbuaa:
@kcudnik , https://dev.azure.com/mssonic/build/_build/results?buildId=354040&view=logs&j=712a2a98-2630-52ac-ed49-4850e5af4e10&t=b2c30a30-98c5-58f9-7d50-fefbf2c245dc
Still has errors.

kcudnik:
this is the issue:
```
checking whether CXX supports -Wno-cast-function-type... yes
checking for sai_query_api_version... no
configure: error: "SAI library libsai.so does not have sai_query_api_version API which is required"
	tail -v -n \+0 config.log
==> config.log <==
This file contains any messages produced by compilers while
```

libsai which is build with that image does not implement sai_query_api_version, in this PR i enabled required field for that sai library method saiversion.h header was introduced in in version v1.9.0 in this commit [0b929737](https://github.com/opencomputeproject/SAI/commit/0b929737) which was on Feb 2022, that SAI library used in that build should already implement that API, lets contact marverll to update library with that api, it's 5min change, implementation looks like this: https://github.com/sonic-net/sonic-sairedis/blob/master/vslib/sai_vs_interfacequery.cpp#L208
```c
sai_status_t sai_query_api_version(
        _Out_ sai_api_version_t *version)
{
    SWSS_LOG_ENTER();

    *version = SAI_API_VERSION;
    return SAI_STATUS_SUCCESS;
}
```

@lguohan fyi

PR #1273, opened 2023-08-09
Build both the regular and RPC version when the RPC profile is enabled

saiarcot895:
When the RPC version needs to be built, the regular version of syncd also needs to be built. Unfortunately, this is not easy to do with the current build infra except for manually rerunning configure, removing the built object files/binaries, and building the new config.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -7,7 +7,7 @@ Standards-Version: 1.0.0
 
 Package: syncd
 Architecture: any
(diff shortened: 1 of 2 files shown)


kcudnik (approved):
not sure if this will have impact on sonic-buildimage

saiarcot895:
It matches the behavior that was there prior to #1194.

PR #1276, opened 2023-08-19
[azp] Attempt to fix swss missing libs

kcudnik:
Swss containter needs dash_api to properly build, based on https://github.com/sonic-net/sonic-swss/pull/2722, attempting to add missing libs


kcudnik:
related: https://github.com/sonic-net/sonic-sairedis/pull/1277

kcudnik:
fixed by https://github.com/sonic-net/sonic-sairedis/pull/1277

PR #1286, opened 2023-09-03
Test [DO NOT MERGE]

kcudnik:


kcudnik:
to test if Azure.sonic-sairedis (BuildSwss amd64) is passing

PR #1289, opened 2023-09-07
Use json.hpp from nlohmann-json-dev instead of swss-common

saiarcot895:
This header file comes from an external package, and a very old version of the header file has been checked into swss-common. This will cause problems for the upcoming Bookworm upgrade.

Change references to the header file to use the Debian package nlohmann-json-dev (added in https://github.com/sonic-net/sonic-buildimage/pull/16308), instead of from swss-common.

### DIFF
--- a/lib/ClientConfig.cpp
+++ b/lib/ClientConfig.cpp
@@ -1,7 +1,7 @@
 #include "ClientConfig.h"
 
 #include "swss/logger.h"
-#include "swss/json.hpp"
+#include <nlohmann/json.hpp>
 
 #include <cstring>
 #include <fstream>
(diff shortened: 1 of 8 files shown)


lib/ClientConfig.cpp
```diff
@@ -1,7 +1,7 @@
 #include "ClientConfig.h"
 
 #include "swss/logger.h"
-#include "swss/json.hpp"
```
kcudnik, line 4: will that json.hpp will be removed from sonic-swss-common repo ?

kcudnik (approved):

lib/ClientConfig.cpp
saiarcot895: Yes, it will be removed soon.

PR #1290, opened 2023-09-08
Install nlohmann-json3-dev package for codeql

saiarcot895:
This is as a result of #1289.


kcudnik (approved):

kcudnik:
should it be this file: https://github.com/sonic-net/sonic-sairedis/blob/master/.github/codeql/codeql-config.yml

saiarcot895:
No, that file doesn't need changing.

The CodeQL checker will fail here because `pull_request_target` is being used (which doesn't include this PR) instead of `pull_request`, so the changes in this PR won't get checked.

PR #1296, opened 2023-09-28
[syncd] Change sai discovery log priority to info

kcudnik:
When multiple asic are present and multiple ports are created then those log lines willpollute logs for no reason


jaganbal-a (approved):

abdosi:
@gechiang for viz.

PR #1297, opened 2023-09-28
[Link event damping] Add concurrent queue.

Ashish1805:
HLD: https://github.com/sonic-net/SONiC/pull/1071

### DIFF
--- a/syncd/ConcurrentQueue.h
+++ b/syncd/ConcurrentQueue.h
@@ -1,0 +1,107 @@
+#pragma once
+
+#include <mutex>
+#include <queue>
+
+#include "swss/logger.h"
+#include "swss/sal.h"
+
+namespace syncd
+{
+    template <class T>
+    class ConcurrentQueue
+    {
+        public:
+
+            explicit ConcurrentQueue(
+                    _In_ size_t queueSizeLimit = UNLIMITED);
+
+            virtual ~ConcurrentQueue()  = default;
+
+            bool enqueue(
+                    _In_ const T& val);
+
+            bool dequeue(
+                    _Out_ T* valOut);
+
+            size_t size();
+
+            bool empty();
+
+        private:
+
+            // Queue size = 0 means there is no limit on queue size.
+            static constexpr size_t UNLIMITED = 0;
+
+            std::mutex m_mutex;
+            std::queue<T> m_queue;
+            size_t m_queueSizeLimit;
+
+            ConcurrentQueue<T>(const ConcurrentQueue<T>&) = delete;
+            ConcurrentQueue<T>& operator=(const ConcurrentQueue<T>&) = delete;
+    };
+
+    template <class T>
+    ConcurrentQueue<T>::ConcurrentQueue(
+              _In_ size_t queueSizeLimit)
+        : m_queueSizeLimit(queueSizeLimit)
+    {
+        SWSS_LOG_ENTER();
+    }
+
+    template <class T>
+    bool ConcurrentQueue<T>::enqueue(
+            _In_ const T& val)
+    {
+        SWSS_LOG_ENTER();
+
+        std::lock_guard<std::mutex> mutex_lock(m_mutex);
+
+        // If the queue exceeds the limit, return false.
+        if ((m_queueSizeLimit == UNLIMITED) || (m_queue.size() < m_queueSizeLimit))
+        {
+            m_queue.push(val);
+            return true;
+        }
+
+        return false;
+    }
+
+    template <class T>
+    bool ConcurrentQueue<T>::dequeue(
+            _Out_ T* valOut)
+    {
+        SWSS_LOG_ENTER();
+
+        std::lock_guard<std::mutex> mutex_lock(m_mutex);
+        if (m_queue.empty())
+        {
+            return false;
+        }
+
+        *valOut = m_queue.front();
+        m_queue.pop();
+
+        return true;
+    }
+
+    template <class T>
+    size_t ConcurrentQueue<T>::size()
+    {
+        SWSS_LOG_ENTER();
+
+        std::lock_guard<std::mutex> mutex_lock(m_mutex);
+
+        return m_queue.size();
+    }
+
+    template <class T>
+    bool ConcurrentQueue<T>::empty()
+    {
+        SWSS_LOG_ENTER();
+
+        std::lock_guard<std::mutex> mutex_lock(m_mutex);
+
+        return m_queue.empty();
+    }
+} // namespace syncd
--- a/tests/aspell.en.pws
+++ b/tests/aspell.en.pws
@@ -50,6 +50,7 @@ IPGs
 IPv
 Inseg
 KEYs
+LLC
 LOGLEVEL
 LOOPBACK
 MACsec
--- a/unittest/syncd/Makefile.am
+++ b/unittest/syncd/Makefile.am
@@ -9,6 +9,7 @@ tests_SOURCES = main.cpp \
                 MockableSaiInterface.cpp \
                 MockHelper.cpp \
 				TestCommandLineOptions.cpp \
+				TestConcurrentQueue.cpp \
 				TestFlexCounter.cpp \
 				TestVirtualOidTranslator.cpp \
 				TestNotificationQueue.cpp \
--- a/unittest/syncd/TestConcurrentQueue.cpp
+++ b/unittest/syncd/TestConcurrentQueue.cpp
@@ -1,0 +1,77 @@
+#include <gtest/gtest.h>
+
+#include "ConcurrentQueue.h"
+
+using namespace syncd;
+
+class ConcurrentQueueTest : public ::testing::Test
+{};
+
+TEST_F(ConcurrentQueueTest, QueueIsEmpty)
+{
+    constexpr size_t queueSize = 5;
+    ConcurrentQueue<int> testQueue(queueSize);
+
+    EXPECT_TRUE(testQueue.empty());
+    EXPECT_EQ(testQueue.size(), 0);
+}
+
+TEST_F(ConcurrentQueueTest, EnqueueSucceeds)
+{
+    constexpr size_t queueSize = 5;
+    ConcurrentQueue<int> testQueue(queueSize);
+
+    EXPECT_TRUE(testQueue.empty());
+
+    EXPECT_TRUE(testQueue.enqueue(1));
+    EXPECT_FALSE(testQueue.empty());
+    EXPECT_EQ(testQueue.size(), 1);
+}
+
+TEST_F(ConcurrentQueueTest, EnqueueFailsIfQueueSizeLimitIsReached)
+{
+    constexpr size_t queueSize = 5;
+    ConcurrentQueue<int> testQueue(queueSize);
+
+    EXPECT_TRUE(testQueue.empty());
+
+    int data = 1;
+    for (size_t idx = 0; idx < queueSize; ++idx)
+    {
+        SCOPED_TRACE(::testing::Message() << "Inserting data at index: " << idx);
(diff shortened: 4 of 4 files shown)


syncd/ConcurrentQueue.h
```diff
@@ -0,0 +1,114 @@
...
+
+  return false;
+}
+
+template <class T>
+bool ConcurrentQueue<T>::Dequeue(T *val_out) {
+  SWSS_LOG_ENTER();
+
+  std::lock_guard<std::mutex> mutex_lock(mutex_);
+  if (queue_.empty()) {
```
kcudnik, line 86: All { should start in new line

syncd/ConcurrentQueue.h
```diff
@@ -0,0 +1,114 @@
...
+  size_t GetQueueSize();
+
+  bool Empty();
+
+ private:
+  // If queue size is specified 0, that will mean there is no limit on queue
+  // size.
+  static constexpr size_t kUnlimited = 0;
+
+  std::mutex mutex_;
```
kcudnik, line 45: All members should be prefixed m_

syncd/ConcurrentQueue.h
```diff
@@ -0,0 +1,114 @@
...
+  std::lock_guard<std::mutex> mutex_lock(mutex_);
+  return queue_.size();
+}
+
+template <class T>
+bool ConcurrentQueue<T>::Empty() {
+  SWSS_LOG_ENTER();
+
+  std::lock_guard<std::mutex> mutex_lock(mutex_);
+  return queue_.empty();
```
kcudnik, line 109: Empty line before review

syncd/LinkEventDamper.cpp
```diff
@@ -0,0 +1,277 @@
...
+  return false;
+}
+
+void LinkEventDamper::ProcessPortStateChange(
+    const sai_port_oper_status_notification_t &data) {
+  SWSS_LOG_ENTER();
+
+  sai_object_id_t vid;
+  bool rid_to_vid_mapping_found =
+      vid_translator_->tryTranslateRidToVid(data.port_id, vid);
```
kcudnik, line 64: Big lump of code totally unreadable

kcudnik (changes requested):
Give up on this review address comments

Ashish1805:
cc: @bhagatgit @eddyk-nvidia @mikeberesford

kcudnik:
> > Unmanageable to review, too many changes, too many files affected in 1 pr, low code quality, not following existing file standards, many new classes, each new class can be a separate code review, new port custom range attributes seems better to be added in SAI repo instead of here, causing to implement custom Metadata where SAI generates that automatically
> 
> Thanks @kcudnik for taking a look.
> 
> On the issue of port custom attribute, this is added in `sairedis.h` in sonic-sairedis as per agreed upon in HLD https://github.com/sonic-net/SONiC/blob/master/doc/link_event_damping/Link-event-damping-HLD.md#sairedis-api
> 
> In the existing sonic-sairedis code, I see there are mix of different standards followed. I couldn't find a guideline or doc for sonic standard code, can you please share them Or if you suggest any existing code file that can be a good reference for sonic standard?
> 
> Thank you.

I keep the standard in this entire repo, there is no specific documents but 90% of files are kept in very good shape, please take a look on more files and point me to the one you were looking to, I pointed you more obvious issues

mikeberesford:
> > > Unmanageable to review, too many changes, too many files affected in 1 pr, low code quality, not following existing file standards, many new classes, each new class can be a separate code review, new port custom range attributes seems better to be added in SAI repo instead of here, causing to implement custom Metadata where SAI generates that automatically
> > 
> > 
> > Thanks @kcudnik for taking a look.
> > On the issue of port custom attribute, this is added in `sairedis.h` in sonic-sairedis as per agreed upon in HLD https://github.com/sonic-net/SONiC/blob/master/doc/link_event_damping/Link-event-damping-HLD.md#sairedis-api
> > In the existing sonic-sairedis code, I see there are mix of different standards followed. I couldn't find a guideline or doc for sonic standard code, can you please share them Or if you suggest any existing code file that can be a good reference for sonic standard?
> > Thank you.
> 
> I keep the standard in this entire repo, there is no specific documents but 90% of files are kept in very good shape, please take a look on more files and point me to the one you were looking to, I pointed you more obvious issues

I'd (strongly) suggest that SONiC adopt a published coding standard, and perhaps we can help with that. As an example (admitting bias here!) Google publishes https://google.github.io/styleguide/ - I'm not suggesting that SONiC adopt this guide, just listing it as an example. Having this will both prevent back-and-forth and will help reduce load on reviewers.

kcudnik:
Yes, we will adopt or create some standard at some point but not at such big reviews, usually anyone is adding couple linea of code and not many classes and changing 45 files, one by one is able to adopt code

syncd/ConcurrentQueue.h
```diff
@@ -0,0 +1,127 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
```
kcudnik, line 7: @lguohan do we allow each file to have different license?

lguohan: no

Ashish1805: Are you saying none of the files that we are trying to upstream can have license? 

cc: @mikeberesford @bhagatgit

kcudnik: I think we have a license for entire repo not per single file

bhagatgit: @lguohan  @kcudnik  Can you provide more details on the licensing. 
On repos that have multiple contributors, does this mean that the contributor that started the repo would add license and others would not ?

Ashish1805: @kcudnik while license issue is being resolved, can you please provide other review comments that you may have so that I can work on those in parallel for this PR and other files that will come in next few PRs? Thank you.

kcudnik: sure, please also fix build errors

syncd/ConcurrentQueue.h
```diff
@@ -0,0 +1,127 @@
...
+
+namespace syncd
+{
+    template <class T>
+    class ConcurrentQueue
+    {
+        public:
+
+            explicit ConcurrentQueue(
+                    size_t queueSizeLimit = m_unlimited);
```
kcudnik, line 30: add _ In _ as other functions have

syncd/ConcurrentQueue.h
```diff
@@ -0,0 +1,127 @@
...
+
+            virtual ~ConcurrentQueue();
```
kcudnik, line 32: if its empty you can use = default;

syncd/ConcurrentQueue.h
```diff
@@ -0,0 +1,127 @@
...
+
+            bool enqueue(
+                    const T& val);
+
+            bool dequeue(
+                    T* valOut);
```
kcudnik, line 38: add _ In_ and _ Out _

syncd/ConcurrentQueue.h
```diff
@@ -0,0 +1,127 @@
...
+
+            size_t getQueueSize();
+
+            bool empty();
+
+        private:
+
+            // If queue size is specified 0, that will mean there is no limit on queue
+            // size.
+            static constexpr size_t m_unlimited = 0;
```
kcudnik, line 48: if it's a static constexpr definition maybe you could make it capital letters as size_t UNLIMITED = 0;

syncd/ConcurrentQueue.h
```diff
@@ -0,0 +1,127 @@
...
+            m_queue.push(val);
+            return true;
+        }
+
+        return false;
+    }
+
+    template <class T>
+    bool ConcurrentQueue<T>::dequeue(
+            T* valOut)
```
kcudnik, line 92: also add _ Out _

syncd/ConcurrentQueue.h
```diff
@@ -0,0 +1,127 @@
...
+
+            virtual ~ConcurrentQueue();
+
+            bool enqueue(
+                    const T& val);
+
+            bool dequeue(
+                    T* valOut);
+
+            size_t getQueueSize();
```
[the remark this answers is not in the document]
Ashish1805: We are using lock_gaurd() inside the size() and empty(), and lock_gaurd() will modify the mutex to lock and unlock it, so making the function const will not work.

syncd/ConcurrentQueue.h
Ashish1805: Thanks. Addressed the review comments. Build issue is because of spell checker (aspellcheck.pl) failing due to it complaining about LLC in "Google LLC" in the copyright header. Not sure why LLC is causing spell check issue though.

Ashish1805:
@kcudnik I was wondering which tool do you use or will recommend to use to format the syncd code during development? Manually formatting the files one by one is time consuming and prone to missing the formatting.

I am mainly looking for formatter that helps to format: 1)  { starts in new line, 2) space formatting for  each line, 3) width for each line for code and comments, etc. Thanks.

kcudnik:
> @kcudnik I was wondering which tool do you use or will recommend to use to format the syncd code during development? Manually formatting the files one by one is time consuming and prone to missing the formatting.
> 
> I am mainly looking for formatter that helps to format: 1) { starts in new line, 2) space formatting for each line, 3) width for each line for code and comments, etc. Thanks.

i dont use any specific tool, i put { in new line manually, and i use formating spacing in vim select code and press "="

syncd/ConcurrentQueue.h
Ashish1805: @kcudnik We are still getting the licensing header issue resolved internally. Given that it is taking time, I have started generating the remaining commits to this PR. Can you please review them too? I dont know if I was supposed to generated separate PR for each commit or all the commits can be generated in same PR. Given these commits are dependent, generating multiple commits in same PRs sounds reasonable. Thanks.

kcudnik: Think you will hav3 still multiple classes and many many code in one pr,  I would advise to have separate pr for specific modules

Ashish1805: Just to confirm, what do you mean exactly by separate PR here? Does it mean each commit should have its own PR, for example: commit 1 in PR https://github.com/sonic-net/sonic-sairedis/pull/XXXX, commit 2 in https://github.com/sonic-net/sonic-sairedis/pull/YYYY, commit 3 in https://github.com/sonic-net/sonic-sairedis/pull/ZZZZ, and so on?

Or you are saying multiple commits in this same PR is ok but each commit should not have multiple classes?

A lot of code depend on each other. For example: commit 2 in this PR depends on commit 1 in the PR. So, if we generate separate separate PR for each commits, they will not compile since dependency will not be present.

kcudnik: no, not the commit, but if you are planning to add many big classes, it make sense to add each class separately, since we squash later all the commits to 1 commit on master, and if you add 40 files in one commit that will be overkill and unmanagable

syncd/ConcurrentQueue.h
```diff
@@ -0,0 +1,107 @@
+#pragma once
```
kcudnik, line 1: Since you are adding only queue here you should change this pr to add concurentqueue etc

kcudnik (changes requested):

kcudnik (approved):

kcudnik:
HLD: https://github.com/sonic-net/SONiC/pull/1071

PR #1288, opened 2023-09-06
[VOQ][saidump] Enhance saidump with new option -r to parser the JSON file and displays/format the right output  

JunhongMao:
#### Why I did it
Fix issue: https://github.com/sonic-net/sonic-buildimage/issues/13561
 The existing saidump use https://github.com/sonic-net/sonic-swss-common/blob/master/common/table_dump.lua script which loops the ASIC_DB more than 5 seconds and blocks other processes access. 

This solution uses the redis-db SAVE option to save the snapshot of DB each time and recover later, instead of looping through each entry in the table.
Related PRs:
   https://github.com/sonic-net/sonic-utilities/pull/2972
   https://github.com/sonic-net/sonic-buildimage/pull/16466

#### How I did it

   To use the Redis-db SAVE option to save the snapshot of DB each time and recover later, instead of looping through each entry in the table and saving it.
```
    (1) Updated platform/broadcom/docker-syncd-brcm-dnx/Dockerfile.j2, install Python library rdbtools into the syncd containter.
    (2) Updated sonic-buildimage/src/sonic-sairedis/saidump/saidump.cpp, add a new option -r, which updates the rdbtools's output-JSON files' format.
    (3) Updated sonic-buildimage/build_debian.sh, to add a new script file: files/scripts/saidump.sh into the host. This shell file does the below steps:
      For each ASIC0, such as ASIC0,

      1. Save the Redis data.
      sudo sonic-db-cli -n asic$1 SAVE > /dev/null

      2. Move dump files to /var/run/redisX/
      docker exec database$1 sh -c "mv /var/lib/redis/dump.rdb /var/run/redis$1/"

      3. Run rdb command to convert the dump files into JSON files
      docker exec syncd$1 sh -c "rdb --command json /var/run/redis$1/dump.rdb | tee /var/run/redis$1/dump.json > /dev/null"

      4. Run saidump -r to update the JSON files' format as same as the saidump before. Then we can get the saidump result in standard output.
      docker exec syncd$1 sh -c "saidump -r /var/run/redis$1/dump.json -m 100"

      5. clear
      sudo rm -f /var/run/redis$1/dump.rdb
      sudo rm -f /var/run/redis$1/dump.json

    (4) Update sonic-buildimage/src/sonic-utilities/scripts/generate_dump, to check the asic db size and if it is larger than xxx entries, then do with REDIS SAVE, otherwise, to do with old method: looping through each entry of Redis DB.
```

#### How to verify it
[...]

PR #1298, opened 2023-10-03
[VOQ][saidump] To move saidump.sh from the sonic-buildimage repo to the sairedis repo

JunhongMao:
#### Why did I do it?

To fix the issue: https://github.com/sonic-net/sonic-buildimage/issues/13561
The existing saidump use https://github.com/sonic-net/sonic-swss-common/blob/master/common/table_dump.lua script which loops the ASIC_DB more than 5 seconds and blocks other processes access.

This solution uses the Redis SAVE command to save the snapshot of DB each time and recover later, instead of looping through each entry in the table.

Related PRs:
https://github.com/sonic-net/sonic-utilities/pull/2972
https://github.com/sonic-net/sonic-buildimage/pull/16466
https://github.com/sonic-net/sonic-sairedis/pull/1288

In addition to the PR:
https://github.com/sonic-net/sonic-sairedis/pull/1288
This PR is to address the below review comments:
To move saidump.sh from the sonic-buildimage repo to the sairedis repo.
https://github.com/sonic-net/sonic-buildimage/pull/16466#discussion_r1341963806

#### Work item tracking
Microsoft ADO (number only):

#### How did I do it?
To use the Redis-db SAVE option to save the snapshot of DB each time and recover later, instead of looping through each entry in the table and saving it.

    1. Updated dockers/docker-base-bullseye/Dockerfile.j2, install Python library rdbtools into the all the docker-base-bullseye containers.

    2. Updated sonic-buildimage/src/sonic-sairedis/saidump/saidump.cpp, add a new option -r, which updates the rdbtools's output-JSON files' format.

    3. To add a new script file: syncd/scripts/saidump.sh into the sairedis repo. This shell script does the following steps:

      For each ASIC, such as ASIC0,

      3.1. Config Redis consistency directory. 
      redis-cli -h $hostname -p $port CONFIG SET dir $redis_dir > /dev/null

      3.2. Save the Redis data.
      redis-cli -h $hostname -p $port SAVE > /dev/null

      3.3. Run rdb command to convert the dump files into JSON files
        rdb --command json $redis_dir/dump.rdb | tee $redis_dir/dump.json > /dev/null

      3.4.  Run saidump -r to update the JSON files' format as same as the saidump before. 
           Then we can get the saidump's result in standard output."
           saidump -r $redis_dir/dump.json -m 100

      3.5. Clear the temporary files.
       rm -f $redis_dir/dump.rdb
       rm -f $redis_dir/dump.json

    4. Update sonic-buildimage/src/sonic-utilities/scripts/generate_dump. To check the asic db size and if it is larger than ROUTE_TAB_LIMIT_DIRECT_ITERATION (with default value 24000) entries, then do with REDIS SAVE, otherwise, to do with old method: looping through each entry of Redis DB.

#### How to verify it
1) On T2 setup with more than 96K routes, execute CLI command -- generate_dump
2) No error should be shown
3) Download the generate_dump result and verify the saidump file after unpacking it.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211
- [x] 202305

- [x] 202205


syncd/scripts/saidump.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+set -e
+
+function debug()
+{
+  /usr/bin/logger "$1"
+}
+
+save_saidump_by_rdb() {
```
kcudnik, line 9: Add new line before { to keep the same for entire file

syncd/scripts/saidump.sh
```diff
@@ -0,0 +1,43 @@
...
+  local filepath="/var/run/redis/sonic-db/database_config.json"
+
+  #Get hostname, port, redis directory
+  local redis_config=$(python3 -c "
+import json
+with open('$filepath') as json_file:
+  data = json.load(json_file)
+  print(data['INSTANCES']['redis']['hostname'], data['INSTANCES']['redis']['port'], data['INSTANCES']['redis']['unix_socket_path'])")
+
+  #split
```
kcudnik, line 19: Add space after hash

syncd/scripts/saidump.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+set -e
+
+function debug()
```
kcudnik, line 4: Not sure if this is necessary, maybe you can use direct logger command instead of this wrapper, no need for explicit path

kcudnik (changes requested):

syncd/scripts/saidump.sh
JunhongMao: Fix it.

syncd/scripts/saidump.sh
JunhongMao: Fix it.

syncd/scripts/saidump.sh
JunhongMao: Fix it.

kcudnik (approved):

gechiang:
@StormLiangMS , @yxieca , please review/approve for 202205 and 202305.
MSFT ADO: 25892380

StormLiangMS:
#1298 depend on #1288, I reverted both, will cherry pick this one after #1288 get separate PR merged to 202305. @JunhongMao @gechiang

PR #1306, opened 2023-10-12
Support TWAMP Light notification in syncd

AlanYoush:
For supporting TWAMP Light, it requires vendor-sai to report TWAMP Light mearsurement data from silicon.

And, so it defines the notification event and getting statistics function in SAI and invokes them in syncd.

### DIFF
--- a/meta/Meta.cpp
+++ b/meta/Meta.cpp
@@ -6729,6 +6729,84 @@ void Meta::meta_sai_on_bfd_session_state_change(
     }
 }
 
+void Meta::meta_sai_on_twamp_session_event_single(
+        _In_ const sai_twamp_session_event_notification_data_t& data)
+{
+    SWSS_LOG_ENTER();
+
+    auto ot = objectTypeQuery(data.twamp_session_id);
+
+    bool valid = false;
+
+    switch (ot)
+    {
+        // TODO hardcoded types, must advance SAI repository commit to get metadata for this
+        case SAI_OBJECT_TYPE_TWAMP_SESSION:
+
+            valid = true;
+            break;
+
+        default:
+
+            SWSS_LOG_ERROR("data.twamp_session_id %s has unexpected type: %s, expected TWAMP_SESSION",
+                    sai_serialize_object_id(data.twamp_session_id).c_str(),
+                    sai_serialize_object_type(ot).c_str());
+            break;
+    }
+
+    // check if all counter ids are in enum range
+    for (uint32_t idx = 0; idx < data.session_stats.number_of_counters; idx++)
+    {
+        if (!sai_metadata_get_enum_value_name(&sai_metadata_enum_sai_twamp_session_stat_t, data.session_stats.counters_ids[idx]))
+        {
+            SWSS_LOG_ERROR("value %d is not in range on sai_twamp_session_stat_t ", data.session_stats.counters_ids[idx]);
+
+            return;
+        }
+    }
+
+    if (valid && !m_oids.objectReferenceExists(data.twamp_session_id))
+    {
+        SWSS_LOG_NOTICE("data.twamp_session_id new object spotted %s not present in local DB (snoop!)",
+                sai_serialize_object_id(data.twamp_session_id).c_str());
+
+        sai_object_meta_key_t key = { .objecttype = (sai_object_type_t)ot, .objectkey = { .key = { .object_id = data.twamp_session_id } } };
+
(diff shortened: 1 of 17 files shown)


meta/Meta.cpp
```diff
@@ -6686,6 +6686,65 @@ void Meta::meta_sai_on_bfd_session_state_change(
...
+
+            valid = true;
+            break;
+
+        default:
+
+            SWSS_LOG_ERROR("data.twamp_session_id %s has unexpected type: %s, expected TWAMP_SESSION",
+                    sai_serialize_object_id(data.twamp_session_id).c_str(),
+                    sai_serialize_object_type(ot).c_str());
+            break;
```
kcudnik, line 6711: just return here ? and valid variable is not needed

meta/Meta.cpp
```diff
@@ -6686,6 +6686,65 @@ void Meta::meta_sai_on_bfd_session_state_change(
...
+        sai_object_meta_key_t key = { .objecttype = (sai_object_type_t)ot, .objectkey = { .key = { .object_id = data.twamp_session_id } } };
+
+        m_oids.objectReferenceInsert(data.twamp_session_id);
+
+        if (!m_saiObjectCollection.objectExists(key))
+        {
+            m_saiObjectCollection.createObject(key);
+        }
+    }
+}
```
kcudnik, line 6728: you can also validate 
sai_twamp_session_state_t session_state;
sai_twamp_session_stats_data_t session_stats;

syncd/NotificationProcessor.cpp
```diff
@@ -548,6 +548,36 @@ void NotificationProcessor::process_on_switch_shutdown_request(
     sendNotification(SAI_SWITCH_NOTIFICATION_NAME_SWITCH_SHUTDOWN_REQUEST, s);
 }
 
+void NotificationProcessor::process_on_twamp_session_event(
+        _In_ uint32_t count,
+        _In_ sai_twamp_session_event_notification_data_t *data)
+{
+    SWSS_LOG_ENTER();
+
+    SWSS_LOG_DEBUG("twamp sessuin state notification count: %u", count);
```
kcudnik, line 557: spelling sessuin

kcudnik (changes requested):

meta/Meta.cpp
AlanYoush: Thanks for comments.
It should check the object type of session_id whether it is TWAMP session. So we add valid variable and set as true when the object type is TWAMP session, otherwise use default value as false. If the valid variable is false, it would just return the error log.

meta/Meta.cpp
AlanYoush: Thanks, that's right.
I have validated 
sai_twamp_session_state_t session_state;
sai_twamp_session_stats_data_t session_stats;

please review again.

kcudnik (approved):

kcudnik:
please resolve merge conflicts

syncd/Syncd.cpp
```diff
@@ -4395,7 +4396,8 @@ void Syncd::performWarmRestartSingleSwitch(
         SAI_SWITCH_ATTR_NAT_EVENT_NOTIFY,
         SAI_SWITCH_ATTR_PORT_STATE_CHANGE_NOTIFY,
         SAI_SWITCH_ATTR_QUEUE_PFC_DEADLOCK_NOTIFY,
-        SAI_SWITCH_ATTR_BFD_SESSION_STATE_CHANGE_NOTIFY
+        SAI_SWITCH_ATTR_BFD_SESSION_STATE_CHANGE_NOTIFY,
+        SAI_SWITCH_ATTR_TWAMP_SESSION_EVENT_NOTIFY
```
kcudnik, line 4400: this entire list could be obtained from sai_metadata_switch_notify_attr, and there will be no need in the future to update that, i will make PR for this after your will merge

AlanYoush: Got it.  Whether I would do nothing before you make PR.

kcudnik: go on proceeded with yours, i will add my on top of it

kcudnik (approved):

AlanYoush:
@kcudnik Thank you very much for approving it. I am wondering if you are able to merge it ? If you can, please help to merge it,  because of PR (https://github.com/sonic-net/sonic-swss/pull/2927) depends on it.

AlanYoush:
@kcudnik Thanks very much

PR #1308, opened 2023-10-17
The meta_sai_validate_fdb_entry() validates the input FDB entry for the set/remove operation. (#1154)

kcudnik:
When there is no matching FDB entry found in the database, then an incorrect SAI error code (SAI_STATUS_INVALID_PARAMETER) is returned. As a port of this fix, the correct SAI error return code (SAI_STATUS_ITEM_NOT_FOUND) is returned.
This Return code can be used by the callers to take approprite actions, if the FDB entry is indeed not available during the set/remove operations.


kcudnik:
cherrypicked from https://github.com/sonic-net/sonic-sairedis/pull/1154

qiluo-msft (approved):

PR #1309, opened 2023-10-18
[syncd] Add pre match logic for pfc storm

kcudnik:
Special case where priority is the same for all acl


syncd/ComparisonLogic.cpp
```diff
@@ -2836,6 +2838,21 @@ void ComparisonLogic::cretePreMatchForAclEntries(
                 continue;
             }
 
+            if (cPrio->getStrAttrValue() == PFC_ACL_RULE_PRIORITY)
+            {
+                // this is special case, all pfc rules to prevent packet storm
+                // will have same priority, we will need to check SAI_ACL_ENTRY_ATTR_FIELD_TC
+
+                auto cTc = cAclEntry->getSaiAttr(SAI_ACL_ENTRY_ATTR_FIELD_TC);
```
vaibhavhd, line 2846: Will this get call fail if the TC field is not present in the ACL?

syncd/ComparisonLogic.cpp
```diff
@@ -2806,6 +2806,8 @@ void ComparisonLogic::cretePreMatchForLagMembers(
     }
 }
 
+#define PFC_ACL_RULE_PRIORITY "999"
```
vaibhavhd, line 2809: @neethajohn can you confirm if hardcoding 999 as a PFC ACL priority will work across branches?

syncd/ComparisonLogic.cpp
```diff
@@ -2836,6 +2838,21 @@ void ComparisonLogic::cretePreMatchForAclEntries(
                 continue;
             }
 
+            if (cPrio->getStrAttrValue() == PFC_ACL_RULE_PRIORITY)
```
vaibhavhd, line 2841: Do we know that only PFC ACLs have the priority of `PFC_ACL_RULE_PRIORITY`? If not, then there is a chance that this block will get executed for other ACLs, and in that case the behavior is indeterministic.

syncd/ComparisonLogic.cpp
```diff
@@ -2836,6 +2838,21 @@ void ComparisonLogic::cretePreMatchForAclEntries(
...
+            {
+                // this is special case, all pfc rules to prevent packet storm
+                // will have same priority, we will need to check SAI_ACL_ENTRY_ATTR_FIELD_TC
+
+                auto cTc = cAclEntry->getSaiAttr(SAI_ACL_ENTRY_ATTR_FIELD_TC);
+                auto tTc = tAclEntry->getSaiAttr(SAI_ACL_ENTRY_ATTR_FIELD_TC);
+
+                if (!cTc || !cTc || cTc->getStrAttrValue() != tTc->getStrAttrValue())
```
vaibhavhd, line 2849: I think you mean to check for `!cTc || !tTc` rather`!cTc || !cTc`?

syncd/ComparisonLogic.cpp
```diff
@@ -2836,6 +2838,21 @@ void ComparisonLogic::cretePreMatchForAclEntries(
...
+                {
+                    // TC attribute is different, not looking for this one
+                    continue;
+                }
+            }
+
             // at this point current and temporary acl entry share the same
             // priority, then we can assume that those objects are the same
```
vaibhavhd, line 2857: I think this comment needs to be updated? Now not just priority, but also TC field may be same.

syncd/ComparisonLogic.cpp
kcudnik: yes, let me fix that

syncd/ComparisonLogic.cpp
kcudnik: no clue, this is hardcoded as well in OA

syncd/ComparisonLogic.cpp
kcudnik: this is hardcoded in OA https://github.com/sonic-net/sonic-swss/blob/f31ccd09b8eafa2d9292e7446fc5fe8d4ec0ad06/orchagent/pfcactionhandler.cpp#L448

syncd/ComparisonLogic.cpp
kcudnik: yes

syncd/ComparisonLogic.cpp
kcudnik: updated

kcudnik:
Special case where priority is the same for all acl

StormLiangMS:
@kcudnik cherry pick to 202305 conflict, could you help to raise separate PR for 202305?

kcudnik:
sure

kcudnik:
here it is: https://github.com/sonic-net/sonic-sairedis/pull/1318 please review

PR #1312, opened 2023-10-24
[Link Event Damping] Add Abstract SelectEventHandler class to handle link event damping timer events.

Ashish1805:
- Handler in the class will be used to handle the timer event of link event dampers.

HLD: sonic-net/SONiC#1071

### DIFF
--- a/syncd/SelectableEventHandler.h
+++ b/syncd/SelectableEventHandler.h
@@ -1,0 +1,19 @@
+#pragma once
+
+namespace syncd
+{
+    // This class implements handler for Selectable events.
+    class SelectableEventHandler
+    {
+        public:
+
+            virtual ~SelectableEventHandler() = default;
+
+            virtual void handleSelectableEvent() = 0;
+
+        protected:
+
+            SelectableEventHandler() = default;
+    };
+
+}  // namespace syncd


syncd/SelectableEventHandler.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include "swss/logger.h"
+#include "swss/sal.h"
+
+namespace syncd
+{
+
```
kcudnik, line 8: empty line not needed

syncd/SelectableEventHandler.h
```diff
@@ -0,0 +1,28 @@
...
+    // This class implements handler for Selectable events.
+    class SelectableEventHandler {
```
kcudnik, line 10: add { at the next line

syncd/SelectableEventHandler.h
```diff
@@ -0,0 +1,28 @@
...
+        public:
+
+            virtual ~SelectableEventHandler()
+            {
+              SWSS_LOG_ENTER();
+            }
```
kcudnik, line 16: if its not a template, move code to cpp file

Ashish1805: Do we see any issue with keeping the destructor here in .h file?

Ashish1805:
Adding @Junchao-Mellanox for review.

syncd/SelectableEventHandler.h
kcudnik: Usually when code is not template we prefer to move cpp code to cpp file, and also add empty intentionally or = default

syncd/SelectableEventHandler.h
```diff
@@ -0,0 +1,28 @@
...
+
+            virtual void handleSelectableEvent() = 0;
+
+        protected:
+
+            SelectableEventHandler()
```
Junchao-Mellanox, line 22: do we really need this?

syncd/SelectableEventHandler.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include "swss/logger.h"
+#include "swss/sal.h"
```
Junchao-Mellanox, line 4: unused include?

syncd/SelectableEventHandler.h
kcudnik: Why constructor is protected ?

kcudnik (changes requested):

Junchao-Mellanox (approved):

Ashish1805:
@kcudnik Please let me know if you have any other comments/concerns. If everything looks good, can you please merge the PR? Thank you.

kcudnik:
You did not answered my questions about protected constructors

kcudnik (approved):

Ashish1805:
> You did not answered my questions about protected constructors

In another comment reply, I had added the explanation:

Strictly speaking constructor is not needed, but constructor was added as protected just to make it clear to clients that they cannot directly instantiate an object of the class.

Agree, this class still be uninstantiable because of the pure virtual method, but to put the emphasize on not trying to directly instantiate this class, protected constructor was kept.

I am inclined to keep constructor as:
protected:
SelectableEventHandler() = default;

Thank you.

PR #1314, opened 2023-10-30
[Link Event Damping] Add link event damping sairedis port attributes.

Ashish1805:
- Add link event damping sairedis port attributes.
- These attributes will be used by OA to set the link event damping algorithm and AIED link event damping config.
- Syncd uses these sairedis port attribute to receive the link event damping algorithm and AIED config to execute syncd-based link event damping.

HLD: sonic-net/SONiC#1071

### DIFF
--- a/lib/sairedis.h
+++ b/lib/sairedis.h
@@ -250,3 +250,58 @@ typedef enum _sai_redis_switch_attr_t
      */
     SAI_REDIS_SWITCH_ATTR_SYNC_OPERATION_RESPONSE_TIMEOUT,
 } sai_redis_switch_attr_t;
+
+/**
+ * @brief Link event damping algorithms.
+ */
+typedef enum _sai_link_event_damping_algorithm_t
+{
+    /** Link event damping algorithm disabled. */
+    SAI_LINK_EVENT_DAMPING_ALGORITHM_DISABLED = 0,
+
+    /** Additive increase exponential decrease based link event damping algorithm. */
+    SAI_LINK_EVENT_DAMPING_ALGORITHM_AIED = 1,
+
+} sai_link_event_damping_algorithm_t;
+
+typedef struct _sai_redis_link_event_damping_algo_aied_config_t
+{
+    /** Max link event damping suppression time (in milliseconds). */
+    sai_uint32_t max_suppress_time;
+
+    /** Link event damping suppress threshold. */
+    sai_uint32_t suppress_threshold;
+
+    /** Link event damping reuse threshold. */
+    sai_uint32_t reuse_threshold;
+
+    /** Link event damping decay half time duration (in milliseconds). */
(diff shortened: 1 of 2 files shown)


Ashish1805:
Adding @Junchao-Mellanox  for review.

Junchao-Mellanox (approved):

lib/sairedis.h
```diff
@@ -250,3 +250,59 @@ typedef enum _sai_redis_switch_attr_t
...
+    SAI_LINK_EVENT_DAMPING_ALGORITHM_DISABLED = 0,
+
+    /** Additive increase exponential decrease based link event damping algorithm */
+    SAI_LINK_EVENT_DAMPING_ALGORITHM_AIED = 1,
+
+} sai_link_event_damping_algorithm_t;
+
+typedef struct _sai_redis_link_event_damping_algo_aied_config_t
+{
+
```
kcudnik, line 269: Extra empty line not needed

Ashish1805:
> please fix aspell:
> 
> ```
> WARNING: Word 'AIED' is misspelled ../lib/sairedis.h
> WARNING: Word 'validonly' is misspelled ../lib/sairedis.h
> FAIL: aspellcheck.pl
> ```

Please let me know if you have any other comments/concerns. If everything looks good, can we merge this PR? Thank you.

kcudnik (approved):

kcudnik:
just one more question here, is this will be used internally by sairedis ? since i see you added  custom range port attributes, or this will interact with SAI via sairedis ?

Ashish1805:
> just one more question here, is this will be used internally by sairedis ? since i see you added custom range port attributes, or this will interact with SAI via sairedis ?

This will be used internally by sairedis, it will not interact with SAI. Basically swss will use these port attributes to set  the  link event damping config and link event damping algo (using sai_port_api->set_port_attribute() API) and these set requests will be consumed by sairedis internally to set the link event damping algo and config in the link event damping logic in syncd.

kcudnik:
Why then this state needs to be held inside sairedis and not OA directly ?

Ashish1805:
> Why then this state needs to be held inside sairedis and not OA directly ?

There was couple of reasons:
- When link events are sent from syncd, there can be multiple listeners to these events notifications coming from syncd (today it is OA but in future other components can listen to these events too) and if OA is doing damping, other listeners will receive the undamped events from syncd.
- Keeping the damping logic in software stack as low as possible will reduce the number of notifications upper layer gets.
- Adding the damping in syncd holds the events before any applications can get them.

We discussed this during HLD review and syncd implementation had advantages over OA implementation.

cc: @bhagatgit @mikeberesford

kcudnik:
ok sounds good

Ashish1805:
> ok sounds good

Can you merge the PR if everything looks good? Thank you.

PR #1318, opened 2023-11-05
[syncd] Add pre match logic for pfc storm (#1309)

kcudnik:
Special case where priority is the same for all acl


kcudnik:
@vaibhavhd can you approve ?

StormLiangMS (approved):
LGTM

PR #1320, opened 2023-11-09
[syncd] Translate removed RID on bfd session state change

kcudnik:
Fixes https://github.com/sonic-net/sonic-buildimage/issues/16992

We will also translate removed RID in notification in case that BFD session was removed, and notification arrived for that session asynchronously


prsunny (approved):

kcudnik:
@dgsudharsan @prsunny added 202305 here: https://github.com/sonic-net/sonic-sairedis/pull/1321

PR #1322, opened 2023-11-16
[Link Event Damping] Serialization/deserialization logic for link event damping port attrs

Ashish1805:
- Adding serialization/deserialization logic for link event damping config so that when libsai is sending link event damping algo and config to syncd for set request, it uses serialization logic to serialize the requested data before sending the set request and when syncd receives the link event damping port attribute set request, it uses the deserialization logic to deserialize the data.

HLD: sonic-net/SONiC#1071

### DIFF
--- a/meta/SaiSerialize.cpp
+++ b/meta/SaiSerialize.cpp
@@ -8,11 +8,20 @@
 #include <inttypes.h>
 #include <vector>
 #include <climits>
+#include <unordered_map>
 
 #include <arpa/inet.h>
 #include <errno.h>
 
 using json = nlohmann::json;
+
+static const std::unordered_map<sai_redis_link_event_damping_algorithm_t, std::string> sai_redis_link_event_damping_algorithm_to_name_map = {
+        {SAI_REDIS_LINK_EVENT_DAMPING_ALGORITHM_DISABLED, "SAI_REDIS_LINK_EVENT_DAMPING_ALGORITHM_DISABLED"},
+        {SAI_REDIS_LINK_EVENT_DAMPING_ALGORITHM_AIED, "SAI_REDIS_LINK_EVENT_DAMPING_ALGORITHM_AIED"}};
+
+static const std::unordered_map<sai_redis_port_attr_t, std::string> sai_redis_port_attr_to_name_map = {
+        {SAI_REDIS_PORT_ATTR_LINK_EVENT_DAMPING_ALGORITHM, "SAI_REDIS_PORT_ATTR_LINK_EVENT_DAMPING_ALGORITHM"},
+        {SAI_REDIS_PORT_ATTR_LINK_EVENT_DAMPING_ALGO_AIED_CONFIG, "SAI_REDIS_PORT_ATTR_LINK_EVENT_DAMPING_ALGO_AIED_CONFIG"}};
 
 static int char_to_int(
         _In_ const char c)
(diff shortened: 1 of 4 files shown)


meta/sai_serialize.h
```diff
@@ -12,13 +12,22 @@ extern "C" {
...
 #include "sairedis.h"
 
+const std::unordered_map<sai_link_event_damping_algorithm_t, std::string> sai_link_event_damping_algorithm_to_name_map = {
+        {SAI_LINK_EVENT_DAMPING_ALGORITHM_DISABLED, "LINK_EVENT_DAMPING_ALGORITHM_DISABLED"},
+        {SAI_LINK_EVENT_DAMPING_ALGORITHM_AIED, "LINK_EVENT_DAMPING_ALGORITHM_AIED"}};
+
+const std::unordered_map<sai_redis_port_attr_t, std::string> sai_redis_port_attr_to_name_map = {
+        {SAI_REDIS_PORT_ATTR_LINK_EVENT_DAMPING_ALGORITHM, "LINK_EVENT_DAMPING_ALGORITHM"},
+        {SAI_REDIS_PORT_ATTR_LINK_EVENT_DAMPING_ALGO_AIED_CONFIG, "LINK_EVENT_DAMPING_ALGO_AIED_CONFIG"}};
+
```
kcudnik, line 30: make this extern, and declare actual code in cpp file

kcudnik: LINK_EVENT_DAMPING_ALGORITHM should be named in full name SAI_REDIS_PORT_ATTR_LINK_EVENT_DAMPING_ALGORITHM

kcudnik: others too

kcudnik:
please fix build errors

meta/sai_serialize.h
Ashish1805: done

meta/SaiSerialize.cpp
```diff
@@ -14,6 +14,14 @@
 
 using json = nlohmann::json;
 
+const std::unordered_map<sai_link_event_damping_algorithm_t, std::string> sai_link_event_damping_algorithm_to_name_map = {
```
kcudnik, line 17: i think sai_link_event_damping_algorithm_t should be renamed to sai_redis_link_event_damping_algorithm_t

kcudnik: since this is not a part of SAI enums

meta/sai_serialize.h
```diff
@@ -12,13 +12,18 @@ extern "C" {
...
 #include <cstring>
 
 #include "swss/logger.h"
 
 #include "sairedis.h"
 
+extern const std::unordered_map<sai_link_event_damping_algorithm_t, std::string> sai_link_event_damping_algorithm_to_name_map;
+
+extern const std::unordered_map<sai_redis_port_attr_t, std::string> sai_redis_port_attr_to_name_map;
+
```
kcudnik, line 26: this is not needed, here, remove, and mark those in cpp files as STATIC

kcudnik: we don't need to expose those to external compilation modules

meta/SaiSerialize.cpp
Ashish1805: done.

meta/sai_serialize.h
Ashish1805: done.

kcudnik (approved):

Ashish1805:
Hi @Junchao-Mellanox , can you please review the PR once you get a chance? Thank you.

PR #1310, opened 2023-10-21
Ledamping 1

Ashish1805:

PR #1323, opened 2023-11-17
[Link Event Damping] Add tracker to track the selectable timers used by link event damper

Ashish1805:
- Link event damper running on each port will have its own SelectableTimer. This class helps create a mapping of SelectableTimer and object that created the SelectableTimer so that when timer is fired, proper EventHandler object can be invoked.

HLD: sonic-net/SONiC#1071


Ashish1805:
Hi @kcudnik `CodeQL / Analyze (cpp) (pull_request_target)` check is failing.

```
/usr/bin/ld: ../../syncd/libSyncd.a(libSyncd_a-PortStateChangeHandler.o): in function `syncd::PortStateChangeHandler::PortStateChangeHandler(std::shared_ptr<swss::SelectableEvent>)':
/home/runner/work/sonic-sairedis/sonic-sairedis/syncd/PortStateChangeHandler.cpp:21: undefined reference to `syncd::PortStateChangeHandler::PORT_STATE_CHANGE_QUEUE_SIZE'
collect2: error: ld returned 1 exit status
```

`PORT_STATE_CHANGE_QUEUE_SIZE` is defined as static and constexpr class data member (https://github.com/sonic-net/sonic-sairedis/blob/master/syncd/PortStateChangeHandler.h#L43). I see sairedis repo uses c++14, so it is necessary to also provide definition for this static data member in .cc file. Like:
```
// Defined in PortStateChangeHandler.cc
constexpr size_t PortStateChangeHandler::PORT_STATE_CHANGE_QUEUE_SIZE;
```

`PortStateChangeHandler.cpp` was added in https://github.com/sonic-net/sonic-sairedis/pull/1310 PR and `CodeQL / Analyze (cpp) (pull_request_target)` check passed in that PR: https://github.com/sonic-net/sonic-sairedis/actions/runs/6885079249. Not sure why it didn't fail during CodeQL check of that PR.

I am not sure what is the protocol here: do we add fix PR for this or revert the culprit PR https://github.com/sonic-net/sonic-sairedis/pull/1310 and merge it again with suggested changes?

kcudnik:
> Hi @kcudnik `CodeQL / Analyze (cpp) (pull_request_target)` check is failing.
> 
> ```
> /usr/bin/ld: ../../syncd/libSyncd.a(libSyncd_a-PortStateChangeHandler.o): in function `syncd::PortStateChangeHandler::PortStateChangeHandler(std::shared_ptr<swss::SelectableEvent>)':
> /home/runner/work/sonic-sairedis/sonic-sairedis/syncd/PortStateChangeHandler.cpp:21: undefined reference to `syncd::PortStateChangeHandler::PORT_STATE_CHANGE_QUEUE_SIZE'
> collect2: error: ld returned 1 exit status
> ```
> 
> `PORT_STATE_CHANGE_QUEUE_SIZE` is defined as static and constexpr class data member (https://github.com/sonic-net/sonic-sairedis/blob/master/syncd/PortStateChangeHandler.h#L43). I see sairedis repo uses c++14, so it is necessary to also provide definition for this static data member in .cc file. Like:
> 
> ```
> // Defined in PortStateChangeHandler.cc
> constexpr size_t PortStateChangeHandler::PORT_STATE_CHANGE_QUEUE_SIZE;
> ```
> 
> `PortStateChangeHandler.cpp` was added in #1310 PR and `CodeQL / Analyze (cpp) (pull_request_target)` check passed in that PR: https://github.com/sonic-net/sonic-sairedis/actions/runs/6885079249. Not sure why it didn't fail during CodeQL check of that PR.
> 
> I am not sure what is the protocol here: do we add fix PR for this or revert the culprit PR #1310 and merge it again with suggested changes?

SAME ISSUE CAUSES SWSS ERROR HERE: https://github.com/sonic-net/sonic-sairedis/pull/1310#issuecomment-1817445036

please figure out why this is causing build error or revert the change !

syncd/SelectablesTracker.h
```diff
@@ -0,0 +1,52 @@
...
+    // corresponding EventHandler objects.
+    class SelectablesTracker
+    {
+        public:
+
+            SelectablesTracker() = default;
+
+            virtual ~SelectablesTracker() = default;
+
+            // Not copyable or movable.
```
kcudnik, line 22: move this to private then

syncd/SelectablesTracker.h
```diff
@@ -0,0 +1,52 @@
...
+            virtual bool selectableIsTracked(
+                    _In_ swss::Selectable *selectable);
+
+            // Gets the EventHandler for a Selectable.
+            virtual SelectableEventHandler *getEventHandlerForSelectable(
+                    _In_ swss::Selectable *selectable);
+
+        private:
+
+            using SelectableFdToEventHandlerMap = std::unordered_map<int, SelectableEventHandler *>;
```
kcudnik, line 47: use shared_ptr instead of pointers

kcudnik (changes requested):
address comments

kcudnik:
change 
```
static constexpr size_t PortStateChangeHandler::PORT_STATE_CHANGE_QUEUE_SIZE;
```
to
```
constexpr size_t PortStateChangeHandler::PORT_STATE_CHANGE_QUEUE_SIZE;
```

first keyword static requires declaration in cpp file since it's considered a field

syncd/SelectablesTracker.h
Ashish1805: Doesn't look like w should be using shared_ptr here. Do you use any benefit of using shared_ptr?

Ashish1805:
Adding @Junchao-Mellanox for review.

syncd/SelectablesTracker.h
kcudnik: yes, benefit is to not lose  trakc of pointers and prevent memory leak, please use shared_ptr, as is used in entire syncd project

syncd/SelectablesTracker.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+    if (selectable == nullptr)
+    {
+        SWSS_LOG_ERROR("Invalid Selectable:Selectable is NULL.");
+
+        return false;
+    }
+
+    int fd = selectable->getFd();
+    if (eventHandler == nullptr)
```
Junchao-Mellanox, line 21: move the check before line 20

kcudnik: althoug he wanted to log fd number in error message, anyway please add empty line before if()

PR #1324, opened 2023-11-20
Fix compilation issue due to PORT_STATE_CHANGE_QUEUE_SIZE undefined reference.

Ashish1805:
- In PR https://github.com/sonic-net/sonic-sairedis/pull/1310, a constexpt static data member was added. Since sairedis repo uses c++14, definition of this data member needs to be provided in .cpp file too.

### DIFF
--- a/syncd/PortStateChangeHandler.cpp
+++ b/syncd/PortStateChangeHandler.cpp
@@ -6,6 +6,8 @@
 
 using namespace syncd;
 
+constexpr size_t PortStateChangeHandler::PORT_STATE_CHANGE_QUEUE_SIZE;
+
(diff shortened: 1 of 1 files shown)


Ashish1805:
@kcudnik I think there is some issue in `CodeQL / Analyze (cpp) (pull_request_target)` check. It doesn't seem to include the current PR when running the check - rather it runs the check on the head of master that was base for this current PR. This will also explain why check didn't fail for PR https://github.com/sonic-net/sonic-sairedis/pull/1310 before merge.

prsunny (approved):

prsunny:
@kcudnik , we may want to force-merge this.

PR #1325, opened 2023-11-23
[syncd] User fdb flush default value from SAI (#1317)

kcudnik:
Previously we were hard-codding default value for fdb flush, and since this value could change in the future, we want to get this form SAI metadata


prsunny (approved):
lgtm

kcudnik:
yes, on this branch seems to be enable, but this change on master passed

StormLiangMS:
remind @kcudnik PR checker failures.

PR #1328, opened 2023-11-28
[syncd] Set inti switch flag to false when doing warm boot

kcudnik:
This flag should be set to false during warm boot. This is a bug.


kcudnik:
This was not catches, since all vendors including virtual switch relay on SAI_WARM_BOOT_READ_FILE to be present and based on the they determine whether do a warm boot

PR #1331, opened 2023-12-06
Add libsai RedisInterface for link event damping.

Ashish1805:
- This supports the link event damping config and algorithm set API.
- Sends the link event damping config from libsai to syncd main thread.

HLD: sonic-net/SONiC#1071

### DIFF
--- a/lib/RedisRemoteSaiInterface.cpp
+++ b/lib/RedisRemoteSaiInterface.cpp
@@ -493,6 +493,83 @@ sai_status_t RedisRemoteSaiInterface::setRedisExtensionAttribute(
     return SAI_STATUS_FAILURE;
 }
 
+sai_status_t RedisRemoteSaiInterface::setLinkEventDampingConfig(
+        _In_ sai_object_type_t objectType,
+        _In_ sai_object_id_t objectId,
+        _In_ const std::vector<swss::FieldValueTuple> &values)
+{
+    SWSS_LOG_ENTER();
+
+    std::string key = sai_serialize_object_type(objectType) + ":" + sai_serialize_object_id(objectId);
+
+    m_communicationChannel->set(key, values, REDIS_ASIC_STATE_COMMAND_DAMPING_CONFIG_SET);
+
+    if (m_syncMode)
+    {
+        swss::KeyOpFieldsValuesTuple kco;
+        auto status = m_communicationChannel->wait(REDIS_ASIC_STATE_COMMAND_DAMPING_CONFIG_SET, kco);
+
+        m_recorder->recordGenericSetResponse(status);
+
+        return status;
+    }
+
+    return SAI_STATUS_SUCCESS;
+}
+
+sai_status_t RedisRemoteSaiInterface::setRedisPortExtensionAttribute(
+        _In_ sai_object_type_t objectType,
+        _In_ sai_object_id_t objectId,
+        _In_ const sai_attribute_t *attr)
+{
+    SWSS_LOG_ENTER();
+
+    if (attr == nullptr)
+    {
+        SWSS_LOG_ERROR("attr pointer is null");
+
+        return SAI_STATUS_INVALID_PARAMETER;
+    }
+
+    std::string str_attr_id = sai_serialize_redis_port_attr_id(
+            static_cast<sai_redis_port_attr_t>(attr->id));
+
+    switch (attr->id)
+    {
+        case SAI_REDIS_PORT_ATTR_LINK_EVENT_DAMPING_ALGORITHM:
+        {
+            std::string str_attr_value = sai_serialize_redis_link_event_damping_algorithm(
+                    static_cast<sai_redis_link_event_damping_algorithm_t>(attr->value.s32));
+
+            return setLinkEventDampingConfig(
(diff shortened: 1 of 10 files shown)


lib/RedisRemoteSaiInterface.cpp
```diff
@@ -485,6 +485,83 @@ sai_status_t RedisRemoteSaiInterface::setRedisExtensionAttribute(
...
+                return SAI_STATUS_INVALID_PARAMETER;
+            }
+
+            std::string str_attr_value = sai_serialize_redis_link_event_damping_aied_config(*config);
+
+            return setLinkEventDampingConfig(
+                    objectType, objectId, {swss::FieldValueTuple(str_attr_id, str_attr_value)});
+        }
+        default:
+            break;
```
kcudnik, line 557: add logging error message that value is not supported

syncd/Syncd.cpp
```diff
@@ -563,6 +566,30 @@ sai_status_t Syncd::processObjectTypeGetAvailabilityQuery(
     return status;
 }
 
+sai_status_t Syncd::processLinkEventDampingConfigSet(
+        _In_ const swss::KeyOpFieldsValuesTuple &kco)
+{
+    SWSS_LOG_ENTER();
+
+    sendLinkEventDampingConfigResponse(SAI_STATUS_NOT_IMPLEMENTED);
```
kcudnik, line 574: you are sending response every time, but SET you implemented is not always synchronous, you added (m_syncMode), so actually sairedis will receive response even in async mode which will break

Ashish1805: Thank you for pointing that out. Done.

Ashish1805:
@Junchao-Mellanox for review.

Ashish1805:
@kcudnik looks like swss pytest is failing - it seems unrelated to my changes. Any idea how to fix the swss pytest?

kcudnik:
those vstests are sometime flaky so dont worry about that

Ashish1805:
@kcudnik I have updated the PR as per review comment. can you please review it when you get a chance? Thank you.

kcudnik:
please fix build erros and code coverage

Ashish1805:
@kcudnik Added the code coverage but I see PR is failing for 2 tests during `Update rsyslog.conf` test. The test failure log doesn't give much info, can you suggest where can I find the helpful debug log to figure out what exactly failed and why?

kcudnik:
> @kcudnik Added the code coverage but I see PR is failing for 2 tests during `Update rsyslog.conf` test. The test failure log doesn't give much info, can you suggest where can I find the helpful debug log to figure out what exactly failed and why?

```
rsyslogd: pidfile '/run/rsyslogd.pid' and pid 18837 already exist.
If you want to run multiple instances of rsyslog, you need to specify
```

in ~/sonic-sairedis/.azure-pipelines/build-template.yml
```
68   - script: |
169       sudo cp azsyslog.conf /etc/rsyslog.conf
170       sudo service rsyslog restart
171     displayName: "Update rsyslog.conf"
```

seems like syslog service restart failed but why, no clue, we would need to look into syslog on pipeline but i dont know how :D maybe update buitld template to do "tail -n 100 /var/log/syslog ? after restaret command ?

lib/RedisRemoteSaiInterface.cpp
```diff
@@ -1792,6 +1874,20 @@ bool RedisRemoteSaiInterface::isRedisAttribute(
...
+        _In_ const sai_attribute_t* attr)
+{
+    SWSS_LOG_ENTER();
+
+    if ((objectType != SAI_OBJECT_TYPE_PORT) || (attr == nullptr) || (attr->id < SAI_PORT_ATTR_CUSTOM_RANGE_START))
+    {
+        return false;
+    }
+
+    return true;
```
kcudnik, line 1888: you can turn this to 1 single line by reversing condition

lib/RedisRemoteSaiInterface.cpp
```diff
@@ -493,6 +493,83 @@ sai_status_t RedisRemoteSaiInterface::setRedisExtensionAttribute(
...
+
+    if (m_syncMode)
+    {
+        swss::KeyOpFieldsValuesTuple kco;
+        auto status = m_communicationChannel->wait(REDIS_ASIC_STATE_COMMAND_DAMPING_CONFIG_SET, kco);
+
+        m_recorder->recordGenericSetResponse(status);
+
+        return status;
+    }
```
kcudnik, line 515: make this api always synchronous, async mode is just for create/remove/set operations all other apis should be synchronous

DendroLabs:
The superseding PR for this one is now open: #1935

It addresses all review feedback from this PR:
- `setLinkEventDampingConfig` is always synchronous (never checks `m_syncMode`), per @kcudnik's review
- `isRedisPortAttribute` simplified to single boolean expression, per @kcudnik's review
- Syncd response is always sent (no `m_enableSyncMode` guard)
- Uses a dedicated response command (`link_event_damping_config_response`) to avoid collision with `GETRESPONSE`
- Clean rebase on current master, no conflicts

@Ashish1805 credited as original author in the PR description.

prsunny:
Closing in favor of https://github.com/sonic-net/sonic-sairedis/pull/1906

PR #1334, opened 2023-12-15
[Link Event Damping] Add per port link event damper class.

Ashish1805:
- Adding per port link event damper class that manages the internal behavior and workings of link event damping logic on a port where link event damping config is enabled.
- This class keeps track of damping timer, current port state, post damping advertised port state and relevant debug counters per port.

HLD: sonic-net/SONiC#1071

### DIFF
--- a/syncd/PortLinkEventDamper.cpp
+++ b/syncd/PortLinkEventDamper.cpp
@@ -1,0 +1,416 @@
+#include "PortLinkEventDamper.h"
+
+#include <algorithm>
+#include <cmath>
+
+#include "Utils.h"
+#include "meta/sai_serialize.h"
+#include "swss/logger.h"
+#include "swss/timestamp.h"
+
+using namespace syncd;
+
+PortLinkEventDamper::PortLinkEventDamper(
+        _In_ std::shared_ptr<NotificationHandler> notificationHandler,
+        _In_ std::shared_ptr<swss::Select> sel,
+        _In_ sai_object_id_t vid,
+        _In_ sai_object_id_t rid,
+        _In_ const sai_redis_link_event_damping_algo_aied_config_t &config)
+    : m_notificationHandler(notificationHandler),
+      m_sel(sel),
+      m_vid(vid),
+      m_rid(rid),
+      m_maxSuppressTimeUsec(uint64_t(config.max_suppress_time) *
+                              MICRO_SECS_PER_MILLI_SEC),
+      m_decayHalfLifeUsec(uint64_t(config.decay_half_life) *
+                            MICRO_SECS_PER_MILLI_SEC),
+      m_suppressThreshold(config.suppress_threshold),
+      m_reuseThreshold(config.reuse_threshold),
+      m_flapPenalty(config.flap_penalty),
+      m_accumulatedPenalty(0),
+      m_penaltyCeiling(std::numeric_limits<uint32_t>::max()),
+      m_dampingConfigEnabled(false),
+      m_dampingActive(false),
+      m_lastPenaltyUpdateTimestamp(0),
+      m_timer(/*interval=*/timespec{.tv_sec = 0, .tv_nsec = 0}),
+      m_timerAddedToSelect(false),
+      m_timerActive(false),
+      m_advertisedState(SAI_PORT_OPER_STATUS_UNKNOWN),
+      m_actualState(SAI_PORT_OPER_STATUS_UNKNOWN),
+      m_stats()
+{
+    SWSS_LOG_ENTER();
+}
+
+PortLinkEventDamper::~PortLinkEventDamper()
+{
+    SWSS_LOG_ENTER();
+
+    // Clean up the resources.
+    if (m_timerAddedToSelect)
+    {
+        cancelTimer();
+        m_sel->removeSelectable(&m_timer);
+
+        SWSS_LOG_DEBUG("Removed timer with fd %d for port vid: %s from select.",
+                       m_timer.getFd(), sai_serialize_object_id(m_vid).c_str());
+    }
+
+    advertisePortState(m_actualState);
+}
+
+void PortLinkEventDamper::setup()
+{
+    SWSS_LOG_ENTER();
+
+    if (configContainsZeroValueParam())
+    {
+        SWSS_LOG_NOTICE(
+                "Damping config is disabled on port vid: %s due to one or more "
+                "received config params being 0.",
+                sai_serialize_object_id(m_vid).c_str());
+
+        m_penaltyCeiling = 0;
+        m_dampingConfigEnabled = false;
+    }
+    else
+    {
+        updatePenaltyCeiling();
+        m_dampingConfigEnabled = true;
+    }
+
+    m_sel->addSelectable(&m_timer);
+    m_timer.stop();
+
+    SWSS_LOG_DEBUG("Added timer with fd %d for port vid: %s to select.",
+                   m_timer.getFd(), sai_serialize_object_id(m_vid).c_str());
+
+    m_timerAddedToSelect = true;
+}
+
+void PortLinkEventDamper::updatePenaltyCeiling()
+{
+    SWSS_LOG_ENTER();
+
+    // If any of config param value is 0, damping config is treated as disabled
+    // and calculation of penalty ceiling is not valid so mark penalty ceiling 0.
+    if (configContainsZeroValueParam())
+    {
+        m_penaltyCeiling = 0;
+
+        return;
+    }
+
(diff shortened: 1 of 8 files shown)


syncd/PortLinkEventDamper.cpp
```diff
@@ -0,0 +1,416 @@
...
+    // Ensure updated accumulated penalty does not cross penalty ceiling.
+    // To avoid integer overflow, first check if enough gap is available.
+    if (m_penaltyCeiling - m_accumulatedPenalty > m_flapPenalty)
+    {
+        m_accumulatedPenalty += m_flapPenalty;
+    }
+    else
+    {
+        m_accumulatedPenalty = m_penaltyCeiling;
+    }
```
kcudnik, line 148: use std::max ?

syncd/PortLinkEventDamper.cpp
```diff
@@ -0,0 +1,416 @@
...
+}
+
+void PortLinkEventDamper::resetTimer(
+        _In_ int64_t time_us)
+{
+    SWSS_LOG_ENTER();
+
+    m_timer.setInterval(
+        /*interval=*/{.tv_sec = (time_us / int64_t(MICRO_SECS_PER_SEC)),
```
kcudnik, line 157: comment in weird place, please remove we know it's setInterval function

syncd/PortLinkEventDamper.cpp
```diff
@@ -0,0 +1,416 @@
...
+{
+    SWSS_LOG_ENTER();
+
+    if (isConfigSameAsRunningConfig(config))
+    {
+        return;
+    }
+
+    m_maxSuppressTimeUsec =
+            uint64_t(config.max_suppress_time) * MICRO_SECS_PER_MILLI_SEC;
```
kcudnik, line 322: join those lines, and why you need to explicitly cast to 64?

syncd/PortLinkEventDamper.h
```diff
@@ -0,0 +1,148 @@
...
+#include "swss/select.h"
+#include "swss/selectabletimer.h"
+
+extern "C" {
+#include "saimetadata.h"
+}
+
+namespace syncd
+{
+
```
kcudnik, line 17: empty line not needed

syncd/PortLinkEventDamper.h
```diff
@@ -0,0 +1,148 @@
...
+    inline uint64_t getCurrentTimeUsecs()
+    {
+        SWSS_LOG_ENTER();
+
+        struct timespec ts;
+        clock_gettime(CLOCK_MONOTONIC, &ts);
+
+        return (ts.tv_sec * MICRO_SECS_PER_SEC) +
+               (ts.tv_nsec / NANO_SECS_PER_MICRO_SEC);
+    }
```
kcudnik, line 31: move code to cpp file

syncd/PortLinkEventDamper.h
```diff
@@ -0,0 +1,148 @@
...
+
+  struct DampingStats
+  {
```
kcudnik, line 34: each struct and class should have it's own file cpp and h

syncd/PortLinkEventDamper.h
```diff
@@ -0,0 +1,148 @@
...
+      // Number of link DOWN events received.
+      uint64_t pre_damping_down_events;
+      // Number of link UP events advertised post damping.
+      uint64_t post_damping_up_events;
+      // Number of link DOWN events advertised post damping.
+      uint64_t post_damping_down_events;
+      // Timestamp for last advertised link up event post damping.
+      std::string last_advertised_up_event_timestamp;
+      // Timestamp for last advertised link down event post damping.
+      std::string last_advertised_down_event_timestamp;
```
kcudnik, line 46: add empty line before each line of comment for readibility or put all comments on the right side

syncd/PortLinkEventDamper.h
```diff
@@ -0,0 +1,148 @@
...
+            sai_port_oper_status_t m_advertisedState;
+
+            // Actual port's oper status in the ASIC.
+            sai_port_oper_status_t m_actualState;
+
+            // Damping related statistics.
+            struct DampingStats m_stats;
+
+            friend class PortLinkEventDamperPeer;
+            friend class LinkEventDamperPeer;
```
kcudnik, line 146: in entire project we don't use friend class, please explain in details why this is exactly what you need and it cant be done in other way?

unittest/syncd/TestPortLinkEventDamper.cpp
```diff
@@ -0,0 +1,1184 @@
...
+#include <vector>
+
+#include "PortLinkEventDamper.h"
+#include "MockNotificationHandler.h"
+
+namespace syncd
+{
+
+// Peer class to access private state and APIs.
+class PortLinkEventDamperPeer
```
kcudnik, line 13: should this be Mock class?

kcudnik: classes in tests should be in separated files

kcudnik (changes requested):

syncd/PortLinkEventDamper.cpp
```diff
@@ -0,0 +1,416 @@
...
+    if (portState != m_advertisedState)
+    {
+        sai_port_oper_status_notification_t data = {.port_id = m_rid,
+                                                    .port_state = portState};
+
+        SWSS_LOG_NOTICE("Advertising [port rid: %s, port oper status: %s]",
+                sai_serialize_object_id(data.port_id).c_str(),
+                sai_serialize_port_oper_status(data.port_state).c_str());
+
+        m_notificationHandler->onPortStateChangePostLinkEventDamping(/*count=*/1, &data);
```
kcudnik, line 295: this is not right, you are manually generating port notification, this breaks SAI design, @lguohan  we need to discuss this whether this is allowed

Ashish1805: This is not manual generation of port event notification exactly rather delayed generation of port events as it is intended with link event damping feature - for the link event damping, the actual port events  will be trapped here  (https://github.com/sonic-net/sonic-sairedis/blob/master/syncd/NotificationHandler.cpp#L164-L172) (NotificationHandler changes will be added in subsequent PRs to do that) and sent to link event damping logic running in syncd main thread where events will be processed as per the damping algo and config. If port events are not needed to be damped or damping duration have elapsed and damped port event needs to be advertised to OA (and other applications), corresponding port event notifications will be generated as the case here.

Since we are doing review of a portion of a code at a time, full logic is not self explanatory here.

Targeted final onPortStateChange() will look like following where port events are trapped and sends for processing. Here m_portStateChangeHandler is of type PortStateChangeHandler (added in https://github.com/sonic-net/sonic-sairedis/pull/1310):
===
void NotificationHandler::onPortStateChange(
        _In_ uint32_t count,
        _In_ const sai_port_oper_status_notification_t *data)
{
    SWSS_LOG_ENTER();

    if (m_portStateChangeHandler == nullptr) {
        auto s = sai_serialize_port_oper_status_ntf(count, data);
        enqueueNotification(SAI_SWITCH_NOTIFICATION_NAME_PORT_STATE_CHANGE, s);
    } else {
        // Send notification to port state change handler for further processing.
        m_portStateChangeHandler->HandlePortStateChangeNotification(count, data);
    }
}

=======

cc: @mikeberesford.

kcudnik: it seems like manual generate, since this code is creating notification data and manually calling onPortStateChangePostLinkEventDamping, which all notifications should come from vendor SAI and not be generated on demand

Ashish1805: That is the intention of the feature. SAI generated link event is trapped by link event damping feature and it will be sent to application as per the damping logic. For example:

1) if damping logic says SAI generated event notification needs to damped, that notification will be suppressed in the syncd and will not be sent to application.
2) if damping logic says no damping needed, then the notification needs to be sent to application using onPortStateChangePostLinkEventDamping(). Since original SAI notification is trapped and processed in the damper logic, we need to manually create the corresponding notification data and send it to application.

Since the original notification from SAI is trapped and processed by damper, we will need to create notification data when notification needs to be sent.

kcudnik: Not fully understand this logic, can you make some kind of flow diagram ? So e I'm not convinced still why we generate manually this notification

DendroLabs:
This PR has been superseded by #1936, which addresses all review feedback, adds syslog on suppress/unsuppress transitions, and adds monitor-only mode (RFC 7196). The flow diagram requested by @kcudnik is included in the new PR description.

PR #1305, opened 2023-10-11
add if statement for module control mode support

dbarashinvd:
#### Why I did it
Feature - CMIS management

##### Work item tracking
- Microsoft ADO **(number only)**:

#### How I did it
update syncd init common script to add module control profile content into sai.profile file

#### How to verify it
manually check sai.profile on a supporting switch (MSFT SKU) and verify module control profile content in it

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

#### Tested branch (Please provide the tested image version)

- [ ] 
- [ ]

PR #1341, opened 2024-01-23
Revert "add if statement for module control mode support"

dbarashinvd:
Reverts sonic-net/sonic-sairedis#1305

this logic is not needed. we'll change sai.profile directly


kcudnik (approved):

keboliu:
@kcudnik would you please help to merge?

PR #1342, opened 2024-01-24
Check MDIO server thread joinable before join the thread

jiahua-wang:
This PR is to address the issue [syncd crash observed during sonic-mgmt reboot tests #17346](https://github.com/sonic-net/sonic-buildimage/issues/17346)

### DIFF
--- a/syncd/MdioIpcServer.cpp
+++ b/syncd/MdioIpcServer.cpp
@@ -465,7 +465,10 @@ void MdioIpcServer::stopMdioThread(void)
     }
 
     m_taskAlive = 0;
-    m_taskThread.join();
+    if(m_taskThread.joinable())
+    {
+        m_taskThread.join();
+    }
     SWSS_LOG_NOTICE("IPC task thread is stopped\n");
 }
 


kcudnik:
Is the root case already known? Was it double joij?

jiahua-wang:
Double join could be an explanation. More likely scenario is that m_taskThread is not joinable because after MdioIpcServer constructor startMdioThread() was not called or because startMdioThread() failed. @kenneth-arista can provide more test details.

kcudnik:
> Double join could be an explanation. More likely scenario is that m_taskThread is not joinable because after MdioIpcServer constructor startMdioThread() was not called or because startMdioThread() failed. @kenneth-arista can provide more test details.

Can you investigate which exactly scenario is that ?

kenneth-arista:
> > Double join could be an explanation. More likely scenario is that m_taskThread is not joinable because after MdioIpcServer constructor startMdioThread() was not called or because startMdioThread() failed. @kenneth-arista can provide more test details.
> 
> Can you investigate which exactly scenario is that ?

Unfortunately, I don't have all the logs when the problem occurred before integrating Jiahua's patch. Prior to the patch, I would often seen several syncd core files after each full sonic-mgmt test runs. I suspect these crashes are triggered in tests involving reboot.

kenneth-arista:
I recall that a syncd crash was observed once during sonic-mgmt `override_config_table` tests.

jiahua-wang:
This patch will address both cases of double join and thread not starting properly.

kcudnik (approved):

rlhui:
@jiahua-wang  - please update the branch? thanks

kcudnik (approved):

jiahua-wang:
@kcudnik @rlhui @judyjoseph  All checks were passed when I created the PR for the first time. After I merged other commits, 2 checks always fail at the place not related to my code change. Should I close this one and create another one with the same code change?

kcudnik:
All passed do you want me to merge?

jiahua-wang:
@kcudnik Please merge.

PR #1362, opened 2024-03-13
Fix flow counter out-of-order issue by notifying counter operations using SelectableChannel

stephenxs:
**What I did**

Fix flow counter out-of-order issue by notifying counter operations using SelectableChannel

Signed-off-by: Stephen Sun <stephens@nvidia.com>

**Why I did it**

Currently, the operations of SAI objects and their counters (if any) are triggered by different channels, which introduces racing conditions:
- the creation and destruction of the objects are notified using the `SelectableChannel`,
- the operations of counters, including starting and stopping polling the counters, are notified by listening to the `FLEX_COUNTER` and `FLEX_COUNTER_GROUP` tables in the `FLEX_COUNTER_DB`
- The orchagent always respects the order when starting/stopping counter-polling (which means to start counter-polling after creating the object and to stop counter-polling before destroying the object) but `syncd` can receive events in a wrong order, eg. it receives destroying an object first and then stopping counter polling on the object, it can poll counter for a non-exist object, which causes errors in vendor SAI.

The new solution is to extend SAI redis attributes on the SAI_SWITCH_OBJECT to notify counter polling. As a result, all the objects and their counters are notified using a unified channel, which is the `SelectableChannel`.

**How I verified it**

Unit test
Manual test
Regressions test

**Details if related**

There are two SAI Redis attributes introduced as below. There are some fields with `const char *` type for each attribute. Passing a field as `nullptr` means not to change it.
- `SAI_REDIS_SWITCH_ATTR_FLEX_COUNTER_GROUP` for counters represented by `FLEX_COUNTER_GROUP` table in the `FLEX_COUNTER_DB`, including the following fields
  - `counter_group_name`, which is the key of the table, representing the group name.
  - `poll_interval`, which is the field `POLL_INTERVAL` of an entry, representing the polling interval of the group.
  - `operation`, which is the field `FLEX_COUNTER_STATUS` of an entry, representing whether the counter polling is enabled for the group
  - `stats_mode`, which is the field `STATS_MODE` of an entry, either `STATS_MODE_READ` or `STATS_MODE_READ_AND_CLEAR`
  - `plugins`, which represents the Lua plugin related to the group
  - `plugin_name`, which is the name of the plugins field. It differs among different groups
- `SAI_REDIS_SWITCH_ATTR_FLEX_COUNTER` for counter groups represented by the `FLEX_COUNTER` table in the `FLEX_COUNTER_DB`, including the following fields
  - `counter_key`, which is the key of the table, with the name convention of `<group-name>:oid:<oid-value>`
  - `counter_ids`, which is a list of counter IDs to be polled for the object
  - `counter_field_name`, which is the name of the counter ID field. It differs among different groups
  - `stats_mode`, which is the field `STATS_MODE` of an entry, either `STATS_MODE_READ` or `STATS_MODE_READ_AND_CLEAR`

Both SAI attributes are terminated by the `RedisRemoteSaiInterface` object in the swss context, which serializes the SAI API call into the selectable channel.
- `REDIS_FLEX_COUNTER_COMMAND_SET_COUNTER_GROUP`: represents the `SET` operation in the `FLEX_COUNTER_GROUP` table 
- `REDIS_FLEX_COUNTER_COMMAND_DEL_COUNTER_GROUP`: represents the `DEL` operation in the `FLEX_COUNTER_GROUP` table 
- `REDIS_FLEX_COUNTER_COMMAND_START_POLL`: represents the `SET` operation in the `FLEX_COUNTER` table 
- `REDIS_FLEX_COUNTER_COMMAND_STOP_POLL`: represents the `DEL` operation in the `FLEX_COUNTER` table

The Syncd will call flex counter functions to handle them on receiving the above-extended commands (representing both SAI extended attributes).

***Gearbox flex counter database***
Pass the Phy OID, an OID of a SAI switch object in syntax, when calling the SAI set API to set the extended attributes. By doing so, the SAI redis objects can choose in which context the SAI API call should be invoked and the corresponding gearbox syncd docker container will handle it.
(ps: THE ORIGINAL GEARBOX FLEX COUNTER IMPLEMENTATION IS BUGGY)

***Context and critical section analysis***
It does not change the critical section hierarchy

***Performance analysis***
The counter operations are handled in the same thread in both the new and old solutions.
In swss, the counter operation was asynchronous in the old solution and is synchronous now, which can introduce a bit more latency. However, as the number of counter operations is small, no performance degradation is observed.


lib/sairedis.h
```diff
@@ -86,6 +86,88 @@ typedef enum _sai_redis_communication_mode_t
...
+     * @brief The name of the filed that represents the Lua plugin
+     */
+    const char *plugin_name;
+
+    /**
+     * @brief The SHA code of the Lua plugin
+     */
+    const char *plugins;
+
+} sai_redis_flex_counter_group_parameter_t;
```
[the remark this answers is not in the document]
stephenxs: I'm OK with using std::string but what do you mean by a separate class? define sai_redis_flex_counter_group_parameter_t as a class?

syncd/Syncd.cpp
```diff
@@ -2436,18 +2468,35 @@ void Syncd::processFlexCounterGroupEvent( // TODO must be moved to go via ASIC c
...
 
-    WatchdogScope ws(m_timerWatchdog, op + ":" + groupName, &kco);
+    processFlexCounterGroupEvent(key, op, values, false);
+}
+
+void Syncd::processFlexCounterGroupEvent( // TODO must be moved to go via ASIC channel queue
+        _In_ const std::string &groupName,
+        _In_ const std::string &op,
+        _In_ const std::vector<swss::FieldValueTuple> &values,
+        _In_ bool operateDb)
```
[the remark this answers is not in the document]
stephenxs: currently, I want to keep both old and new approaches.
- `operateDb == true` means it is notified by SelectableChannel. In this function, it should add it to the `FLEX_COUNTER_DB`.
- `operateDb == false` means it is notified by `ConsumerTable` and the items have been already in the `FLEX_COUNTER_DB`, no need to add it to db in this function.

lib/RedisRemoteSaiInterface.cpp
```diff
@@ -485,6 +491,75 @@ sai_status_t RedisRemoteSaiInterface::setRedisExtensionAttribute(
...
+
+    if (flexCounterGroupParam->operation != nullptr)
+    {
+        entries.push_back({FLEX_COUNTER_STATUS_FIELD, flexCounterGroupParam->operation});
+    }
+
+    m_recorder->recordGenericSet(key, entries);
+    m_communicationChannel->set(key, entries, REDIS_FLEX_COUNTER_COMMAND_COUNTER_GROUP);
+
+    return waitForResponse(SAI_COMMON_API_SET);
```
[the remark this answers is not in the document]
stephenxs: in the backend (syncd daemon), both actual SET and counter operations use the same channel, so there should be no issue.

lib/sairedis.h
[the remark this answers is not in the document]
stephenxs: but it doesn't look like a convention to use c++ types in this file?

lib/RedisRemoteSaiInterface.cpp
```diff
@@ -485,6 +491,75 @@ sai_status_t RedisRemoteSaiInterface::setRedisExtensionAttribute(
     return SAI_STATUS_FAILURE;
 }
 
+sai_status_t RedisRemoteSaiInterface::notifyCounterGroupOperations(
+        _In_ sai_object_id_t objectId,
+        _In_ const sai_attribute_t *attr)
```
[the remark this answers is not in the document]
stephenxs: fixed

lib/RedisRemoteSaiInterface.cpp
```diff
@@ -485,6 +491,75 @@ sai_status_t RedisRemoteSaiInterface::setRedisExtensionAttribute(
...
+{
+    std::vector<swss::FieldValueTuple> entries;
+    sai_redis_flex_counter_group_parameter_t *flexCounterGroupParam = reinterpret_cast<sai_redis_flex_counter_group_parameter_t*>(attr->value.ptr);
```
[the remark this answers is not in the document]
stephenxs: fixed

lib/RedisRemoteSaiInterface.cpp
```diff
@@ -485,6 +491,75 @@ sai_status_t RedisRemoteSaiInterface::setRedisExtensionAttribute(
...
+
+    m_recorder->recordGenericSet(key, entries);
+    m_communicationChannel->set(key, entries, REDIS_FLEX_COUNTER_COMMAND_COUNTER_GROUP);
+
+    return waitForResponse(SAI_COMMON_API_SET);
+}
+
+sai_status_t RedisRemoteSaiInterface::notifyCounterOperations(
+        _In_ sai_object_id_t objectId,
+        _In_ const sai_attribute_t *attr)
```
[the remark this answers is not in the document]
stephenxs: fixed

lib/sairedis.h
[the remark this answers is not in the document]
kcudnik: if you want to achieve attribute look like, use sai_s8list_t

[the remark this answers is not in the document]
stephenxs: ok. I saw an example using s8list. will change it in this way.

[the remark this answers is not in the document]
stephenxs: fixed

kcudnik:
please satisfy code coverage test

lib/RedisRemoteSaiInterface.cpp
```diff
@@ -485,6 +493,114 @@ sai_status_t RedisRemoteSaiInterface::setRedisExtensionAttribute(
     return SAI_STATUS_FAILURE;
 }
 
+bool RedisRemoteSaiInterface::isSaiS8ListValidString(const sai_s8_list_t &s8list, const char* hint)
```
kcudnik, line 496: Missing _ in _ in arrguments, also one argument per line

lib/RedisRemoteSaiInterface.cpp
```diff
@@ -485,6 +493,114 @@ sai_status_t RedisRemoteSaiInterface::setRedisExtensionAttribute(
...
+    SWSS_LOG_ENTER();
+
+    if (flexCounterParam == nullptr || !isSaiS8ListValidString(flexCounterParam->counter_key, "COUNTER_KEY"))
+    {
+        SWSS_LOG_ERROR("Invalid parameters when handling counter operation");
+        return SAI_STATUS_FAILURE;
+    }
+
+    std::vector<swss::FieldValueTuple> entries;
+    std::string key((const char*)flexCounterParam->counter_key.list);
```
kcudnik, line 579: Country missing ? Potential memory garage read

lib/RedisRemoteSaiInterface.cpp
```diff
@@ -485,6 +493,114 @@ sai_status_t RedisRemoteSaiInterface::setRedisExtensionAttribute(
...
+    std::string command;
+
+    if (isSaiS8ListValidString(flexCounterParam->counter_ids, "COUNTER_IDS") && isSaiS8ListValidString(flexCounterParam->counter_field_name, "COUNTER_FIELD_NAME"))
+    {
+        entries.emplace_back(std::string((const char*)flexCounterParam->counter_field_name.list, flexCounterParam->counter_field_name.count),
```
kcudnik, line 584: For all string that youncreate its a lot of code so please create function that will take parametry and return string already with also checking whether string is vslid inside to avoid so much copyright pastę here for entire code, you can also add entires as in oit param and if valid it could internally emplace

lib/RedisRemoteSaiInterface.cpp
```diff
@@ -485,6 +493,114 @@ sai_status_t RedisRemoteSaiInterface::setRedisExtensionAttribute(
...
+    {
+        SWSS_LOG_ERROR("Invalid parameters when handling counter operation");
+        return SAI_STATUS_FAILURE;
+    }
+
+    std::vector<swss::FieldValueTuple> entries;
+    std::string key((const char*)flexCounterParam->counter_key.list);
+    std::string command;
+
+    if (isSaiS8ListValidString(flexCounterParam->counter_ids, "COUNTER_IDS") && isSaiS8ListValidString(flexCounterParam->counter_field_name, "COUNTER_FIELD_NAME"))
```
kcudnik, line 582: Please replace all COUNTERS IDS and other co stanowi string with proporcje defines or constexpr

lib/RedisRemoteSaiInterface.h
```diff
@@ -343,6 +343,17 @@ namespace sairedis
                     _In_ sai_object_id_t objectId,
                     _In_ const sai_attribute_t *attr);
 
+            bool isSaiS8ListValidString(const sai_s8_list_t &s8list,
```
kcudnik, line 346: Missing _ in _

syncd/Syncd.cpp
```diff
@@ -380,6 +382,46 @@ sai_status_t Syncd::processSingleEvent(
     if (op == REDIS_ASIC_STATE_COMMAND_OBJECT_TYPE_GET_AVAILABILITY_QUERY)
         return processObjectTypeGetAvailabilityQuery(kco);
 
+    if (op == REDIS_FLEX_COUNTER_COMMAND_START_POLL)
```
kcudnik, line 385: Each precious if have dedicated function, please do same here

syncd/Syncd.cpp
```diff
@@ -2436,18 +2478,35 @@ void Syncd::processFlexCounterGroupEvent( // TODO must be moved to go via ASIC c
...
+    processFlexCounterGroupEvent(key, op, values, false);
+}
+
+void Syncd::processFlexCounterGroupEvent( // TODO must be moved to go via ASIC channel queue
+        _In_ const std::string &groupName,
+        _In_ const std::string &op,
+        _In_ const std::vector<swss::FieldValueTuple> &values,
+        _In_ bool operateDb)
+{
+    WatchdogScope ws(m_timerWatchdog, op + ":" + groupName);
```
kcudnik, line 2494: Why we suddenly need wathdog here, wathdog was used only for SAI apps here we dont have SAI callz

syncd/Syncd.cpp
```diff
@@ -2469,8 +2528,18 @@ void Syncd::processFlexCounterEvent( // TODO must be moved to go via ASIC channe
...
+    processFlexCounterEvent(key, op, values, false);
+}
+
+void Syncd::processFlexCounterEvent( // TODO must be moved to go via ASIC channel queue
+        _In_ const std::string &key,
+        _In_ const std::string &op,
+        _In_ const std::vector<swss::FieldValueTuple> &values,
+        _In_ bool operateDb)
+{
+    WatchdogScope ws(m_timerWatchdog, op + ":" + key);
```
kcudnik, line 2542: Same here with wathdog

syncd/Syncd.cpp
```diff
@@ -2484,27 +2553,42 @@ void Syncd::processFlexCounterEvent( // TODO must be moved to go via ASIC channe
...
 
     if (!m_translator->tryTranslateVidToRid(vid, rid))
     {
-        SWSS_LOG_WARN("port VID %s, was not found (probably port was removed/splitted) and will remove from counters now",
-                sai_serialize_object_id(vid).c_str());
-
-        op = DEL_COMMAND;
+        if (operateDb)
+        {
+            SWSS_LOG_THROW("port VID %s, was not found (probably port was removed/splitted) and will remove from counters now",
```
kcudnik, line 2567: Why we need to throw  here? It will crash, and fail in remove counter from. Flex list is not critical and syncd can stuull run, please remove

kcudnik (changes requested):

syncd/Syncd.cpp
```diff
@@ -2469,8 +2518,18 @@ void Syncd::processFlexCounterEvent( // TODO must be moved to go via ASIC channe
...
 
-    WatchdogScope ws(m_timerWatchdog, op + ":" + key, &kco);
+    processFlexCounterEvent(key, op, values, false);
+}
+
+void Syncd::processFlexCounterEvent( // TODO must be moved to go via ASIC channel queue
+        _In_ const std::string &key,
+        _In_ const std::string &op,
+        _In_ const std::vector<swss::FieldValueTuple> &values,
+        _In_ bool operateDb)
```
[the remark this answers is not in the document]
stephenxs: change it to a more meaningful name. `fromAsicChannel`

lib/RedisRemoteSaiInterface.cpp
stephenxs: fixed

lib/RedisRemoteSaiInterface.cpp
stephenxs: fixed

lib/RedisRemoteSaiInterface.cpp
stephenxs: fixed

lib/RedisRemoteSaiInterface.cpp
stephenxs: remove those const strings as they do not bring to many benefits. the const strings are for error messages but even without them a developer can easily find where goes wrong.

lib/RedisRemoteSaiInterface.h
stephenxs: fixed

syncd/Syncd.cpp
stephenxs: Removed.

syncd/Syncd.cpp
stephenxs: Removed

syncd/Syncd.cpp
stephenxs: removed.

kcudnik (approved):

stephenxs:
Hi @kcudnik 
can you help to merge it?
thanks

PR #1363, opened 2024-03-26
[saiproxy] Add SAI proxy library

kcudnik:
Will be needed for DASH application

SAI proxy will forward every SAI api call to server


kcudnik:
Will be needed for DASH application

SAI proxy will forward every SAI api call to server

PR #1365, opened 2024-04-02
Use sonic-common agent pool to build swss.

liushilongbuaa:


kcudnik:
please provide description and fix build process ?

liushilongbuaa:
> please provide description and fix build process ?

Convert to draft. Because the original agent pool seems good.
I run 5 jobs on master branch. They all passed. Will keep monitoring it.

PR #1366, opened 2024-04-02
Add dash SAI support for build chain of syncd

Pterosaur:
Add an option to link the dash SAI for syncd.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -41,6 +41,11 @@ AC_ARG_WITH(sai,
 esac])
 AM_CONDITIONAL(SAIVS,   test x$sai = xvs)
 
+AC_ARG_WITH(dashsai,
+    [ --with-dashsai         compiling against DASH SAI],
+    [AM_CONDITIONAL(DASHSAI,   true)],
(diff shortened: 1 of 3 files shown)


kcudnik (approved):

kcudnik:
im not sure whether we should have each specific parameter for specific SAI to add expected libraries, or to add EXTRA_LIBSAI_LDFLAGS in the console before typing make, so configure should pick that up from environment set variables

Pterosaur:
> im not sure whether we should have each specific parameter for specific SAI to add expected libraries, or to add EXTRA_LIBSAI_LDFLAGS in the console before typing make, so configure should pick that up from environment set variables


I understand that this PR is not an elegant way to deal with the dash SAI, and it seems like a temporary fix. There are many codes in sonic-buildimage that depend on vs. It's difficult to make the vs support dual SAI or create a new sku in a short time. However, I hope we will eventually have a new SKU like virtual DPU.

kcudnik:
There are no temporary fixes, it will propagate to scripts and stay forever since different versions will not be compatible in the future

PR #1372, opened 2024-04-15
Fix SSCI parameter when creating MACsec tunnels on Bookworm

saiarcot895:
While Bookworm's iproute2 has built-in support for the MACsec XPN protocols, the SSCI argument is passed in in a different way compared to the patch that we had. Instead of taking in a number, a hex string is taken in instead.

Update the code in sairedis to pass in a hex string for SSCI instead of a number.

Also simplify the endian conversion for the SCI by using `htobe64`, so that we don't need `#ifdef`'s checking for endianness. It's technically non-standard, but since this will be compiled for Linux, that should be fine.

### DIFF
--- a/vslib/SwitchStateBaseMACsec.cpp
+++ b/vslib/SwitchStateBaseMACsec.cpp
@@ -706,11 +706,11 @@ sai_status_t SwitchStateBase::loadMACsecAttrFromMACsecSA(
         // So, this conversion is useful to convert SSCI from the host order to network order.
         //
         // Starting with Debian Bookworm (iproute2 6.1), ssci is interpreted as a hex string,
-        // so this needs to convert the ssci integer to a hex string and account for encoding.
-        auto ssci = attr->value.u32;
+        // so this needs to convert the ssci integer to a hex string, and doesn't need to change
+        // the encoding at this point.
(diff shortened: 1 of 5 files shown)


vslib/SwitchStateBaseMACsec.cpp
```diff
@@ -708,7 +704,15 @@ sai_status_t SwitchStateBase::loadMACsecAttrFromMACsecSA(
 
         // The Linux kernel directly uses ssci to XOR with the salt that is network order,
         // So, this conversion is useful to convert SSCI from the host order to network order.
-        macsecAttr.m_ssci = htonl(attr->value.u32);
+        //
+        // Starting with Debian Bookworm (iproute2 6.1), ssci is interpreted as a hex string,
+        // so this needs to convert the ssci integer to a hex string and account for encoding.
+        auto ssci = attr->value.u32;
```
kcudnik, line 710: why you need extra value if this is used only once ?

saiarcot895: Good point. Updated the PR so that the extra variable isn't created, and updated the comment.

kcudnik (approved):

Pterosaur (approved):

PR #1377, opened 2024-05-09
[syncd]: Generate syncd_dash

Pterosaur:
Make original syncd-vs and syncd-dash co-exist.


kcudnik (approved):

kcudnik:
what's the reason to make this separate binaries ?

Pterosaur:
> what's the reason to make this separate binaries ?

Guohan suggests to keep the dash sai and virtual sai in ONE image and we can toggle them at the runtime instead of compiling time.

lguohan:
should b bmv2 sai, not dash sai. @Pterosaur

PR #1378, opened 2024-05-16
Populate CHASSIS_APP_DB when flex counters are read

viveksrao-arista:
When flex counters are read, if the counter is for a VOQ, populate chassis app db for this VOQ so that on the supervisor it can be used to calculate aggregate stats across all VOQs for a port.


syncd/FlexCounter.cpp
```diff
@@ -31,6 +31,13 @@ static const std::string ATTR_TYPE_PG = "Priority Group Attribute";
 static const std::string ATTR_TYPE_MACSEC_SA = "MACSEC SA Attribute";
 static const std::string ATTR_TYPE_ACL_COUNTER = "ACL Counter Attribute";
 
+// --> Needed for agg voq stats
+static sai_queue_api_t *sai_queue_api = NULL;
+static swss::DBConnector configDb("CONFIG_DB", 0);
+static swss::Table cfgDeviceMetaDataTable(&configDb, CFG_DEVICE_METADATA_TABLE_NAME);
+static swss::DBConnector counterDb("COUNTERS_DB", 0);
+// <-- agg voq stats end
+
```
kcudnik, line 40: please remove all global variables, all connectsion for db should come from parametrs to the class not had coded

syncd/FlexCounter.cpp
```diff
@@ -31,6 +31,13 @@ static const std::string ATTR_TYPE_PG = "Priority Group Attribute";
 static const std::string ATTR_TYPE_MACSEC_SA = "MACSEC SA Attribute";
 static const std::string ATTR_TYPE_ACL_COUNTER = "ACL Counter Attribute";
 
+// --> Needed for agg voq stats
+static sai_queue_api_t *sai_queue_api = NULL;
```
kcudnik, line 35: no api should be uses, VendorCalss shoud be used

syncd/FlexCounter.cpp
```diff
@@ -511,8 +519,72 @@ class CounterContext : public BaseCounterContext
         }
     }
 
+    void addVoqEntryToChassisAppDb(_In_ const sai_object_id_t rid,
```
kcudnik, line 522: first parameter should be also in separate line

syncd/FlexCounter.cpp
```diff
@@ -511,8 +519,72 @@ class CounterContext : public BaseCounterContext
...
+                                   _In_ const sai_object_id_t vid,
+                                   _In_ std::vector<swss::FieldValueTuple> &counterValues,
+                                   _Out_ swss::Table *appDbCountersTable ) {
```
kcudnik, line 525: { to new line, plese forllow code quality

syncd/FlexCounter.cpp
```diff
@@ -511,8 +519,72 @@ class CounterContext : public BaseCounterContext
...
+                SWSS_LOG_ERROR( "sai get_queue_attribute failed for SAI_QUEUE_ATTR_INDEX."
+                      " status %d rid 0x%" PRIx64 "vid 0x%" PRIx64, status, rid, vid) ;
+                return;
+             }
+             auto iface = counterDb.hget(COUNTERS_PORT_NAME_VOQ_MAP, sai_serialize_object_id(vid));
+             if (!iface)
+             {
+                SWSS_LOG_ERROR("Cannot query port name for queue. rid 0x%" PRIx64 "vid 0x%" PRIx64, rid, vid);
+                return;
+             }
```
kcudnik, line 553: add empty line before each if and after each }

syncd/FlexCounter.cpp
```diff
@@ -363,6 +370,7 @@ void deserializeAttr(
     sai_deserialize_acl_counter_attr(name, attr);
 }
 
+
 template <typename StatType>
 class CounterContext : public BaseCounterContext
```
kcudnik, line 375: this should be in separate file 1 class per one file cpp

syncd/FlexCounter.cpp
```diff
@@ -511,8 +519,72 @@ class CounterContext : public BaseCounterContext
...
+          if (status != SAI_STATUS_SUCCESS ) {
+             SWSS_LOG_ERROR( "sai get_queue_attribute failed for SAI_QUEUE_ATTR_TYPE. "
+                   "status %d rid 0x%" PRIx64 "vid 0x%" PRIx64, status, rid, vid);
+             return;
+          }
+          if (attr.value.s32 == SAI_QUEUE_TYPE_UNICAST_VOQ) {
+             attr.id = SAI_QUEUE_ATTR_INDEX;
+             status = sai_queue_api->get_queue_attribute(rid, 1, &attr);
+             if (status != SAI_STATUS_SUCCESS ) {
+                SWSS_LOG_ERROR( "sai get_queue_attribute failed for SAI_QUEUE_ATTR_INDEX."
```
kcudnik, line 544: why there is space after ( ? remove

syncd/FlexCounter.cpp
```diff
@@ -1014,6 +1097,14 @@ FlexCounter::FlexCounter(
     m_enable = false;
     m_isDiscarded = false;
 
+    if (!sai_queue_api) {
+       sai_status_t status;
+       status = sai_api_query(SAI_API_QUEUE, (void **)&sai_queue_api);
```
kcudnik, line 1102: do not use gloal apis ! Vendor class should be used!

syncd/FlexCounter.cpp
```diff
@@ -1396,6 +1513,16 @@ void FlexCounter::flexCounterThreadRunFunction()
...
     swss::Table countersTable(&pipeline, COUNTERS_TABLE, true);
 
+    string switchType;
+    swss::Table *appDbCountersTable = NULL;
+    getCfgSwitchType(switchType);
+    if (switchType == "voq")
+    {
+       swss::DBConnector appDb("CHASSIS_APP_DB", 0, true);
+       swss::RedisPipeline appDbPipeline(&appDb);
+       appDbCountersTable = new swss::Table(&appDbPipeline, "COUNTERS", true);
```
kcudnik, line 1523: don't use new operator !

kcudnik (changes requested):
fix coments

syncd/FlexCounter.cpp
viveksrao-arista: Sorry, I do not understand. I need a handle `sai_queue_api` so that I can query queue stats like this ` status = sai_queue_api->get_queue_attribute(rid, 1, &attr);`. I am not aware of an alternative to this. What is VendorClass? How to use it? Can you point me to an example?

syncd/FlexCounter.cpp
viveksrao-arista: This code is not added by me.

syncd/FlexCounter.cpp
viveksrao-arista: Well that's string concatenation and is used to keep the line within a reasonable length. Helps when editing code in terminals and editors that do not support line length of say for eg more than 85 characters or so. Good practice to make your code readable across various editors and terminals. Linux mandates a line length of 85 chars. Not sure what Sonic enforces. Good to follow none the less.

syncd/FlexCounter.cpp
viveksrao-arista: Why? I need a pointer and not a reference to an object here.

kcudnik: use shared_ptr

syncd/FlexCounter.cpp
kcudnik: please keep same style as rest of the file

syncd/FlexCounter.cpp
kcudnik: use VendorSai class

syncd/FlexCounter.cpp
viveksrao-arista: Fixed.

syncd/FlexCounter.cpp
viveksrao-arista: Done.

syncd/FlexCounter.cpp
viveksrao-arista: Done.

syncd/FlexCounter.cpp
viveksrao-arista: Done.

syncd/FlexCounter.cpp
viveksrao-arista: Done

syncd/FlexCounter.cpp
viveksrao-arista: Done

syncd/FlexCounter.cpp
viveksrao-arista: Done

syncd/FlexCounter.cpp
viveksrao-arista: Done

syncd/FlexCounter.cpp
```diff
@@ -511,8 +512,82 @@ class CounterContext : public BaseCounterContext
...
     }
 
+    void addVoqEntryToChassisAppDb(
+            _In_ const sai_object_id_t rid,
+            _In_ const sai_object_id_t vid,
+            _In_ std::vector<swss::FieldValueTuple> &counterValues,
+            _In_ swss::DBConnector &countersDb,
+            _In_ std::shared_ptr<swss::Table> cfgDeviceMetaDataTable,
+            _Out_ std::shared_ptr<swss::Table> appDbCountersTable ) {
+       // We need to update CHASSIS_APP_DB only if it is a VOQ counter. So first check
```
kcudnik, line 522: please move { to new lina and add empty line every after } to make code readable

syncd/FlexCounter.cpp
```diff
@@ -995,10 +1080,41 @@ class AttrContext : public CounterContext<AttrType>
...
+            {
+               // Chassis based system
+               appDbCountersTable->set(sai_serialize_object_id(vid), values, "");
+            }
         }
     }
 };
 
+void FlexCounter::getCfgSwitchType(
+         _In_ std::string &switch_type)
```
kcudnik, line 1093: you are changing output variable, yo ushould use _ Out _, or better make function as return std::string instead of modyfying input parameter

syncd/FlexCounter.cpp
```diff
@@ -995,10 +1080,41 @@ class AttrContext : public CounterContext<AttrType>
...
+   }
+   catch(const std::system_error& e)
+   {
+      SWSS_LOG_ERROR("System error: %s", e.what());
+      switch_type = "switch";
+   }
+
+   if (switch_type != "voq" && switch_type != "fabric" && switch_type != "chassis-packet" && switch_type != "switch" && switch_type != "dpu")
+   {
+      SWSS_LOG_ERROR("Invalid switch type %s configured", switch_type.c_str());
```
kcudnik, line 1111: this seems really fishy here, why you need to check all those variables? this should be more unified?

syncd/FlexCounter.cpp
```diff
@@ -995,10 +1080,41 @@ class AttrContext : public CounterContext<AttrType>
...
         }
     }
 };
 
+void FlexCounter::getCfgSwitchType(
+         _In_ std::string &switch_type)
+{
+   try
+   {
+      if (!m_cfgDeviceMetaDataTable->hget("localhost", "switch_type", switch_type))
```
kcudnik, line 1097: is this connecting for localhost database to get this ? this seems a hacky way to do that, also all variables are hard coded, those should be defined somewhere with comments

syncd/FlexCounter.cpp
```diff
@@ -1014,6 +1130,18 @@ FlexCounter::FlexCounter(
...
+
+    getCfgSwitchType(m_switchType);
+
+    if (m_switchType == "voq")
+    {
+       m_appDb = std::make_shared<swss::DBConnector>("CHASSIS_APP_DB", 0, true);
+       m_appDbPipeline = std::make_shared<swss::RedisPipeline>(m_appDb.get());
+       m_appDbCountersTable = std::make_shared<swss::Table>(m_appDbPipeline.get(), "COUNTERS", true);
+    }
+
```
kcudnik, line 1144: all databases names are hardcoded, they should be obrained from defines or config header

syncd/FlexCounter.cpp
viveksrao-arista: Copying this function from https://github.com/sonic-net/sonic-swss/blob/master/orchagent/main.cpp#L160

syncd/FlexCounter.cpp
viveksrao-arista: Done

syncd/FlexCounter.cpp
viveksrao-arista: Done

syncd/FlexCounter.cpp
viveksrao-arista: Pretty much every access to database is hardcoded in FlexCounter.cpp. We probably should change all of them in a separate pull request to not mix things up.

syncd/FlexCounter.cpp
viveksrao-arista: Same as above

vivekverma-arista:
We will not be changing sonic-sairedis as per the new design, therefore closing this pull request.

Updated HLD: [#1587](https://github.com/sonic-net/SONiC/pull/1587)

PR #1381, opened 2024-05-21
[submodule] Update SAI submodule to v1.14

kcudnik:
Will include adding new POE api

### DIFF
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -90,6 +90,7 @@ libsairedis_la_SOURCES = \
 						 sai_redis_ars.cpp \
 						 sai_redis_ars_profile.cpp \
 						 sai_redis_twamp.cpp \
+						 sai_redis_poe.cpp \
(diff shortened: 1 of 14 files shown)


kcudnik:
@prince https://github.com/sonic-net/sonic-swss/pull/3154 could this be the issue we see here on this repo? many PRs are failing in similar way, almost all tests fail in connection issue

prsunny:
> @prsunny [sonic-net/sonic-swss#3154](https://github.com/sonic-net/sonic-swss/pull/3154) could this be the issue we see here on this repo? many PRs are failing in similar way, almost all tests fail in connection issue

we've another issue on swss thats being tracked. But this could be an issue as well

kcudnik:
@prsunny @rlhui i made separate pr to fix pipeline: https://github.com/sonic-net/sonic-sairedis/pull/1382 just to see if this will fix the issue

PR #1385, opened 2024-05-27
Support bulk counter init for hackathon

stephenxs:
for hackathon only
no ut.


stephenxs:
created it by mistake.

PR #1386, opened 2024-05-29
[saiproxy] Implement all APIs except bulk

kcudnik:
Initial implementaion of all APIs except bulk and notifications required for sai proxy feature


Pterosaur:
Hi @r12f , will the saiproxy only be used in the KVM dpu or real physical platform as well?

kcudnik:
> Hi @r12f , will the saiproxy only be used in the KVM dpu or real physical platform as well?

this should not matter, its proxy it will not care

PR #1395, opened 2024-06-13
[saisubmodule] Update SAI submodule to latest master

kcudnik:
Will include SAI metadata methods to update notification pointers which will do that automatically for all notifications reducing number of work we need to update that each time


r12f:
thanks Kamil!

PR #1399, opened 2024-07-04
optimize for best candidate find

inspurSDN:
why I did:
    in best candidate, it looking for all items of this type to search
    for the most seamful item, when their is too many vlan member, we
    can wait very long.
what I did:
    since VLAN and VLAN MEMBER is unique, so if we find one with all attr
    matched, we break and do not compare others.

### DIFF
--- a/syncd/BestCandidateFinder.cpp
+++ b/syncd/BestCandidateFinder.cpp
@@ -1886,6 +1886,9 @@ std::shared_ptr<SaiObj> BestCandidateFinder::findCurrentBestMatchForGenericObjec
                 soci.equal_attributes);
 
         candidateObjects.push_back(soci);
+        if (soci.equal_attributes == attrs.size()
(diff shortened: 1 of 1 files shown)


syncd/BestCandidateFinder.cpp
```diff
@@ -1886,6 +1886,9 @@ std::shared_ptr<SaiObj> BestCandidateFinder::findCurrentBestMatchForGenericObjec
                 soci.equal_attributes);
 
         candidateObjects.push_back(soci);
+        if (soci.equal_attributes == attrs.size()
+            && (object_type == SAI_OBJECT_TYPE_VLAN || object_type == SAI_OBJECT_TYPE_VLAN_MEMBER))
```
kcudnik, line 1890: This is object specific, please remove it from this place and add object specific logic for finding best candidatte

kcudnik (changes requested):
Address comments

PR #1400, opened 2024-07-04
after apply_view, use malloc_trim to release memory to system

inspurSDN:
why I did:
    after warm-reboot with all vlan members, syncd use memory up to 1.8g,
    and can not reduce to its normal state.
what I did:
    this come from the glibc behavior, in the main heap, it almost not release
    memory to system, in apply view, huge memory used for compare logic, and
    do not use it again. so I add malloc_trim to release it.

### DIFF
--- a/syncd/Syncd.cpp
+++ b/syncd/Syncd.cpp
@@ -36,6 +36,8 @@
 #include <iterator>
 #include <algorithm>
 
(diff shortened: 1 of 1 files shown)


kcudnik:
How many clan members are in your case ? How did you counted that memory ?

PR #1401, opened 2024-07-04
Use temp view also in fast-reboot

arfeigin:
syncd_init_common.sh is checking fast-reboot by reading /proc/cmdline. However, /proc/cmdline will not change after config reload. So, consider a case like fast-reboot -> config reload. In the config reload process, syncd_init_common.sh will also treat it as fast-reboot and probably enter wrong logic.

### DIFF
--- a/syncd/scripts/syncd_init_common.sh
+++ b/syncd/scripts/syncd_init_common.sh
@@ -26,10 +26,8 @@ else
     CMD_ARGS=
 fi
 
-# Use temporary view between init and apply except when in fast-reboot
-if [[ "$(cat /proc/cmdline)" != *"SONIC_BOOT_TYPE=fast-reboot"* ]]; then
-    CMD_ARGS+=" -u"
-fi
+# Use temporary view between init view and apply view
+CMD_ARGS+=" -u"
 
 # Create a folder for SAI failure dump files
 mkdir -p /var/log/sai_failure_dump/


arfeigin:
@kcudnik Can you please help reviewing this PR?

Junchao-Mellanox (approved):

arfeigin:
@vaibhavhd Can you please reviewing this PR?

kcudnik (approved):

Junchao-Mellanox:
Hi @vaibhavhd , could you please help review this?

syncd/scripts/syncd_init_common.sh
```diff
@@ -26,10 +26,8 @@ else
     CMD_ARGS=
 fi
 
-# Use temporary view between init and apply except when in fast-reboot
-if [[ "$(cat /proc/cmdline)" != *"SONIC_BOOT_TYPE=fast-reboot"* ]]; then
```
vaibhavhd, line 30: We can still keep the check but change the condition. There is a system wide fast-reboot flag that is now used.
May be using that flag is better than 1) removing the check 2) using cmdline which does not change.

Besides, in the current change it is not just fixing the problem with config reload (after fast-reboot). This PR will change the behavior of the fast-reboot as well.

vaibhavhd (commented):
Please update the PR with the test results.

stepanblyschak:
@vaibhavhd The change is in our internal 202405 branch and running in regression for few month now with no issues related to fast-reboot

vaibhavhd (approved):

bingwang-ms:
@arfeigin Can you please sync with master branch?

dprital:
@bingwang-ms , Can you please merge and cherry pick to 202405 ?

PR #1404, opened 2024-07-19
[vslib] Add support for PoE feature

SerhiyBoikoPLV:
Add PoE virtual switch implementation to vslib
Add a python script to spawn the PoE syncd process


syncd/SaiSwitch.cpp
```diff
@@ -150,7 +150,12 @@ sai_switch_type_t SaiSwitch::getSwitchType() const
         attr.value.s32 = SAI_SWITCH_TYPE_NPU;
     }
 
-    SWSS_LOG_INFO("switch type: '%s'", (attr.value.s32 == SAI_SWITCH_TYPE_NPU ? "SAI_SWITCH_TYPE_NPU" : "SAI_SWITCH_TYPE_PHY"));
+    if (attr.value.s32 == SAI_SWITCH_TYPE_NPU)
```
kcudnik, line 153: Use sai serialize of switch type

syncd/scripts/poesyncdmgrd
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env python3
+
+import os
+import signal
+import subprocess
```
kcudnik, line 5: Why this script is needed who will be calling it ?

vslib/SwitchConfig.cpp
```diff
@@ -96,6 +105,7 @@ bool SwitchConfig::parseSwitchType(
                 SAI_VALUE_VS_SWITCH_TYPE_BCM81724,
                 SAI_VALUE_VS_SWITCH_TYPE_BCM56850,
                 SAI_VALUE_VS_SWITCH_TYPE_BCM56971B0,
+                SAI_VALUE_VS_SWITCH_TYPE_POE_VS,
```
kcudnik, line 108: Vs at the end is not needed you already have vs in name before

vslib/SwitchConfig.h
```diff
@@ -28,6 +28,8 @@ namespace saivs
 
         SAI_VS_SWITCH_TYPE_NVDA_MBF2H536C,
 
+        SAI_VS_SWITCH_TYPE_POE_VS,
```
kcudnik, line 31: Remove vs at the end

syncd/scripts/poesyncdmgrd
SerhiyBoikoPLV: This is meant to be called by the supervisord process instead of spawning syncd (poesyncd) directly. The main purpose of the script is to check if PoE is supported on the platform and if we should be running syncd at all.

syncd/SaiSwitch.cpp
SerhiyBoikoPLV: added `sai_serialize_switch_type()`

kcudnik:
Please write some unittests to satisfy code coverage

vslib/SwitchPoE.cpp
```diff
@@ -0,0 +1,346 @@
...
+    attr.id = SAI_POE_PORT_ATTR_ADMIN_ENABLED_STATE;
+    attr.value.booldata = false;
+    CHECK_STATUS(set(SAI_OBJECT_TYPE_POE_PORT, object_id, &attr));
+
+    /* if not set in create_internal then use default values*/
+    if (!power_limit)
+    {
+        attr.id = SAI_POE_PORT_ATTR_POWER_LIMIT;
+        attr.value.u32 = 0;
+        CHECK_STATUS(set(SAI_OBJECT_TYPE_POE_PORT, object_id, &attr));
```
praveenraja1, line 183: This is just for my understanding, 

I was under the impression that this code would eventually interact with PoE controller, could you point me to code where that happens.

SerhiyBoikoPLV: This is an implementation for the virtual switch. There is no poe cotroller to communicate with, so we only report dummy data.
When vendors are writing sai for their physical devices and handling SAI_POE_* attributes then they will be communicating with the poe controller.

matiAlfaro (approved):
LGTM

kcudnik (dismissed):

SerhiyBoikoPLV:
@kcudnik , could you please approve the semgrep workflow? Do you have any other comments that prevent this PR from being merged?

Cahb:
Rebased on current master and pushed. Reviving this, open to re-review.

PR #1413, opened 2024-08-12
[Build] - Reduce libsairedis build time by 16m

mikemallin:
# Add -fno-var-tracking-assignments to pysairedis_wrap.cpp when built with GCC

This improves compilation time by a total of 16m by disabling the GCC feature "Variable Tracking Assingments."

The trade-off for disabling this feature is reduced debuggability if anyone needs to go through the object file pysairedis_wrap.o with a debugger. Please see the background below for details.

This was identified via the following issue in the SWIG project:
- https://github.com/swig/swig/issues/2384#issuecomment-1596308842

This is a performance issue in GCC specifically, not CLANG.

# Results:
On a server with the following specifications, the build time for libsairedis went from 20m23s down to 3m50s.

    CPU: Intel(R) Xeon(R) Gold 6430
    RAM: 1Tb DDR5
    OS: Ubuntu 22.04
    g++: g++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0

On an Apple MacBook Pro (with a Debian build VM), the build time for libsairedis also dropped significantly from 17m43s down to 3m46s

    CPU: Apple M3 Pro
    RAM: 18Gb
    OS: Debian 12 Bookworm
    g++: g++ (Debian 12.2.0-14) 12.2.0

# Background:
- https://gcc.gnu.org/wiki/Var_Tracking_Assignments
- https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/6/html/developer_guide/ch-debug-vta#ch-debug-vta


kcudnik (dismissed):

mikemallin:
@kcudnik I've fixed the build issue, would it be possible to approve again? Thanks!

kcudnik (approved):

PR #1416, opened 2024-08-21
[nvidia] Skip SAI discovery on ports

stepanblyschak:
Given that modern systems have lots of ports, performing SAI discovery takes very long time, e.g. (8 sec) for 256 port system. This has a big impact of fast-boot downtime and the discovery itself is not required for Nvidia platform fast-boot.

Same applies to Nvidia fastfast-boot (aka warm-boot), yet needs to be tested separately.


kcudnik (changes requested):
this will lead to inconsistency ASIC_DB vs what's on device, which will later on lead to crash

syncd/SaiSwitchInterface.h
```diff
@@ -89,7 +89,8 @@ namespace syncd
 
             virtual void onPostPortCreate(
                     _In_ sai_object_id_t port_rid,
-                    _In_ sai_object_id_t port_vid) = 0;
+                    _In_ sai_object_id_t port_vid,
+                    _In_ bool discoverPortObjects = true) = 0;
```
kcudnik, line 93: this is very strict to ports, if we decide later on to do something similar on other objects then this is not optimal solution

syncd/Syncd.cpp
```diff
@@ -5296,3 +5296,23 @@ syncd_restart_type_t Syncd::handleRestartQuery(
...
+
+bool Syncd::shouldDiscoverPortObjects() const
+{
+    SWSS_LOG_ENTER();
+
+#ifdef SKIP_SAI_PORT_DISCOVERY_ON_FAST_BOOT
+    const bool discoverPortObjectsInFastBoot = false;
+#else
+    const bool discoverPortObjectsInFastBoot = true;
+#endif
```
kcudnik, line 5308: fast boot cak be initiated after code was compiled which then this check will be hardcoded

kcudnik: also there are no tests for testing this code

stepanblyschak: > fast boot cak be initiated after code was compiled which then this check will be hardcoded

This was the intention

syncd/SaiSwitchInterface.h
stepanblyschak: The function is meant to be used on ports. Considering current approach, I assume there will be onPostXCreate() functions for other object types. Then, if needed, they can accept a boolean flag in the same way. This is simple and gives required granularity.

stepanblyschak:
> this will lead to inconsistency ASIC_DB vs what's on device, which will later on lead to crash

@kcudnik Yes, could you suggest a test to cause a crash?

kcudnik:
> > this will lead to inconsistency ASIC_DB vs what's on device, which will later on lead to crash
> 
> @kcudnik Yes, current design leads to performance problems on devices with lots of ports (could be 512, 1024 and more - tens of thousands keys to insert to ASIC_DB on init). Could you suggest a test to cause a crash?

Discovery Is done only once at switch create, how long it takes in your case ? We need to have a full view of device since later on warm boo will try to delete objects which were not discovered at current stage it's hard to predict when cras will happen, but it will be 99% on warm boot

stepanblyschak:
@kcudnik I got your point, please note that Nvidia does not use standard warm-reboot flow in syncd, instead uses fast-fast-boot mode which is quite similar to fast-boot. No comparison logic is involved on Nvidia, therefore in my testing I did not observe crashes even after doing consecutive fast-reboot and warm-reboot.

I attach the log of discovery during fast-reboot. I measured 4.8 sec on 202405 dedicated to discovery which is 16% of time budget for fast-reboot. It is only a 256 port system and we expect the time to increase linearly with the number of ports which is growing with new HWSKUs.

[logs.txt](https://github.com/user-attachments/files/17935003/logs.txt)

kcudnik:
Logs are very consuming time please check without logging, also maybe there is also room on your reiver to optimize speed to return objects on ports ?

kcudnik:
fropm logs you pasted, it seems like you are creating those posts explicitly?  why thsoe ports are not existing already when switch is created ?

lguohan:
i am a little bit confuse, the PR title is about fast-boot. @kcudnik , while you are talking about warm reboot. what does this code impact?

kcudnik:
it does not matter whether is warm boot or fast boot this wiil cause inconsestincy in db and later on crash

syncd/Syncd.cpp
kcudnik: this should be runtime check

syncd/SaiSwitch.cpp
```diff
@@ -948,46 +948,50 @@ void SaiSwitch::redisUpdatePortLaneMap(
 
 void SaiSwitch::onPostPortCreate(
         _In_ sai_object_id_t port_rid,
-        _In_ sai_object_id_t port_vid)
+        _In_ sai_object_id_t port_vid,
+        _In_ bool discoverPortObjects)
```
kcudnik, line 952: i feel like this eintrie change in this function is overcomplicated, it sholud be something like this:

```C++

if (object_type(oid) == SAI_OBJECT_TYPE_PORT && shouldSkipPorts)
  continue;

Comment (stepanblyschak) [reply syncd/SaiSwitch.cpp:952]: @kcudnik 
Regarding ```object_type(oid) == SAI_OBJECT_TYPE_PORT```, the function is called ```onPostPortCreate``` so unless someone is calling it on object other than port I don't think this check is needed.

Do you mean early return? Like:

```c++
redisUpdatePortLaneMap(port_rid);

if (!discoverPortObjects)
{
    return;
}

...
```

syncd/Syncd.cpp
stepanblyschak: @kcudnik What is the benefit of runtime check here? Syncd is compiled per platform and on Nvidia we do not want to run discovery. We know this at compile time.

syncd/SaiSwitch.cpp
kcudnik: ah, i thoung you also modify discover process, since it will also discover all objects on all ports, so i guess on cold boot you only need onpostportcreate, but this could still crash on next fast-boot

please do couple of fst-boot to fast-boot reboots with your patch to see if this will wrok

kcudnik: looking at taht code, you only need to modify SaiDiscovery process with flag to ignore port discovery, no else code is needed to be changed

kcudnik: and ig you look on master, in SaiDiscovery.cpp file at line 34, you can actually pass new flag - to not discover port objects over VendorSaiOptions class to not  forward all bool arguments to discover ports, if you want

stepanblyschak: @kcudnik Thanks, looking into VendorSaiOptions, that is new in master. Will check whether I can utilize it to avoid parameter passing.

What I tested few month ago was 202311 with this patch:
cold boot -> fast boot -> fast boot
cold boot -> fast boot -> warm boot

Will double check on latest master.

kcudnik: Just simple reboot maybe not enough, I'm maybe this required: cold boot, fast boot, query ingress priority groups and assign something like all or query port queues, ten warm boot and again warm boot etc, there must be operation that will query objects that were not queried in discovery

kcudnik: And this dlag should be command line options like skip port discovery with just couple line changes I a pro o used in this comment

stepanblyschak: > Just simple reboot maybe not enough, I'm maybe this required: cold boot, fast boot, query ingress priority groups and assign something like all or query port queues, ten warm boot and again warm boot etc, there must be operation that will query objects that were not queried in discovery

@kcudnik This is how the test was done. Orchagent queries PGs, queues that were not queried during discovery - https://github.com/sonic-net/sonic-swss/blob/master/orchagent/portsorch.cpp#L5738




> And this dlag should be command line options like skip port discovery with just couple line changes I a pro o used in this comment

The command line option means someone can disable it at runtime and get broken fast boot, we would like to avoid that.

kcudnik: yes, since we don want to skip port ddetection on other platforms than yours

kcudnik: and you can disable those ports in init script for your platform only

stepanblyschak: @kcudnik Platform is known at compile time, syncd is compiled differently for different platforms. My change as is right now should not affect other platforms. This change purpose is to improve startup time, however with platform detection done in script I will add some additional CPU cycles for that. Even though it is very small, on some lower systems the init scripts execution time is worse, that's why I am leaning towards moving all to compile time if possible.

kcudnik: if it's know at compile time, put this:

```c
#ifdef nvidia
if (object_type(oid) == SAI_OBJECT_TYPE_PORT)
  continue;
#endif
```
in si discovery

kcudnik: also add  log warn message, that discovery port was disabled on nvidia platform

kcudnik (approved):

kcudnik:
You could add this message in syncd constructor

syncd/SaiDiscovery.cpp
```diff
@@ -60,6 +60,13 @@ void SaiDiscovery::discover(
 
     sai_object_type_t ot = m_sai->objectTypeQuery(rid);
 
+#ifdef SKIP_SAI_PORT_DISCOVERY
```
kcudnik, line 63: Just small wu3stion, this check should be done at line 90 after putting port o jest id to discovered set

syncd/SaiSwitch.cpp
stepanblyschak: @kcudnik Yes, but not on get operations, since this is not required for fast/warm boot for us. Priority is on set operations.

syncd/SaiDiscovery.cpp
stepanblyschak: Fixed

kcudnik (approved):

stepanblyschak:
@kcudnik Can be merged?

kcudnik (approved):

stepanblyschak:
Test fails on:

```
[ RUN      ] VendorSaiTest.bulk_flow_entry
Assertion failed: _owned.empty () (src/own.cpp:197)
/bin/bash: line 6: 25568 Aborted                 ${dir}$tst
```

Didn't reproduce locally, it looks like the assert is from zmq lib. Will restart to see if issue persists.

kcudnik:
Given that modern systems have lots of ports, performing SAI discovery takes very long time, e.g. (8 sec) for 256 port system. This has a big impact of fast-boot downtime and the discovery itself is not required for Nvidia platform fast-boot.

Same applies to Nvidia fastfast-boot (aka warm-boot), yet needs to be tested separately.

PR #1424, opened 2024-09-26
Adding vpp platform

yue-fred-gao:
# why
VPP is a new platform of SONiC. Here is the HLD: https://github.com/sonic-net/sonic-platform-vpp/blob/main/docs/HLD/SONICVPP-HLD.md. Currently, VPP is built by patching itself on top of VS platform, which requires it carrying a lot of diffs. It is very difficult to upgrade SONiC because patch may have a lot of conflicts. We will make VPP a platform as any other platforms to avoid patching. 

# What this change is doing
VPP SAI layer is built on top of vslib. It uses some classes from sonic-sairedis/meta so we need to export the header files from libsaimetata-dev.
This PR also adds some vpp specific logic to syncd_init_common.sh before starting syncd.


lguohan (approved):

lguohan:
lgtm. @kcudnik  to check.

debian/libsaimetadata-dev.install
```diff
@@ -1,5 +1,12 @@
 meta/sai*.h usr/include/sai
 meta/Sai*.h usr/include/sai
+meta/Meta.h usr/include/sai
+meta/AttrKeyMap.h usr/include/sai
+meta/MetaKeyHasher.h usr/include/sai
+meta/OidRefCounter.h usr/include/sai
+meta/PortRelatedSet.h usr/include/sai
+meta/Notification*.h usr/include/sai
+meta/Globals.h usr/include/sai
```
kcudnik, line 9: not sure if this is good idea to make those headers part of this package

kcudnik:
i will need some meeting to discuss vpp platform to understand what is the long term goal here

yue-fred-gao:
> i will need some meeting to discuss vpp platform to understand what is the long term goal here

Sounds good. When will be a good time for you?

lguohan:
vpp sai is the same concept as dash sai, instead of bmv2 implementation, this is dataplane implementation on top of vpp. the difference between bmv2 v.s vpp is the vpp will implementation API more like traditional switches asic, while bmv2 is on the dpu.

 https://github.com/sonic-net/DASH/tree/main/dash-pipeline/SAI/src

kcudnik:
it seems for me like vpp is concept of another virtual switch, my question is why we need another virtual switch next to sairedis-VS and DASH bmv2 implementation? why we need to keep track and maintain of 3 virtual switches ?

yue-fred-gao:
> it seems for me like vpp is concept of another virtual switch, my question is why we need another virtual switch next to sairedis-VS and DASH bmv2 implementation? why we need to keep track and maintain of 3 virtual switches ?

vpp is a very popular data plane with rich features. I am not familiar with vs but Riff and Jing have done some research evaluating the vpp and vs platform. @r12f and @zjswhhh , can you please share your research? bmv2 is a completely different data plane. It implements DPU while vpp and vs implement NPU functionalities.

zjswhhh:
Hi @kcudnik - sonic vs platform only supports control plane. We wanted to be able to do data plane validate with virtual testbeds too, that was the intention for adding vpp platform. 

> i will need some meeting to discuss vpp platform to understand what is the long term goal here

zjswhhh (approved):
lgtm

debian/libsaimetadata-dev.install
kcudnik: why you need those headers be present here ?

yue-fred-gao: sonic-vpp is morphed from sonic-vs. Similar to vs, it uses classes from sonic-sairedis/meta so we need to export the header files from libsaimetata-dev.

kcudnik: ok, let see how it will go

yue-fred-gao: Thanks. What's the next step for this PR?

kcudnik (approved):

debian/libsaimetadata-dev.install
kcudnik: i signed off

yue-fred-gao: Thanks!

PR #1426, opened 2024-09-28
[202405] update the SAI  submodule in sonic-sairedis

arlakshm:
Update the SAI submodule for 202405 to pickup the fix for https://github.com/sonic-net/sonic-buildimage/issues/19972


arlakshm:
@ysmanman and @saksarav-nokia for viz..

kcudnik (approved):

PR #1430, opened 2024-10-10
[syncd] Add workaround for port error status notification

kcudnik:
Adding new field in port status notification is breaking
change and can cause some issues when getting notification
from older SAI library, this will add workaround to handle
that case

### DIFF
--- a/syncd/NotificationHandler.cpp
+++ b/syncd/NotificationHandler.cpp
@@ -38,6 +38,13 @@ void NotificationHandler::setApiVersion(
     m_apiVersion = apiVersion;
 }
 
+sai_api_version_t NotificationHandler::getApiVersion() const
+{
+    SWSS_LOG_ENTER();
+
+    return m_apiVersion;
+}
+
 void NotificationHandler::setSwitchNotifications(
         _In_ const sai_switch_notifications_t& switchNotifications)
 {
--- a/syncd/NotificationHandler.cpp
+++ b/syncd/NotificationHandler.cpp
@@ -1,4 +1,5 @@
 #include "NotificationHandler.h"
+#include "Workaround.h"
 #include "sairediscommon.h"
 
 #include "swss/logger.h"
@@ -10,8 +11,10 @@
 using namespace syncd;
 
 NotificationHandler::NotificationHandler(
-        _In_ std::shared_ptr<NotificationProcessor> processor):
-    m_processor(processor)
+        _In_ std::shared_ptr<NotificationProcessor> processor,
+        _In_ sai_api_version_t apiVersion):
+    m_processor(processor),
+    m_apiVersion(apiVersion)
 {
     SWSS_LOG_ENTER();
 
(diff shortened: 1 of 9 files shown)


kcudnik:
I'm just creating this PR to see what i miss from build, i will later split this to 2 PR's as advancing SAI submodule and then commiting actual workaround

syncd/VendorSai.cpp
```diff
@@ -1300,6 +1354,52 @@ sai_status_t VendorSai::bulkRemove(
             object_statuses);
 }
 
+sai_status_t VendorSai::bulkRemove(
```
prgeor, line 1357: @kcudnik I see this change is not relevant as per the PR title. Can this be kept outside this PR and added in another PR?

syncd/VendorSai.cpp
```diff
@@ -998,6 +998,60 @@ sai_status_t VendorSai::bulkCreate(
             object_statuses);
 }
 
+sai_status_t VendorSai::bulkCreate(
```
prgeor, line 1001: @kcudnik this change be added as another PR as this is NOT relevant as per the PR title

syncd/VendorSai.cpp
kcudnik: read this comment: https://github.com/sonic-net/sonic-sairedis/pull/1430#issuecomment-2405155952, this is 1st pr: https://github.com/sonic-net/sonic-sairedis/pull/1431

syncd/VendorSai.cpp
kcudnik: read this comment: https://github.com/sonic-net/sonic-sairedis/pull/1430#issuecomment-2405155952, this is 1st pr: https://github.com/sonic-net/sonic-sairedis/pull/1431

prgeor:
@kcudnik could you please check the build failure.

kcudnik:
Yes, already fixed in latest pr #1431 but swss repo tests are failing because of sai backward compatibility break, I already fixed that and posted pr on swss repo pleased follow thread on #1431

kcudnik:
SAI was updated in separate PR: https://github.com/sonic-net/sonic-sairedis/pull/1431

syncd/NotificationHandler.cpp
```diff
@@ -93,7 +111,9 @@ void NotificationHandler::onPortStateChange(
 {
     SWSS_LOG_ENTER();
 
-    auto s = sai_serialize_port_oper_status_ntf(count, data);
+    auto ntfdata = Workaround::convertPortOperStatusNotification(count, data, m_apiVersion);
+
+    auto s = sai_serialize_port_oper_status_ntf((uint32_t)ntfdata.size(), ntfdata.data());
```
prgeor, line 116: @kcudnik I think you need to parse the new field `sai_port_error_status_t` in sai_serialize_port_oper_status_ntf() otherwise this new field will not be posted to redis DB and OA will not see this field in the notification

prgeor (approved):

kcudnik:
@prgeor why there was force push to my branch from you?

PR #1432, opened 2024-10-15
[syncd] Create neighbor entries before next hop

kcudnik:
During warm boot scenario make sure we will create all neighbor entries before next hop that have the same IP address as neighbor entry. In Broadcom platform neighbor entry needs to be created before next hop with the same ip address otherwise create next hop will fail (hardware limitation?).


kcudnik:
ready to review

lguohan (approved):

PR #1448, opened 2024-10-28
Rename file name to fit case insensitive file system.

liushilongbuaa:
Rename file to avoid clone issue in case insensitive file system.
This caused auto-cherry-pick don't work.


kcudnik:
Codeql is failing on Jason.hpp include, I think this branch is missing one of updates where we switched to use nlohmann package directly

kcudnik:
This needs to be cherry picked

PR #1456, opened 2024-11-07
[DASH] Add meter rule bulk support in syncd

mukeshmv:
Add VendorSai bulk quad oid api support for DASH SAI meter rule object.


mukeshmv:
> Please provide more description

Added description

kcudnik (approved):

PR #1458, opened 2024-11-11
[vslib] SAI_KEY_VS_OPER_SPEED_IS_CONFIGURED_SPEED, SAI_PORT_ATTR_HOST_TX_READY_STATUS support

ishidawataru:
This PR adds two features to `vslib`.

- `SAI_KEY_VS_OPER_SPEED_IS_CONFIGURED_SPEED`: when `true`, `SAI_PORT_ATTR_SPEED` returns the configured speed instead of the value retrieved via [`/sys/class/net/<name>/speed`](https://github.com/sonic-net/sonic-sairedis/blob/master/vslib/SwitchStateBaseHostif.cpp#L892-L893). 
  - fixes https://github.com/sonic-net/sonic-buildimage/issues/19735

- `SAI_PORT_ATTR_HOST_TX_READY_STATUS`: always returns `true`. Required to support running `xcvrd` in the VS env.
  - ref: https://github.com/sonic-net/SONiC/pull/1849/files#diff-6f3e95e6c57a3edc2e30e1f13edb9fd9a32a0db44e1035ac1f0b1b9a191762a5R46

### DIFF
--- a/vslib/SwitchStateBase.cpp
+++ b/vslib/SwitchStateBase.cpp
@@ -1313,6 +1313,11 @@ sai_status_t SwitchStateBase::create_ports()
         attr.value.u32 = DEFAULT_VLAN_NUMBER;
 
         CHECK_STATUS(set(SAI_OBJECT_TYPE_PORT, port_id, &attr));
+
+        attr.id = SAI_PORT_ATTR_HOST_TX_READY_STATUS;
+        attr.value.u32 = SAI_PORT_HOST_TX_READY_STATUS_READY;
+
+        CHECK_STATUS(set(SAI_OBJECT_TYPE_PORT, port_id, &attr));
     }
 
     return SAI_STATUS_SUCCESS;
@@ -1775,6 +1780,11 @@ sai_status_t SwitchStateBase::create_port_dependencies(
 
     attr.id = SAI_PORT_ATTR_ADMIN_STATE;
     attr.value.booldata = false;
+
+    CHECK_STATUS(set(SAI_OBJECT_TYPE_PORT, port_id, &attr));
+
+    attr.id = SAI_PORT_ATTR_HOST_TX_READY_STATUS;
+    attr.value.u32 = SAI_PORT_HOST_TX_READY_STATUS_READY;
 
     CHECK_STATUS(set(SAI_OBJECT_TYPE_PORT, port_id, &attr));
 
@@ -2504,6 +2514,7 @@ sai_status_t SwitchStateBase::refresh_read_only(
                  */
 
             case SAI_PORT_ATTR_OPER_STATUS:
+            case SAI_PORT_ATTR_HOST_TX_READY_STATUS:
                 return SAI_STATUS_SUCCESS;
 
             case SAI_PORT_ATTR_FABRIC_ATTACHED:
--- a/vslib/SwitchStateBase.cpp
+++ b/vslib/SwitchStateBase.cpp
@@ -2339,7 +2339,12 @@ sai_status_t SwitchStateBase::refresh_port_oper_speed(
     }
     else
     {
-        if (!vs_get_oper_speed(port_id, attr.value.u32))
+        if (m_switchConfig->m_useConfiguredSpeedAsOperSpeed)
+        {
+            attr.id = SAI_PORT_ATTR_SPEED;
+            CHECK_STATUS(get(SAI_OBJECT_TYPE_PORT, port_id, 1, &attr));
+        }
+        else if (!vs_get_oper_speed(port_id, attr.value.u32))
         {
             return SAI_STATUS_FAILURE;
         }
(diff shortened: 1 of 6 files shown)


vslib/saivs.h
```diff
@@ -56,6 +56,16 @@ extern "C" {
...
 
+/**
+ * @def SAI_KEY_VS_OPER_SPEED_IS_CONFIGURED_SPEED
+ *
+ * Bool flag, (true/false). If set to true, SAI_PORT_ATTR_OPER_SPEED returns
+ * the SAI_PORT_ATTR_SPEED instead of querying the speed of veth
+ *
+ * By default this flag is set to false.
+ */
+#define SAI_KEY_VS_OPER_SPEED_IS_CONFIGURED_SPEED "SAI_VS_OPER_SPEED_IS_CONFIGURED_SPEED"
```
kcudnik, line 67: this name is a little bit confusing, it suggest that this is a question or value to read and not an config value

why we need this actually ?

also does this only applies when tap device is in use?

vslib/SwitchStateBase.cpp
```diff
@@ -2260,7 +2270,10 @@ sai_status_t SwitchStateBase::refresh_port_oper_speed(
     }
     else
     {
-        if (!vs_get_oper_speed(port_id, attr.value.u32))
+        if (m_switchConfig->m_operSpeedIsConfiguredSpeed) {
+            attr.id = SAI_PORT_ATTR_SPEED;
+            CHECK_STATUS(get(SAI_OBJECT_TYPE_PORT, port_id, 1, &attr));
+        } else if (!vs_get_oper_speed(port_id, attr.value.u32))
```
kcudnik, line 2276: refactor to move { to new line and else if also to new line like rest of the code

vslib/saivs.h
ishidawataru: > this name is a little bit confusing,

How about `SAI_KEY_VS_USE_CONFIGURED_SPEED_AS_OPER_SPEED`?

> why we need this actually ?

I'm testing DPB on the VS environment.
To check that DPB is working as expected, it'd be nice to see the `show interface status` command shows the expected interface speed.

```
admin@sonic:~$ show interface status
  Interface    Lanes    Speed    MTU    FEC    Alias    Vlan    Oper    Admin                                             Type    Asym PFC
-----------  -------  -------  -----  -----  -------  ------  ------  -------  -----------------------------------------------  ----------
  Ethernet0  1,2,3,4     400G   9100    N/A   Eth0/0  routed    down     down  QSFP-DD Double Density 8X Pluggable Transceiver         N/A
admin@sonic:~$ sudo config interface breakout Ethernet0 "2x200G"
Do you want to Breakout the port, continue? [y/N]: y

Running Breakout Mode : 1x400G 
Target Breakout Mode : 2x200G

Ports to be deleted : 
 {
    "Ethernet0": "400000"
}
Ports to be added : 
 {
    "Ethernet0": "200000",
    "Ethernet2": "200000"
}
Breakout process got successfully completed.
Please note loaded setting will be lost after system reboot. To preserve setting, run `config save`.
admin@sonic:~$ show interface status
  Interface    Lanes    Speed    MTU    FEC    Alias    Vlan    Oper    Admin                                             Type    Asym PFC
-----------  -------  -------  -----  -----  -------  ------  ------  -------  -----------------------------------------------  ----------
  Ethernet0      1,2     200G   9100    N/A   Eth0/0  routed    down     down  QSFP-DD Double Density 8X Pluggable Transceiver         N/A
  Ethernet2      3,4     200G   9100    N/A   Eth0/1  routed    down     down                                              N/A         N/A
```

> also does this only applies when tap device is in use?

I don't think so.  [`SwitchStateBase::refresh_port_oper_speed`](https://github.com/sonic-net/sonic-sairedis/blob/master/vslib/SwitchStateBase.cpp#L2246) doesn't care `SAI_VS_HOSTIF_USE_TAP_DEVICE` configuration.

kcudnik: Changed name looks good

kcudnik (approved):

ishidawataru:
@kcudnik I have addressed the review comments and rebased onto the master branch. Could you please take a look again?

kcudnik (approved):

ishidawataru:
@kcudnik Thank you for approving the PR. Could you also merge this? I'm planning to open another PR in the sonic-buildimage repo and want to include this fix in it.

kcudnik:
There are 4 pr which were reverted on master in buldimage because broadcom issues, can we wait to next week after that we will be reverted to avoid conlicts

ishidawataru:
@kcudnik I rebased the branch onto the latest master branch. Could you take a look?

This PR is required to implement [this HLD](https://github.com/sonic-net/SONiC/pull/1849#issuecomment-2600919814), which is planned to be included in the 202505 release.

kcudnik (approved):

PR #1473, opened 2024-11-24
Merge vector packet processing (vpp) from sonic-platform-vpp with virtual switch 

kcudnik:
Initial attempt


yue-fred-gao (commented):
This looks good in general. But I think we should wait vpp pipeline ready and verify vpp is able to compile and run.

kcudnik:
> This looks good in general. But I think we should wait vpp pipeline ready and verify vpp is able to compile and run.

i just made build from source on azp in sairedis just for test: https://github.com/sonic-net/sonic-sairedis/pull/1486 if this will succeed, thatn we can just copy that to azpipeline on sonic-platform-vpp if noowne is working on that yet

kcudnik:
seems like vpp build succeeded on azure pipeline: https://dev.azure.com/mssonic/build/_build/results?buildId=724631&view=logs&j=d755766c-c33e-5abe-260c-8d11f335b1a1&t=f2559076-9bdb-5064-4af2-83c32f011787 it took a little bit over 1 hour, first build fialed, nto sure why, but second one after restart succeeded :)

vslib/Makefile.am
```diff
@@ -1,9 +1,35 @@
...
+			   -Wno-conversion \
+			   -Wno-unused-variable \
+			   -Wno-sign-compare \
+			   -Wno-shadow \
+			   -Wno-discarded-qualifiers \
+			   -Wno-redundant-decls \
+			   -Wno-cast-align \
+			   -Wno-pointer-sign \
+			   -Wno-stringop-truncation \
+			   -Wno-cast-qual
```
kcudnik, line 22: @yue-fred-gao can you look at those warnings? i disabled those since *c files in vppxlate/*c developer produce a lot of warnings which are treated as errors by our convention, and for many of them, i don't know what was actual developer intention, and without any unittests provided on that i don't want to change any of this code

@yue-fred-gao will you be able to help to eliminate those warnings ?

also what is a purpose of vppxlate directory? is there any purpose of that on other repo? or this is specifically designed to for vpp in sairedis/buildimage ? is there specific requirement that this code needs to be in C and not cpp ?

yue-fred-gao: I created an issue (https://github.com/sonic-net/sonic-platform-vpp/issues/149) to track this. 
vppxlate is the vpp API adaptation layer. It is specifically for vpp. There is some history that I don't know but I guess vpp APIs are C code so C is used here.

kcudnik:
artifacts vpp used from here: https://github.com/sonic-net/sonic-sairedis/pull/1521

kcudnik:
first build did not succeed because of wrong directory: https://github.com/sonic-net/sonic-platform-vpp/pull/159

.azure-pipelines/build-swss-template.yml
```diff
@@ -138,6 +145,33 @@ jobs:
...
+      sudo dpkg -i download/libvppinfra-dev_*_${{ parameters.arch }}.deb
+      sudo dpkg -i download/libvppinfra_*_${{ parameters.arch }}.deb
+      sudo dpkg -i download/vpp_*_${{ parameters.arch }}.deb
+      sudo dpkg -i download/vpp-crypto-engines_*_${{ parameters.arch }}.deb
+      sudo dpkg -i download/vpp-dbg_*_${{ parameters.arch }}.deb
+      sudo dpkg -i download/vpp-dev_*_${{ parameters.arch }}.deb
+      sudo dpkg -i download/vpp-plugin-core_*_${{ parameters.arch }}.deb
+      sudo dpkg -i download/vpp-plugin-devtools_*_${{ parameters.arch }}.deb
+      sudo dpkg -i download/vpp-plugin-dpdk_*_${{ parameters.arch }}.deb
+      sudo dpkg -i download/python3-vpp-api_*_${{ parameters.arch }}.deb
```
lguohan, line 171: suggestion is to add these dependency here. https://github.com/sonic-net/sonic-buildimage/tree/master/platform/vs/docker-sonic-vs

.azure-pipelines/build-docker-sonic-vs-template.yml
```diff
@@ -85,6 +85,35 @@ jobs:
...
+      sudo dpkg -i $(Build.ArtifactStagingDirectory)/download/vpp-crypto-engines_*_${{ parameters.arch }}.deb
+      sudo dpkg -i $(Build.ArtifactStagingDirectory)/download/vpp-dbg_*_${{ parameters.arch }}.deb
+      sudo dpkg -i $(Build.ArtifactStagingDirectory)/download/vpp-dev_*_${{ parameters.arch }}.deb
+      sudo dpkg -i $(Build.ArtifactStagingDirectory)/download/vpp-plugin-core_*_${{ parameters.arch }}.deb
+      sudo dpkg -i $(Build.ArtifactStagingDirectory)/download/vpp-plugin-devtools_*_${{ parameters.arch }}.deb
+      sudo dpkg -i $(Build.ArtifactStagingDirectory)/download/vpp-plugin-dpdk_*_${{ parameters.arch }}.deb
+      sudo dpkg -i $(Build.ArtifactStagingDirectory)/download/python3-vpp-api_*_${{ parameters.arch }}.deb
+    workingDirectory: $(Build.ArtifactStagingDirectory)
+    displayName: "Install sonic platform-vpp packages"
+    condition: eq('${{ parameters.arch }}', 'amd64')
```
lguohan, line 116: we should not touch the template here. we should modify the https://github.com/sonic-net/sonic-buildimage/tree/master/platform/vs/docker-sonic-vs to include vpp in docker-sonic-vs  doker.

kcudnik: how then this sairedis repo will pull that image ?

yue-fred-gao:
I copy the libsaivs.so from libsaivs_1.0.0_amd64.deb to a sonic-vpp setup to replace the libsai.so in syncd. With a few tweaks, it is able to start syncd with the vs/vpp combined libsai. Ping is working between 2 back 2 back sonic-vpp setup. Here is what I did from sonic.

docker cp libsaivs.so.0 syncd:/usr/lib/x86_64-linux-gnu/libsai.so.0.0.0

run below commands from syncd container:
echo "sed -i 's/SAI_VS_SWITCH_TYPE_BCM56850/SAI_VS_SWITCH_TYPE_VPP/g' /etc/sai.d/sai.profile">> /usr/bin/start.sh
sed -i '/config_syncd_vpp()/{:a;n;/}/b;s|\$HWSKU_DIR/sai\.profile|/etc/sai.d/sai.profile|;ba}' /usr/bin/syncd_init_common.sh

Restart everything with sudo config reload -y.

kcudnik:
does this change also requires to install vpp itself on image ? how did you achieve that ?

yue-fred-gao:
> does this change also requires to install vpp itself on image ? how did you achieve that ?

This is just an experiment. It still uses sonic-vpp image built with platform/vpp submodule. After sonic vpp is booted, replacing libsai in syncd with libsai built from this PR. In the next baby step, we can build vpp syncd container using the libsaivs.so from here. Is libsaivs.so only built under PLATFORM=vs?

yue-fred-gao (approved):

liuh-80:
Hi @kcudnik , the sonic-swss repo PR validation now failed because vpp missing:

Preparing to unpack .../libsaivs-dev_1.0.0_amd64.deb ...
Unpacking libsaivs-dev (1.0.0) ...
dpkg: dependency problems prevent configuration of libsaivs:
 libsaivs depends on libvppinfra; however:
  Package libvppinfra is not installed.
 libsaivs depends on vpp; however:
  Package vpp is not installed.

kcudnik:
> Hi @kcudnik , the sonic-swss repo PR validation now failed because vpp missing:
> 
> Preparing to unpack .../libsaivs-dev_1.0.0_amd64.deb ... Unpacking libsaivs-dev (1.0.0) ... dpkg: dependency problems prevent configuration of libsaivs: libsaivs depends on libvppinfra; however: Package libvppinfra is not installed. libsaivs depends on vpp; however: Package vpp is not installed.

ahh, i installed this here https://github.com/sonic-net/sonic-sairedis/blob/master/.azure-pipelines/build-swss-template.yml#L148

and the same could be done in swss repo, i can make PR to fix this, but we will have a bigger problem here, since now since sairedis is depending on vpp, we would need need to propagate that in all packages and docker up to buildimage to install vpp, even if it's not required, for example there is no need to install vpp in production images

i think this would need to be discussed and addressed, maybe solution would be to build 2 version of sairedis in pipeline with and without vpp and export artifacts of both ? definitely needs a discussion

PR #1477, opened 2024-12-02
Turn on SAI attribute version validation

bradh352:
As per https://github.com/sonic-net/sonic-buildimage/issues/20725 attribute validation must be enabled otherwise failures will be observed.   This prevents the ASIC from coming online on broadcom and likely other vendors.

Signed-off-by: Brad House

### DIFF
--- a/syncd/scripts/syncd_init_common.sh
+++ b/syncd/scripts/syncd_init_common.sh
@@ -29,6 +29,10 @@ fi
 # Use temporary view between init view and apply view
 CMD_ARGS+=" -u"
 
+# Enable SAI attribute version checking, required currently for Broadcom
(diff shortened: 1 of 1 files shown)


kcudnik (approved):

bradh352:
I'm assuming the test failures are unrelated to this PR

bradh352:
I saw they reverted the SAI updates to branch 202411, but then didn't reapply after branching?  seems odd

kcudnik:
> I saw they reverted the SAI updates to branch 202411, but then didn't reapply after branching? seems odd

Those chcnages were reverted on master because of brcm issue, I'm not sure if they will be applied on 202411, @yxieca can you confirm?

bradh352:
well, more interested when they'll bring back to master .... i doubt will go into 202411

PR #1492, opened 2025-01-03
[syncd] Update log level for bulk api

jianyuewu:
Log level should be warning instead of error, because functionality is fine. Latter code will fall back to use single api instead of bulk api, when syncd support bulk but sai doesn't support it.

**Background**
When syncd added support for bulk APIs like bulk set, but sai didn't implement it, this error will be printed:
"processBulkOidSet: bulk set api is not implemented or not supported, object_type = SAI_OBJECT_TYPE_QUEUE", and it will fail the cases. Actually the print should be warning, as in current code, bulk set fall back to single set in latter code, so sai will still receive single set APIs, so functionality is actually OK.

**What I did**
Update log level from error to warning.

**How to verify it**
When bulk set api is added in syncd, but sai did not implement it, only warning logs will be printed, instead of error logs.


Junchao-Mellanox (approved):

stephenxs (approved):

stepanblyschak (approved):

kcudnik (dismissed):

kcudnik:
you would need to add test coverage for this

jianyuewu:
Hello, Kamil, thank you, understand, now adding the UTs.

kcudnik (dismissed):

jianyuewu:
> Your tests are failing please fix

Thanks for remind, I'll check it.

jianyuewu:
I'll try in local firstly, seems some difference in local result.

kcudnik:
```
[ RUN      ] SyncdTest.processEventBulkOperations
TestSyncd.cpp:325: Failure
Actual function call count doesn't match EXPECT_CALL(consumer, empty())...
         Expected: to be called twice
           Actual: never called - unsatisfied and active
unknown file: Failure
C++ exception with description ":- sai_deserialize_object_id: invalid oid 0" thrown in the test body.
[  FAILED  ] SyncdTest.processEventBulkOperations (1003 ms)
[ RUN      ] SyncdTest.processEventBulkOperationsSuccess
TestSyncd.cpp:396: Failure
Actual function call count doesn't match EXPECT_CALL(consumer, empty())...
         Expected: to be called twice
           Actual: never called - unsatisfied and active
unknown file: Failure
C++ exception with description ":- sai_deserialize_object_id: invalid oid 0" thrown in the test body.
[  FAILED  ] SyncdTest.processEventBulkOperationsSuccess (1003 ms)
```

jianyuewu:
> ```
> [ RUN      ] SyncdTest.processEventBulkOperations
> TestSyncd.cpp:325: Failure
> Actual function call count doesn't match EXPECT_CALL(consumer, empty())...
>          Expected: to be called twice
>            Actual: never called - unsatisfied and active
> unknown file: Failure
> C++ exception with description ":- sai_deserialize_object_id: invalid oid 0" thrown in the test body.
> [  FAILED  ] SyncdTest.processEventBulkOperations (1003 ms)
> [ RUN      ] SyncdTest.processEventBulkOperationsSuccess
> TestSyncd.cpp:396: Failure
> Actual function call count doesn't match EXPECT_CALL(consumer, empty())...
>          Expected: to be called twice
>            Actual: never called - unsatisfied and active
> unknown file: Failure
> C++ exception with description ":- sai_deserialize_object_id: invalid oid 0" thrown in the test body.
> [  FAILED  ] SyncdTest.processEventBulkOperationsSuccess (1003 ms)
> ```

Thank you, Kamil, I'll check the errors.

jianyuewu:
UT updated and code rebased to latest commit.

jianyuewu:
Now all tests passed.

kcudnik (dismissed):

syncd/tests/TestSyncdMlnx.cpp
```diff
@@ -78,7 +78,7 @@ void syncdMlnxWorkerThread()
     commandLineOptions->m_enableTempView = false;
     commandLineOptions->m_disableExitSleep = true;
     commandLineOptions->m_enableUnittests = false;
-    commandLineOptions->m_enableSaiBulkSupport = true;
+    commandLineOptions->m_enableSaiBulkSupport = false;
```
kcudnik, line 81: why this change ?

stephenxs: just to leverage the pipeline to check whether coverage can pass in this way. we will have a formal solution if it can.
it's not easy for us to check coverage locally.

kcudnik: you can actually do it in docker, it  will cost you some time to setup this locally, and then you can run gcovr following steps taht are located in *yml files

i don;t have this steps locally, but i usually write unittests for code that i add/change

stephenxs: thanks. indeed. we usually toggle breakpoints and check whether breakpoints are reached.
but sometimes even if the breakpoints are reached, coverage reports uncovered..
we will try steps in yaml.

jianyuewu:
Looks inside this function in official code, there is some bug:

bool VirtualOidTranslator::tryTranslateRidToVid(
        _In_ sai_object_id_t rid,
        _Out_ sai_object_id_t &vid)
{
...
    auto it = m_rid2vid.find(vid); // Maybe m_rid2vid.find(**rid**); ?
...
@kcudnik @stephenxs Local UT seems failed here, could I update it in some other ticket or is it an issue?

jianyuewu:
Got this error:TestSyncd.cpp:476: Failure
Actual function call count doesn't match EXPECT_CALL(*channel, empty())...
         Expected: to be called at least once
           Actual: never called - unsatisfied and active
unknown file: Failure
C++ exception with description ":- postPortRemove: expected rid oid:0x0 to be present in RIDTOVID" thrown in the test body.
[Thread 0x7ffff6cc86c0 (LWP 29131) exited]
[  FAILED  ] SyncdTest.BulkRemoveNotImplemented (34118 ms)
Yes, maybe this error not related with m_rid2vid, as I later updated it, seems no impact.

jianyuewu:
> > Got this error:TestSyncd.cpp:476: Failure Actual function call count doesn't match EXPECT_CALL(*channel, empty())... Expected: to be called at least once Actual: never called - unsatisfied and active unknown file: Failure C++ exception with description ":- postPortRemove: expected rid oid:0x0 to be present in RIDTOVID" thrown in the test body. [Thread 0x7ffff6cc86c0 (LWP 29131) exited] [ FAILED ] SyncdTest.BulkRemoveNotImplemented (34118 ms) Yes, maybe this error not related with m_rid2vid, as I later updated it, seems no impact.
> 
> is taht error locally only ? since on pipeline all tests pass ok

Yes, that is local only, seems caused by that port is already removed, but mock still return valid, will add some condition to tell mock that port is already deleted.

kcudnik:
This is not my code since I did not added mock methods, but mock can return independent values than actual function, c
So maybe entire test case is wrong

kcudnik (approved):

jianyuewu:
> > Now coverage passed:) Missing: 0 lines Coverage: 100% Threshold: 80%
> 
> It's ironic since changing 1 line of code and adding 10 of unit tests will pass code coverage since unit tests are also included on code coverage 🤣 and the test don't need to do actual testing 😂

Ah, yes, that is really tricky😂

dprital:
@jianyuewu , Please provide direct PR to 202411 as this one got conflicts.

jianyuewu:
> @jianyuewu , Please provide direct PR to 202411 as this one got conflicts.

@dprital Thanks for remind, backport PR to 202411 created:
https://github.com/sonic-net/sonic-sairedis/pull/1532

PR #1496, opened 2025-01-09
[syncd] Support bulk set in INIT_VIEW mode

stepanblyschak:
Support bulk set in INIT_VIEW mode.

### DIFF
--- a/syncd/Syncd.cpp
+++ b/syncd/Syncd.cpp
@@ -921,7 +921,6 @@ sai_status_t Syncd::processBulkQuadEventInInitViewMode(
     {
         case SAI_COMMON_API_BULK_CREATE:
         case SAI_COMMON_API_BULK_REMOVE:
-        case SAI_COMMON_API_BULK_SET:
 
             if (info->isnonobjectid)
             {
@@ -960,6 +959,14 @@ sai_status_t Syncd::processBulkQuadEventInInitViewMode(
 
                     return SAI_STATUS_SUCCESS;
             }
+
+        case SAI_COMMON_API_BULK_SET:
+
+            sendApiResponse(api, SAI_STATUS_SUCCESS, (uint32_t)statuses.size(), statuses.data());
+
+            syncUpdateRedisBulkQuadEvent(api, statuses, objectType, objectIds, strAttributes);
+
+            return SAI_STATUS_SUCCESS;
 
         case SAI_COMMON_API_BULK_GET:
             SWSS_LOG_THROW("GET bulk api is not implemented in init view mode, FIXME");
--- a/syncd/Syncd.cpp
+++ b/syncd/Syncd.cpp
@@ -962,6 +962,19 @@ sai_status_t Syncd::processBulkQuadEventInInitViewMode(
 
         case SAI_COMMON_API_BULK_SET:
 
+            switch (objectType)
+            {
+                case SAI_OBJECT_TYPE_SWITCH:
+                case SAI_OBJECT_TYPE_SCHEDULER_GROUP:
+
+                    SWSS_LOG_THROW("%s is not supported in init view mode",
+                            sai_serialize_object_type(objectType).c_str());
+
+                default:
+
+                    break;
+            }
+
             sendApiResponse(api, SAI_STATUS_SUCCESS, (uint32_t)statuses.size(), statuses.data());
 
             syncUpdateRedisBulkQuadEvent(api, statuses, objectType, objectIds, strAttributes);
--- a/syncd/tests/TestSyncdBrcm.cpp
+++ b/syncd/tests/TestSyncdBrcm.cpp
@@ -506,7 +506,18 @@ TEST_F(SyncdBrcmTest, nexthopBulkTest)
     }
 }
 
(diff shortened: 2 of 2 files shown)


syncd/Syncd.cpp
```diff
@@ -961,6 +960,14 @@ sai_status_t Syncd::processBulkQuadEventInInitViewMode(
                     return SAI_STATUS_SUCCESS;
             }
 
+        case SAI_COMMON_API_BULK_SET:
```
kcudnik, line 963: description says hat you support PORT and IPG, but you also supported here SWITCH and SCHEDULER, code is not duing what's expected. please remove at least SWITCH support.

PS. why do you need SET support on init view on those objects ? when initview is compiled, all objects should already be known and its attributes so no SET actions should be added to those excluded objects

kcudnik (changes requested):
address comments

syncd/Syncd.cpp
kcudnik: also if you are changing this, please add recodring TESTS in initview if this scenario will pass

stepanblyschak: @kcudnik Changed the title. Got this error when configuring PGs in warm start in orchagent on VS:
```
ERR #syncd: :- run: Runtime error: :- processBulkQuadEventInInitViewMode: SAI_OBJECT_TYPE_INGRESS_PRIORITY_GROUP is not supported in init view mode
```

I don't see any restrictions on non-bulk set in init view, neither for PGs, ports nor switch.
Unit tests are present and passes coverage requirement.

kcudnik: i dont remember why i put those restrictions on the first place, dont remember exactly now, but they were there for a reason, SWITCH should be remove from this SET, since if you are doing CREATE_SWITCH then all attributes should be posted already to that create function, you can change other later on. switch is special case, since for example you can set some ACL etc, and those objects may not exists before you create switch

syncd/tests/TestSyncdBrcm.cpp
```diff
@@ -543,6 +554,33 @@ TEST_F(SyncdBrcmTest, portBufferBulkSet)
...
+
+        // create switch
+
+        attrs[0].id = SAI_SWITCH_ATTR_INIT_SWITCH;
+        attrs[0].value.booldata = true;
+
+        status = m_sairedis->create(SAI_OBJECT_TYPE_SWITCH, &switchId, SAI_NULL_OBJECT_ID, 1, attrs);
+        ASSERT_EQ(status, SAI_STATUS_SUCCESS);
+    }
+
```
kcudnik, line 583: you are changing previous implementaiton of the test case, please create new one, also you are not testing SET operation for all 4 objects you allowed SET, for swithc, port, ipg, schedueler, there are not tests for that

syncd/Syncd.cpp
stepanblyschak: @kcudnik ok, added exception for switch and scheduler group back

syncd/tests/TestSyncdBrcm.cpp
stepanblyschak: @kcudnik It's not changing previous test case, there are two distinct test cases generated by gtest.
This is exactly how I wanted to have it. You see, I added bulk support in previous PR, tested on Nvidia platform that does not use init view in warm boot and didn't realize there's a restriction in bulk set in init view comparing to regular set. I didn't want to duplicate the test for init view since all operations I do after initial view are tested in init view as well. This way, I make sure that whatever else I add to the test is also covered for init view instead of remembering to explicitly add it to a separate test. It is exactly how orchagent will work, it will replay exactly the same configuration again but in init view in warm start.

kcudnik: ok i digged into INSTANTIATE_TEST_SUITE_P and now i see what's happening, but it seems weird approach to me, since i didn't generate tests in this way, its basically adding if/case or switch(x) based on test params given to instantiate, i myself usually write tests that test specific functionality that was added/touched in a separate test scenario, giving that if/case statement just confuses me and make test case more hard to read

imo test flow should be as simple as possible and test 1 thing and 1 thing only, and should contain as low as possible on/off code like in this case

i assume you wanted to save code duplication for test case, but i think this king of approach would make more sense to change just param value like for example 5 diffveret param values for specifig given attributes, would generate 5 different taste cases, but without on/off entire block of code inside unittest

kcudnik:
since you touched that switch case segment: https://dev.azure.com/mssonic/build/_build/results?buildId=743098&view=codecoverage-tab code coverage is not passing, please add test cases that will cover those too

syncd/tests/TestSyncdBrcm.cpp
stepanblyschak: @kcudnik Ok, I added separate tests

stepanblyschak:
> since you touched that switch case segment: https://dev.azure.com/mssonic/build/_build/results?buildId=743098&view=codecoverage-tab code coverage is not passing, please add test cases that will cover those too

@kcudnik done

kcudnik (approved):

stepanblyschak:
@kcudnik Can be merged?

PR #1497, opened 2025-01-10
fix SAI_ATTR_CONDITION_TYPE_AND for MANDATORY_ON_CREATE condition

baorliu:
when check MANDATORY_ON_CREATE condition, the SAI_ATTR_CONDITION_TYPE_AND support is missing.

### DIFF
--- a/meta/Meta.cpp
+++ b/meta/Meta.cpp
@@ -3817,6 +3817,7 @@ sai_status_t Meta::meta_generic_validation_create(
         // this is conditional attribute, check if it's required
 
         bool any = false;
+        bool all = true;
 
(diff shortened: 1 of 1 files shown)


meta/Meta.cpp
```diff
@@ -3817,6 +3817,7 @@ sai_status_t Meta::meta_generic_validation_create(
         // this is conditional attribute, check if it's required
 
         bool any = false;
+        bool all = true;
```
kcudnik, line 3820: maybe it's time to use this https://github.com/opencomputeproject/SAI/blob/master/meta/saimetadatautils.h#L194

PR #1499, opened 2025-01-14
Fix for credo gbsyncd warm-reboot

byu343:
To support credo gbsyncd warm-reboot, we found several issues to be fixed in sonic-sairedis as follows:
1) More attributes are needed for the create_switch of credo gbsyncd warm-boot. The fix here is to pass all the attributes when SAI_SWITCH_ATTR_TYPE is SAI_SWITCH_TYPE_PHY.
2) firstRun being true disallows the start of warm-reboot. It was set to true when no switch attributes is found under the key of HIDDEN* in ASIC_DB or GB_ASIC_DB. However, all such HIDDEN attributes are unrelated to the credo gbsyncd, so hasNoHiddenKeysDefined is always true for gbsyncd. The fix here is to determine the first run also based on no switch objects having been created in ASIC_DB or GB_ASIC_DB.
3) SAI_SWITCH_ATTR_DEFAULT_TRAP_GROUP is not supported by credo gbsyncd and unrelated to it. Skip the steps related to it when it is not available

### DIFF
--- a/syncd/ComparisonLogic.cpp
+++ b/syncd/ComparisonLogic.cpp
@@ -61,9 +61,26 @@ ComparisonLogic::ComparisonLogic(
 
     // TODO needs to be removed and done in generic
 
-    m_current->m_defaultTrapGroupRid     = m_switch->getSwitchDefaultAttrOid(SAI_SWITCH_ATTR_DEFAULT_TRAP_GROUP);
-    m_temp->m_defaultTrapGroupRid        = m_switch->getSwitchDefaultAttrOid(SAI_SWITCH_ATTR_DEFAULT_TRAP_GROUP);
+    try
+    {
+        m_current->m_defaultTrapGroupRid     = m_switch->getSwitchDefaultAttrOid(SAI_SWITCH_ATTR_DEFAULT_TRAP_GROUP);
+    }
+    catch(const std::exception &e)
+    {
+        SWSS_LOG_WARN("Failed to to obtain SAI_SWITCH_ATTR_DEFAULT_TRAP_GROUP in current view");
+        m_current->m_defaultTrapGroupRid = SAI_NULL_OBJECT_ID;
+    }
 
+    try
+    {
+        m_temp->m_defaultTrapGroupRid        = m_switch->getSwitchDefaultAttrOid(SAI_SWITCH_ATTR_DEFAULT_TRAP_GROUP);
+    }
+    catch(const std::exception &e)
+    {
+        SWSS_LOG_WARN("Failed to to obtain SAI_SWITCH_ATTR_DEFAULT_TRAP_GROUP in temp view");
+        m_temp->m_defaultTrapGroupRid = SAI_NULL_OBJECT_ID;
+
+    }
     auto seed = (unsigned int)std::time(0);
 
     SWSS_LOG_NOTICE("srand seed for switch %s: %u", sai_serialize_object_id(m_switch->getVid()).c_str(), seed);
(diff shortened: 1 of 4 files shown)


kcudnik:
This needs to be discussed, nice passing more attributes in warm boot make no sense, by design warm boot should have the same attributes after warm boot. Exception is attributes pointers which changes because of memory allocation for syncd process and  hardware info to point which switch should be instantiated. If you are changing this then there is some issue inside libsai in vendor code for warm boot

syncd/ComparisonLogic.cpp
```diff
@@ -61,9 +61,26 @@ ComparisonLogic::ComparisonLogic(
 
     // TODO needs to be removed and done in generic
 
-    m_current->m_defaultTrapGroupRid     = m_switch->getSwitchDefaultAttrOid(SAI_SWITCH_ATTR_DEFAULT_TRAP_GROUP);
-    m_temp->m_defaultTrapGroupRid        = m_switch->getSwitchDefaultAttrOid(SAI_SWITCH_ATTR_DEFAULT_TRAP_GROUP);
+    try
```
kcudnik, line 64: Why tis is modified?

syncd/RedisClient.cpp
```diff
@@ -691,6 +691,15 @@ bool RedisClient::hasNoHiddenKeysDefined() const
     return keys.size() == 0;
 }
 
+bool RedisClient::hasNoSwitchDefined() const
```
kcudnik, line 694: Why there would be no switch? What is purpose of this? You have flag isveryfirstrun

syncd/Syncd.cpp
```diff
@@ -4645,7 +4645,16 @@ void Syncd::performWarmRestartSingleSwitch(
     sai_attribute_t *attrList = list.get_attr_list();
 
     uint32_t attrCount = list.get_attr_count();
-
+    bool isPhy = false;
+    for (uint32_t idx = 0; idx < attrCount; idx++)
```
kcudnik, line 4649: There is a function for that in sai utils

syncd/Syncd.cpp
```diff
@@ -4662,7 +4671,7 @@ void Syncd::performWarmRestartSingleSwitch(
          * new process could be loaded at different address space.
          */
 
-        if (id == SAI_SWITCH_ATTR_SWITCH_HARDWARE_INFO || meta->attrvaluetype == SAI_ATTR_VALUE_TYPE_POINTER)
+        if (isPhy || id == SAI_SWITCH_ATTR_SWITCH_HARDWARE_INFO || meta->attrvaluetype == SAI_ATTR_VALUE_TYPE_POINTER)
```
kcudnik, line 4674: You can pass all attributes, look at my comment in main thread

syncd/Syncd.cpp
```diff
@@ -5103,7 +5112,7 @@ bool Syncd::isVeryFirstRun()
      * this is first run, let's query HIDDEN ?
      */
 
-    bool firstRun = m_client->hasNoHiddenKeysDefined();
+    bool firstRun = m_client->hasNoHiddenKeysDefined() && m_client->hasNoSwitchDefined();
```
kcudnik, line 5115: You already have flag isveryfirst run, why this is modified?

kcudnik (changes requested):

syncd/ComparisonLogic.cpp
byu343: SAI_SWITCH_ATTR_DEFAULT_TRAP_GROUP is not supported by Credo gbsyncd.
This is to avoid raise exception for SAI_SWITCH_ATTR_DEFAULT_TRAP_GROUP not supported

syncd/Syncd.cpp
byu343: firstRun being true will disallow the start of warm-reboot.
It was set to true when no switch attributes is found under the key of HIDDEN* in ASIC_DB or GB_ASIC_DB. However, all such attributes under HIDDEN are unrelated to the credo gbsyncd, so hasNoHiddenKeysDefined is always true for gbsyncd.
The change here is to determine the first run also based on no switch objects has been created in ASIC_DB or GB_ASIC_DB.

PR #1509, opened 2025-01-24
[sairedis/syncd] Implement bulk get support

stepanblyschak:
DEPENDS ON https://github.com/sonic-net/sonic-swss-common/pull/966.

Implement bulk get support for SAIRedis, Syncd, SaiPlayer and VSlib.


meta/Meta.cpp
```diff
@@ -1280,9 +1280,57 @@ sai_status_t Meta::bulkGet(
...
+
+        sai_object_meta_key_t meta_key = { .objecttype = object_type, .objectkey = { .key = { .object_id  = object_id[idx] } } };
+
+        vmk.push_back(meta_key);
+
+        status = meta_generic_validation_get(meta_key, attr_count[idx], attr_list[idx]);
+
+        // FIXME: This macro returns on failure.
+        // When mode is SAI_BULK_OP_ERROR_MODE_IGNORE_ERROR we should continue instead of return.
+        // This issue exists for all bulk operations.
```
kcudnik, line 1318: should we fix it here for this api ? since GET operation is specific, so maybe some attributes maybe not implemented and they will definitely return failure

meta/SaiSerialize.cpp
```diff
@@ -90,24 +90,27 @@ static sai_status_t transfer_list(
 {
     SWSS_LOG_ENTER();
 
-    if (countOnly || dst_element.count == 0)
```
kcudnik, line 93: why this logic is changed ?

saiplayer/SaiPlayer.cpp
```diff
@@ -2354,6 +2354,39 @@ sai_status_t SaiPlayer::handle_bulk_object(
...
+                attr_lists[idx] = attributes[idx]->get_attr_list();
+            }
+
+            status = m_sai->bulkGet(object_type,
+                                            object_count,
+                                            ids.data(),
+                                            attr_counts.data(),
+                                            attr_lists.data(),
+                                            mode,
+                                            statuses.data()); // STATUS_NOT_EXECUTED
```
kcudnik, line 2383: why this comment means not executed ?

saiplayer/SaiPlayer.cpp
```diff
@@ -2522,17 +2629,17 @@ int SaiPlayer::replay()
 
     SWSS_LOG_NOTICE("using file: %s", filename.c_str());
 
-    std::ifstream infile(filename);
```
kcudnik, line 2525: why this logic is changed ? not need for this variable to be member of class

saiplayer/SaiPlayer.h
```diff
@@ -259,6 +260,8 @@ namespace saiplayer
 
             std::shared_ptr<CommandLineOptions> m_commandLineOptions;
 
+            std::ifstream m_infile;
```
kcudnik, line 263: why this is moved to member ?

syncd/Syncd.cpp
```diff
@@ -908,12 +912,12 @@ sai_status_t Syncd::processBulkQuadEventInInitViewMode(
...
     SWSS_LOG_ENTER();
 
+    const auto objectCount = static_cast<uint32_t>(objectIds.size());
+
     std::vector<sai_status_t> statuses(objectIds.size());
 
-    for (auto &a: statuses)
-    {
-        a = SAI_STATUS_SUCCESS;
-    }
```
kcudnik, line 916: why this is removeD ?

unittest/lib/TestRedisRemoteSaiInterface.cpp
```diff
@@ -4,27 +4,3 @@
 #include <gtest/gtest.h>
 
 using namespace sairedis;
-
-TEST(RedisRemoteSaiInterface, bulkGet)
```
kcudnik, line 8: please don't remove tests, implement correct TEST to validate your code

unittest/vslib/TestSwitchStateBase.cpp
```diff
@@ -69,7 +69,7 @@ TEST_F(SwitchStateBaseTest, switchHashGet)
     attr.id = SAI_HASH_ATTR_NATIVE_HASH_FIELD_LIST;
     attr.value.s32list.list = nullptr;
     attr.value.s32list.count = 0;
-    ASSERT_EQ(m_ss->get(SAI_OBJECT_TYPE_HASH, sai_serialize_object_id(ecmpHashOid), 1, &attr), SAI_STATUS_SUCCESS);
+    ASSERT_EQ(m_ss->get(SAI_OBJECT_TYPE_HASH, sai_serialize_object_id(ecmpHashOid), 1, &attr), SAI_STATUS_BUFFER_OVERFLOW);
```
kcudnik, line 72: why this suddenly chhanges status ?

kcudnik (changes requested):
address comments

saiplayer/SaiPlayer.cpp
stepanblyschak: Removed comment

meta/SaiSerialize.cpp
stepanblyschak: Will extract this change in a different PR

unittest/vslib/TestSwitchStateBase.cpp
stepanblyschak: Will extract this change in a different PR

meta/Meta.cpp
stepanblyschak: Could you please clarify? Attribute may be not implemented with SET as well.

syncd/Syncd.cpp
stepanblyschak: status is not SUCCESS for get operation

saiplayer/SaiPlayer.cpp
stepanblyschak: to use m_infile in processBulk method

unittest/lib/TestRedisRemoteSaiInterface.cpp
stepanblyschak: I was not sure what this test actually checks. It seems the test was added to satisfy code coverage rather then check functionality. As for functioanlity check and code coverage I added tests to TestSyncdBrcm and added player tests.

syncd/Syncd.cpp
kcudnik: this was added here becasue of init view mode since operatiom must be success for craete/remove/set

stepanblyschak: Right, so the new code looks like this:

```
const sai_status_t initialObjectStatus = api != SAI_COMMON_API_BULK_GET ? SAI_STATUS_SUCCESS : SAI_STATUS_NOT_EXECUTED;
statuses.assign(statuses.size(), initialObjectStatus);
```

where it is STATUS_NOT_EXECUTED for get and STATUS_SUCCESS for create/set/remove

saiplayer/SaiPlayer.cpp
kcudnik: i see that, but why the reason ? since this is local variable no need to upgrade it to member

syncd/Syncd.cpp
```diff
@@ -962,7 +966,47 @@ sai_status_t Syncd::processBulkQuadEventInInitViewMode(
...
         case SAI_COMMON_API_BULK_GET:
-            SWSS_LOG_THROW("GET bulk api is not implemented in init view mode, FIXME");
+            if (info->isnonobjectid)
+            {
+                /*
+                * Those objects are user created, so if user created ROUTE he
+                * passed some attributes, there is no sense to support GET
+                * since user explicitly know what attributes were set, similar
+                * for other non object id types.
+                */
```
kcudnik, line 976: but thoe objects may have readonly attributes that can dynamically change and user may want to query those, for example SAI_ROUTE_ENTRY_ATTR_IP_ADDR_FAMILY

syncd/VendorSai.cpp
```diff
@@ -657,11 +657,35 @@ sai_status_t VendorSai::bulkGet(
...
+    MUTEX();
     SWSS_LOG_ENTER();
+    VENDOR_CHECK_API_INITIALIZED();
+
+    sai_bulk_object_get_attribute_fn ptr;
+
+    switch (object_type)
+    {
+        case SAI_OBJECT_TYPE_PORT:
+            ptr = m_apis.port_api->get_ports_attribute;
```
kcudnik, line 669: since you are adding bulk support, add all apis here that support bulk ?

saiplayer/SaiPlayer.cpp
stepanblyschak: SaiPlayer operates on input stream, the input stream is feeding multiple functions with data throuhgout this operation. From that point of view the lifetime of the input stream needs to match the lifetime of SaiPlayer, thus fits to be a member variable. I did this change for ergonomics primaraly - not having to pass the input stream explicitelly everywhere it is needed in downstream code. Besides, passing an input stream explicitelly implies a member function may operate on different input streams throughout SaiPlayer's lifetime which is not the case.
I can make a function parameter though if you prefer so.

syncd/Syncd.cpp
stepanblyschak: This comment is inherited from processQuadInInitViewModeGet(). I do not want to make the behaviour incosistent between get vs. get bulk thus I support only objects

syncd/VendorSai.cpp
stepanblyschak: Same as for SET API. I add here only what I tested and have unit tests for (to satisfy coverage requirement). If anyone wants to add more APIs they can be added seperatelly in which case a seperate unit testing code needs to be written.

unittest/lib/TestRedisRemoteSaiInterface.cpp
stepanblyschak: @kcudnik Please share your thoughts on what should be in this test case instead and if needed at all since the code is covered by a set of other tests.

syncd/Syncd.cpp
stepanblyschak: @kcudnik can be resolved?

stepanblyschak:
Tests are failing due to build using older swss-common which is missing PR https://github.com/sonic-net/sonic-swss-common/pull/966.

Test is using swss-common built here - https://dev.azure.com/mssonic/build/_build/results?buildId=757465 which is from a previous commit.

Waiting for a new swss-common build

volodymyrsamotiy (approved):

kcudnik:
please r esolve conflicts

unittest/lib/TestRedisRemoteSaiInterface.cpp
kcudnik: if this is not your tests, then why you are removing it ?

kcudnik: if  you added code that changes ouyput of this tests, that now it can succeed, then please update test

stepanblyschak: @kcudnik Yes, the output of this call changes from returning SAI_STATUS_NOT_IMPLEMENTED to throwing an exception ```waitForBulkGetResponse: wrong number of statuses, got 0, expected 1``` after a 60 sec timeout because no one is responding, so I don't get the reason for such a test. We have bulkGet covered by player test and syncd test

kcudnik: please update this test to run correctly

kcudnik (dismissed):

kcudnik:
since that PR is implementing Bulk Support, in test file: ~/sr/unittest/lib/TestClientServerSai.cpp there is this test:

213 SAIREDIS_DECLARE_EVERY_ENTRY(TEST_ENTRY)
214
215 #define TEST_BULK_ENTRY(OT,ot)                                                                                               \
216 TEST(ClientServerSai, bulk_ ## OT)                                                                                           \
217 {                                                                                                                            \
218     auto css = std::make_shared<ClientServerSai>();                                                                          \
219     sai_ ## ot ## _t e[2] = {};                                                                                              \
220     EXPECT_EQ(SAI_STATUS_SUCCESS, css->apiInitialize(0, &test_services));                                                       \
221     EXPECT_NE(SAI_STATUS_SUCCESS, css->bulkCreate(0, e, nullptr, nullptr, SAI_BULK_OP_ERROR_MODE_IGNORE_ERROR, nullptr));    \
222     EXPECT_NE(SAI_STATUS_SUCCESS, css->bulkSet(2, e, nullptr, SAI_BULK_OP_ERROR_MODE_IGNORE_ERROR, nullptr));                \
223     EXPECT_NE(SAI_STATUS_SUCCESS, css->bulkRemove(2, e, SAI_BULK_OP_ERROR_MODE_IGNORE_ERROR, nullptr));                      \
224                                                                                                                              \
225     auto ss = std::make_shared<ClientServerSai>();                                                                           \
226     EXPECT_EQ(SAI_STATUS_SUCCESS, ss->apiInitialize(0, &test_client_services));                                                 \
227     EXPECT_NE(SAI_STATUS_SUCCESS, ss->bulkCreate(0, e, nullptr, nullptr, SAI_BULK_OP_ERROR_MODE_IGNORE_ERROR, nullptr));     \
228     EXPECT_NE(SAI_STATUS_SUCCESS, ss->bulkSet(0, e, nullptr, SAI_BULK_OP_ERROR_MODE_IGNORE_ERROR, nullptr));                 \
229     EXPECT_NE(SAI_STATUS_SUCCESS, ss->bulkRemove(0, e, SAI_BULK_OP_ERROR_MODE_IGNORE_ERROR, nullptr));                       \
230 }

which expects that all bulk apis will fail, and seems like the changes to bulk PR causing this test to fail with:

[ RUN      ] ClientServerSai.DIRECTION_LOOKUP_ENTRY
Assertion failed: _owned.empty () (src/own.cpp:197)
/bin/bash: line 6: 27699 Aborted                 (core dumped) ${dir}$tst
FAIL: tests

[@Stepan Blyshchak](mailto:stepanb@nvidia.com) can you check that locally ?

kcudnik:
i narrowed this down to : https://gitea.atria-soft.org/generic-library/libzmq/src/commit/092ad50b0c32174cdffdd45b5516b50237a8315d/src/own.cpp
this line:

Assertion failed: _owned.empty () (src/own.cpp:197)

is from zmq library, but beats me why this assert is crashing 😕

kcudnik:
some of the tests are not passing:
```
Replay BCM56850/port_bulk_get.rec
  what():  :- waitForBulkGetResponse: wrong number of statuses, got 0, expected 2
player BCM56850/port_bulk_get.rec: exitcode: 134:
FAIL: BCM56850.pl
```

this seems like implementation bulk error

stepanblyschak:
@kcudnik I am familiar with this failure, it requires sonic-swss-common PR https://github.com/sonic-net/sonic-swss-common/pull/966. The problem is - there is no new sonic-swss-common build that includes this PR as I pointed out in previous comment:

> Tests are failing due to build using older swss-common which is missing PR [sonic-net/sonic-swss-common#966](https://github.com/sonic-net/sonic-swss-common/pull/966).
> 
> Test is using swss-common built here - https://dev.azure.com/mssonic/build/_build/results?buildId=757465 which is from a previous commit.
> 
> Waiting for a new swss-common build

There's no new sonic-swss-common build for few weeks because of different failures in sonic-swss-common (one of which was the mentioned assertion in zmq lib) and I do not have a way to restart sonic-swss-common build:


image

kcudnik:
i tried to compile this on my own branch here, and im getting multiple errors of bulkGet fails:
https://dev.azure.com/mssonic/build/_build/results?buildId=779564&view=logs&jobId=83516c17-6666-5250-abde-63983ce72a49&j=83516c17-6666-5250-abde-63983ce72a49&t=e270ae33-f7d3-59b1-6dcf-c4951de3d6cc

please refine your PR, put all your changes on top of master instead of merging everything, and please refine your code so all unittests will pass

stepanblyschak:
@kcudnik I see, vs sai needs a set of tests to be updated. For the purpose to have bulk get support in sairedis I do not require vslib to support bulk, so will try to do that in different PR. That part I revert from this PR

stepanblyschak:
sairedis/syncd unit tests for bulk get support are passing now, looking into adding vslib bulk get back and pass corresponding unit tests

stepanblyschak:
@kcudnik I added/fixed all required unit tests, also rebased changes on top of latest master and all checks seems to pass, could you please check and merge?

liat-grozovik:
@kcudnik could you please help to approve and merge?

kcudnik (approved):

PR #1514, opened 2025-01-28
Fix pipeline errors related to rsyslogd and libswsscommon installation

saiarcot895:
On arm64 (and maybe sometimes amd64), rsyslogd appears to need a second or two to actually fully exit. The current code expects it to exit practically instantly. Add a sleep of 2 seconds to give it some time. Also enable some logging so that the commands being run can be seen.

Also, fix an error related to libswsscommon not getting installed due to new dependencies being added. Solve this by using apt install to install the package, which brings in any necessary dependencies.

### DIFF
--- a/.azure-pipelines/build-template.yml
+++ b/.azure-pipelines/build-template.yml
@@ -166,7 +166,10 @@ jobs:
   - script: |
       set -ex
       sudo cp azsyslog.conf /etc/rsyslog.conf
+      cat /run/rsyslogd.pid
       sudo pkill -F /run/rsyslogd.pid
+      ps -ef
+      sleep 5
       ps -ef
       sudo rsyslogd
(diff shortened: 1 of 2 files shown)


r12f:
build passed! this looks working!

the test is failing due to another infra issue causing by dependency, not sure if we should use another PR to fix that issue and get this one merged first.

image

r12f (approved):
either way, I am signing off this one first.

saiarcot895:
@kcudnik Are you able to force-merge this PR? This fixes the rsyslogd issues in the PR checkers and fixes a libswsscommon installation failure.

stephenxs:
@saiarcot895 @r12f  could you help to cherry-pick this PR to 202411? I see a similar issue in that branch

PR #1505, opened 2025-01-22
[syncd] Move logSet logGet under mutex to prevent race condition

kcudnik:
Fixes: https://github.com/sonic-net/sonic-buildimage/issues/21180

Mutex is added to protect m_logLevelMap when doing logSet from multiple thread

PR #1518, opened 2025-02-05
[Rebase for #1505][syncd] Move logSet logGet under mutex to prevent race condition

yejianquan:
Original PR: https://github.com/sonic-net/sonic-sairedis/pull/1505
Author: @kcudnik 

Create this PR to rebase #1505 to the newest master branch

Fixes: https://github.com/sonic-net/sonic-buildimage/issues/21180

Mutex is added to protect m_logLevelMap when doing logSet from multiple thread


yejianquan:
This PR is for bumping up the progress, we will build new chassis image before this Friday, and want to include this fix.
@kcudnik  @abdosi for viz

yejianquan:
#1505 rebased already

PR #1457, opened 2024-11-08
Optimize counter polling interval by making it more accurate

stephenxs:
**What I did**

Optimize the counter-polling performance in terms of polling interval accuracy

1. Enable bulk counter-polling to run at a smaller chunk size
   There is one counter-polling thread for each counter group. All such threads can compete for the critical sections at the vendor SAI level, which means a counter-polling thread can wait for a critical section if another thread has been in it, which introduces latency for the waiting counter group.
   An example is the competition between the PFC watchdog and the port counter groups.
   The port counter group contains many counters and is polled in a bulk mode which takes a relatively longer time. The PFC watchdog counter group contains only a few counters but is polled at a short interval. Sometimes, PFC watchdog counters need to wait before polling, which makes the polling interval inaccurate and prevents the PFC storm from being detected in time.
   To resolve this issue, we can reduce the chunk size of the port counter group. The port counter group polls the counters of all ports in a single bulk operation by default. By using a smaller chunk size, it polls the counters in several bulk operations with each polling counter of a subset (whose size <= `chunk size`) of all ports.
   By doing so, the port counter group stays in the critical section for a shorter time and the PFC watchdog is more likely to be scheduled to poll counters and detect the PFC storm in time.

2. Collect the time stamp immediately after vendor SAI API returns.
   Currently, many counter groups require a Lua plugin to execute based on polling interval, to calculate rates, detect certain events, etc.
   Eg. For PFC watchdog counter group to PFC storm. In this case, the polling interval is calculated based on the difference of time stamps between the `current` and `last` poll to avoid deviation due to scheduling latency. However, the timestamp is collected in the Lua plugin which is several steps after the SAI API returns and is executed in a different context (redis-server). Both introduce even larger deviations. To overcome this, we collect the timestamp immediately after the SAI API returns.

Depends on 
1. https://github.com/sonic-net/sonic-sairedis/pull/1519
2. https://github.com/sonic-net/sonic-swss/pull/3391

**Why I did it**

**How I verified it**

Run regression test and observe counter-polling performance.

A comparison test shows very good results if we put any/or all of the above optimizations.

**Details if related**

For 2, each counter group contains more than one counter context based on the type of objects. counter context is mapped from (group, object type). But the counters fetched from different counter groups will be pushed into the same entry for the same objects.
eg. PFC_WD group contains counters of ports and queues. PORT group contains counters of ports. QUEUE_STAT group contains counters of queues.
[...]

PR #1519, opened 2025-02-05
Define bulk chunk size and bulk chunk size per counter ID

stephenxs:
Define bulk chunk size and bulk chunk size per counter ID.
This is to resolve the VS test failure in #1457, which is caused by loop dependency.
In PR #1457, new fields `bulk_chunk_size` and `bulk_chunk_size_per_prefix` have been introduced to `sai_redis_flex_counter_group_parameter_t` whose instances are initialized by orchagent.
However, the orchagent is still compiled with the old sairedis header, which prevents both new fields from being uninitialized which in turn fails vs test.

We have to split this PR into two:
1. #1519 which updates the header sairedis.h only. the motivation is to compile swss(orchagent) with both new fields initiated.
2. #1457 contains all the rest of code

The order to merge:
1. #1519
2. https://github.com/sonic-net/sonic-swss/pull/3391
3. #1457

### DIFF
--- a/lib/sairedis.h
+++ b/lib/sairedis.h
@@ -140,6 +140,20 @@ typedef struct _sai_redis_flex_counter_group_parameter_t
      */
     sai_s8_list_t plugins;
 
+    /**
+     * @brief The bulk chunk size of the counter group
+     *
+     * It should be a number representing the bulk chunk size.
+     */
+    sai_s8_list_t bulk_chunk_size;
+
+    /**
+     * @brief The bulk counter prefix map the counter group
+     *
+     * It should be a string representing bulk chunk size of each sub counter group.
+     */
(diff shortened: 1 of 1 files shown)


stephenxs:
Hi @kcudnik 
As mentioned in the description, we are suffering loop dependency in VS test of counter optimization PRs. I have to split some definitions into this PR to break the loop and pass the VS test.
Would you please approve and merge the PR so that we can move forward with swss PR https://github.com/sonic-net/sonic-swss/pull/3391 and then the rest of sairedis PR https://github.com/sonic-net/sonic-sairedis/pull/1457?
Thank you.

kcudnik (approved):

r12f:
already included in 202411.

PR #1521, opened 2025-02-05
[azp] Download and install sonic-platform-vpp artifacts

kcudnik:
Download and install vpp deb packages so sairedis could be build with VPP enabled


kcudnik:
Download and install vpp deb packages so sairedis could be build with VPP enabled

PR #1522, opened 2025-02-05
Test new test case

yejianquan:


kcudnik:
this test is already added in original PR

PR #1507, opened 2025-01-23
Revert "Revert back to SAI version 1 15 (#1481)"

prabhataravind:
This reverts commit 4f7eed0aac95288077fa9959dc61eacd2f3938d6.

PR #1531, opened 2025-02-11
Update SAI submodule pointer to v1.15.4

BYGX-wcr:
Update SAI submodule pointer to v1.15.4.


meta/Meta.cpp
```diff
@@ -6686,14 +6665,24 @@ void Meta::meta_sai_on_port_state_change_single(
 
     auto ot = objectTypeQuery(data.port_id);
 
-    bool valid = isPortObjectIdValid(ot);
```
kcudnik, line 6689: this is not part of prefix entry

meta/Meta.cpp
```diff
@@ -7113,46 +7102,3 @@ void Meta::populate(
         }
     }
 }
-
-bool Meta::isPortObjectIdValid(
```
kcudnik, line 7117: this is not part of prefix entry

meta/Meta.cpp
```diff
@@ -7113,46 +7102,3 @@ void Meta::populate(
...
-                return true;
-        }
-
-        return false;
-    }
-
-    SWSS_LOG_THROW("port_id member not found on sai_port_oper_status_notification");
-}
-
-std::vector<std::string> Meta::getValidPortObjectTypes()
```
kcudnik, line 7143: this is not part of prefix entry

meta/SaiInterface.cpp
```diff
@@ -378,20 +378,3 @@ sai_status_t SaiInterface::clearStats(
 
     return SAI_STATUS_NOT_IMPLEMENTED;
 }
-
-std::shared_ptr<SaiOptions> SaiInterface::getOptions(
```
kcudnik, line 382: this is not part of prefix entry

meta/SaiOptions.h
```diff
@@ -1,11 +0,0 @@
-#pragma once
-
-namespace sairedis
```
kcudnik, line 3: this is not part of prefix entry

syncd/AttrVersionChecker.cpp
```diff
@@ -1,107 +0,0 @@
-#include "AttrVersionChecker.h"
-
-#include "swss/logger.h"
```
kcudnik, line 3: this is not part of prefix entry

kcudnik (changes requested):
a lot of removed code is not a aprt of compression prefix entry, is that necessary ro remove ?

kcudnik:
@yxieca @lguohan i think this is something similar we already made to revert my changes, and seems like this is very similar change, do we need to do this again ?

yxieca:
> @yxieca @lguohan i think this is something similar we already made to revert my changes, and seems like this is very similar change, do we need to do this again ?
> 
> take a look here: #1507 we already get back at some point to master SAI, why we are doing this agian?

I believe we agreed that the public master/202411 will use SAI header version 1.15.3. Which matches BRCM SAI version 12.

BYGX-wcr:
@kcudnik , I tested yesterday. The build could not pass except that I reverted all these changes in sonic-sairedis.

@rlhui , @r12f , @abdosi

kcudnik:
i'm not sure what changed it looks like running tests by vssyncd in tests directory is not generating gcovr, or it is generating but not counted to cover those linse, there are tests in /tests/BCM* that actually cover this single reiniter :/ maybe gcovr was updated since nobody was touching those tests, i encountered this in my latest pr too

BYGX-wcr:
> i'm not sure what changed it looks like running tests by vssyncd in tests directory is not generating gcovr, or it is generating but not counted to cover those linse, there are tests in /tests/BCM* that actually cover this single reiniter :/ maybe gcovr was updated since nobody was touching those tests, i encountered this in my latest pr too

My understanding is that this line in single reiniter should also be covered in unittest/syncd/TestSyncd.cpp because the syncd was set to cold reboot which would trigger the hard renit. Isn't that the case?

abdosi (approved):

meta/Meta.cpp
BYGX-wcr: yes, but we need to revert it too. Otherwise, the build with SAI v1.15.4 won't pass.

meta/Meta.cpp
BYGX-wcr: Ditto

meta/Meta.cpp
BYGX-wcr: Ditto

meta/Meta.h
```diff
@@ -315,11 +315,6 @@ namespace saimeta
             static bool is_ipv6_mask_valid(
                     _In_ const uint8_t* mask);
 
-            static bool isPortObjectIdValid(
```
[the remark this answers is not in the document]
BYGX-wcr: Ditto

meta/SaiInterface.cpp
BYGX-wcr: Ditto

meta/SaiOptions.h
BYGX-wcr: Ditto

syncd/AttrVersionChecker.cpp
BYGX-wcr: Ditto

r12f (approved):

kcudnik (approved):

kcudnik:
after everything will be settled with buildimage, pleae bring those changes back and move SAI to master

BYGX-wcr:
Closed since it is no longer needed.

PR #1536, opened 2025-02-19
Ignore ref counts when deleting DASH objects

theasianpianist:


kcudnik:
this seems like a bad idea and in long run it will cause crash and inconsistency, please provide description

theasianpianist:
> this seems like a bad idea and in long run it will cause crash and inconsistency, please provide description

Updated description

kcudnik:
i dont quite understand, that you can request to delete all entries even if they are still in use, this will definitely cause inconsistency error, since some attributes will still have attributes SET to that value you deleted, and this will cause in long rung syncd to crash because of inconsistency and reference count error. 

The change you are making is not solving actual issue, its one step to collapse everything, later on you will have  to make changes in syncd and allow to remove objects which are in use and you will have no choice to crash.

PR #1540, opened 2025-02-21
Fix counter issues on VS platform

stephenxs:
Fix two day 1 counter issues on VS platform which have been exposed recently.

1. Use switch object to detect counter capability in flex counter. This is to fix the following error message
    ```
    E               2025 Feb 20 15:43:21.906623 vlab-03 ERR syncd#syncd: :- queryStatsCapability: parameter switchId oid:0x100000001 object type is SAI_OBJECT_TYPE_PORT, but expected SAI_OBJECT_TYPE_SWITCH
    E               
    E               2025 Feb 20 15:43:21.910428 vlab-03 ERR syncd#syncd: :- queryStatsCapability: parameter switchId oid:0x1500000003 object type is SAI_OBJECT_TYPE_QUEUE, but expected SAI_OBJECT_TYPE_SWITCH
    ```

2. Support PFC counters on VS platform
    Previously it didn't check counter capability before polling PFC counters by mistake.

3. Do not poll counters for a group without any counter supported by the system in bulk creation

### DIFF
--- a/syncd/FlexCounter.cpp
+++ b/syncd/FlexCounter.cpp
@@ -480,7 +480,7 @@ class CounterContext : public BaseCounterContext
     std::map<std::string, uint32_t> m_counterChunkSizeMapFromPrefix;
 
 protected:
-    sai_object_id_t m_switchId;
+    sai_object_id_t m_switchId = SAI_NULL_OBJECT_ID;
 
 public:
     typedef CounterIds<StatType> CounterIdsType;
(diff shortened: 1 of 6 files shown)


syncd/FlexCounter.cpp
```diff
@@ -479,6 +479,9 @@ class CounterContext : public BaseCounterContext
 {
     std::map<std::string, uint32_t> m_counterChunkSizeMapFromPrefix;
 
+protected:
+    sai_object_id_t m_switchId;
```
kcudnik, line 483: this dont need to be a member, move it to function locally, since you are using this in 1 place

stephenxs: but I do not need to fetch it every time it is called since the switch id isn't changed

kcudnik (approved):

stephenxs:
cherry-pick conflict due to dependency on #1527

r12f:
hi @stephenxs , do you mind to help create a manual cherry pick for this change? looks like a lot of conflicts in this PR:

image

PR #1542, opened 2025-02-26
ICMP_ECHO_SESSION meta and notification layer changes

harjotsinghpawra:
Added support for ICMP_ECHO_SESSION in following areas of sonic-sairedis

meta layer
syncd
Notification Handling
added unittest to test notification and attribute handling

### DIFF
--- a/meta/Makefile.am
+++ b/meta/Makefile.am
@@ -36,6 +36,7 @@ libsaimeta_la_SOURCES = \
 				NotificationSwitchShutdownRequest.cpp \
 				NotificationSwitchStateChange.cpp \
 				NotificationBfdSessionStateChange.cpp \
+				NotificationIcmpEchoSessionStateChange.cpp \
 				NotificationHaSetEvent.cpp \
 				NotificationHaScopeEvent.cpp \
 				NotificationTwampSessionEvent.cpp \
--- a/meta/Meta.cpp
+++ b/meta/Meta.cpp
@@ -7055,6 +7055,56 @@ void Meta::meta_sai_on_ha_scope_event(
     }
 }
 
+void Meta::meta_sai_on_icmp_echo_session_state_change_single(
+        _In_ const sai_icmp_echo_session_state_notification_t& data)
+{
+    SWSS_LOG_ENTER();
+
+    auto ot = objectTypeQuery(data.icmp_echo_session_id);
+
+    bool valid = isIcmpEchoSessionObjectIdValid(ot);
+
+    if(!valid){
+        SWSS_LOG_ERROR("data.icmp_echo_session_id %s has unexpected type: %s, expected %s",
+                sai_serialize_object_id(data.icmp_echo_session_id).c_str(),
+                sai_serialize_object_type(ot).c_str(),
+                boost::algorithm::join(getValidIcmpEchoSessionObjectTypes(), ",").c_str());
+    }
+
+    if (valid && !m_oids.objectReferenceExists(data.icmp_echo_session_id))
+    {
+        SWSS_LOG_NOTICE("data.icmp_echo_session_id new object spotted %s not present in local DB (snoop!)",
+                sai_serialize_object_id(data.icmp_echo_session_id).c_str());
+
+        sai_object_meta_key_t key = { .objecttype = ot, .objectkey = { .key = { .object_id = data.icmp_echo_session_id } } };
+
+        m_oids.objectReferenceInsert(data.icmp_echo_session_id);
+
+        if (!m_saiObjectCollection.objectExists(key))
+        {
+            m_saiObjectCollection.createObject(key);
+        }
+    }
+}
+
+void Meta::meta_sai_on_icmp_echo_session_state_change(
+        _In_ uint32_t count,
+        _In_ const sai_icmp_echo_session_state_notification_t *data)
+{
+    SWSS_LOG_ENTER();
+
+    if (count && data == NULL)
+    {
+        SWSS_LOG_ERROR("sai_icmp_echo_session_state_notification_t pointer is NULL but count is %u", count);
+        return;
+    }
+
+    for (uint32_t i = 0; i < count; ++i)
+    {
+        meta_sai_on_icmp_echo_session_state_change_single(data[i]);
+    }
+}
+
 void Meta::meta_sai_on_twamp_session_event_single(
         _In_ const sai_twamp_session_event_notification_data_t& data)
 {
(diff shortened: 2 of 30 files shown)


meta/Makefile.am
```diff
@@ -36,6 +36,7 @@ libsaimeta_la_SOURCES = \
 				NotificationSwitchShutdownRequest.cpp \
 				NotificationSwitchStateChange.cpp \
 				NotificationBfdSessionStateChange.cpp \
+                NotificationIcmpEchoSessionStateChange.cpp \
```
kcudnik, line 39: correct ident

meta/Meta.cpp
```diff
@@ -6906,6 +6906,65 @@ void Meta::meta_sai_on_bfd_session_state_change(
...
+{
+    SWSS_LOG_ENTER();
+
+    auto ot = objectTypeQuery(data.icmp_echo_session_id);
+
+    bool valid = false;
+
+    switch (ot)
+    {
+        // TODO hardcoded types, must advance SAI repository commit to get metadata for this
```
kcudnik, line 6920: i think this is already done, SAI repo was advanced enough to get that metadata, please use it here

proxylib/Proxy.cpp
```diff
@@ -1203,6 +1204,19 @@ void Proxy::onBfdSessionStateChange(
...
+        _In_ const sai_icmp_echo_session_state_notification_t *data)
+{
+    SWSS_LOG_ENTER();
+
+    std::string s = sai_serialize_icmp_echo_session_state_ntf(count, data);
+
+    sendNotification(SAI_SWITCH_NOTIFICATION_NAME_ICMP_ECHO_SESSION_STATE_CHANGE, s);
+}
+
+
```
kcudnik, line 1218: extra empty lines not needed

unittest/meta/Makefile.am
```diff
@@ -26,6 +26,7 @@ tests_SOURCES = \
 				TestNotificationSwitchAsicSdkHealthEvent.cpp \
 				TestNotificationSwitchStateChange.cpp \
 				TestNotificationBfdSessionStateChange.cpp \
+                TestNotificationIcmpEchoSessionStateChange.cpp \
```
kcudnik, line 29: correct ident

kcudnik (changes requested):

meta/SaiSerialize.cpp
```diff
@@ -2537,6 +2545,32 @@ std::string sai_serialize_bfd_session_state_ntf(
...
+        SWSS_LOG_THROW("icmp_echo_session _state pointer is null");
+    }
+
+    json j = json::array();
+
+    for (uint32_t i = 0; i < count; ++i)
+    {
+        json item;
+
+        item["icmp_echo"] = sai_serialize_object_id(icmp_echo_session_state[i].icmp_echo_session_id);
```
manamand2020, line 2565: This needs to be "icmp_echo_session_id"

manamand2020 (changes requested):
Fix sai_serialize_icmp_echo_session_state_ntf()

meta/SaiSerialize.cpp
zjswhhh: Is UT covering this function? No?

meta/Meta.cpp
harjotsinghpawra: Hi @kcudnik  can I use https://github.com/sonic-net/sonic-sairedis/blob/master/meta/Meta.cpp#L1754 -> Meta::meta_sai_validate_oid this function dose all the verification it just dose not check objectRefrences which i am not sure why. But can use this directly?

kcudnik: 1. this function is by desing, i was proposing something like isPortObjectIdValid, please check on meta_sai_on_port_state_change_single

harjotsinghpawra: Fixed in updated diff

proxylib/Proxy.cpp
harjotsinghpawra: Fixed in updated diff

unittest/meta/Makefile.am
harjotsinghpawra: Fixed in updated diff

meta/Meta.cpp
```diff
@@ -6906,6 +6906,65 @@
...
+            valid = true;
+            break;
+
+        default:
+
+            SWSS_LOG_ERROR("data.icmp_echo_session_id %s has unexpected type: %s, expected ICMP_ECHO_SESSION",
+                    sai_serialize_object_id(data.icmp_echo_session_id).c_str(),
+                    sai_serialize_object_type(ot).c_str());
+            break;
+    }
```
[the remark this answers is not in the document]
harjotsinghpawra: Fixed in updated diff

meta/Meta.cpp
harjotsinghpawra: Hi @kcudnik  i am trying to debug bool Meta::isIcmpEchoSessionObjectIdValid function if its properly returning true and false. Is there a way i can attach gdb or put logs to debug when i am running unittest using make check rather than loading it on a device and configuring?

kcudnik: Yes not related to your change

harjotsinghpawra: @kcudnik Hi Kamil i realized while testing the oid i used is conflicting  with SAI_OBJECT_TYPE_PORT , i see that in Legacy test we create new oid and use that , but going onward is there specific way or place to create oid for unittests i am unable to create a valid oid for SAI_OBJECT_TYPE_ICMP_ECHO_SESSION ?

kcudnik: you can see https://github.com/sonic-net/sonic-sairedis/blob/master/lib/VirtualObjectIdManager.cpp how oids are constructed if you need for unittests

kcudnik:
please resolve conflicts

kcudnik (dismissed):

harjotsinghpawra:
FYI Squashing the commits to single commit

kcudnik:
> @kcudnik Hi kamil , I squashed all my commits to single commit. Now code coverage is getting failed , i guess there was some bug when i pushed incremental changes it caused coverage to pass . to pass this coverage i need to test the syncd notification code . I dont see any other similar test . Nevertheless to pass code coverage do i need to put my state change test code in unitest/syncd Notification processor and Notification Handler . But i thought they were to do generic test of notification processor and handler not specific notifications . What would be the best place to put this test code to get proper code coverage or are we not supposed to test it

yea, you would need to put test code in unitests/syncd/ and corresponding files for notification handler and notification processor, i see that you also have some uncovered lines in Meta.cpp and DummySaiInterface.cpp, also SaiSerialize is not covered, those should be fairly easy to implement 

and tests in NotificationProcessor and NotificationHandler are alrady written for other notifications in TestNotificationHandler.cpp and TestNotificationProcessor.cpp

kcudnik:
please resolve conflicts

harjotsinghpawra:
@kcudnik i made no changes in FlexCounter section. its failing for me even without my commit also. i see last commit in sonic-sairedis being merged having same failure . I have added all the test cases and in my previous run my code coverage was reaching 92%. I resynced my fork and raised new consolidated commit now i see all runs stuck and failing on this FlexCOunter.Bulksize test case with which my code is not related. Can you please help me review it again with added test case and help me to get it merged .thnaks for all you help

kcudnik:
that bulk test counter fail is also failing on master build, but only on ASAN build, other builds and tests are passing fine, can you rebase your changes to master and push force ?

harjotsinghpawra:
@kcudnik Hi kamil, i think you re-triggered the checks now everything is passing nice

PR #1546, opened 2025-02-27
Update to support SAI BFD_SESSION_OFFLOAD_TYPE.

dypet:
Adding support for SAI_SWITCH_ATTR_SUPPORTED_IPV4_BFD_SESSION_OFFLOAD_TYPE and SAI_SWITCH_ATTR_SUPPORTED_IPV6_BFD_SESSION_OFFLOAD_TYPE in vslib. This is needed by BfdOrch in sonic-swss to choose between software BFD or regular (offloaded) BFD.

Default will be to return SAI_BFD_SESSION_OFFLOAD_TYPE_FULL, as this was the default behavior previously when this value was not checked. 

Added capability to return SAI_BFD_SESSION_OFFLOAD_TYPE_NONE by setting SAI_VS_BFD_OFFLOAD_SUPPORTED=false in sai.profile for testing and in cases where software BFD is desired.

### DIFF
--- a/vslib/SwitchConfig.cpp
+++ b/vslib/SwitchConfig.cpp
@@ -17,7 +17,8 @@ SwitchConfig::SwitchConfig(
     m_bootType(SAI_VS_BOOT_TYPE_COLD),
     m_switchIndex(switchIndex),
     m_hardwareInfo(hwinfo),
-    m_useTapDevice(false)
+    m_useTapDevice(false),
+    m_bfdOffload(true)
 {
     SWSS_LOG_ENTER();
 
@@ -152,4 +153,17 @@ bool SwitchConfig::parseUseTapDevice(
     }
 
     return false;
+}
+
+bool SwitchConfig::parseBfdOffloadSupported(
+    _In_ const char* bfdOffloadSupportedStr)
+{
+    SWSS_LOG_ENTER();
+
+    if (bfdOffloadSupportedStr)
+    {
+        return strcmp(bfdOffloadSupportedStr, "true") == 0;
+    }
+
+    return true;
 }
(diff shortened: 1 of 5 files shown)


vslib/SwitchConfig.cpp
```diff
@@ -17,7 +17,8 @@ SwitchConfig::SwitchConfig(
     m_bootType(SAI_VS_BOOT_TYPE_COLD),
     m_switchIndex(switchIndex),
     m_hardwareInfo(hwinfo),
-    m_useTapDevice(false)
+    m_useTapDevice(false),
+    m_bfdOffload(true)
```
kcudnik, line 21: why this is enabled by default ?

vslib/SwitchStateBase.cpp
```diff
@@ -1055,6 +1055,20 @@ sai_status_t SwitchStateBase::set_switch_default_attributes()
...
+    }
+    attr.value.u32list.count = 1;
+    attr.value.u32list.list = list;
+    CHECK_STATUS(set(SAI_OBJECT_TYPE_SWITCH, m_switch_id, &attr));
+
+    attr.id = SAI_SWITCH_ATTR_SUPPORTED_IPV6_BFD_SESSION_OFFLOAD_TYPE;
+    attr.value.u32list.count = 1;
+    attr.value.u32list.list = list;
+    CHECK_STATUS(set(SAI_OBJECT_TYPE_SWITCH, m_switch_id, &attr));
+
```
kcudnik, line 1071: add empty lines before if and after } and before and after CHECK_STATUS

vslib/SwitchConfig.cpp
dypet: BFD_SESSION_OFFLOAD_TYPE was not checked in sonic-swss previously, and default behavior in BfdOrch assumed that offload was supported. Defaulting to true to stay consistent with this previous behavior/testing, and allowing it to be set false to use or test the newly introduced software BFD (no offload) as needed.

vslib/SwitchStateBase.cpp
dypet: Updated

kcudnik (dismissed):

kcudnik (approved):

dypet:
Hi @kcudnik , I don't have the permissions to run the remaining checks - can they be started?

PR #1547, opened 2025-02-28
fix build not link libnl-3.so

nanfengnan1:


kcudnik:
everything compiles fine without this change, and this PR will not be merged

PR #1548, opened 2025-03-05
[syncd] Move log set function after api initialize

kcudnik:
It may happen that in case of virtual switch, logset will be called before api initialize, which will log error messages, since VS was not initialized yet, and logSet in VS requires api to be initialized before call


stephenxs (approved):

prabhataravind:
hi @kcudnik, can this be merged? 
Once this is merged, I can include this PR and restart PR checks for https://github.com/sonic-net/sonic-buildimage/pull/22019

PR #1559, opened 2025-03-18
[discovery] speed up sai discovery on ports

stepanblyschak:
The performance issue observed in port discovery part (as explained in https://github.com/azure/sonic-sairedis/pull/1416) addressed by optimizing discovery process. We get from 4.7 sec to 0.8 sec and pretty much bandwidth limited by redis-server performance (redis-benchmark shows 0.7 sec for 30K set operations).

Strategy:
1. Don't get SAI attributes on ports that are just created that have a known default null oid or empty list value
2. Allocate VIDs for all discovered objects in one redis command INCRBY
3. Use Redis pipeline for set operations

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -272,11 +272,20 @@ AC_MSG_ERROR("SAI headers API version and library version mismatch")])],
 CXXFLAGS="$SAVED_FLAGS"
 ])])
 
-AM_COND_IF([SYNCD], [
-SAVED_FLAGS="$CXXFLAGS"
-CXXFLAGS="-Xlinker --no-as-needed -lsai -I$srcdir/SAI/inc -I$srcdir/SAI/experimental -I$srcdir/SAI/meta"
-AC_CHECK_FUNCS(sai_bulk_object_clear_stats sai_bulk_object_get_stats)
-CXXFLAGS="$SAVED_FLAGS"
+AM_COND_IF([SAIVS],
+[
+    AC_DEFINE([HAVE_SAI_BULK_OBJECT_CLEAR_STATS], [1], [])
+    AC_DEFINE([HAVE_SAI_BULK_OBJECT_GET_STATS], [1], [])
+    AC_DEFINE([HAVE_SAI_QUERY_STATS_ST_CAPABILITY], [1], [])
+    AC_DEFINE([HAVE_SAI_TAM_TELEMETRY_GET_DATA], [1], [])
+],
+[
+    AM_COND_IF([SYNCD], [
+    SAVED_FLAGS="$CXXFLAGS"
+    CXXFLAGS="-Xlinker --no-as-needed -lsai -I$srcdir/SAI/inc -I$srcdir/SAI/experimental -I$srcdir/SAI/meta"
+    AC_CHECK_FUNCS(sai_bulk_object_clear_stats sai_bulk_object_get_stats sai_query_stats_st_capability sai_tam_telemetry_get_data)
+    CXXFLAGS="$SAVED_FLAGS"
+    ])
 ])
 
 AC_DEFINE([SAIREDIS_GIT_REVISION],
--- a/meta/Makefile.am
+++ b/meta/Makefile.am
@@ -36,6 +36,8 @@ libsaimeta_la_SOURCES = \
 				NotificationSwitchShutdownRequest.cpp \
 				NotificationSwitchStateChange.cpp \
 				NotificationBfdSessionStateChange.cpp \
+				NotificationHaSetEvent.cpp \
+				NotificationHaScopeEvent.cpp \
 				NotificationTwampSessionEvent.cpp \
 				NotificationPortHostTxReadyEvent.cpp \
 				NotificationTamTelTypeConfigChange.cpp \
--- a/meta/Meta.cpp
+++ b/meta/Meta.cpp
@@ -6937,6 +6937,124 @@ void Meta::meta_sai_on_bfd_session_state_change(
     }
 }
 
+void Meta::meta_sai_on_ha_set_event_single(
+    _In_ const sai_ha_set_event_data_t& data)
+{
+    SWSS_LOG_ENTER();
+
+    auto ot = objectTypeQuery(data.ha_set_id);
+
+    bool valid = false;
+
+    switch ((int)ot)
+    {
+        // TODO hardcoded types, must advance SAI repository commit to get metadata for this
+        case SAI_OBJECT_TYPE_HA_SET:
+
+            valid = true;
+            break;
+
+        default:
+
+            SWSS_LOG_ERROR("data.ha_set_id %s has unexpected type: %s, expected HA_SET",
+                    sai_serialize_object_id(data.ha_set_id).c_str(),
+                    sai_serialize_object_type(ot).c_str());
+            break;
+    }
+
+    if (valid && !m_oids.objectReferenceExists(data.ha_set_id))
+    {
+        SWSS_LOG_NOTICE("data.ha_set_id new object spotted %s not present in local DB (snoop!)",
+                sai_serialize_object_id(data.ha_set_id).c_str());
+
+        sai_object_meta_key_t key = { .objecttype = ot, .objectkey = { .key = { .object_id = data.ha_set_id } } };
+
+        m_oids.objectReferenceInsert(data.ha_set_id);
+
+        if (!m_saiObjectCollection.objectExists(key))
+        {
+            m_saiObjectCollection.createObject(key);
+        }
+    }
+}
+
+void Meta::meta_sai_on_ha_set_event(
+        _In_ uint32_t count,
+        _In_ const sai_ha_set_event_data_t *data)
+{
+    SWSS_LOG_ENTER();
+
+    if (count && data == NULL)
+    {
+        SWSS_LOG_ERROR("sai_ha_set_event_data_t pointer is NULL but count is %u", count);
+        return;
+    }
+
+    for (uint32_t i = 0; i < count; ++i)
+    {
+        meta_sai_on_ha_set_event_single(data[i]);
+    }
+}
+
+void Meta::meta_sai_on_ha_scope_event_single(
+        _In_ const sai_ha_scope_event_data_t& data)
+{
+    SWSS_LOG_ENTER();
+
+    auto ot = objectTypeQuery(data.ha_scope_id);
+
+    bool valid = false;
+
+    switch ((int)ot)
+    {
+        // TODO hardcoded types, must advance SAI repository commit to get metadata for this
+        case SAI_OBJECT_TYPE_HA_SCOPE:
+
+            valid = true;
+            break;
+
+        default:
+
+            SWSS_LOG_ERROR("data.ha_scope_id %s has unexpected type: %s, expected HA_SCOPE",
+                    sai_serialize_object_id(data.ha_scope_id).c_str(),
+                    sai_serialize_object_type(ot).c_str());
+            break;
+    }
+
+    if (valid && !m_oids.objectReferenceExists(data.ha_scope_id))
+    {
+        SWSS_LOG_NOTICE("data.ha_scope_id new object spotted %s not present in local DB (snoop!)",
+                sai_serialize_object_id(data.ha_scope_id).c_str());
+
+        sai_object_meta_key_t key = { .objecttype = ot, .objectkey = { .key = { .object_id = data.ha_scope_id } } };
+
+        m_oids.objectReferenceInsert(data.ha_scope_id);
+
+        if (!m_saiObjectCollection.objectExists(key))
+        {
+            m_saiObjectCollection.createObject(key);
+        }
+    }
+}
+
+void Meta::meta_sai_on_ha_scope_event(
+        _In_ uint32_t count,
+        _In_ const sai_ha_scope_event_data_t *data)
+{
+    SWSS_LOG_ENTER();
+
+    if (count && data == NULL)
+    {
+        SWSS_LOG_ERROR("sai_ha_scope_event_data_t pointer is NULL but count is %u", count);
+        return;
+    }
+
+    for (uint32_t i = 0; i < count; ++i)
+    {
+        meta_sai_on_ha_scope_event_single(data[i]);
+    }
+}
+
 void Meta::meta_sai_on_twamp_session_event_single(
         _In_ const sai_twamp_session_event_notification_data_t& data)
 {
--- a/meta/Meta.h
+++ b/meta/Meta.h
@@ -245,6 +245,14 @@ namespace saimeta
                     _In_ uint32_t count,
                     _In_ const sai_bfd_session_state_notification_t *data);
 
+            void meta_sai_on_ha_set_event(
+                    _In_ uint32_t count,
+                    _In_ const sai_ha_set_event_data_t *data);
+
+            void meta_sai_on_ha_scope_event(
+                    _In_ uint32_t count,
+                    _In_ const sai_ha_scope_event_data_t *data);
+
             void meta_sai_on_port_host_tx_ready_change(
                     _In_ sai_object_id_t port_id,
                     _In_ sai_object_id_t switch_id,
@@ -278,6 +286,12 @@ namespace saimeta
 
             void meta_sai_on_bfd_session_state_change_single(
                     _In_ const sai_bfd_session_state_notification_t& data);
+
+            void meta_sai_on_ha_set_event_single(
+                    _In_ const sai_ha_set_event_data_t& data);
+
+            void meta_sai_on_ha_scope_event_single(
+                    _In_ const sai_ha_scope_event_data_t& data);
 
             void meta_sai_on_twamp_session_event_single(
                     _In_ const sai_twamp_session_event_notification_data_t& data);
--- a/meta/NotificationFactory.cpp
+++ b/meta/NotificationFactory.cpp
@@ -10,6 +10,8 @@
 #include "NotificationTwampSessionEvent.h"
 #include "NotificationPortHostTxReadyEvent.h"
 #include "NotificationTamTelTypeConfigChange.h"
+#include "NotificationHaSetEvent.h"
+#include "NotificationHaScopeEvent.h"
 #include "sairediscommon.h"
 
 #include "swss/logger.h"
@@ -48,6 +50,12 @@ std::shared_ptr<Notification> NotificationFactory::deserialize(
 
     if (name == SAI_SWITCH_NOTIFICATION_NAME_BFD_SESSION_STATE_CHANGE)
         return std::make_shared<NotificationBfdSessionStateChange>(serializedNotification);
+
+    if (name == SAI_SWITCH_NOTIFICATION_NAME_HA_SET_EVENT)
+        return std::make_shared<NotificationHaSetEvent>(serializedNotification);
+
+    if (name == SAI_SWITCH_NOTIFICATION_NAME_HA_SCOPE_EVENT)
+        return std::make_shared<NotificationHaScopeEvent>(serializedNotification);
 
     if (name == SAI_SWITCH_NOTIFICATION_NAME_TWAMP_SESSION_EVENT)
         return std::make_shared<NotificationTwampSessionEvent>(serializedNotification);
--- a/meta/NotificationHaScopeEvent.cpp
+++ b/meta/NotificationHaScopeEvent.cpp
@@ -1,0 +1,78 @@
+#include "NotificationHaScopeEvent.h"
+
+#include "swss/logger.h"
+
+#include "meta/sai_serialize.h"
+
+using namespace sairedis;
+
+NotificationHaScopeEvent::NotificationHaScopeEvent(
+        _In_ const std::string& serializeNotification):
+    Notification(
+        SAI_SWITCH_NOTIFICATION_TYPE_HA_SCOPE_EVENT,
+        serializeNotification),
+    m_haScopeEventData(nullptr)
+{
+    SWSS_LOG_ENTER();
+
+    sai_deserialize_ha_scope_event_ntf(
+        serializeNotification,
+        m_count,
+        &m_haScopeEventData);
+}
+
+NotificationHaScopeEvent::~NotificationHaScopeEvent()
+{
+    SWSS_LOG_ENTER();
+
+    sai_deserialize_free_ha_scope_event_ntf(m_count, m_haScopeEventData);
+}
+
+sai_object_id_t NotificationHaScopeEvent::getSwitchId() const
+{
+    SWSS_LOG_ENTER();
+
+    // Ha scope event does not have switch id
+    return SAI_NULL_OBJECT_ID;
+}
+
+sai_object_id_t NotificationHaScopeEvent::getAnyObjectId() const
+{
+    SWSS_LOG_ENTER();
+
+    if (m_haScopeEventData == nullptr)
+    {
+        return SAI_NULL_OBJECT_ID;
+    }
+
+    for (uint32_t i = 0; i < m_count; ++i)
+    {
+        if (m_haScopeEventData[i].ha_scope_id != SAI_NULL_OBJECT_ID)
+        {
+            return m_haScopeEventData[i].ha_scope_id;
+        }
+    }
+
+    return SAI_NULL_OBJECT_ID;
+}
+
+void NotificationHaScopeEvent::processMetadata(
+        _In_ std::shared_ptr<saimeta::Meta> meta) const
+{
+    SWSS_LOG_ENTER();
+
+    meta->meta_sai_on_ha_scope_event(
+            m_count,
+            m_haScopeEventData);
+}
+
+void NotificationHaScopeEvent::executeCallback(
+        _In_ const sai_switch_notifications_t& switchNotifications) const
+{
+    SWSS_LOG_ENTER();
+
+    if (switchNotifications.on_ha_scope_event)
+    {
+        switchNotifications.on_ha_scope_event(m_count, m_haScopeEventData);
+    }
+}
--- a/meta/NotificationHaScopeEvent.h
+++ b/meta/NotificationHaScopeEvent.h
@@ -1,0 +1,35 @@
+#pragma once
+
+#include "Notification.h"
+
+namespace sairedis
+{
+    class NotificationHaScopeEvent:
+        public Notification
+    {
+        public:
+
+            NotificationHaScopeEvent(
+                    _In_ const std::string& serializedNotification);
+
+            virtual ~NotificationHaScopeEvent();
+
+        public:
+
+            virtual sai_object_id_t getSwitchId() const override;
+
+            virtual sai_object_id_t getAnyObjectId() const override;
+
+            virtual void processMetadata(
+                    _In_ std::shared_ptr<saimeta::Meta> meta) const override;
+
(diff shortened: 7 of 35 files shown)


kcudnik:
> Strategy:
> 
> 1. Don't get SAI attributes on ports that are just created that have a known default null oid or empty list value
this helps to verify if any of attribute is implemented or returns anything, this helps to discover vendor bugs

> 2. Allocate VIDs for all discovered objects in one redis command INCRBY
not sure how this was implemented, but usually there is less than 2k objects that are discovered, then this should be really fast

> 3. Use Redis pipeline for set operations
will review that

4.7seconds it's not that much, especially that dataplane traffic is not affected by this at all

why you are making this change ? who requested that ? what part of sonic is impacted by this ?

i rather want to have all discovery working fine and discover everything than risking that we will have inconsistent state and crash later on

stepanblyschak:
> > Strategy:
> > 
> > 1. Don't get SAI attributes on ports that are just created that have a known default null oid or empty list value
> >    this helps to verify if any of attribute is implemented or returns anything, this helps to discover vendor bugs
> 
> > 2. Allocate VIDs for all discovered objects in one redis command INCRBY
> >    not sure how this was implemented, but usually there is less than 2k objects that are discovered, then this should be really fast
> 
> > 3. Use Redis pipeline for set operations
> >    will review that
> 
> 4.7seconds it's not that much, especially that dataplane traffic is not affected by this at all

This time affects cold boot and fast boot traffic downtime (no traffic until create_ports() returns for sure) as well as warm boot control plane downtime (since host interfaces can be created only after creaing the switch and ports).
4.7 for 256 ports, it scales linearly with number of ports. ~16% (256) and ~31% (512 ports). Most of the time it's redis operations, not SAI. It may take more time than init the switch or create ports in SAI.
This is why I did https://github.com/sonic-net/sonic-sairedis/pull/1416 previously

> why you are making this change ? who requested that ? what part of sonic is impacted by this ?

Fast-boot downtime requirement.

> i rather want to have all discovery working fine and discover everything than risking that we will have inconsistent state and crash later on

I am looking into test failures, appreciate if you suggest flows to test manually or I can add more test cases to ensure no degradation.

kcudnik:
> This time affects cold boot and fast boot traffic downtime (no traffic until create_ports() returns for sure) as well as warm boot control plane downtime (since host interfaces can be created only after creaing the switch and ports). 4.7 for 256 ports, it scales linearly with number of ports. ~16% (256) and ~31% (512 ports). Most of the time it's redis operations, not SAI. It may take more time than init the switch or create ports in SAI. This is why I did #1416 previously
> 
> > why you are making this change ? who requested that ? what part of sonic is impacted by this ?
> 
> Fast-boot downtime requirement.
> I am looking into test failures, appreciate if you suggest flows to test manually or I can add more test cases to ensure no degradation.

im thinking on scenario where there will be bug on vendor device, at this point it would be hard to simulate, probably a special test case would need to be written on virtual switch to simulate bug for this specific scenario

i'm more thinking of delegating those ports discovery after switch configuration, but question is when to do that ? after apply view?

lib/OidIndexGenerator.h
```diff
@@ -15,6 +15,7 @@ namespace sairedis
         public:
 
             virtual uint64_t increment() = 0;
+            virtual uint64_t incrementBy(uint64_t count) = 0;
```
kcudnik, line 18: the idea here is that this function returns index, so this function should return list of new indexes

kcudnik: also fix convention, each param in new line with In prefx

lib/RedisVidIndexGenerator.cpp
```diff
@@ -25,6 +27,16 @@ uint64_t RedisVidIndexGenerator::increment()
     return m_dbConnector->incr(m_vidCounterName); // "VIDCOUNTER"
 }
 
+uint64_t RedisVidIndexGenerator::incrementBy(uint64_t count)
```
kcudnik, line 30: also fix convention, each param in new line with In prefx

lib/VirtualObjectIdManager.cpp
```diff
@@ -259,6 +259,56 @@ sai_object_id_t VirtualObjectIdManager::allocateNewObjectId(
...
+    {
+        SWSS_LOG_THROW("object type of switch %s is %s, should be SWITCH",
+                sai_serialize_object_id(switchId).c_str(),
+                sai_serialize_object_type(switchObjectType).c_str());
+    }
+
+    uint32_t switchIndex = static_cast<uint32_t>(SAI_REDIS_GET_SWITCH_INDEX(switchId));
+
+    uint64_t lastObjectIndex = m_oidIndexGenerator->incrementBy(count);
+    uint64_t firstObjectIndex = lastObjectIndex - count + 1;
```
kcudnik, line 295: i would move that functionality to incrementBy and return list of new indexes for user, in this way there is no need to do taht every time of usage

meta/NumberOidIndexGenerator.cpp
```diff
@@ -18,6 +18,13 @@ uint64_t NumberOidIndexGenerator::increment()
     return ++m_index;
 }
 
+uint64_t NumberOidIndexGenerator::incrementBy(uint64_t count)
```
kcudnik, line 21: also fix convention, each param in new line with In prefx

meta/NumberOidIndexGenerator.h
```diff
@@ -17,6 +17,7 @@ namespace saimeta
         public:
 
             virtual uint64_t increment() override;
+            virtual uint64_t incrementBy(uint64_t count) override;
```
kcudnik, line 20: also fix convention, each param in new line with In prefx

saiasiccmp/Makefile.am
```diff
@@ -19,7 +19,7 @@ saiasiccmp_SOURCES = main.cpp
 saiasiccmp_CPPFLAGS = $(CODE_COVERAGE_CPPFLAGS)
 saiasiccmp_CXXFLAGS = $(DBGFLAGS) $(AM_CXXFLAGS) $(CXXFLAGS_COMMON) $(CODE_COVERAGE_CXXFLAGS)
 saiasiccmp_LDADD = libAsicCmp.a \
-				   -ldl -lhiredis -lswsscommon -lpthread -L$(top_srcdir)/meta/.libs -lsaimetadata -lsaimeta -lzmq \
-				   $(top_srcdir)/syncd/libSyncd.a $(top_srcdir)/lib/libSaiRedis.a $(CODE_COVERAGE_LIBS)
+				   -ldl -lpthread -L$(top_srcdir)/meta/.libs -lsaimetadata -lsaimeta -lzmq \
+				   $(top_srcdir)/syncd/libSyncd.a $(top_srcdir)/lib/libSaiRedis.a -lhiredis -lswsscommon $(CODE_COVERAGE_LIBS)
```
kcudnik, line 23: why this is changed ?

saiasiccmp/SaiSwitchAsic.cpp
```diff
@@ -279,8 +279,8 @@ std::set<sai_object_id_t> SaiSwitchAsic::getWarmBootDiscoveredVids() const
 }
 
 void SaiSwitchAsic::onPostPortCreate(
-        _In_ sai_object_id_t port_rid,
-        _In_ sai_object_id_t port_vid)
+        _In_ sai_object_id_t* port_rids,
+        _In_ size_t count)
```
kcudnik, line 283: function naming suggests 1 port and returns multiple ports

syncd/RedisClient.cpp
```diff
@@ -231,6 +231,32 @@ void RedisClient::setDummyAsicStateObject(
     m_dbAsic->hset(strKey, "NULL", "NULL");
 }
 
+void RedisClient::setDummyAsicStateObject(
+        _In_ sai_object_id_t* objectVid,
```
kcudnik, line 235: name this objectVids, and it probably be better if  this would be const std::vector<sai_object_id_t>

syncd/RedisClient.cpp
```diff
@@ -717,6 +743,36 @@ void RedisClient::insertVidAndRid(
     m_dbAsic->hset(RIDTOVID, strRid, strVid);
 }
 
+void RedisClient::insertVidAndRid(
+        _In_ sai_object_id_t* vids,
+        _In_ sai_object_id_t* rids,
+        _In_ size_t count)
```
kcudnik, line 749: name is for single vid and rid, change to vid/rids, and also use std::vector maybe ?

syncd/RedisClient.cpp
```diff
@@ -741,6 +797,57 @@ sai_object_id_t RedisClient::getVidForRid(
     return vid;
 }
 
+void RedisClient::getVidForRid(
+        _In_ const sai_object_id_t* rids,
+        _Out_ sai_object_id_t* vids,
+        _In_ size_t count)
+{
```
kcudnik, line 804: naming - rids/vids, also vectors ?

syncd/RedisClient.cpp
```diff
@@ -741,6 +797,57 @@ sai_object_id_t RedisClient::getVidForRid(
...
+
+    for (size_t idx = 0; idx < ctx->elements; idx++)
+    {
+        auto element = ctx->element[idx];
+
+        if (element->type == REDIS_REPLY_STRING)
+        {
+            sai_deserialize_object_id(element->str, vids[idx]);
+        }
+        else
```
kcudnik, line 844: in single version of this function null is returned when hget returns nullptr, in this case what is REDIS_REPLY_XXX  value? maybe only one specific should be considered, and other should throw since redis can reply multple errors

syncd/SaiDiscovery.cpp
```diff
@@ -157,7 +159,10 @@ void SaiDiscovery::discover(
                  * create, we don't need to query this attribute.
                  */
 
-                //continue;
+                if (m_flags & SaiDiscoveryFlags::SkipDefaultEmptyAttributes)
+                {
+                    continue;
+                }
```
kcudnik, line 165: this seems to affect all attributes on all objects, please limit this scope to only Port

syncd/SaiDiscovery.cpp
```diff
@@ -234,7 +239,10 @@ void SaiDiscovery::discover(
...
```
kcudnik, line 245: same here limit to port only

syncd/SaiDiscovery.cpp
```diff
@@ -289,6 +297,15 @@ std::set<sai_object_id_t> SaiDiscovery::discover(
...
 
+    return discover(&startRid, 1);
+}
+
+std::set<sai_object_id_t> SaiDiscovery::discover(
+        _In_ sai_object_id_t* rids,
+        _In_ size_t count)
+{
+    SWSS_LOG_ENTER();
+
```
kcudnik, line 308: why here will be multiple rids to start ? call function in line 287 multiple times in your loop, don't put loop here

syncd/SaiDiscovery.h
```diff
@@ -9,8 +9,24 @@
...
+        None = 0,
+        SkipDefaultEmptyAttributes = 1 << 0,
+    };
+
+    inline SaiDiscoveryFlags operator|(SaiDiscoveryFlags a, SaiDiscoveryFlags b)
+    {
+        SWSS_LOG_ENTER();
+
+        return static_cast<SaiDiscoveryFlags>(static_cast<int>(a) | static_cast<int>(b));
+    }
```
kcudnik, line 28: should this be in SaiDiscovery class ?

syncd/SaiSwitch.cpp
```diff
@@ -371,6 +371,19 @@ void SaiSwitch::redisSetDummyAsicStateForRealObjectId(
     m_client->setDummyAsicStateObject(vid);
 }
 
+void SaiSwitch::redisSetDummyAsicStateForRealObjectIds(
+        _In_ sai_object_id_t* rids,
+        _In_ size_t count) const
```
kcudnik, line 376: consider vector as input, you create that internally anyway

syncd/SaiSwitchInterface.h
```diff
@@ -91,8 +91,8 @@ namespace syncd
             virtual std::set<sai_object_id_t> getWarmBootNewDiscoveredVids();
 
             virtual void onPostPortCreate(
-                    _In_ sai_object_id_t port_rid,
-                    _In_ sai_object_id_t port_vid) = 0;
+                    _In_ sai_object_id_t* port_rid,
+                    _Out_ size_t count) = 0;
```
kcudnik, line 96: function naming dont correspond to parameters

syncd/SaiSwitch.h
```diff
@@ -190,13 +190,13 @@ namespace syncd
...
-             * queues, ipgs and scheduler groups that belong to new created port,
+             * Performs actions needed after ports creation. Will discover new
+             * queues, ipgs and scheduler groups that belong to new created ports,
              * and updated ASIC DB accordingly.
              */
             virtual void onPostPortCreate(
-                    _In_ sai_object_id_t port_rid,
-                    _In_ sai_object_id_t port_vid) override;
+                    _In_ sai_object_id_t* port_rid,
+                    _Out_ size_t count) override;
```
kcudnik, line 199: function naming dont correspond to parameters

syncd/VirtualOidTranslator.cpp
```diff
@@ -147,6 +147,79 @@ sai_object_id_t VirtualOidTranslator::translateRidToVid(
     return vid;
 }
 
+void VirtualOidTranslator::translateRidToVid(
```
kcudnik, line 150: please analyse funcion in line 60, this method here dont translare removed objects

kcudnik (changes requested):
please address comments and consider using std::vector for input arrays instead of pointers and count

lib/OidIndexGenerator.h
stepanblyschak: Done

lib/RedisVidIndexGenerator.cpp
stepanblyschak: Done

lib/VirtualObjectIdManager.cpp
stepanblyschak: Done

meta/NumberOidIndexGenerator.cpp
stepanblyschak: Done

saiasiccmp/Makefile.am
stepanblyschak: get this error after my changes
```
libtool: link: g++ -g -I../SAI/inc -I../SAI/experimental -I../SAI/meta -I../meta -I../lib -I../syncd -ansi -fPIC -pipe -std=c++14 -Wall -Wcast-align -Wcast-qual -Wconversion -Wdisabled-optimization -Werror -Wextra -Wfloat-equal -Wformat=2 -Wformat-nonliteral -Wformat-security -Wformat-y2k -Wimport -Winit-self -Wno-inline -Winvalid-pch -Wmissing-field-initializers -Wmissing-format-attribute -Wmissing-include-dirs -Wmissing-noreturn -Wno-aggregate-return -Wno-padded -Wno-switch-enum -Wno-unused-parameter -Wpacked -Wpointer-arith -Wredundant-decls -Wshadow -Wstack-protector -Wstrict-aliasing=3 -Wswitch -Wswitch-default -Wunreachable-code -Wunused -Wvariadic-macros -Wwrite-strings -Wno-switch-default -Wconversion -Wno-psabi -Wcast-align=strict -g -O2 -o .libs/saiasiccmp saiasiccmp-main.o  libAsicCmp.a -ldl -lhiredis -lswsscommon -lpthread -L../meta/.libs /sonic/src/sonic-sairedis/meta/.libs/libsaimetadata.so /sonic/src/sonic-sairedis/meta/.libs/libsaimeta.so -lzmq ../syncd/libSyncd.a ../lib/libSaiRedis.a
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/12/../../../x86_64-linux-gnu/libswsscommon.so: undefined reference to symbol 'redisGetReply'
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/12/../../../x86_64-linux-gnu/libhiredis.so: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:512: saiasiccmp] Error 1
```

linker order should be dependency library comes after usage.

saiasiccmp/SaiSwitchAsic.cpp
stepanblyschak: Done

syncd/RedisClient.cpp
stepanblyschak: Done

syncd/RedisClient.cpp
stepanblyschak: Done

syncd/RedisClient.cpp
stepanblyschak: Done

syncd/SaiDiscovery.cpp
stepanblyschak: Not related to objec type. e.g. discovery after warm boot should be without this flag - attributes could be populated with objects before in old life.
If object is just created/cold boot/fast boot - no need to query most default value attributes.
It depends on the context saidiscovery used for, so user should pass a flag.

syncd/SaiDiscovery.cpp
stepanblyschak: discover(rid) changes log level. I find it somewhat slow (in comparison to all the gets) when called in a loop.
Changing log level in a loop is an unusual use case to optimize sai driver for, therefore prefer discover multiple at once. Multi rids version can also take adventage of bulk SAI calls if neccessary.

syncd/SaiSwitchInterface.h
stepanblyschak: Done

syncd/VirtualOidTranslator.cpp
stepanblyschak: I don't pass removedObjects = true in the code, would requires coverage by some code that actually uses it

syncd/RedisClient.cpp
stepanblyschak: (ptr + size) - more generic, static array, dynamic array, ring buffers, etc.
Ideally, std::span - but it's in c++20

syncd/RedisClient.cpp
stepanblyschak: done

lib/OidIndexGenerator.h
```diff
@@ -2,6 +2,9 @@
 
 #include <stdint.h>
 
+#include <saitypes.h> // For SAL annotations
```
kcudnik, line 5: use sai.h and all sai should be included by using extern "C"

kcudnik: or use #include "swss/sal.h"

lib/VirtualObjectIdManager.cpp
```diff
@@ -259,6 +259,55 @@ sai_object_id_t VirtualObjectIdManager::allocateNewObjectId(
     return objectId;
 }
 
+void VirtualObjectIdManager::allocateNewObjectIds(
+        _In_ sai_object_id_t switchId,
+        _In_ sai_object_type_t* objectTypes,
```
kcudnik, line 264: this must be const

lib/VirtualObjectIdManager.cpp
```diff
@@ -259,6 +259,55 @@ sai_object_id_t VirtualObjectIdManager::allocateNewObjectId(
...
+        _Out_ sai_object_id_t* oids,
```
kcudnik, line 265: oids should be lats parameter, and count should be before objectTypes parameter

saiasiccmp/SaiSwitchAsic.h
```diff
@@ -67,9 +67,9 @@ namespace saiasiccmp
 
             virtual std::set<sai_object_id_t> getWarmBootDiscoveredVids() const override;
 
-            virtual void onPostPortCreate(
-                    _In_ sai_object_id_t port_rid,
-                    _In_ sai_object_id_t port_vid) override;
+            virtual void onPostPortsCreate(
+                    _In_ const sai_object_id_t* port_rids,
+                    _In_ size_t count) override;
```
kcudnik, line 72: count should before table pointer

syncd/RedisClient.cpp
```diff
@@ -231,6 +231,32 @@ void RedisClient::setDummyAsicStateObject(
     m_dbAsic->hset(strKey, "NULL", "NULL");
 }
 
+void RedisClient::setDummyAsicStateObjects(
+        _In_ const sai_object_id_t* objectVids,
+        _In_ size_t count)
```
kcudnik, line 236: count should be first, please change that in all files you touched

syncd/SaiDiscovery.cpp
kcudnik: your scope is to port objects, please limit to port

kcudnik (changes requested):
please address comments

syncd/SaiDiscovery.h
stepanblyschak: Put inside class

syncd/SaiDiscovery.cpp
stepanblyschak: Added a check to limit usage to port

lib/OidIndexGenerator.h
stepanblyschak: swss/sal.h is what I was looking for

lib/VirtualObjectIdManager.cpp
stepanblyschak: done

lib/VirtualObjectIdManager.cpp
stepanblyschak: done

syncd/RedisClient.cpp
stepanblyschak: done

stepanblyschak:
FC test passes in my environment on this branch, will restart tests

```
[ RUN      ] FlexCounter.bulkCounter
[       OK ] FlexCounter.bulkCounter (13665 ms)
[ RUN      ] FlexCounter.bulkChunksize
[       OK ] FlexCounter.bulkChunksize (9470 ms)
[ RUN      ] FlexCounter.counterIdChange
[       OK ] FlexCounter.counterIdChange (6302 ms)
[ RUN      ] FlexCounter.addRemoveDashMeterCounter
[       OK ] FlexCounter.addRemoveDashMeterCounter (1224 ms)
[ RUN      ] FlexCounter.noSupportedDashMeterCounter
[       OK ] FlexCounter.noSupportedDashMeterCounter (0 ms)
[ RUN      ] FlexCounter.noEniDashMeterCounter
[       OK ] FlexCounter.noEniDashMeterCounter (0 ms)
[----------] 12 tests from FlexCounter (65320 ms total)
```

stepanblyschak:
@kcudnik Test passed, please review

lib/VirtualObjectIdManager.cpp
kcudnik: still objectType pointer is not const

lib/VirtualObjectIdManager.h
```diff
@@ -67,6 +67,17 @@ namespace sairedis
...
 
+            /**
+             * @brief Allocate multiple object ids on a given switch.
+             *
+             * Throws when object type is switch.
+             */
+            void allocateNewObjectIds(
+                    _In_ sai_object_id_t switchId,
+                    _In_ size_t count,
+                    _In_ sai_object_type_t* objectType,
```
kcudnik, line 78: objectType pointer must be const

syncd/SaiDiscovery.cpp
```diff
@@ -289,6 +297,15 @@ std::set<sai_object_id_t> SaiDiscovery::discover(
 {
     SWSS_LOG_ENTER();
 
+    return discover(&startRid, 1);
+}
+
+std::set<sai_object_id_t> SaiDiscovery::discover(
+        _In_ const sai_object_id_t* rids,
+        _In_ size_t count)
```
kcudnik, line 305: count should be first parameter to keep consistent with all project

syncd/SaiSwitch.cpp
```diff
@@ -980,15 +993,17 @@ void SaiSwitch::redisUpdatePortLaneMap(
...
+        _In_ size_t count,
+        _In_ const sai_object_id_t* port_rids)
 {
     SWSS_LOG_ENTER();
 
-    SaiDiscovery sd(m_vendorSai);
+    SWSS_LOG_TIMER("discovering objects after creating ports");
 
-    auto discovered = sd.discover(port_rid);
+    SaiDiscovery sd(m_vendorSai, SaiDiscovery::Flags::SkipDefaultEmptyAttributes);
```
kcudnik, line 1004: ideally i would like to have this parameter as a command line option to enable

kcudnik: but i guess we can keep it for now until something breaks

syncd/SaiSwitch.cpp
```diff
@@ -1002,29 +1017,19 @@ void SaiSwitch::onPostPortCreate(
...
 
-    for (sai_object_id_t rid: discovered)
-    {
-        /*
-         * We also could thing of optimizing this since it's one call to redis
-         * per rid, and probably this should be ATOMIC.
-         *
-         * NOTE: We are also storing read only object's here, like default
-         * virtual router, CPU, default trap group, etc.
-         */
```
kcudnik, line 1022: is this operation now ATOMIC operation that you skipped this comment ?

syncd/Syncd.cpp
```diff
@@ -4023,6 +4027,23 @@ void Syncd::snoopGetOid(
         return;
     }
 
+    /*
+    * Check if object was previously discovered on this switch, then no need to update ASIC_STATE.
+    */
```
kcudnik, line 4032: comment alignment ident is not consistent

syncd/Syncd.cpp
```diff
@@ -4023,6 +4027,23 @@ void Syncd::snoopGetOid(
...
+        sai_object_id_t rid = SAI_NULL_OBJECT_ID;
+        if (m_translator->tryTranslateVidToRid(vid, rid))
+        {
+            const auto switchVid = VidManager::switchIdQuery(vid);
+            if (m_switches[switchVid]->isDiscoveredRid(rid))
+            {
+                // Already discovered object.
+                return;
+            }
+        }
```
kcudnik, line 4044: add empty line before each if, also since rid is out paremeter, you dont need to initialize it

kcudnik (changes requested):

lib/VirtualObjectIdManager.h
stepanblyschak: my bad, fixed now

syncd/SaiDiscovery.cpp
stepanblyschak: done

syncd/SaiSwitch.cpp
stepanblyschak: Re-added comment. redis pipeline is not atomic

syncd/Syncd.cpp
stepanblyschak: done

bingwang-ms:
@kcudnik Can you help merge this PR if no other concern?

kcudnik (approved):

PR #1566, opened 2025-03-29
Enable zmq mode in syncd for smartswitch DPUs

prabhataravind:
* Enable zmq mode by default if device type in $SYNCD_VARS is SmartSwitchDPU
* Provide a context configuration to syncd with the appropriate context keys
* This goes along with the following changes:
    - https://github.com/sonic-net/sonic-buildimage/pull/21940
    - https://github.com/sonic-net/sonic-swss/pull/3547

#### Why I did it
`To improve the performance between orchagent and syncd for smartswitch DPUs`
##### Work item tracking
- Microsoft ADO **30024522**:

#### How I did it
`By enabling zmq sync communication mechanism between orchagent and syncd for smartswitch DPUs`
#### How to verify it
`Verified by confirming that syncd is started with "-z zmq_sync" with additional context configuration as shown below:`

`root@sonic:/# ps aux | grep orch
root          20  0.5  0.0 1972900 27908 pts/0   Sl   07:00   0:11 /usr/bin/syncd -u -z zmq_sync -x /usr/share/sonic/hwsku/context_config.json -B null -l -p /tmp/sai.profile -w 180000000`


kcudnik:
notifications will not work in OA becasue of this: https://github.com/sonic-net/sonic-swss/issues/1145

prabhataravind:
> notifications will not work in OA becasue of this: [sonic-net/sonic-swss#1145](https://github.com/sonic-net/sonic-swss/issues/1145)

@kcudnik, for dpu, we care only about port state change notification and we plan to forward such notifications to redis channel that portsorch listens to. See corresponding orchagent change here: https://github.com/sonic-net/sonic-swss/pull/3547. This is far from perfect but it works for now. We plan to use this only for dpus. @prsunny for viz.

kcudnik:
All notifications will not work and forwarding it to redisnwhile zmq is used it's workaround and maintainer of swiss should fix this, I do not agree on making workaround on that

prsunny (approved):
Syncd offline with Kamil and using this only for DPU usecase with a temporary work-around in orchagent. Longterm fix to be planned.

kcudnik (approved):

PR #1567, opened 2025-04-01
Move time stamp to time stamp table in counter database to avoid frequently update the counter table

stephenxs:
Move the timestamp out of the COUNTER table to avoid updating the table too frequently and overwhelming the telemetry system.
Fixes https://github.com/sonic-net/sonic-buildimage/issues/22201

### DIFF
--- a/syncd/FlexCounter.cpp
+++ b/syncd/FlexCounter.cpp
@@ -1317,7 +1317,7 @@ private:
         }
 
         // First generate the key, then replace spaces with underscores to avoid issues when Lua plugins handle the timestamp
-        std::string timestamp_key = "PFC_WD_" + m_name + "_time_stamp";
+        std::string timestamp_key = m_instanceId + "_" + m_name + "_time_stamp";
         std::replace(timestamp_key.begin(), timestamp_key.end(), ' ', '_');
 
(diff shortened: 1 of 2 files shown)


syncd/FlexCounter.cpp
```diff
@@ -1303,11 +1303,18 @@ class CounterContext : public BaseCounterContext
             {
                 values.emplace_back(serializeStat(ctx.counter_ids[j]), std::to_string(ctx.counters[i * ctx.counter_ids.size() + j]));
             }
-            values.emplace_back(m_instanceId + "_time_stamp", std::to_string(time_stamp));
```
kcudnik, line 1306: this was previously done in a loop, and now it's done outside loop only once, this logic is different now is that intended?

stephenxs: Yes. All items share the identical timestamp. Previously we write it in each record but it overwhelmed gNMI. Now we move it out of the counter table and use a single record representing it

kcudnik (approved):

kcudnik (approved):

kcudnik (approved):

stephenxs:
Hi @kcudnik 
Could you please merge this PR?
Thanks

PR #1569, opened 2025-04-02
Fix invalid polling of port counters

rick-arista:
In some cases when bulk polling is unsupported the fallback logic to single polling doesn't individually check objects can be polled


r12f (dismissed):

syncd/FlexCounter.cpp
```diff
@@ -929,15 +929,21 @@ class CounterContext : public BaseCounterContext
...
-                    auto it_vid = m_objectIdsMap.find(vids[i]);
-                    if (it_vid != m_objectIdsMap.end())
-                    {
-                        // Remove and re-add if vid already exists
-                        m_objectIdsMap.erase(it_vid);
-                    }
+                    auto rid = rids[i];
+                    auto vid = vids[i];
+                    std::vector<uint64_t> stats(counter_ids.size());
+                    if (collectData(rid, counter_ids, effective_stats_mode, false, stats)) {
```
stephenxs, line 935: Hi
Just to confirm.
We have ruled out single-call-unsupported counters by object type in line [888-895](https://github.com/sonic-net/sonic-sairedis/pull/1569/files#diff-6b568ec42486b69121856646d4bc61b4a9083d12227a5420a1b3b1da93ef24fbR888-L895), which means all the counters in counter_ids should be supported for single call by the object type.
But we still need to try calling `collectData` for all the objects in the RIDs/VIDs array to rule out the objects that do not support the counters. Does this mean that for that object type, some counters support single polls, and some don't?
Thanks

syncd/FlexCounter.cpp
```diff
@@ -929,15 +929,21 @@ class CounterContext : public BaseCounterContext
...
+                        auto it_vid = m_objectIdsMap.find(vid);
+                        if (it_vid != m_objectIdsMap.end())
+                        {
+                            // Remove and re-add if vid already exists
+                            m_objectIdsMap.erase(it_vid);
+                        }
+
+                        auto counter_data = std::make_shared<CounterIds<StatType>>(rid, counter_ids);
+                        m_objectIdsMap.emplace(vid, counter_data);
+                    }
```
stephenxs, line 946: You may want to move the log message in line [949](https://github.com/sonic-net/sonic-sairedis/pull/1569/files#diff-6b568ec42486b69121856646d4bc61b4a9083d12227a5420a1b3b1da93ef24fbL943) inside the `if (collectData(rid, counter_ids, effective_stats_mode, false, stats))` branch and also add message indicating failure of `collectData`
Otherwise, it is misleading if it reports `fallback to single call` but it doesn't insert the vid to `m_objectIdsMap`.

rick-arista: Agreed, fixed.

syncd/FlexCounter.cpp
rick-arista: My understanding is that the check on lines 888-895 use `rid[0]` as a reference and assumes all other RID/VID objects are homogeneous. On some SKUs I'm seeing continuous single poll failures for a small number of rids. These objects are just being checked and skipped here.

A more granular fix might be to iterate the rids and group them by their set of supported counters, then continue with the bulk/single poll logic.

stephenxs (approved):

r12f (approved):

kcudnik (approved):

PR #1582, opened 2025-04-17
[Meta]: Fix type check for TAM object in meta_sai_on_tam_tel_type_config_change

Pterosaur:


kcudnik (approved):

kcudnik:
can we also add unittests for this if they dont exist yet?

PR #1583, opened 2025-04-18
Add support for ha event notifications

zjswhhh:
Adding sairedis and syncd support to process and send back ha set and ha scope events.

Sign-off: Jing Zhang zhangjing@microsoft.com

### DIFF
--- a/meta/Meta.cpp
+++ b/meta/Meta.cpp
@@ -6946,7 +6946,7 @@ void Meta::meta_sai_on_ha_set_event_single(
 
     bool valid = false;
 
-    switch (ot)
+    switch ((int)ot)
     {
         // TODO hardcoded types, must advance SAI repository commit to get metadata for this
         case SAI_OBJECT_TYPE_HA_SET:
@@ -7005,7 +7005,7 @@ void Meta::meta_sai_on_ha_scope_event_single(
 
     bool valid = false;
 
-    switch (ot)
+    switch ((int)ot)
     {
         // TODO hardcoded types, must advance SAI repository commit to get metadata for this
         case SAI_OBJECT_TYPE_HA_SCOPE:
--- a/meta/Meta.cpp
+++ b/meta/Meta.cpp
@@ -6938,20 +6938,28 @@ void Meta::meta_sai_on_bfd_session_state_change(
 }
 
 void Meta::meta_sai_on_ha_set_event_single(
-        _In_ const sai_ha_set_event_data_t& data)
+    _In_ const sai_ha_set_event_data_t& data)
 {
     SWSS_LOG_ENTER();
 
     auto ot = objectTypeQuery(data.ha_set_id);
 
-    bool valid = isHaSetObjectIdValid(ot);
+    bool valid = false;
 
-    if (!valid)
+    switch (ot)
     {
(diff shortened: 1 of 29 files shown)


meta/Meta.cpp
```diff
@@ -6937,6 +6937,124 @@ void Meta::meta_sai_on_bfd_session_state_change(
...
+{
+    SWSS_LOG_ENTER();
+
+    auto ot = objectTypeQuery(data.ha_set_id);
+
+    bool valid = false;
+
+    switch (ot)
+    {
+        // TODO hardcoded types, must advance SAI repository commit to get metadata for this
```
kcudnik, line 6951: you already have this commit, metadata can be used to support that, check other notifications implementation in this file how to do this

kcudnik (approved):

meta/Meta.cpp
zjswhhh: Hi @kcudnik - I checked the latest libsaimetadata on the current head, but didn't see the metadata related to HA notifications? Is it expected?

kcudnik: i was thinking about similar to this: https://github.com/sonic-net/sonic-sairedis/blob/master/meta/Meta.cpp#L6768
please follow https://github.com/sonic-net/sonic-sairedis/blob/master/meta/Meta.cpp#L7232

zjswhhh: Hi @kcudnik - yes I did with commit https://github.com/sonic-net/sonic-sairedis/pull/1583/commits/7dbe8e589034d749283a33bdd0eea1fe04969d08

But the metadata is missing, hence the build couldn't pass.

zjswhhh: Hi @kcudnik - even if I revert the commit: 
image

The ha types are in the extension headers, is that why? 
https://github.com/opencomputeproject/SAI/blob/2f583d4c5477db13a3680cf2e8e622971d8f4fe2/experimental/saitypesextensions.h#L73

kcudnik:
here is  the problem:
```
Meta.cpp:6952:9: error: case value '536870929' not in enumerated type '_sai_object_type_t' [-Werror=switch]
 6952 |         case SAI_OBJECT_TYPE_HA_SET:
      |         ^~~~
Meta.cpp: In member function 'void saimeta::Meta::meta_sai_on_ha_scope_event_single(const sai_ha_scope_event_data_t&)':
Meta.cpp:7011:9: error: case value '536870930' not in enumerated type '_sai_object_type_t' [-Werror=switch]
 7011 |         case SAI_OBJECT_TYPE_HA_SCOPE:
      |         ^~~~
```

since switch is on enum object, you will need to make it switch((int)xxx), since experimental object types are not actually part of sai_object_type_t

meta/Meta.cpp
kcudnik: You need to cast to int before switch

zjswhhh: Hi @kcudnik  - done already. Please check on the latest commit.

kcudnik (dismissed):

zjswhhh:
Hi @kcudnik - help merge please?

kcudnik:
code coverage shows 4% covered?

zjswhhh:
Hi @kcudnik  - the coverage shows 100% now. Please help sign-off and merge. Great thanks!

kcudnik (approved):

PR #1603, opened 2025-05-22
[az] Use ubuntu-latest in vmImage

kcudnik:
Image ubuntu-20.04 is no longer available


kcudnik:
fixed here: https://github.com/sonic-net/sonic-sairedis/pull/1605#event-17860011582

PR #1604, opened 2025-05-26
Setup tc redirect for Loopback interface

AkeelAli:
Previously, the loopback interface would be deleted and recreated as a tap interface with the VPP linux-cp plugin. During this time, route installation may fail and result in some routes stuck in queued state. This affected a number of testcases that expect the routes to be installed. The failures may appear intermittent and the tests unstable depending on which routes were not successfully programmed. One way to verify if you encountered this problem is if this command's output is not empty, showing some routes are stuck in queue: `vtysh  -c "show ip route" | grep q`.

In this PR, I removed the old way of recreating the loopback and instead used a tc redirect from the tap to the original loopback. I no longer see the queued routes with this change. I did a few config reloads, re-launched the topo too, and the routes appear to be successfully installed each time. The sonic-mgmt sanity suite was also run a number of times with this change and showed improvement in the stability of some tests (including bgp and acl).

### DIFF
--- a/meta/Meta.cpp
+++ b/meta/Meta.cpp
@@ -14,7 +14,7 @@
 
 // TODO add validation for all oids belong to the same switch
 
-#define MAX_LIST_COUNT 0x1000
+#define MAX_LIST_COUNT (0x1<<24) // 16M
 
 #define CHECK_STATUS_SUCCESS(s) { if ((s) != SAI_STATUS_SUCCESS) return (s); }
 
--- a/meta/NotificationTamTelTypeConfigChange.cpp
+++ b/meta/NotificationTamTelTypeConfigChange.cpp
@@ -26,7 +26,7 @@ sai_object_id_t NotificationTamTelTypeConfigChange::getAnyObjectId() const
 {
     SWSS_LOG_ENTER();
 
-    return SAI_NULL_OBJECT_ID;
+    return m_tam_id;
 }
 
 void NotificationTamTelTypeConfigChange::processMetadata(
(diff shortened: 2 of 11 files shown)


vslib/vpp/SwitchVppHostif.cpp
```diff
@@ -103,6 +104,36 @@ int SwitchVpp::promisc(
...
+        _In_ const std::string& tap,
+        _In_ const std::string& hostIfname)
+{
+    SWSS_LOG_ENTER();
+
+    int ret;
+    std::stringstream cmd;
+    std::string res;
+
+    cmd << "tc qdisc add dev " << tap << " ingress";
```
yue-fred-gao, line 117: what will happen if has been done before? For example, if we restart syncd, it looks like we don't remove the tc filter. Once syncd is restarted, will it add tc filter again?

vslib/vpp/SwitchVppRif.cpp
```diff
@@ -1256,32 +1228,20 @@ sai_status_t SwitchVpp::vpp_add_lpb_intf_ip_addr (
...
-        std::string addr = prefix.getIp().to_string();
-        ret = configureLoopbackInterface(true/*lpb_add*/, hostIfname, addr, prefix.getMaskLength());
-        if (ret != 0)
-        {
-            SWSS_LOG_ERROR("Failed to configure loopback interface add");
-        }
-        SWSS_LOG_DEBUG("configure_lcp_interface vpp_name:%s sonic_name:%s prefix:%s",
-            vppIfName.c_str(), hostIfname.c_str(), prefix.to_string().c_str());
+        // add tc filter to redirect traffic from tap to Loopback
+        CHECK_STATUS(add_tc_filter_redirect(tap, hostIfname));
```
yue-fred-gao, line 1244: Do we need to remove tc filter when the loopback interface is deleted?

yue-fred-gao (approved):

yue-fred-gao (approved):

kcudnik (approved):

PR #1605, opened 2025-05-27
Fix PR testing issue caused by deprecation of Ubuntu 20.04

wangxin:
After Ubuntu 20.04 was deprecated, it caused multiple failures of PR testing:
1. Any job requires Ubuntu 20.04 image will fail. No agent can be allocated to run Ubuntu 20.04. This change updated the pipelines to use Ubuntu 22.04.
2. Dependent upstream pipelines have been updated to use Ubuntu 22.04. Some build/test code are not updated accordingly, so downloading artifacts from upstream pipelines failed. This change fixed the upstream pipeline name and artifacts name.

This change also deprecated a parameter `sonic_buildimage_ubuntu20_04` which is not really necessary.

The agent pool for running VStest was upgraded to Ubuntu 22.04 too. The step for installing dependent packages need to be updated accordingly.

### DIFF
--- a/.azure-pipelines/test-docker-sonic-vs-template.yml
+++ b/.azure-pipelines/test-docker-sonic-vs-template.yml
@@ -67,17 +67,31 @@ jobs:
 
   - script: |
       set -ex
+      # install packages for vs test
+      sudo pip3 install pytest flaky exabgp docker redis
+
+      # install packages for kvm test
+      sudo apt-get update
+      sudo apt-get -o DPkg::Lock::Timeout=600 install -y libvirt-clients \
+          qemu \
+          openvswitch-switch \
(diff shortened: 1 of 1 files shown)


kcudnik:
i have pr for this: https://github.com/sonic-net/sonic-sairedis/pull/1603

kcudnik:
but seems like your have also updated images i missed

prabhataravind (changes requested):
libyang0.16 is not available in 22.04. This needs to be replaced with libyang-dev instead.

wangxin:
> libyang0.16 is not available in 22.04. This needs to be replaced with libyang-dev instead.

Fixed.

PR #1613, opened 2025-06-13
[meta] do not fail bulk operations if MODE_IGNORE_ERROR

Ndancejic:
bulk operations do not continue when SAI_BULK_OP_ERROR_MODE_IGNORE_ERROR is set and an error occurs. Create new check that considers the mode when validating bulk operations.

fixes a crash in https://github.com/sonic-net/sonic-buildimage/issues/21243 because route entries that are not removed in syncd by bulker are assumed to be removed in orchagent. Then when the nexthops are attempting to be removed, orchagent crashes due to OBJECT_IN_USE.

### DIFF
--- a/meta/Meta.cpp
+++ b/meta/Meta.cpp
@@ -18,6 +18,11 @@
 
 #define CHECK_STATUS_SUCCESS(s) { if ((s) != SAI_STATUS_SUCCESS) return (s); }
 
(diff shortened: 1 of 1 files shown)


Ndancejic:
Hi @kcudnik, if you approve of the changes can you help merge?

bingwang-ms:
@Ndancejic Can you raise a PR to 202411?

PR #1632, opened 2025-07-24
Add missing <array> include to TestSyncdBrcm.cpp

saiarcot895:
With the future base image upgrade to Trixie and changes to the Bookworm build, the <array> header is not indirectly included anymore. Explicitly add it here, since std:array is used in this file.


Gfrom2016 (approved):
lgtm

PR #1641, opened 2025-08-11
Skip setting SAI_TUNNEL_ATTR_ENCAP_TTL_MODE on comparisonLogic

Ryangwaite:
This is a draft. I'll decline this and raise a PR for master once i'm happy with the implementation and that it is fixing the bug.

### DIFF
--- a/tests/aspell.en.pws
+++ b/tests/aspell.en.pws
@@ -21,6 +21,7 @@ CHARDATA
 COLDVIDS
 CONST
 CRM
+CSP
 CreateObject
 DEI
 Decrement


syncd/Workaround.cpp
```diff
@@ -70,6 +71,15 @@ bool Workaround::isSetAttributeWorkaround(
         return true;
     }
 
+    if (doingComparisonLogic) {
+        if (objectType == SAI_OBJECT_TYPE_TUNNEL && attrId == SAI_TUNNEL_ATTR_ENCAP_TTL_MODE) {
+            SWSS_LOG_WARN("setting %s failed: %s, not all platforms support this attribute when doing comparison logic of 202411->202505 warm upgrade",
```
vaibhavhd, line 76: Should you be checking for if this upgrade is cross branch? Your logic will prevent these SET operations irrespective of the upgrade path. But your log here assumes a different intention.

vaibhavhd:
I think this change should go to master, and then cherry picked to 202505. This should stay until the setting can be handlded by SAI/SDK implementation by the chip vendors who take comparison-logic path.

PR #1657, opened 2025-09-18
[vpp] Fix error reply handlers of vpp api

super-jiying:
What I did
Fix for #1656

How I did it
Modify reply handlers of "WANT_INTERFACE_EVENTS_REPLY", "WANT_BFD_EVENTS_REPLY", and "BFD_UDP_ENABLE_MULTIHOP_REPLY" in the SaiVppXlate.c file.


yue-fred-gao:
The change looks good. Can you please share what test have you done?

PR #1661, opened 2025-10-07
Fix syncd crash due to attr obj list not initialized

lolyu:
## Issue description
SYNCD crash is observed in warm-reboot in swss-update PR build, the crash is due to the `SAI_SWITCH_ATTR_TAM_OBJECT_ID` attribute object list is not properly initialized in `warm_update_switch`.

## Root cause
The swss update includes PR [[orchagent]: HFTOrch init](https://github.com/sonic-net/sonic-swss/pull/3759/files#top), which creates the SAI TAM object.

## Fix
Let's initialize the object list explicitly.


prsunny:
@Pterosaur , @zjswhhh, @mramezani95 for viz

Issue #1667, opened 2025-10-17
[vslib] MACsec interface creation fails when send_sci=false


vikram-nexthop:
### Bug Description
MACsec interface creation commands fails on VM/VS with send_sci=false and SCI combination. 

```
admin@VM:~$ sudo ip link add link "eth0" name "macsec_eth1" type macsec  sci 226b54b065000001 encrypt  on  send_sci off cipher GCM-AES-128
RTNETLINK answers: Invalid argument
```

#### Steps to Reproduce
Configure MACsec with send_sci=false

#### Expected Behavior
MACsec interface should be created successfully regardless of send_sci setting.

### Root Cause
The current implementation always includes SCI and send_sci parameters in ip link add command.
Linux kernel rejects SCI with send_sci=false combination with invalid argument error [drivers/net/macsec.c](https://github.com/torvalds/linux/blob/7ea30958b3054f5e488fa0b33c352723f7ab3a2a/drivers/net/macsec.c#L4098).

#### Code Location
[sonic-sairedis/vslib/MACsecManager.cpp - create_macsec_egress_sc() function](https://github.com/sonic-net/sonic-sairedis/blob/master/vslib/MACsecManager.cpp#L384)

### Proposed Fix
Implement two-step approach:

1. Create interface with send_sci=on to set SCI.
2. Switch to send_sci=off if required.

### Impact
Blocks MACsec deployment in scenarios requiring SCI transmission disabled
Affects VM/VS testing environments

### Related Issue(s)
https://github.com/sonic-net/sonic-buildimage/issues/12991

PR #1668, opened 2025-10-17
[vslib] MACsec interface creation command fails on VM/VS with send_sci=false and SCI combination.

vikram-nexthop:
### What I did
Fix for #1667 

### How I did it
Modified MACsec interface creation logic to use a two-step approach:
1. Always create the MACsec interface with explicit SCI specification (default send_sci=on)
2. When send_sci=false is required, add a second command to switch the interface to send_sci=off mode.

### How to verify it
1. Run MACsec tests with profile parameted send_sci set to false.
2. Verify interface creation commands in syslog:
```bash
2025 Oct 15 16:26:50.911039 VM01T3 NOTICE syncd#syncd: :- create_macsec_egress_sc: /sbin/ip link add link "eth1" name "macsec_eth1" type macsec  sci 226b54b065000001 encrypt  on  cipher GCM-AES-128 && ip link set link "eth1" name "macsec_eth1" type macsec  send_sci off && ip link set dev "macsec_eth1" up
```
3. Check that TX SCI and send_sci are correctly set using command 'ip macsec show'
```bash
admin@VM01T3:~$ ip macsec show
83: macsec_eth1: protect on validate strict sc off sa off encrypt on send_sci off end_station off scb off replay off 
    cipher suite: GCM-AES-256, using ICV length 16
    TXSC: 226b54b065000001 on SA 0
        0: PN 1534, state on, key 8912df4fd8ea3f1547e02a4410f89996
    RXSC: 5c0758de98020001, state on
        0: PN 1597, state on, key 8912df4fd8ea3f1547e02a4410f89996
    offload: off 
```

### DIFF
--- a/vslib/vpp/SwitchVpp.cpp
+++ b/vslib/vpp/SwitchVpp.cpp
@@ -1234,6 +1234,49 @@ sai_status_t SwitchVpp::setPort(
     return set_internal(SAI_OBJECT_TYPE_PORT, sid, attr);
 }
 
+sai_status_t SwitchVpp::setLag(
+        _In_ sai_object_id_t lagId,
+        _In_ const sai_attribute_t* lagAttr)
+{
+    SWSS_LOG_ENTER();
+
+    auto attr_type = sai_metadata_get_attr_by_id(SAI_LAG_ATTR_INGRESS_ACL, 1, lagAttr);
+
+    if (attr_type != NULL)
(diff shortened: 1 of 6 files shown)


Pterosaur (dismissed):
lgtm

vikram-nexthop:
requesting approval from the maintainer.

PR #1669, opened 2025-10-18
[DASH] Fix FlexCounter bulk get DASH ENI meter class range

mukeshmv:
The Flexcounter periodic poll for the DASH bulk meter stats uses wrong meter class range (0 to max-1). The correct meter class range is (1 to max) since class 0 denotes unmetered traffic. This PR fixes the range in the Flexcounter bulk stats request.


KrisNey-MSFT:
@prabhataravind to assign reviewer if possible

prabhataravind (approved):

PR #1671, opened 2025-10-22
[202505] Revert "[meta] do not fail bulk operations if MODE_IGNORE_ERROR (#1647)"

Ndancejic:
This reverts commit 8645610df0962f5ad78aca5dbd4adae9a503e2a5.

Reverts bulker mode flag logic. This causes a crash in duplicate route test due to the processing of bulk creates.


prsunny:
@yejianquan @StormLiangMS for viz. This was the offending PR for 202505 crash which is reverted.

PR #1675, opened 2025-10-28
Reduce severity of log for unallowed enum value

croos12:
Reduces the warning for !sai_metadata_is_allowed_enum_value from ERR to WARN to support SAI custom header extensions.

Addresses https://github.com/sonic-net/sonic-buildimage/issues/23467


dgsudharsan (approved):

dgsudharsan:
@lolyu @kcudnik  Can you please review?

PR #1676, opened 2025-10-28
Revert "[meta] do not fail bulk operations if MODE_IGNORE_ERROR (#1613)"

Ndancejic:
This commit caused failures in test_duplicate_route.

This reverts commit fb5652bd59c4f2df1ef9106728a9776bbb40a83a.


lolyu (commented):
what's the plan for the crm failure after reverting this PR?

PR #1680, opened 2025-10-31
vpp: Handle routes using link local nexthop

yue-fred-gao:
### why
When we add a route, the current code doesn't provide interface to vl_api_ip_route_add_del but only provides nexthop address. It works with if the nexthop address is unique. However, in link-local case, the nexthop address is not unique and only valid with the egress interface. vpp fails to find the correct nexthop and interface.

### what this PR does
SAI provides SAI_NEXT_HOP_ATTR_ROUTER_INTERFACE_ID. We can get the egress interface from this attribute and provide it to vpp in the vl_api_ip_route_add_del. This removes the ambiguity in the link-local case.


dypet (approved):
LGTM

PR #1685, opened 2025-11-04
update sai version to 17.1

lakshmi-nexthop:


lakshmi-nexthop:
@tjchadaga I added some UT to improve coverage. But there are 3 issues with this PR. 
1) Currently there is no sai attr with value type of SAI_ATTR_VALUE_TYPE_PRBS_BIT_ERROR_RATE. Thats why meta/SaiSerialize.cpp:537 is not covered as part of UT
2) Currently there are no sai attrs with type SAI_ATTR_VALUE_TYPE_UINT16_RANGE that are writable. the existing sai attrs are all read only. And hence there is no coverage for this value type in meta/Meta.cpp:3705, 4337
3) Flex counter UT test is failing in upstream, I am not able to repro this locally. Not sure if it is timing related.

tjchadaga (dismissed):

tjchadaga (approved):

lakshmi-nexthop:
> @lakshmi-nexthop can you manually raise a PR with these changes targeting the 202511 branch?

ack. will do today.

PR #1693, opened 2025-11-12
[alpinevs] Add config option for Alpine

sreeiyer-1:
For the Alpine platform, adding a case for 'alpinevs' in the config script

HLD - https://github.com/sonic-net/SONiC/blob/master/doc/alpine/alpine_hld.md
Repo - https://github.com/sonic-net/sonic-alpine

### DIFF
--- a/syncd/FlexCounter.cpp
+++ b/syncd/FlexCounter.cpp
@@ -206,6 +206,7 @@ struct BulkStatsContext
     std::vector<uint64_t> counters;
     std::string name;
     uint32_t default_bulk_chunk_size;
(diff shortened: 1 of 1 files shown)


sreeiyer-1:
Hi @yue-fred-gao , the pipeline build is now passing after the environment is fixed. semgrep waits indefinitely but that is apparently an issue that can happen for the first submit  that one makes in a repo (something to do with semgrep not able to find a base view to get the diff). In any case, semgrep analyses only the source code, and ignores scripts, and the only change here is in a script. If this is okay, can you please merge this PR.

PR #1697, opened 2025-11-15
[lib] Make ZMQ buffer size adjustable and set it to 4MB by default

croos12:
Adds zmqResponseBufferSize variables to ZeroMQ Channel classes to reduce the size of ZeroMQ Servers in memory. Only ServerSai <-> Syncd channel will use 64 MB ZMQ Buffer. All other buffers reduced back to 4 MB to be the same as before commit [8123a9820cc330675147b5a99cfd81ea8df39271](https://github.com/sonic-net/sonic-sairedis/commit/8123a9820cc330675147b5a99cfd81ea8df39271)


croos12:
@lolyu the issue with setting the buffer to 64 MB is that not every service needs 64MB. This doubles the steady-state memory consumption of orchagent (which does not need that much memory)

PR #1698, opened 2025-11-17
Fix issues to support custom SAI header

Junchao-Mellanox:
**Why I did this**

SAI starts to support customer SAI header a few months ago. During integration with custom SAI header, I found some issues. This PR is to fix them. The PR also add a custom flag to VID and RID.

**How I did this**

1. sonic-sairedis should compile with custom headers
2. sonic vs sai should install custom headers to /usr/bin/sai
3. add a custom flag in VID and RID


Junchao-Mellanox:
Hi @lguohan , could you please suggest who may help review this PR? Thanks!

lguohan:
what is the use case for custom SAI header? can you help to elaborate more?

PR #1710, opened 2025-12-02
[ci] Use 1ES agent pool to provide more disk when building swss.

liushilongbuaa:
Build swss needs more disk space. Default agent pool only has 4 GB disk in slave container.
Use self customized pool instead.


Pterosaur:
> @Pterosaur , do you know these test error?

I'm not aware of that. 
@lolyu @Gfrom2016 could you please help to take a look?

prsunny (approved):

PR #1723, opened 2025-12-10
Fix deadlock between syncd and orchagent syncd during initialization failure

DavidZagury:
When syncd requests a shutdown, orchagent may be blocked waiting for a response to an init view (or other NOTIFY) command. Since syncd stops processing commands while waiting for the shutdown response, orchagent never receives its response and cannot acknowledge the shutdown request - resulting in a deadlock.

This fix adds the selectable channel to the select loop during shutdown-wait mode and handles incoming commands appropriately:
- NOTIFY commands receive a SAI_STATUS_FAILURE response to unblock the waiting orchagent
- Other commands are logged and ignored

This prevents orchagent from hanging until timout when syncd is failing.

This fix - https://github.com/sonic-net/sonic-buildimage/issues/24799

### DIFF
--- a/.azure-pipelines/build-swss-template.yml
+++ b/.azure-pipelines/build-swss-template.yml
@@ -18,9 +18,6 @@ parameters:
   type: number
   default: 60
 
-- name: sonic_slave
-  type: string
-
 - name: debian_version
   type: string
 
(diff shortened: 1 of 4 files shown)


r12f:
Requested @prsunny to review. Looks to be a very nice (better) to have fix.

prabhataravind:
@DavidZagury lgtm. Please address copilot comments or close them if not applicable.

DavidZagury:
> @DavidZagury lgtm. Please address copilot comments or close them if not applicable.

@prabhataravind I addressed the copilot comments.

PR #1732, opened 2025-12-19
vpp: support binding multiple ACL tables by priority

yue-fred-gao:
### why
currently vpp doesn't support binding multiple ACL tables. Each table is appended with default permit-all rules. With multiple tables, this may cause acl matched by such rules and skip the actual rule to make in the tables after this one.

### what this PR does
1. remove the default permit-all rules for each table
2. If a table is empty, create a dummy rule that won't match any traffic because vpp doesn't allow empty table. The dummy rule matches dest-ip to 0.0.0.0/32
3. sort all the tables by priority in the table group. vpp doesn't support parallel matching
4. added catch-all acl group to the end. vpp default behavior of no match is drop but sonic is accept.
5. Fix sonic-vpp crashing due to race condition during stats pull. If the interface to get stats has been removed, stat_segment_ls_r returns null.

### DIFF
--- a/vslib/vpp/SwitchVpp.h
+++ b/vslib/vpp/SwitchVpp.h
@@ -622,6 +622,9 @@ namespace saivs
             std::map<sai_object_id_t, std::list<sai_object_id_t>> m_acl_tbl_grp_ports_map;
             std::map<sai_object_id_t, vpp_ace_cntr_info_t> m_ace_cntr_info_map;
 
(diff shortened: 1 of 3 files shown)


dypet (approved):
LGTM

AkeelAli (approved):
I would just add a comment in the PR description about the changes in SaiVppStats.c.

lguohan (approved):

PR #1743, opened 2026-01-17
Use matching slave container from the branch we're building against

saiarcot895:


tjchadaga (approved):

vmittal-msft:
@saiarcot895 we don't need this for 202511 ? if so, please remove these labels

saiarcot895:
Included as part of #1746.

PR #1450, opened 2024-11-04
Do not enter vendor SAI critical section for counter polling/clearing operations

stephenxs:
**What I did**

Do not enter the vendor SAI critical section for counter-polling/-clearing operations.

**Why I did it**

There are two types of threads in syncd:
1. The syncd main thread in which the create/destroy/set/get SAI APIs are called
2. Flex counter threads in which counter-polling/-clearing APIs are called

The critical section in vendor SAI was introduced to protect vendors' SAI from being re-entered, which prevents the flex counter threads from running concurrently, and introduces latency for one flex counter thread when it's waiting for the critical section.
It is not necessary to enter a section in counter-polling/clearing operations since the objects' state won't be changed in that API.

**How I verified it**

Run regression test and observe counter-polling performance.

**Details if related**

PR #1498, opened 2025-01-13
Revert "Do not enter vendor SAI critical section for counter polling/clearing operations (#1450)"

stephenxs:
Revert "Do not enter vendor SAI critical section for counter polling/clearing operations (#1450)"

This reverts commit 0317b16e4fdccf769c77a7b164ff90c1365c7d4e.

PR #1748, opened 2026-01-21
Do not enter vendor SAI critical section for counter polling/clearing operations

croos12:
There are two types of threads in syncd:

The syncd main thread in which the create/destroy/set/get SAI APIs are called Flex counter threads in which counter-polling/-clearing APIs are called The critical section in vendor SAI was introduced to protect vendors' SAI from being re-entered, which prevents the flex counter threads from running concurrently, and introduces latency for one flex counter thread when it's waiting for the critical section. It is not necessary to enter a section in counter-polling/clearing operations since the objects' state won't be changed in that API.

Reapplies commit from https://github.com/sonic-net/sonic-sairedis/pull/1450


dgsudharsan (approved):

stephenxs (approved):

tjchadaga:
@croos12 , @dgsudharsan - this change was previously reverted due to this - https://github.com/sonic-net/sonic-sairedis/pull/1498#issuecomment-2589184116.  Could you please help add an explanation of why this is not an issue anymore? Please also include what testing has been done

dgsudharsan:
@tjchadaga We had an issue with our SAI library implementation that threw errors. This was fixed with our SAI library that is currently present in master branch. We have ran the entire suite of sonic-mgmt tests to ensure that no errors are seen across the tests

prsunny (approved):

tjchadaga (approved):

PR #1774, opened 2026-02-19
[master] Smart Counter Poll to allow counters to work properly on Broadcom platforms.

justin-wong-ce:
#### Summary
This is change is to address https://github.com/sonic-net/sonic-sairedis/issues/1753 and the master implementation of the HLD added in https://github.com/sonic-net/SONiC/pull/2190.

`FlexCounter.cpp` assumes all ports support the same counter capabilities. This causes issues on most Broadcom platform switches as there are different types of ports on a switch that does not support the same set of counters.

Fix by dynamically discovering what each interface is capable of during initialization of `syncd`.

For more detials please refer to the above issue and HLD.

#### Testing
Master testing is also done, but due to frequent changes in master tests fail on other issues depending which commit is used for testing. However, none of the issues seem related to this change.

More comprehensive testing is done on the 202511 stable branch:
Testing is done on both a `Arista-7060X6-16PE-384C-B-O128S2` and `Arista-7260CX3-D108C8` on 202511 with the tests:
```
sonic-mgmt/tests/dhcp_relay/test_dhcp_counter_stress.py
sonic-mgmt/tests/drop_packets/test_drop_counters.py
sonic-mgmt/tests/drop_packets/test_configurable_drop_counters.py
sonic-mgmt/tests/gnmi/test_gnmi_countersdb.py
sonic-mgmt/tests/snmp/test_snmp_queue_counters.py
```

A full `sonic-mgmt` test suite run has also been ran (202511). There are no notable fallout compared to `sonic-mgmt` runs without this change. This is ran on both XGS and DNX Broadcom platforms.

**EDIT: Jul 13, 2026**
**202605 Testing**
Additional 202605 testing has been done.
Comparing full 202605 `sonic-mgmt` test results on TH5 (using `Arista-7060X6-64PE`) with and without these changes, there is no difference in the test results.

#### Performance Impact:
All logic change is only done in the counter initialisation stage of FlexCounters.cpp - there is no polling logic change at all. 
**Therefore, any performance impact is limited to any new execution of `syncd` - i.e. reboot / `config reload` / `systemctl restart`.**

Tested on several topologies:
image

This impact seems reasonable for what this offers. 

The fastest operation that would cause a `syncd` restart is a `systemctl restart swss`, that is a operation that spans minutes. The worst real life scenario on a HwSKU with older CPU and many interfaces takes additional ~17 seconds. New, high interface-count HwSKUs only takes an extra ~2 seconds.

As for memory usage - the impact is on the order of kilobytes, the impact is negligible as the system has GBs of RAM.


justin-wong-ce:
WIP - adding unit test

justin-wong-ce:
CI failing on dependency installation step:
```
Install dependencies
...
+ sudo rsyslogd
rsyslogd: error during config processing: omfile: chown for file '/var/log/auth.log' failed: Operation not permitted [v8.2302.0 try https://www.rsyslog.com/e/2207 ]
+ cat /etc/apt/sources.list
cat: /etc/apt/sources.list: Permission denied
```

justin-wong-ce:
CI failing on a different setup error in `Install sonic platform-vpp packages`:
```
dpkg: error processing package libvppinfra (--install):
 dependency problems - leaving unconfigured
Processing triggers for libc-bin (2.36-9+deb12u14) ...
Errors were encountered while processing:
 libvppinfra

```

syncd/FlexCounter.cpp
```diff
@@ -1119,6 +1462,45 @@ class CounterContext : public BaseCounterContext
...
+        if (!referenced)
+        {
+            // Remove from sorted list so updateSupportedCounterGroups won't match
+            m_counterGroupsSorted.erase(
+                std::remove_if(m_counterGroupsSorted.begin(), m_counterGroupsSorted.end(),
+                    [removedGroupIdx](const CounterGroupRef& ref) { return ref.idx == removedGroupIdx; }),
+                m_counterGroupsSorted.end());
+
+            // Clear the group data (can't erase from vector without invalidating other indices)
+            m_supportedCounterGroups[removedGroupIdx].clear();
```
lolyu, line 1498: this is a per-cycle leak. so across repeated churn, the vector `m_supportedCounterGroups` accumulates empty-set tombstomes forever.

We can keep the freed group indices into a vector, and reuse those freed group indices first when creating groups.

syncd/FlexCounter.cpp
```diff
@@ -1159,9 +1543,24 @@ class CounterContext : public BaseCounterContext
                                         kv.second->getStatsMode() == SAI_STATS_MODE_READ_AND_CLEAR) ? SAI_STATS_MODE_READ_AND_CLEAR : SAI_STATS_MODE_READ;
             }
 
-            std::vector<uint64_t> stats(statIds.size());
+            std::vector<uint64_t> stats(statIds.size(), 0);
             if (!collectData(rid, statIds, effective_stats_mode, true, stats))
             {
+                if (m_failedPolls.find({rid, vid}) == m_failedPolls.end())
```
lolyu: didn't see `m_failedPolls` cleanup on the object removal

syncd/FlexCounter.cpp
```diff
@@ -1539,6 +2008,191 @@ class CounterContext : public BaseCounterContext
...
+                    newCounters.insert(intersectedCounters.begin(), intersectedCounters.end());
+
+                    // If vid already has assigned counter group, merge the two groups if dont_clear flag is set
+                    if (m_objectSupportedCountersGroupMap.count(vid) && dont_clear_support_counter)
+                    {
+                            std::set<StatType> oldGroup = m_supportedCounterGroups[m_objectSupportedCountersGroupMap[vid]];
+                            newCounters.insert(oldGroup.begin(), oldGroup.end());
+                    }
+
+                    m_objectSupportedCountersGroupMap[vid] = m_supportedCounterGroups.size();
```
lolyu: The counter group creation is not consistent across this PR, please add a single helper like the following to formalize the group creation:

```cpp
size_t addGroup(std::set<StatType> s)
{
    size_t idx = m_supportedCounterGroups.size();   // captured ONCE, pre-push — the only truth
    m_supportedCounterGroups.push_back(std::move(s));
    m_counterGroupsSorted.push_back(makeCounterGroupRef(idx, m_supportedCounterGroups[idx].size()));
    sortCounterGroups();                            // the 4 duplicated sort-lambdas, factored out
    return idx;
}
```

syncd/FlexCounter.cpp
lolyu: **N1 — semantics + a latent off-by-one.** Two problems here:

1. **Repeated `operator[]` lookups.** This block looks up `m_failedPolls[{rid, vid}]` separately in the `find`, the `else if` condition, the `+=`, and the log arg — 3-4 hashed lookups (each constructing a `pair` key) per failed poll. And `operator[]` default-inserts, so it's only safe because `find` gates it.

2. **The tier boundaries are off.** With the current `< 3` / `== 3` structure, trace the stored value: it reaches the `== 3` ERROR branch when the stored count is 3 (the *4th* failure), but the DEBUG branch already printed "2 times"/"3 times" for counts 2 and 3 — so the ERROR says "more than 3 times" on the 4th, while "3 times" was logged on the 3rd. The intent is reasonable but the boundary is confusing and fragile.

Both fixed by a single lookup with explicit tiers:
```cpp
auto [it, inserted] = m_failedPolls.try_emplace({rid, vid}, 0);
uint32_t n = ++it->second;
if (n == 1)      SWSS_LOG_DEBUG(..."1 time"...);
else if (n <= 3) SWSS_LOG_DEBUG(..."%d times"..., n);
else if (n == 4) SWSS_LOG_ERROR(..."more than 3 times"...);
```
One lookup, no default-insert footgun, boundary explicit. (Pairs naturally with the `m_failedPolls` cleanup I flagged separately.)

syncd/FlexCounter.cpp
```diff
@@ -1539,6 +2008,191 @@ class CounterContext : public BaseCounterContext
...
+        {
+            if (!always_check_supported_counters)
+            {
+                SWSS_LOG_NOTICE("Ignore checking of supported counters");
+                return;
+            }
+            // If there is already a counter group for this vid, and that counter group contains a counter that is not
+            // included in counter_ids, then it means we are handling a different set of counters.
+            // Remove the previous group and mappings to this vid
+            std::set<StatType> existingGroup = m_supportedCounterGroups[m_objectSupportedCountersGroupMap[vid]];
```
lolyu: **N2 — unnecessary full copy of a `std::set`.** `existingGroup` deep-copies the whole set but is only iterated read-only just below. Should be a const reference:
```cpp
const std::set<StatType>& existingGroup = m_supportedCounterGroups[m_objectSupportedCountersGroupMap[vid]];
```
Same pattern repeats for `oldGroup` (2087), `prevGroup`/`currGroup` (2120/2122) in the merge branches — all are inserted *from*, never mutated, so references suffice. Minor per-site, but it's a repeated convention miss in a hot init path.

syncd/FlexCounter.cpp
```diff
@@ -538,7 +538,21 @@ class CounterContext : public BaseCounterContext
...
+    {
+        size_t idx;
+        size_t size;
+    };
+    CounterGroupRef makeCounterGroupRef(size_t idx, size_t size)
+    {
+        SWSS_LOG_ENTER();
+        return {idx, size};
+    }
+    std::set<StatType>* getSortedCGRef(CounterGroupRef const& cgr)
```
lolyu: **N3 — mutable pointer into internal storage from a logically-const lookup, plus a perf nit.** `getSortedCGRef` returns a non-const `std::set<StatType>*` into `m_supportedCounterGroups`, but the callers only read through it. Handing out a mutable interior pointer from a "get" helper lets a caller rewrite a group's contents and silently break the index/sort invariants. Suggest returning `const std::set<StatType>&` (add a separately-named mutable accessor only if some caller truly needs to mutate).

Also: `SWSS_LOG_ENTER()` in this one-line accessor adds logging overhead on the per-group init loop — inconsistent with the "no SWSS_LOG_ENTER for perf" rationale already used on the sort lambdas. Drop it here.

syncd/FlexCounter.cpp
```diff
@@ -1539,6 +2008,191 @@ class CounterContext : public BaseCounterContext
...
+                SWSS_LOG_NOTICE("Ignore checking of supported counters");
+                return;
+            }
+            // If there is already a counter group for this vid, and that counter group contains a counter that is not
+            // included in counter_ids, then it means we are handling a different set of counters.
+            // Remove the previous group and mappings to this vid
+            std::set<StatType> existingGroup = m_supportedCounterGroups[m_objectSupportedCountersGroupMap[vid]];
+            for (auto &counter : existingGroup)
+            {
+                if (find(counter_ids.begin(), counter_ids.end(), counter) == counter_ids.end())
```
lolyu: **N4 — unqualified `find` relies on ADL and is inconsistent with the rest of the file.** This resolves to `std::find`, but everywhere else std algorithms are qualified (`std::set_intersection`, `std::sort`, `std::remove_if`, `std::set_difference`). An unqualified `find` next to a class that could later gain a `find` member is also an overload-resolution trap. Qualify it `std::find`.

syncd/FlexCounter.cpp
```diff
@@ -904,8 +1033,224 @@ class CounterContext : public BaseCounterContext
...
+                        continue;
+                    }
+                    std::vector<uint64_t> stats(intf_counter_ids.size());
+                    if (!collectData(rid, intf_counter_ids, effective_stats_mode, false, stats))
+                    {
+                        SWSS_LOG_ERROR("%s RID %s VID %s can't provide the statistic",  m_name.c_str(),
+                               sai_serialize_object_id(rid).c_str(), sai_serialize_object_id(vid).c_str());
+                        throw std::runtime_error("Test counter poll failed on populating m_objectIdsMap");
+                    }
+                    auto it_vid = m_objectIdsMap.find(vid);
```
lolyu: **N5 — this erase-then-emplace block is duplicated 4× and is reducible to one line.** The same shape — `find(vid)` → conditional `erase` → `make_shared<CounterIds>` → `emplace` (with this exact `// Remove and re-add if vid already exists` comment) — appears at lines ~1132, ~1227, ~1341, ~1435. `std::map::operator[]` already overwrites an existing entry, so the find/erase dance is itself redundant. Extract a helper:
```cpp
void resetSingleObjectPolling(sai_object_id_t vid, sai_object_id_t rid,
                              const std::vector<StatType>& ids)
{
    m_objectIdsMap[vid] = std::make_shared<CounterIds<StatType>>(rid, ids);
}
```
This dedups all four sites and simplifies each to a single call. Same spirit as the `addGroup` helper I suggested for group creation.

syncd/FlexCounter.cpp
```diff
@@ -554,6 +568,121 @@ class CounterContext : public BaseCounterContext
...
+        if (!supportBulk)
+        {
+            auto counter_data = std::make_shared<CounterIds<StatType>>(rid, supportedIds);
+            // TODO: use if const expression when cpp17 is supported
+            if (HasStatsMode<CounterIdsType>::value)
+            {
+                counter_data->setStatsMode(instance_stats_mode);
+            }
+            m_objectIdsMap.emplace(vid, counter_data);
+        }
```
lolyu, line 655: **N6 — prefix-partition / bulk-chunk dispatch is duplicated between the single and bulk add paths.** The `m_counterChunkSizeMapFromPrefix.empty()` if/else — `createCounterBulkChunkSizePerPrefixPartition` + per-prefix sort + `getBulkStatsContext`/`addBulkStatsContext` — appears nearly verbatim here and in `bulkAddObjectWithCounterGroups` (~25 lines each). The two differ only in single-vs-bulk context creation. Factor the partition-and-dispatch into a helper that takes the supported-id vector and a small callable for the context step, so the chunking logic lives in one place.

lolyu (commented):
**Review:** https://github.com/sonic-net/sonic-sairedis/pull/1774 (second pass)

Follow-up to my earlier comments (tombstone free-list, `m_failedPolls` cleanup, `addGroup` helper). These are **new** issues, focused on C++ semantics, conventions, and duplication — none overlaps the three above. I read the full file at `3eb7802`, not just the diff.

Highest-value items inline: **N1** (the `m_failedPolls` tier logic has a real off-by-one *and* does 3-4 map lookups per failed poll) and **N5** (the erase-then-emplace `m_objectIdsMap` block is copy-pasted 4×, and is reducible to a single `operator[]` assignment).

Lower-priority polish not worth individual inline threads:
- **N7** — the `effective_stats_mode` `HasStatsMode<...> ? ... : m_groupStatsMode` ladder is triplicated (addObject / bulkAddObject / collectData); extract `computeEffectiveStatsMode()`.
- **N8** — map-membership idiom is mixed across the file: `.count(vid)`, `.count(vid) == 0`, and `.find(vid) != end()` all used for the same presence check. Pick one (`.count` reads cleanly pre-C++20).
- **N9** — the merge branches around lines 2087-2091 / 2120-2133 are over-indented one level vs the surrounding 8-space style (looks like a tab/space mix); a clang-format pass would normalize.

Nice work overall — the feature is sound and well-tested; these are about tightening the new code before it lands.

justin-wong-ce:
> **Review:** #1774 (second pass)
> 
> Follow-up to my earlier comments (tombstone free-list, `m_failedPolls` cleanup, `addGroup` helper). These are **new** issues, focused on C++ semantics, conventions, and duplication — none overlaps the three above. I read the full file at `3eb7802`, not just the diff.
> 
> Highest-value items inline: **N1** (the `m_failedPolls` tier logic has a real off-by-one _and_ does 3-4 map lookups per failed poll) and **N5** (the erase-then-emplace `m_objectIdsMap` block is copy-pasted 4×, and is reducible to a single `operator[]` assignment).
> 
> Lower-priority polish not worth individual inline threads:
> 
> * **N7** — the `effective_stats_mode` `HasStatsMode<...> ? ... : m_groupStatsMode` ladder is triplicated (addObject / bulkAddObject / collectData); extract `computeEffectiveStatsMode()`.
> * **N8** — map-membership idiom is mixed across the file: `.count(vid)`, `.count(vid) == 0`, and `.find(vid) != end()` all used for the same presence check. Pick one (`.count` reads cleanly pre-C++20).
> * **N9** — the merge branches around lines 2087-2091 / 2120-2133 are over-indented one level vs the surrounding 8-space style (looks like a tab/space mix); a clang-format pass would normalize.
> 
> Nice work overall — the feature is sound and well-tested; these are about tightening the new code before it lands.

Thanks for the review, I will look into it.

syncd/FlexCounter.cpp
justin-wong-ce: As mentioned before the SWSS_LOG_ENTER() is required by the build.

I think I will remove this helper and directly reference the vector in where this function is called. Readability will drop but a perf tradeoff is not worth it.

syncd/FlexCounter.cpp
justin-wong-ce: I dont think a helper function with only one line is a good idea especially with the build `SWSS_LOG_ENTER` CI check bringing in a performance hit.

Will replace this logic with the one-liner in the suggested helper function instead.

syncd/FlexCounter.cpp
justin-wong-ce: The original intention is to completely remove the fallback method `addObject()` (and bulk equivalent) once `addObjectWithCounterGroups()` (and builk equivalent) is shown to be stable.

Helper functions are also avoided since each function would require a `SWSS_LOG_ENTER()` to fulfil the CI check requirements.

Will leave it separated since there is a good chance the old method will be removed.

syncd/FlexCounter.cpp
justin-wong-ce: Done

syncd/FlexCounter.cpp
justin-wong-ce: Added

syncd/FlexCounter.cpp
justin-wong-ce: Done

syncd/FlexCounter.cpp
justin-wong-ce: Done with C++11 syntax

syncd/FlexCounter.cpp
justin-wong-ce: Done

syncd/FlexCounter.cpp
justin-wong-ce: Done

justin-wong-ce:
Running the full `sonic-mgmt` test suite on 202511 to verify no surprise regressions at the moment, will update this thread when it is done.

Running the subset of counter related `sonic-mgmt` tests shows no issues.

@lolyu

justin-wong-ce:
> Running the full `sonic-mgmt` test suite on 202511 to verify no surprise regressions at the moment, will update this thread when it is done.
> 
> Running the subset of counter related `sonic-mgmt` tests shows no issues.
> 
> @lolyu

Aside from an unrelated regression, there are no notable regressions from running `sonic-mgmt` with a `syncd` built with the changes from this PR.

StormLiangMS:
@justin-wong-ce - I re-reviewed the latest head. The earlier cleanup comments look mostly addressed and CI is green, but I still have two correctness questions before approval.

1. In `bulkAddObjectWithCounterGroups()`, we build the bulk context from the largest supported counter group and add all VIDs. If a VID has no supported group, it can still be included in that bulk context, and the fallback path uses `m_objectSupportedCountersGroupMap[vid]`, which default-inserts group 0. Could this still poll unsupported counters for management/no-capability ports? Should we build bulk contexts per actual group membership and skip VIDs without a group, using `find()` instead of `operator[]`?

2. For re-adding an existing VID with the same or a superset-compatible counter list, `addObjectWithCounterGroups()` updates group state but does not remove/overwrite existing single or bulk polling state in all cases (`emplace()` can leave the old entry, and bulk context add skips duplicate VIDs). Could we remove the VID from existing contexts before installing the new grouped state, matching legacy `addObject()` behavior?

lolyu (approved):
**Review:** https://github.com/sonic-net/sonic-sairedis/pull/1774 (re-review at `7627bd9`)

Thanks for the thorough revision — I verified every change against the full source at this head, not just the diff. All nine findings across my earlier rounds are resolved or consciously deferred with good reason. Approving.

**The two real bugs and the latent one — all fixed:**
- **Tombstone leak → free-list.** `addGroup` (1425) reuses `m_freeGroupIndices` before appending; `cleanupCounterGroupMapping` reclaims the index on last-reference removal (1499). The vector no longer accumulates permanent empty-set tombstones across churn. 👍
- **`m_failedPolls` leak → fixed.** `removeObject` now `erase({rid, vid})` before dropping the object (1514), so re-added ports get an honest DEBUG→ERROR ramp again.
- **`m_failedPolls` tier off-by-one → fixed.** Now `uint32_t n = ++m_failedPolls[{rid, vid}]` — a single lookup with clean `n==1` / `n<=3` / `n==4` tiers (the ERROR fires exactly once). The C++11 `++operator[]` adaptation is functionally equivalent to the `try_emplace` I suggested.

**Convention / duplication:**
- **`addGroup` helper** now owns group creation at all five sites (index captured once, pre-push) — the four-spelling index inconsistency is now unwritable, and the sort-lambda is factored in.
- **`std::set` copies → `const&`** (2008/2067/2080/2082); **unqualified `find` → `std::find`** (2011).
- **The 4× erase-then-emplace block** collapsed to a single `m_objectIdsMap[vid] = std::make_shared<...>` at each site — cleaner than the helper I proposed.
- **`getSortedCGRef`** removed entirely, which resolves the mutable-interior-pointer concern more completely than constifying it would have.

**On the `SWSS_LOG_ENTER` build constraint:** fair point, and well handled. Resolving the dedup/encapsulation items by *inlining* (removing the accessor, using `operator[]` directly) rather than adding helper functions is the right synthesis — it achieves the cleanup without paying the per-call logging overhead the CI check imposes. Noted and agreed.

**One optional follow-up (non-blocking):** the bulk-vs-single dispatch duplication you're leaving in place pending removal of the fallback path is a reasonable call — consider a `// TODO: remove fallback path once addObjectWithCounterGroups is proven stable` marker so that intent is visible in-tree to the next reader.

Nice work — the feature was sound to begin with, and the new code is now tight. LGTM.

StormLiangMS:
Hi @justin-wong-ce for the cherrypick, could you test with 202605 and update in the description?

justin-wong-ce:
> Hi @justin-wong-ce for the cherrypick, could you test with 202605 and update in the description?

I have ran the changes with 202511, 202605 sonic-mgmt is in progress - I expect more noise with that due to 202605 being newly branched.

I have manually created a cherry-pick here: #1956  - I will post further updates directly there.

vaibhavhd:
202605 tests are in progress. Cherry pick will not be done until testing is complete with results.

justin-wong-ce:
FYI @lolyu @vaibhavhd 
The full `sonic-mgmt` testing on 202605 has completed. I have ran the tests on TH5 using `Arista-7060X6-64PE`.
Comparing the test results with and without these changes, I don't see any difference in the test results.

I have updated the PR description as well.

PR #1780, opened 2026-02-25
[meta] Continue bulk operations if MODE_IGNORE_ERROR and skip entries that fail validation

markx-arista:
[PR1613](https://github.com/sonic-net/sonic-sairedis/pull/1613) was proposed to ignore errors during bulk operations if SAI_BULK_OP_ERROR_MODE_IGNORE_ERROR is set. But it was reverted as it caused regression in test_duplicate_route.py because validation errors are ignored and some SAI can successfully add duplicated route. Later, createObject will be called on already existed entries in meta_generic_validation_post_create, and throws a fatal runtime_error exception.
This fix removes entries that fail validation checks before passing to SAI during bulk operations with SAI_BULK_OP_ERROR_MODE_IGNORE_ERROR, and populates object_statuses with validation results for those entries.

### DIFF
--- a/meta/Meta.cpp
+++ b/meta/Meta.cpp
@@ -18,6 +18,16 @@
 
 #define CHECK_STATUS_SUCCESS(s) { if ((s) != SAI_STATUS_SUCCESS) return (s); }
 
+#define CHECK_STATUS_SUCCESS_BULK(s,m,os)                                                       \
+{                                                                                               \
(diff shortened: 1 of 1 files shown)


Ndancejic (approved):
LGTM

lolyu:
Hi @markx-arista, can you fix the conflict?

PR #1805, opened 2026-03-23
Skip BVI/vxlan_tunnel interfaces in asyncIntfStateUpdate

rejithomas-arista:
VPP-internal interfaces (BVI, vxlan_tunnel) have no Linux TAP pair. Skip port state notification for these to prevent ERR logs that cause loganalyzer failures in VxLAN ECMP tests.
Also add prefix to error messages and guard against empty interface name.

### DIFF
--- a/vslib/vpp/SwitchVppRif.cpp
+++ b/vslib/vpp/SwitchVppRif.cpp
@@ -479,6 +479,14 @@ sai_status_t SwitchVpp::asyncIntfStateUpdate(const char *hwif_name, bool link_up
 {
     SWSS_LOG_ENTER();
 
+    // BVI and vxlan_tunnel interfaces are VPP-internal and have no Linux TAP pairs.
+    // Skip port state notification for these interfaces.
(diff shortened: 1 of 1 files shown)


rejithomas-arista:
@yue-fred-gao could you review this?

vslib/vpp/SwitchVppRif.cpp
```diff
@@ -479,6 +479,14 @@ sai_status_t SwitchVpp::asyncIntfStateUpdate(const char *hwif_name, bool link_up
 {
     SWSS_LOG_ENTER();
 
+    // BVI and vxlan_tunnel interfaces are VPP-internal and have no Linux TAP pairs.
+    // Skip port state notification for these interfaces.
+    if (strncmp(hwif_name, "bvi", 3) == 0 || strncmp(hwif_name, "vxlan_tunnel", 12) == 0)
```
lolyu, line 484: can we add `ipip` here, as the ipinip tunnel doesn't have the tap pair?

PR #1809, opened 2026-03-23
Increase syncd watchdog timeout to 60s for VPP

rejithomas-arista:
VPP processes route entries synchronously via individual API calls (ip_route_add_del). During BGP reconvergence, bulk route programming can exceed the default 30s watchdog timeout, causing false ERR logs.

Set to 60s for now to allow tests to pass. This may need further tuning based on topology size and route scale.

### DIFF
--- a/syncd/scripts/syncd_init_common.sh
+++ b/syncd/scripts/syncd_init_common.sh
@@ -504,6 +504,10 @@ vpp_api_check()
 config_syncd_vpp()
 {
     CMD_ARGS+=" -l -p $HWSKU_DIR/sai_vpp.profile"
+    # VPP processes bulk route entries synchronously (one VPP API call per route).
+    # A bulkset of ~1000 routes can take ~50 seconds, exceeding the default 30s
+    # watchdog. Increase to 60s to avoid false watchdog ERR logs.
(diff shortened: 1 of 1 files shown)


yue-fred-gao:
I fixed the performance issue with PR https://github.com/sonic-net/sonic-sairedis/pull/1794. It can add 42000 in about 17 seconds. It breaks it into multiple batches with bulk api. So 30 seconds should be enough. I think this is not needed anymore.

rejithomas-arista:
Any plans for cherry picking PR https://github.com/sonic-net/sonic-sairedis/pull/1794. to 202511 branch? Should I go ahead and do that?

yue-fred-gao:
> Any plans for cherry picking PR #1794. to 202511 branch? Should I go ahead and do that?

We don't have such plan. Do you need it in 202511?

rejithomas-arista:
> > Any plans for cherry picking PR #1794. to 202511 branch? Should I go ahead and do that?
> 
> We don't have such plan. Do you need it in 202511?

I was using 202511 for some scale testing. I’ll cherry-pick the patch into my private branch for now, which should be sufficient. I’ll close this review.

PR #1821, opened 2026-03-30
[syncd] Add config_db.json fallback for saithrift

vishal-nokia:
#### Why I did it

Some HWSKUs use `hwsku.json` instead of `port_config.ini` to define port configurations. 
When saithrift is enabled, syncd currently hardcodes the port map path to `$HWSKU_DIR/port_config.ini`, which causes failures on platforms that don't ship a `port_config.ini`. 
This is part of the broader effort to clean up port_config.ini dependencies (see [sonic-buildimage#24494](https://github.com/sonic-net/sonic-buildimage/issues/24494)).

#### How I did it

- Updated `syncd_init_common.sh` to use to `/etc/sonic/config_db.json` when `port_config.ini` is absent and saithrift is enabled.
- Extended PortMapParser to parse the PORT table from JSON files to build the lane-to-port map.
- Added unit tests for the new JSON parsing path.

#### How to verify it

Remove `port_config.ini` and run the sonic-mgmt QoS tests (`sonic-mgmt/tests/qos/`).

### DIFF
--- a/syncd/PortMapParser.cpp
+++ b/syncd/PortMapParser.cpp
@@ -2,12 +2,26 @@
 
 #include "swss/logger.h"
 
+#include <nlohmann/json.hpp>
+
 #include <iostream>
 #include <fstream>
 #include <sstream>
 #include <cstring>
 
 using namespace syncd;
+
+static bool hasJsonExtension(const std::string& filename)
+{
+    SWSS_LOG_ENTER();
+
+    const std::string ext = ".json";
+
+    if (filename.size() < ext.size())
+        return false;
+
+    return filename.compare(filename.size() - ext.size(), ext.size(), ext) == 0;
+}
 
 // TODO: introduce common config format for SONiC
 std::shared_ptr<PortMap> PortMapParser::parsePortMap(
@@ -22,6 +36,11 @@ std::shared_ptr<PortMap> PortMapParser::parsePortMap(
         SWSS_LOG_NOTICE("no port map file, returning empty port map");
 
         return portMap;
+    }
+
+    if (hasJsonExtension(portMapFile))
+    {
+        return parsePortMapFromJson(portMapFile);
     }
 
     std::ifstream portmap(portMapFile);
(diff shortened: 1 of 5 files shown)


Pavan-Nokia (approved):

rawal01 (dismissed):
lgtm

syncd/PortMapParser.cpp
```diff
@@ -64,3 +83,67 @@ std::shared_ptr<PortMap> PortMapParser::parsePortMap(
...
+
+        std::string lanes = portData["lanes"].get<std::string>();
+
+        std::istringstream iss(lanes);
+        std::string lane_str;
+        std::set<int> lane_set;
+
+        while (getline(iss, lane_str, ','))
+        {
+            lane_set.insert(stoi(lane_str));
```
bingwang-ms: Please add `try-catch` to catch invalid lane numbers and exit with error code.

syncd/PortMapParser.cpp
```diff
@@ -24,6 +38,11 @@ std::shared_ptr<PortMap> PortMapParser::parsePortMap(
         return portMap;
     }
 
+    if (hasJsonExtension(portMapFile))
```
bingwang-ms, line 41: If both `config_db.json` and `port_config.ini` exist, but `lanes` is inconsistent, which one will be taken?

bingwang-ms:
@kperumalbfn @developfast Please help review. This could impact dynamic breakout as well. Thanks

syncd/PortMapParser.cpp
vishal-nokia: Added `try-catch` around `stoi` to catch invalid lane values. It now logs the offending value and port name before exiting.

syncd/PortMapParser.cpp
vishal-nokia: `port_config.ini` is always preferred when present. The `elif` in `syncd_init_common.sh` ensures `config_db.json` is only consulted when the `ini` file is absent, so only one of the two is ever used. The ordering was a deliberate design choice to ensure zero impact on existing platforms, as per our discussion in the weekly meeting.
This fallback targets platforms that use generic HWSKU and do not ship a `port_config.ini`. On those platforms, `config_db.json` is used to build the port map.
Regarding dynamic breakout, if a platform supports generic HWSKU but still ships `port_config.ini`, and a breakout is configured at runtime, the static `ini` would be picked over the updated `config_db.json`, which could indeed cause a mismatch. 
The expectation is that platforms adopting generic HWSKU will not ship `port_config.ini`, so `config_db.json` is always used and reflects the current breakout state.

bingwang-ms: I suggest taking `lanes` from config_db if both `config_db.json` and `port_config.ini` are present. In this way we can always override the `lanes` by `config_db`.

Pavan-Nokia: okay, we can make this switch in design. Please be aware that this will impact the code flow for all existing platforms as config_db.json is always present. (we were trying to avoid this)
We will update the PR once regression testing is complete.

kperumalbfn: @Pavan-Nokia @vishal-nokia Will be better to give priority to config_db.json instead of port_config.ini as that configs will determine the final port names, lanes and port index. This is a requirement for generic SKU and port breakout.

syncd/scripts/syncd_init_common.sh
```diff
@@ -690,7 +690,15 @@ config_syncd()
     set_start_type
 
     if [ ${ENABLE_SAITHRIFT} == 1 ]; then
-        CMD_ARGS+=" -r -m $HWSKU_DIR/port_config.ini"
+        CMD_ARGS+=" -r"
+        if [ -f $HWSKU_DIR/port_config.ini ]; then
+            CMD_ARGS+=" -m $HWSKU_DIR/port_config.ini"
+        elif [ -f /etc/sonic/config_db.json ]; then
```
bingwang-ms, line 696: If `config_db.json` has higher priority, then this shouldn't be `elif`

syncd/PortMapParser.cpp
bingwang-ms: Hi @Pavan-Nokia, could you please share the offline discussion in this PR as well? Thanks

Pavan-Nokia: Hi, 
As discussed, offline.
This block of code is gated by ENABLE_SAITHRIFT=1 and only runs in the SAI-thrift test docker which is mainly used for QOS testing. The saithrift is not compiled as part of buildimage and would not impact any production scenario. 
This is the reason a decision was made to take the least disruptive approach that does not impact the code flow for any of the existing platforms.
Mainly multi-asic platforms would be impacted if the default is changed to config_db.json

PR #1825, opened 2026-03-31
vpp: multi L2 VXLAN

sambenko:
Following PR https://github.com/sonic-net/sonic-sairedis/pull/1761, this one aims to add functionality of creating/deleting multiple L2 VXLAN tunnels based on VLAN VNI pairs, something which didn't work in the aforementioned base L2 VXLAN support PR.

Root cause: The code that creates L2 VXLAN tunnels looped through VNI-to-VLAN mapper entries but had a break after the first one. Only one VPP tunnel was ever created per remote VTEP, regardless of how many VNIs were configured.

This patch introduces 3 fixes:

1. _**Multi-VNI creation**_ - Removed the break and looped through all mapper entries, creating a separate VPP tunnel per VNI.
2. _**Remove path fix**_ - During full VXLAN teardown, SONiC deletes map entries before the tunnel itself. The remove function walked those same map entries to find which VPP tunnels to clean up and by the time it ran, the entries were already gone, so nothing got cleaned up. Fixed by iterating the internal m_l2_tunnel_map directly and matching by IP address instead of relying on mapper entries.
3. _**Late mapper entry handling**_ - BGP could trigger tunnel creation before all VNI map entries were in the SAI database, which is a race condition. Whichever VNI wasn't in the DB yet would never get a VPP tunnel. Fixed by hooking into tunnel map entry creation - when a new entry is added, it checks if a tunnel already exists and creates the missing VPP tunnel on the spot.


sambenko:
Hey @yue-fred-gao, this is a patch to support multi L2 vxlan tunnel with a small refactor of tunnel creation as well to allow late map entry additions. Can someone review it? Thank you.

vslib/vpp/TunnelManager.cpp
```diff
@@ -515,126 +569,196 @@ TunnelManager::create_l2_vxlan_tunnel(
...
-        SWSS_LOG_ERROR("Failed to remove tunnel sw_if %u from BD %u",
-            tunnel_data.sw_if_index, tunnel_data.vlan_id);
-        // Continue with tunnel deletion anyway
-    } else {
-        SWSS_LOG_NOTICE("Removed tunnel sw_if %u from BD %u",
-            tunnel_data.sw_if_index, tunnel_data.vlan_id);
+    // Only handle VNI_TO_VLAN_ID mapper type (L2 VXLAN)
+    auto mapper_obj = m_switch_db->get_sai_object(SAI_OBJECT_TYPE_TUNNEL_MAP,
+        sai_serialize_object_id(tunnel_map_oid));
+    if (!mapper_obj) {
```
yue-fred-gao: This is an error if db is in inconsistent state. We should log an error message.

vslib/vpp/TunnelManager.cpp
```diff
@@ -515,126 +569,196 @@ TunnelManager::create_l2_vxlan_tunnel(
...
 {
     SWSS_LOG_ENTER();
 
-    auto it = m_l2_tunnel_map.find(tunnel_oid);
-    if (it == m_l2_tunnel_map.end()) {
-        // Not an L2 tunnel we manage, could be L3 or P2MP, skip silently
-        SWSS_LOG_NOTICE("Tunnel %s not in L2 tunnel map, skipping removal",
-            sai_serialize_object_id(tunnel_oid).c_str());
-        return SAI_STATUS_SUCCESS;
+    uint32_t vni = 0;
```
yue-fred-gao, line 606: According to SAI definition, tunnel map entry has tunnel map type: https://github.com/opencomputeproject/SAI/blob/9ca8312f2f78c35051344fc36c5ab0ab3e1e8d62/inc/saitunnel.h#L107. I think we should check the type first and skip if it is not SAI_TUNNEL_MAP_TYPE_VNI_TO_VLAN_ID. The code will be more readable.

vslib/vpp/TunnelManager.cpp
```diff
@@ -515,126 +569,196 @@ TunnelManager::create_l2_vxlan_tunnel(
...
+    // Already have a VPP tunnel for this VNI
+    if (m_l2_tunnel_map.find(vni) != m_l2_tunnel_map.end()) {
+        return SAI_STATUS_SUCCESS;
     }
 
-    SWSS_LOG_NOTICE("Removed L2 VXLAN tunnel %s (sw_if=%u, VNI=%u, VLAN=%u)",
-        sai_serialize_object_id(tunnel_oid).c_str(),
-        tunnel_data.sw_if_index, tunnel_data.vni, tunnel_data.vlan_id);
+    // Find P2P tunnels that reference this mapper as a decap mapper
+    auto& tunnel_hash = m_switch_db->m_objectHash.at(SAI_OBJECT_TYPE_TUNNEL);
```
yue-fred-gao: you can call get_child_objs from tunnel_mapper object with type SAI_OBJECT_TYPE_TUNNEL to find all the tunnel objects that are referring to this tunnel_mapper.

vslib/vpp/TunnelManager.cpp
```diff
@@ -515,126 +569,196 @@ TunnelManager::create_l2_vxlan_tunnel(
...
+    return SAI_STATUS_SUCCESS;
+}
 
-    TunnelVPPData tunnel_data;
-    tunnel_data.vni = vni;
-    tunnel_data.src_ip = src_ip;
-    tunnel_data.dst_ip = dst_ip;
-    tunnel_data.vlan_id = vlan_id;  // Store for cleanup
+sai_status_t
+TunnelManager::remove_l2_vxlan_tunnel(
```
yue-fred-gao: You added handle_l2_vxlan_tunnel_map_entry to process new map entry after tunnel is created. I think we should handle map entry deleted before tunnel by removing the corresponding vpp tunnels instead of cleaning up tunnels when SAI tunnel is deleted. Removing a tunnel map entry alone without deleting tunnel might be a valid use case.

vslib/vpp/TunnelManager.cpp
```diff
@@ -676,12 +676,16 @@ TunnelManager::handle_l2_vxlan_tunnel_map_entry(
 {
     SWSS_LOG_ENTER();
 
+    int32_t tunnel_map_type = -1;
     uint32_t vni = 0;
     uint16_t vlan_id = 0;
     sai_object_id_t tunnel_map_oid = SAI_NULL_OBJECT_ID;
 
     for (uint32_t i = 0; i < attr_count; i++) {
```
yue-fred-gao: There is SaiCachedObject which is a subclass of SaiObject but built from attr_list. It provides same functions like get_attr, get_parent_object, get_child_objects etc as SaiObject. For attr_list in an update operation, it merges the new attributes and existing attributes to provide a holistic view. You can consider to use it.

vslib/vpp/TunnelManager.cpp
```diff
@@ -515,102 +569,231 @@ TunnelManager::create_l2_vxlan_tunnel(
...
+
+    attr.id = SAI_TUNNEL_ATTR_ENCAP_DST_IP;
+    if (tunnel_obj->get_attr(attr) != SAI_STATUS_SUCCESS) {
+        SWSS_LOG_NOTICE("No ENCAP_DST_IP, skipping removal");
+        return SAI_STATUS_SUCCESS;
+    }
+    sai_ip_address_t dst_ip = attr.value.ipaddr;
+
+    // Iterate m_l2_tunnel_map directly — don't rely on mapper entries
+    // which may already be deleted during full teardown
```
yue-fred-gao: Is it still needed since you are handling map_entry delete? I wander if it is a valid use case to delete tunnel before map_entry are deleted.

sambenko: I kept the iteration there as kind of a safety net for cases when tunnel remove arrives without individual map_entry removes, especially in case of warm reboot or syncd restart. Does it make sense in those particular cases?

yue-fred-gao: sonic-vpp doesn't support warm boot. warm boot means restarting sonic while keeping data plane intact. sonic-vpp's dataplane is vpp process, which is running inside sonic (syncd container). For the same reason, syncd restart will restart vpp data plane as well. Deleting tunnel before tunnel map entries is not supported sequence in sonic. Below are the object dependencies. 
TUNNEL_MAP_ENTRY -> TUNNEL_MAP       (via SAI_TUNNEL_MAP_ENTRY_ATTR_TUNNEL_MAP)
TUNNEL                         -> TUNNEL_MAP       (via SAI_TUNNEL_ATTR_{ENCAP,DECAP}_MAPPERS)
Here is the orchagent code deleting vxlan tunnel.
https://github.com/zjswhhh/sonic-swss/blob/e6f134fbdda2a51608e948878faa62696a4e3565/orchagent/vxlanorch.cpp#L812. It deletes tunnel first then tunnel_mappers. To delete tunnel mappers, the tunnel map entries have to be deleted first. Otherwise, it violates above dependency. So the map entries must have been deleted before tunnel.

sambenko: Thanks for the explanation, you're right, at that time, I didn't realize that vpp dataplane resets when syncd restarts. I'll remove the unnecessary iteration as soon as I get on the PC.

yue-fred-gao (approved):
lgtm

sambenko:
@yue-fred-gao hey, checks have passed, I think it's ready to be merged now.

PR #1819, opened 2026-03-27
Enable async ASIC_DB writes for Southbound-ZMQ

inder-nexthop:
#### Why I did it
We are extending southbound ZMQ support (currently only used for DPU switches) to regular NPU switches for improved route offloading performance. However, PR #1694 introduced `DisabledRedisClient` that prevents syncd from writing ASIC state to Redis when ZMQ is enabled. On NPU switches we still want ASIC_DB populated (for warm boot, diagnostics, `show` tooling, etc.), but without the synchronous Redis writes blocking the ZMQ data path.

#### How I did it
Added a dedicated `ZmqRedisClient` (subclass of `RedisClient`) that performs ASIC_DB writes asynchronously on a background thread via `swss::AsyncDBUpdater`, instead of extending/mutating `RedisClient` itself.

Added a new opt-in knob, `-R` / `--asyncRec`, driven by `SYSTEM_DEFAULTS|async_rec` (`status: enabled`) in CONFIG_DB and plumbed through `syncd_init_common.sh`.

Redis client selection (in `Syncd`):
- **DPU + ZMQ**: `DisabledRedisClient` — no Redis writes, maintains PR #1694 behavior.
- **NPU + ZMQ + `async_rec` enabled**: `ZmqRedisClient`,  async ASIC_DB writes that persist ASIC state without blocking the ZMQ data path.
- **Everything else** (including ZMQ active but `async_rec` disabled): `RedisClient` synchronous writes (traditional behavior).

`ZmqRedisClient` uses field-merge (`HSET`) semantics for create/set operations and `DEL` for removes, matching the synchronous `RedisClient` (which only `hset`/`hmset`s and never deletes on a write). This avoids wiping an object's other attributes when a single field is updated.

> **Dependency:** requires sonic-swss-common #1217, which adds `HSET_COMMAND` to `AsyncDBUpdater` (merge-only write). This PR will not compile until that change is merged and the submodule is bumped.

Added unit tests: `ZmqRedisClient` create/set/remove behavior including field-merge regression tests, `Syncd` client-selection tests across DPU/NPU/`async_rec`/context-config combinations, and `--asyncRec` command-line parsing.

#### How to verify it

Used the route benchmark script under `sonic-mgmt/tests/scripts/route_programming_benchmark.py`.
Also tested that the routes are written to ASIC_STATE table with a scale of 500k routes:

```
[...]

PR #1835, opened 2026-04-07
Runtime knob for Southbound ZMQ

inder-nexthop:
#### Why I did it
Enable runtime-toggleable ZMQ communication between orchagent and syncd (southbound) via a CONFIG_DB knob.

#### How I did it

1. Changed default ZMQ endpoints in ContextConfig from ipc:///tmp/zmq_ep to tcp://127.0.0.1:5555 / tcp://127.0.0.1:5556 so orchagent↔syncd ZMQ works out-of-the-box without a hwsku-specific context_config.json.
2. Added an m_loadedFromJson flag to ContextConfig, set by ContextConfigContainer::loadFromFile(). This lets us distinguish "no config file present (defaults)" from "config file explicitly disables ZMQ".
3. In both RedisRemoteSaiInterface (orchagent side) and Syncd (syncd side): when -z zmq_sync is requested via cmdline but a loaded context_config.json sets zmq_enable=false, fall back to Redis sync. This allows context_config.json to act as an explicit override.
4. Scoped DisabledRedisClient (write-only-to-ZMQ) to DPU switches only. Non-DPU ZMQ retains a real Redis client so the ASIC_DB mirror is still populated for tools and observability.
5. syncd_init_common.sh reads orch_southbound_zmq_enabled from SYNCD_VARS (generated from CONFIG_DB DEVICE_METADATA) and passes -z zmq_sync when true.

#### Behavior changes

Documenting silent behavior changes for reviewers:

1. **Default ZMQ endpoints changed** from `ipc:///tmp/zmq_ep` / `ipc:///tmp/zmq_ntf_ep` to `tcp://127.0.0.1:5555` / `tcp://127.0.0.1:5556`. The doc comment in `lib/sairedis.h` for `SAI_REDIS_COMMUNICATION_MODE_ZMQ_SYNC` is updated to match.
2. **`-z zmq_sync` cmdline flag is no longer absolute.** When `context_config.json` is loaded and sets `zmq_enable=false` for the context, both syncd and sairedis fall back to `REDIS_SYNC`. The JSON is now authoritative.
3. **`-s` flag interaction with JSON ZMQ.** The old code rewrote `m_commandLineOptions->m_redisCommunicationMode` to `ZMQ_SYNC` when `-s` was passed and `m_zmqEnable` was true. The new code skips that block when `m_loadedFromJson && m_zmqEnable`, so the field retains its initial value. Net transport is unchanged (ZMQ is still picked via the later `m_zmqEnable` check), but the cmdline-options state field is no longer rewritten.
4. **`DisabledRedisClient` is now DPU-only.** Non-DPU ZMQ deployments retain a real `RedisClient`, so the ASIC_DB mirror is still populated for tools and observability.
5. **State alignment after fallback.** When sairedis falls back to `RedisChannel` because JSON disables ZMQ, `m_redisCommunicationMode` is now set to `REDIS_SYNC` so the member variable matches the actual channel.

#### How to verify it
Ran route benchmark test with both configurations to confirm:

- Default behavior (knob disabled) remains unchanged
- ZMQ mode activates correctly when knob is enabled
- Routes program successfully through the full pipeline in both modes

#### Related PRs:

https://github.com/sonic-net/sonic-buildimage/pull/26638
https://github.com/sonic-net/sonic-swss/pull/4556

### DIFF
--- a/lib/ContextConfig.cpp
+++ b/lib/ContextConfig.cpp
@@ -18,8 +18,9 @@ ContextConfig::ContextConfig(
     m_dbFlex(dbFlex),
     m_dbState(dbState),
     m_zmqEnable(false),
-    m_zmqEndpoint("ipc:///tmp/zmq_ep"),
-    m_zmqNtfEndpoint("ipc:///tmp/zmq_ntf_ep")
+    m_loadedFromJson(false),
+    m_zmqEndpoint("tcp://127.0.0.1:5555"),
+    m_zmqNtfEndpoint("tcp://127.0.0.1:5556")
 {
     SWSS_LOG_ENTER();
 
--- a/lib/RedisRemoteSaiInterface.cpp
+++ b/lib/RedisRemoteSaiInterface.cpp
@@ -81,16 +81,19 @@ sai_status_t RedisRemoteSaiInterface::apiInitialize(
     m_redisCommunicationMode = SAI_REDIS_COMMUNICATION_MODE_REDIS_ASYNC;
     m_zmqResponseBufferSize = SAI_ZMQ_DEFAULT_RESPONSE_BUFFER_SIZE;
 
-    if (m_contextConfig->m_zmqEnable)
+    if (m_contextConfig->m_loadedFromJson && m_contextConfig->m_zmqEnable)
     {
+        // context_config.json is authoritative: lock this context to ZMQ at init
+        SWSS_LOG_NOTICE("context %u: JSON zmq_enable=true, creating ZMQ channel at init",
+                m_contextConfig->m_guid);
+
         m_communicationChannel = std::make_shared<ZeroMQChannel>(
                 m_contextConfig->m_zmqEndpoint,
                 m_contextConfig->m_zmqNtfEndpoint,
                 std::bind(&RedisRemoteSaiInterface::handleNotification, this, _1, _2, _3),
                 m_zmqResponseBufferSize);
 
-        SWSS_LOG_NOTICE("zmq enabled, forcing sync mode");
-
+        m_redisCommunicationMode = SAI_REDIS_COMMUNICATION_MODE_ZMQ_SYNC;
         m_syncMode = true;
     }
     else
@@ -369,14 +372,24 @@ sai_status_t RedisRemoteSaiInterface::setRedisExtensionAttribute(
 
         case SAI_REDIS_SWITCH_ATTR_REDIS_COMMUNICATION_MODE:
 
-            m_redisCommunicationMode = (sai_redis_communication_mode_t)attr->value.s32;
-
-            if (m_contextConfig->m_zmqEnable)
+            // JSON zmq_enable=true: transport locked to ZMQ at init; ignore attr
+            if (m_contextConfig->m_loadedFromJson && m_contextConfig->m_zmqEnable)
             {
-                SWSS_LOG_NOTICE("zmq enabled via context config");
+                if (attr->value.s32 == SAI_REDIS_COMMUNICATION_MODE_ZMQ_SYNC)
+                {
+                    SWSS_LOG_NOTICE("context %u: JSON zmq_enable=true, comm-mode attr ZMQ_SYNC matches; ignoring redundant request",
+                            m_contextConfig->m_guid);
+                }
+                else
+                {
+                    SWSS_LOG_WARN("context %u: JSON zmq_enable=true enforces ZMQ_SYNC but caller requested comm-mode %d; ignoring",
+                            m_contextConfig->m_guid, attr->value.s32);
+                }
 
-                m_redisCommunicationMode = SAI_REDIS_COMMUNICATION_MODE_ZMQ_SYNC;
+                return SAI_STATUS_SUCCESS;
             }
+
+            m_redisCommunicationMode = (sai_redis_communication_mode_t)attr->value.s32;
 
             m_communicationChannel = nullptr;
 
@@ -415,11 +428,32 @@ sai_status_t RedisRemoteSaiInterface::setRedisExtensionAttribute(
                     return SAI_STATUS_SUCCESS;
 
                 case SAI_REDIS_COMMUNICATION_MODE_ZMQ_SYNC:
+                    // If context_config.json explicitly set zmq_enable=false,
+                    // respect it and fall back to RedisChannel
+                    if (m_contextConfig->m_loadedFromJson && !m_contextConfig->m_zmqEnable)
+                    {
+                        SWSS_LOG_NOTICE("context %u: zmq_enable=false in context config, falling back to Redis sync",
+                                m_contextConfig->m_guid);
+
+                        // Keep m_redisCommunicationMode aligned with the actual channel
+                        m_redisCommunicationMode = SAI_REDIS_COMMUNICATION_MODE_REDIS_SYNC;
+
+                        m_syncMode = true;
+
+                        m_communicationChannel = std::make_shared<RedisChannel>(
+                                m_contextConfig->m_dbAsic,
+                                std::bind(&RedisRemoteSaiInterface::handleNotification, this, _1, _2, _3));
+
+                        m_communicationChannel->setResponseTimeout(m_responseTimeoutMs);
+
+                        m_communicationChannel->setBuffered(false);
+
+                        return SAI_STATUS_SUCCESS;
+                    }
+
+                    SWSS_LOG_NOTICE("ZMQ sync mode enabled for context %u", m_contextConfig->m_guid);
 
                     m_contextConfig->m_zmqEnable = true;
-
-                    // main communication channel was created at initialize method
-                    // so this command will replace it with zmq channel
 
                     m_communicationChannel = std::make_shared<ZeroMQChannel>(
                             m_contextConfig->m_zmqEndpoint,
@@ -428,8 +462,6 @@ sai_status_t RedisRemoteSaiInterface::setRedisExtensionAttribute(
                             m_zmqResponseBufferSize);
 
                     m_communicationChannel->setResponseTimeout(m_responseTimeoutMs);
-
-                    SWSS_LOG_NOTICE("zmq enabled, forcing sync mode");
 
                     m_syncMode = true;
 
--- a/syncd/Syncd.cpp
+++ b/syncd/Syncd.cpp
@@ -22,6 +22,8 @@
 #include "swss/tokenize.h"
 #include "swss/notificationproducer.h"
 #include "swss/exec.h"
+#include "swss/dbconnector.h"
+#include "swss/table.h"
 
 #include "meta/sai_serialize.h"
 #include "meta/ZeroMQSelectableChannel.h"
@@ -82,16 +84,8 @@ Syncd::Syncd(
         SWSS_LOG_THROW("no context config defined at global context %u", m_commandLineOptions->m_globalContext);
     }
 
-    if (m_contextConfig->m_zmqEnable && m_commandLineOptions->m_enableSyncMode)
-    {
-        SWSS_LOG_NOTICE("disabling command line sync mode, since context zmq enabled");
-
-        m_commandLineOptions->m_enableSyncMode = false;
-
-        m_commandLineOptions->m_redisCommunicationMode = SAI_REDIS_COMMUNICATION_MODE_ZMQ_SYNC;
-    }
-
-    if (m_commandLineOptions->m_enableSyncMode)
+    if (m_commandLineOptions->m_enableSyncMode
+            && !(m_contextConfig->m_loadedFromJson && m_contextConfig->m_zmqEnable))
     {
         SWSS_LOG_WARN("enable sync mode is deprecated, please use communication mode, FORCING redis sync mode");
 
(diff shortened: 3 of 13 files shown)


inder-nexthop:
@deepak-singhal0408 Southbound ZMQ config knob PR is ready for review.

syncd/scripts/syncd_init_common.sh
```diff
@@ -38,11 +38,15 @@ mkdir -p /var/log/sai_failure_dump/
 # CMD_ARGS+=" -l"
 
 # Set zmq mode by default for smartswitch DPU
+# Otherwise, check if southbound ZMQ is enabled for route offloading
 # Otherwise, set synchronous mode if it is enabled in CONFIG_DB
 SYNC_MODE=$(echo $SYNCD_VARS | jq -r '.synchronous_mode')
 SWITCH_TYPE=$(echo $SYNCD_VARS | jq -r '.switch_type')
+SOUTHBOUND_ZMQ=$(sonic-db-cli CONFIG_DB hget "DEVICE_METADATA|localhost" "orch_southbound_route_zmq_enabled")
```
vganesan-nokia: We can use the method of using swss_vars.j2 to get orch_southbound_route_zmq_enabled similiar to how we get SWITCH_TYPE and SYNC_MODE

syncd/scripts/syncd_init_common.sh
```diff
@@ -38,11 +38,15 @@ mkdir -p /var/log/sai_failure_dump/
...
 if [ "$SWITCH_TYPE" == "dpu" ]; then
     CMD_ARGS+=" -z zmq_sync -x /usr/share/sonic/hwsku/context_config.json"
+elif [ "$SOUTHBOUND_ZMQ" == "true" ]; then
```
vganesan-nokia, line 48: For SOUTHBOUND_ZMQ == true and SWITCH_TYPE== dpu we do the same thing. It will be good if we combine them to single  "if" statement 
if [ "$SWITCH_TYPE" == "dpu" ] || [ "$SOUTHBOUND_ZMQ" == "true" ]; then
    CMD_ARGS+=" -z zmq_sync -x /usr/share/sonic/hwsku/context_config.json"

syncd/scripts/syncd_init_common.sh
inder-nexthop: Done

deepak-singhal0408:
@inder-nexthop please add more UT coverage.

deepak-singhal0408:
@inder-nexthop could you check the CI failure? thanks,

deepak-singhal0408 (dismissed):

deepak-singhal0408:
@prsunny @prabhataravind, please help review/sign-off? thanks,

syncd/Syncd.cpp
```diff
@@ -155,8 +163,14 @@ Syncd::Syncd(
...
+    swss::Table deviceMetadataTable(&configDb, "DEVICE_METADATA");
+    std::string switchType;
+    deviceMetadataTable.hget("localhost", "switch_type", switchType);
+
+    bool isDpuSwitch = switchType == "dpu";
 
-    if (m_contextConfig->m_zmqEnable && !isVirtualSwitch)
+    if (m_contextConfig->m_zmqEnable && isDpuSwitch && !isVirtualSwitch)
     {
         m_client = std::make_shared<DisabledRedisClient>();
```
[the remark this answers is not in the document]
inder-nexthop: There are tests for DisabledRedisClient and there is a follow on PR for async asic_db writes that has more unit test coverage I think that we should leave it as is in this PR. 

PR #1819  for reference.

lib/ContextConfig.cpp
```diff
@@ -18,8 +18,9 @@ ContextConfig::ContextConfig(
     m_dbFlex(dbFlex),
     m_dbState(dbState),
     m_zmqEnable(false),
-    m_zmqEndpoint("ipc:///tmp/zmq_ep"),
-    m_zmqNtfEndpoint("ipc:///tmp/zmq_ntf_ep")
+    m_loadedFromJson(false),
+    m_zmqEndpoint("tcp://127.0.0.1:5555"),
+    m_zmqNtfEndpoint("tcp://127.0.0.1:5556")
```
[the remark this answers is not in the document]
inder-nexthop: Will update the comment in sairedis.h header to reflect the defaults changing.

syncd/scripts/syncd_init_common.sh
```diff
@@ -41,8 +41,11 @@ mkdir -p /var/log/sai_failure_dump/
 # Otherwise, set synchronous mode if it is enabled in CONFIG_DB
 SYNC_MODE=$(echo $SYNCD_VARS | jq -r '.synchronous_mode')
 SWITCH_TYPE=$(echo $SYNCD_VARS | jq -r '.switch_type')
+SOUTHBOUND_ZMQ=$(echo $SYNCD_VARS | jq -r '.orch_southbound_zmq_enabled')
 if [ "$SWITCH_TYPE" == "dpu" ]; then
     CMD_ARGS+=" -z zmq_sync -x /usr/share/sonic/hwsku/context_config.json"
+elif [ "$SOUTHBOUND_ZMQ" == "true" ]; then
+    CMD_ARGS+=" -z zmq_sync"
```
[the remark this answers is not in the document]
inder-nexthop: Added the file existence check for SOUTHBOUND_ZMQ config knob only, there are checks in `ContextConfigContainer.cpp` that log an error before falling back to default config values that I don't want to supress in the DPU case.

syncd/scripts/syncd_init_common.sh
```diff
@@ -41,8 +42,14 @@ mkdir -p /var/log/sai_failure_dump/
 # Otherwise, set synchronous mode if it is enabled in CONFIG_DB
 SYNC_MODE=$(echo $SYNCD_VARS | jq -r '.synchronous_mode')
 SWITCH_TYPE=$(echo $SYNCD_VARS | jq -r '.switch_type')
+SOUTHBOUND_ZMQ=$(echo $SYNCD_VARS | jq -r '.orch_southbound_zmq_enabled')
 if [ "$SWITCH_TYPE" == "dpu" ]; then
-    CMD_ARGS+=" -z zmq_sync -x /usr/share/sonic/hwsku/context_config.json"
+    CMD_ARGS+=" -z zmq_sync -x $CONTEXT_CONFIG_FILE"
```
vganesan-nokia, line 47: It will be good if we check the existence of contect config file before setting this -x option (similar to what is done for when southbound zmq is enabled)

inder-nexthop: `ContextConfigContainer.cpp` has logging for missing context config, do we want to suppress that in the DPU case? Is context config optional for DPU?

vganesan-nokia: O.k. thanks for the clarification. If the correct zmq endpoint configs can be made available by other means (like default hard coded), then we may not strictly need the context_config file (even for the southbound zmq enable case). However,if the context_config.json is the only way to get the end point config for dpu and if the context_config file does not exist we may have to notify this.

deepak-singhal0408 (dismissed):

lib/RedisRemoteSaiInterface.cpp
```diff
@@ -369,15 +372,17 @@ sai_status_t RedisRemoteSaiInterface::setRedisExtensionAttribute(
...
             {
-                SWSS_LOG_NOTICE("zmq enabled via context config");
+                SWSS_LOG_NOTICE("context %u: JSON zmq_enable=true, ignoring comm-mode attr %d",
+                        m_contextConfig->m_guid, attr->value.s32);
 
-                m_redisCommunicationMode = SAI_REDIS_COMMUNICATION_MODE_ZMQ_SYNC;
+                return SAI_STATUS_SUCCESS;
             }
 
+            m_redisCommunicationMode = (sai_redis_communication_mode_t)attr->value.s32;
```
prabhataravind, line 392: In the loadedFromJson && !zmqEnable fallback inside case ZMQ_SYNC, m_redisCommunicationMode could be set to ZMQ_SYNC in this line but the channel created later below is a RedisChannel. This is confusing and error prone from a maintenance point of view.

syncd/Syncd.cpp
```diff
@@ -104,11 +98,25 @@ Syncd::Syncd(
...
+        if (m_contextConfig->m_loadedFromJson && !m_contextConfig->m_zmqEnable)
+        {
+            SWSS_LOG_NOTICE("context %u: zmq_enable=false in context config, falling back to Redis sync",
+                    m_contextConfig->m_guid);
 
-        m_contextConfig->m_zmqEnable = true;
+            m_enableSyncMode = true;
 
-        m_enableSyncMode = true;
+            m_commandLineOptions->m_redisCommunicationMode = SAI_REDIS_COMMUNICATION_MODE_REDIS_SYNC;
```
prabhataravind, line 110: If JSON disables ZMQ but orchagent passes -z zmq_sync, syncd coerces to REDIS_SYNC but orchagent's sairedis layer keeps ZMQ_SYNC. The two sides of the channel would attempt different transports. The intent in this PR to keep context json as the authority is applied only on the syncd side. The behavior needs to be consistent on both sides.

lib/RedisRemoteSaiInterface.cpp
```diff
@@ -415,11 +420,29 @@ sai_status_t RedisRemoteSaiInterface::setRedisExtensionAttribute(
                     return SAI_STATUS_SUCCESS;
 
                 case SAI_REDIS_COMMUNICATION_MODE_ZMQ_SYNC:
+                    // If context_config.json explicitly set zmq_enable=false,
+                    // respect it and fall back to RedisChannel
+                    if (m_contextConfig->m_loadedFromJson && !m_contextConfig->m_zmqEnable)
```
prabhataravind, line 433: gRedisCommunicationMode is an orchagent global set from the -z command line flag. It is never updated when sairedis silently falls back to Redis. Notification handlers in notifications.cpp check gRedisCommunicationMode == ZMQ_SYNC to decide whether to forward events via NotificationProducer to ASIC_DB. After a silent fallback, they'd take the wrong path, so notifications to orchagent (like fdb or port notifications) will not work.

lib/RedisRemoteSaiInterface.cpp
```diff
@@ -369,15 +372,17 @@ sai_status_t RedisRemoteSaiInterface::setRedisExtensionAttribute(
...
         case SAI_REDIS_SWITCH_ATTR_REDIS_COMMUNICATION_MODE:
 
-            m_redisCommunicationMode = (sai_redis_communication_mode_t)attr->value.s32;
-
-            if (m_contextConfig->m_zmqEnable)
+            // JSON zmq_enable=true: transport locked to ZMQ at init; ignore attr
+            if (m_contextConfig->m_loadedFromJson && m_contextConfig->m_zmqEnable)
             {
-                SWSS_LOG_NOTICE("zmq enabled via context config");
+                SWSS_LOG_NOTICE("context %u: JSON zmq_enable=true, ignoring comm-mode attr %d",
```
prabhataravind: When loadedFromJson && zmqEnable and orchagent sends a comm-mode attr other than ZMQ_SYNC, the check ignores it with a NOTICE log. IMO, this should be as follows:

When orchagent requests a different mode → misconfiguration, warn
When orchagent requests ZMQ_SYNC → redundant but consistent, notice

syncd/Syncd.cpp
```diff
@@ -82,16 +84,8 @@ Syncd::Syncd(
         SWSS_LOG_THROW("no context config defined at global context %u", m_commandLineOptions->m_globalContext);
     }
 
-    if (m_contextConfig->m_zmqEnable && m_commandLineOptions->m_enableSyncMode)
-    {
-        SWSS_LOG_NOTICE("disabling command line sync mode, since context zmq enabled");
-
-        m_commandLineOptions->m_enableSyncMode = false;
-
-        m_commandLineOptions->m_redisCommunicationMode = SAI_REDIS_COMMUNICATION_MODE_ZMQ_SYNC;
```
prabhataravind, line 91: The old code promoted "-s" to zmq mode when m_zmqEnable was true but with the new changes, communication mode will remain at the default value. This is a silent behavior change. Please document all these silent behavioral changes in the description.

lib/RedisRemoteSaiInterface.cpp
inder-nexthop: Fixed at the fallback branch in lib/RedisRemoteSaiInterface.cpp. m_redisCommunicationMode is now set to SAI_REDIS_COMMUNICATION_MODE_REDIS_SYNC before constructing the RedisChannel so member state matches the actual channel. Test coverage added in unittest/lib/TestRedisRemoteSaiInterface.cpp::zmqSyncFallsBackToRedisSyncWhenJsonDisablesZmq.

syncd/Syncd.cpp
inder-nexthop: The sairedis-side fallback in lib/RedisRemoteSaiInterface.cpp mirrors the syncd coercion, so the channel transport is consistent on both sides when JSON disables ZMQ. The cross-component drift you describe is the orchagent process global gRedisCommunicationMode, addressed in companion sonic-net/sonic-swss#4556.

lib/RedisRemoteSaiInterface.cpp
inder-nexthop: Addressed in companion sonic-net/sonic-swss#4556. Orchagent now reads context_config.json in initSaiRedis() and demotes gRedisCommunicationMode to REDIS_SYNC before sending the comm-mode attr, so notifications.cpp routes events through the same transport sairedis selects.

lib/RedisRemoteSaiInterface.cpp
inder-nexthop: Split the log into two branches: NOTICE when the caller request matches the JSON-locked ZMQ_SYNC, WARN when the caller requests a different mode while JSON enforces ZMQ (misconfiguration).

syncd/Syncd.cpp
inder-nexthop: Updated the PR description with a new "Behavior changes" section enumerating the silent state changes: default ZMQ endpoints, -z zmq_sync no longer absolute, -s flag no longer rewriting communication mode when JSON enables ZMQ, DisabledRedisClient now DPU-only, and the m_redisCommunicationMode alignment after fallback.

prabhataravind:
@inder-nexthop please don't force push after reviews. It becomes difficult to understand incremental changes.

prabhataravind (approved):
Thanks for addressing the comments. I left a comment in the swss PR #4556.

deepak-singhal0408 (approved):

PR #1839, opened 2026-04-08
Fix wrong acl index used in bind

yue-fred-gao:
### why 
We saw below message. 
2026 Apr  8 05:14:38.971705 tb3-c0-dut1 NOTICE syncd#syncd#syncd: :- aclDefaultCreate: Shared default permit ACL created with swindex 0
2026 Apr  8 05:14:38.987285 tb3-c0-dut1 NOTICE syncd#syncd#syncd: :- aclBindUnbindPort: Default ACL swindex 1 bound to port etp1
vpp_acl_add_replace returns acl swindex 0 but when it was bound to a port, it has changed to 1. The 2 function calls are synchronous and in the same thread. The problem is there are some function in vppxlate are still using W. W is a vpp macro that returns at the first message it received in the binary api socket. If there is unsolicited message, like interface status event, W will return before it process the actual response of its API call. The response will be left in the buffer and when the next time W is called, it will be processed. The above issue is due to this. Some earlier API call using W didn't take the response. When vpp_acl_add_replace is executed, W in the function will get the earlier response so swindex was not the correct ACL index. When vpp_acl_interface_bind_unbind was called, it executed the handler of vpp_acl_add_replace, which set acl swindex to the actual index 1. But vpp_acl_interface_bind_unbind already used 0 to bind the acl to port.

### what this PR does
1. Change all W to WR, which correctly handles unsolicited messages
2. Fix Wong message registration WANT_BFD_EVENTS, WANT_INTERFACE_EVENTS and BFD_UDP_ENABLE_MULTIHOP_REPLY
3. Change vppxlate to use SWSS logger so we can dynamically enable/disable INFO logs from vppxlate
4. Fix vpp_update_router_interface as it doesn't need to remove vrf
5. vpp_router_interface_remove_vrf should remove all IP addresses first otherwise vpp returns VNET_API_ERROR_ADDRESS_FOUND_FOR_INTERFACE (-144)

### DIFF
--- a/vslib/Makefile.am
+++ b/vslib/Makefile.am
@@ -89,6 +89,7 @@ libSaiVS_a_SOURCES +=\
 					  vpp/SwitchVppNbr.cpp \
 					  vpp/SwitchVppSRv6.cpp \
 					  vpp/TunnelManager.cpp \
+					  vpp/SaiVppLog.cpp \
 					  vpp/vppxlate/SaiVppXlate.c \
 					  vpp/vppxlate/SaiVppStats.c \
 					  vpp/vppxlate/SaiAclStats.c \
--- a/vslib/vpp/SaiVppLog.cpp
+++ b/vslib/vpp/SaiVppLog.cpp
@@ -1,0 +1,62 @@
+#include "SaiVppLog.h"
+
+#include "swss/logger.h"
+
+#include <cstdio>
+#include <cstdarg>
+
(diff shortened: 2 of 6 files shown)


yue-fred-gao:
verified using PR https://github.com/sonic-net/sonic-buildimage/pull/26207. All vpp test passed (https://elastictest.org/scheduler/testplan/69d7d9ae2e61bab317c44262)

dypet (approved):

PR #1846, opened 2026-04-14
Fix the wrong RID to VID mapping for south bound notifications with ZMQ

saksarav-nokia:
Fixes the issue reported in https://github.com/sonic-net/sonic-buildimage/issues/26757 (break out ports are not coming up when south bound ZMQ is enabled)

In the current code when the ZMQ is enabled, VID2RID and RID2VID tables are not created in ASIC_DB.
When the breakout ports are created, orchagent bulks them and sairedis creates the Vids for all ports in the bulk (processBulkOidCreate), calls the SAI vendor API to create the port in SAI/ASIC and then updates the m_rid2vid & m_vid2rid maps.
In the same bulk process processBulkOidCreate->onPostPortsCreate (Added as part of https://github.com/sonic-net/sonic-sairedis/pull/1559), it discovers all the new RIDs, creates new VIDs again for these VIDs (please note getVidsForRids called from translateRidsToVids returns SAI_NULL_OBJECT_ID when ZMQ is enabled and it does not check the m_rid2vid map) and updates the m_rid2vid & m_vid2rid maps. However, the first VID added in the map (created in processBulkOidCreate)  is the one passed to orchagent . So Northbound API calls are working. However, when the south bound notifications are received from SAI (ex: port state change notification), the RID to VID mapping fails due to the bug in translateRidToVid (added as part of https://github.com/sonic-net/sonic-sairedis/pull/727) and also the VID creation twice in onPostPortsCreate.
This PR addresses these 2 issues and the RID to VID and also VID to RID translations are happening correctly now and the ports are coming up.

### DIFF
--- a/syncd/VirtualOidTranslator.cpp
+++ b/syncd/VirtualOidTranslator.cpp
@@ -39,7 +39,7 @@ bool VirtualOidTranslator::tryTranslateRidToVid(
         return true;
     }
 
-    auto it = m_rid2vid.find(vid);
+    auto it = m_rid2vid.find(rid);
 
     if (it != m_rid2vid.end())
     {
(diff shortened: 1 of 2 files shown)


saksarav-nokia:
@deepak-singhal0408 for your viz

deepak-singhal0408:
Thanks @saksarav-nokia .. changes look good to me..

Can you also help add unit tests for DisabledRedisClient case, for both tryTranslateRidToVid and translateRidsToVids.

saksarav-nokia:
> Thanks @saksarav-nokia .. changes look good to me..
> 
> Can you also help add unit tests for DisabledRedisClient case, for both tryTranslateRidToVid and translateRidsToVids.

sure. I will add them

deepak-singhal0408 (approved):
Thanks @saksarav-nokia ..

PR #1847, opened 2026-04-14
[dash] Cleanup ENI counters on ENI deletion

theasianpianist:
- When an ENI counter is deleted, also remove counter entries associated with that ENI to prevent storing stale data
- Added a test case to verify ENI and DASH meter counter data are deleted when the parent ENI is removed


prabhataravind (approved):

theasianpianist:
closing as duplicate of https://github.com/sonic-net/sonic-sairedis/pull/1836/changes

PR #1848, opened 2026-04-14
[syncd]: Fall back to TCP connection when unix socket path is not defined

theasianpianist:
## Description
When FlexCounter connects to COUNTERS_DB, it previously always used a unix socket connection. If the target database has no unix socket path configured, this would fail (e.g. syncd is running on a DPU and the target DB is in the remote redis instance on the NPU, the remote redis instance is only reachable via TCP connection)

## Changes
- At `FlexCounter` construction time, check `SonicDBConfig::getDbSock()` for the target DB. If the socket path is empty, set `m_isTcpConn = true`.
- Pass `m_isTcpConn` to all 6 `DBConnector` instantiation sites across `FlexCounter` and its counter context classes (`PortPhyAttrContext`, `PortPhySerdesAttrContext`, `DashMeterCounterContext`).

### DIFF
--- a/syncd/FlexCounter.cpp
+++ b/syncd/FlexCounter.cpp
@@ -14,6 +14,7 @@
 
 #include "swss/redisapi.h"
 #include "swss/tokenize.h"
+#include "swss/dbconnector.h"
 
 using namespace syncd;
 using namespace std;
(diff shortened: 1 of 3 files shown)


prabhataravind (approved):
lgtm. Please add more details to the description as to when the target DB might have no valid sock path.

prsunny:
@saiarcot895 , could you help with merge?

saiarcot895 (approved):

PR #1849, opened 2026-04-15
Update SAI header to v1.18.1

tjchadaga:
Update SAI header to v1.18.1 and add supporting serialization changes


meta/SaiSerialize.cpp
```diff
@@ -297,10 +297,10 @@ sai_status_t transfer_attribute(
...
             break;
 
-//        case SAI_ATTR_VALUE_TYPE_UINT16_LIST:
-//            RETURN_ON_ERROR(transfer_list(src_attr.value.u16list, dst_attr.value.u16list, countOnly));
-//            break;
-//
+        case SAI_ATTR_VALUE_TYPE_UINT16_LIST:
+            RETURN_ON_ERROR(transfer_list(src_attr.value.u16list, dst_attr.value.u16list, countOnly));
+            break;
+
```
lakshmi-nexthop, line 303: @tjchadaga do you know why the support for this UINT16_LIST is added now? I dont see any new attributes that are using this type?

tjchadaga: @lakshmi-nexthop - this is just to support the value type in case it is used by an attribute in the future. Do you see an issue with adding this here?

lakshmi-nexthop: no issue. but I was wondering about the coverage for this code if the value type is not used currently (as theres no attr of this value type).

tjchadaga: only the serialization/deserialization is added for now. Rest of the meta create, set, remove handling can be done once we have a reference to this value type. Coverage currently look ok. Could you please help complete the review?

lakshmi-nexthop (approved):
As long as the test coverage check passes (due to adding SAI_ATTR_VALUE_TYPE_UINT16_LIST), its fine to merge.

PR #1850, opened 2026-04-15
[syncd][bcm] Enable SAI bulk mode in config_syncd_bcm

vpandian-nokia:
## Summary
- Gate `syncd -l` in `config_syncd_bcm()` based on Broadcom ASIC subtype so SAI bulk mode is enabled on XGS and skipped on DNX.
- Keep the change localized to the Broadcom syncd init path and leave other ASIC handlers unchanged.
- DNX skips `syncd -l` because bulk next-hop-group/ECMP member operations are not broadly supported across DNX platforms.

## Test plan
- Build image successfully with the updated `syncd/scripts/syncd_init_common.sh`.
- Boot DUT with the new image.
- Verify live `syncd` command line includes `-l` on XGS platforms and does not include `-l` on DNX platforms.
- Verify `syncd` and `swss` containers remain stable after boot.
- Verify no immediate startup regressions in `docker logs syncd` and `docker logs swss`.
- Run normal route-scale / convergence workload on the DUT.


vpandian-nokia:
@deepak-singhal0408

deepak-singhal0408:
@rlhui @arlakshm for viz.

vpandian-nokia:
Additional DUT performance data for this change:

The measured gain with this change is in the range of about 8% to 10% on a DUT with AMD EPYC 3251 8-Core 2.5GHz and 16 GB DRAM, using route-download workloads of 300K IPv4 or 300K IPv6 routes.

deepak-singhal0408:
@prsunny @prabhataravind for viz.
@arista-nwolfe @eswaran-nexthop for viz.

rlhui:
please confirm sonic-mgmt tests/pfcwd related are passing with this change

deepak-singhal0408:
@vpandian-nokia could you provide more data on sonic-mgmt Nightly tests with this change? And also the CI is failing. please help check.

vpandian-nokia:
> @vpandian-nokia could you provide more data on sonic-mgmt Nightly tests with this change? And also the CI is failing. please help check.

Thanks for the note. We are working on the `sonic-mgmt` / nightly validation for this change, including the earlier `pfcwd`-related ask, and will share the results once they are available.

We are also checking the current CI failure in parallel. The `/azp run` trigger appears to require additional permissions here. Could you please help rerun the Azure pipeline for this PR so we can confirm whether the failure reproduces?

vpandian-nokia:
> > @vpandian-nokia could you provide more data on sonic-mgmt Nightly tests with this change? And also the CI is failing. please help check.
> 
> Thanks for the note. We are working on the `sonic-mgmt` / nightly validation for this change, including the earlier `pfcwd`-related ask, and will share the results once they are available.
> 
> We are also checking the current CI failure in parallel. The `/azp run` trigger appears to require additional permissions here. Could you please help rerun the Azure pipeline for this PR so we can confirm whether the failure reproduces?

@deepak-singhal0408 : The rerun shows the current failure is in `BuildTrixie amd64` during `Run sonic sairedis unit tests`, with `FlexCounter.bulkChunksize` failing. From the unit-test output, the failure is due to repeated `object_count` expectation mismatches within that test.

I also checked recent `master`, and there has been a recent failure in the same `BuildTrixie amd64` unit-test path there as well, although on a different `FlexCounter` test. Since this PR only changes the Broadcom `config_syncd_bcm()` path in `syncd_init_common.sh`, the current Trixie unit-test failure does not appear obviously attributable to this PR.

arlakshm:
@vpandian-nokia, @deepak-singhal0408, as we discussed in the chassis subgroup, could you add some data to determine if the delay in port up/down notifications from sync to orchagent has any impact, especially during device boot-up?

vpandian-nokia:
> @vpandian-nokia, @deepak-singhal0408, as we discussed in the chassis subgroup, could you add some data to determine if the delay in port up/down notifications from sync to orchagent has any impact, especially during device boot-up?

Port up/down notifications follow a different path. Port state change notifications are handled through the SAI notification callback path in `syncd` and then forwarded to `orchagent`, rather than through the bulk SAI operation path.

In the current `syncd` implementation, `syncd -l` is consumed for bulk SAI operations toward vendor SAI and the ASIC, for example bulk route programming requests from `orchagent`. So it does not change the syncd-to-orchagent port up/down notification path.

vpandian-nokia:
@deepak-singhal0408  Latest CI update for this PR:

At present, the failing job is `Azure.sonic-sairedis (Test vstest)`, while `BuildTrixie amd64` is passing.

The change in this PR is narrowly scoped to `syncd/scripts/syncd_init_common.sh`, specifically to enable the `-l` option for Broadcom. Based on the current failure location, the failing path does not appear to directly intersect with the code modified in this PR.

Accordingly, I do not see clear evidence that the observed CI failure is attributable to the change under review here. The present signal appears more consistent with a separate CI/test issue that should be investigated independently.

prsunny (approved):

vpandian-nokia:
> @vpandian-nokia, @deepak-singhal0408, as we discussed in the chassis subgroup, could you add some data to determine if the delay in port up/down notifications from sync to orchagent has any impact, especially during device boot-up?

@arlakshm Port up/down notifications do follow a different functional path. Port state change notifications are handled through the SAI notification callback path in `syncd` and then forwarded to `orchagent`, rather than through the bulk SAI operation path.

In the current `syncd` implementation, `syncd -l` is used for bulk SAI operations toward the vendor SAI and the ASIC. So it does not change the functional `syncd`-to-`orchagent` port up/down notification path directly.

From the current `syncd` implementation, port-state notifications and main ASIC programming work are synchronized through the same `syncd` mutex, so there is no obvious interleaving advantage in the non-bulk case.

In both cases, `syncd` processes both notification handling and ASIC programming work under the same serialized path. The main difference is what happens while that path is held: with `syncd -l`, `syncd` can issue vendor bulk SAI calls for supported batched operations, whereas without `-l`, it still processes the same batch while holding the same mutex, but falls back to handling the entries one by one until the batch is drained.

So based on the implementation, `syncd -l` is more likely to reduce `syncd`-side processing overhead than worsen port-state notification servicing relative to the non-bulk path.

vpandian-nokia:
@deepak-singhal0408,  `sonic-mgmt` nightly validation for this change is still in progress.

The preliminary results helped uncover a missing support case for bulk mode on DNX: bulk next-hop-group/ECMP member operations are not supported there. The latest update in `syncd_init_common.sh` addresses this by gating `syncd -l` based on Broadcom ASIC subtype, keeping it enabled for XGS and skipping it for DNX.
CSP case `CS00012462593` was also raised with Broadcom for the DNX bulk next-hop-group/ECMP support issue.

The full `sonic-mgmt` nightly results will be shared once they are available.

deepak-singhal0408 (approved):
LGTM

prsunny (approved):

prsunny:
@lolyu , would you help merge?

PR #1851, opened 2026-04-15
Ignore SAI_VPP_ERR_VALUE_EXIST from sw_interface_ip6_enable_disable

yue-fred-gao:
### why
sw_interface_ip6_enable_disable returns SAI_VPP_ERR_VALUE_EXIST if ipv6 already enabled on the interface. This could happen in a race condition when SONIC sets ipv6 address in TAP, for example Ethernet1. linux-cp plugin receives the netlink event and enables ipv6 on the PHY interface.

### what this PR does
1. Ignore SAI_VPP_ERR_VALUE_EXIST error from sw_interface_ip6_enable_disable
2. Define SAI_VPP_ERR_ code to match the errors defined in vnet/api_error.h


yue-fred-gao:
The issue was due to a misconfiguration in startup.conf that enables interface creation via netlink.

PR #1858, opened 2026-04-18
vpp: get interface speed from vpp

yue-fred-gao:
### why 
currently saivpp doesn't support get interface speed. It returns NOT_IMPLEMENTED

### what this PR does
interface speed is available in vpp vl_api_sw_interface_dump_t api. We save it in a map with sw_if_index as key. When syncd queries port speed, get the speed from the map if it is available. Otherwise, returns configured rate or 0 if the port is down.

### DIFF
--- a/vslib/vpp/vppxlate/SaiVppXlate.c
+++ b/vslib/vpp/vppxlate/SaiVppXlate.c
@@ -473,6 +473,7 @@ static uintptr_t get_index_ptr (uint32_t idx)
 
 vat_main_t vat_main;
 uword *interface_name_by_sw_index = NULL;
+uword *link_speed_by_sw_index = NULL;
 
 f64
 vat_time_now (vat_main_t * vam)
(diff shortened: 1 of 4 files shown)


vslib/vpp/vppxlate/SaiVppXlate.c
```diff
@@ -2733,6 +2739,32 @@ int interface_get_state (const char *hwif_name, bool *link_is_up)
...
+int vpp_get_interface_speed (const char *hwif_name, uint32_t *speed)
+{
+    vat_main_t *vam = &vat_main;
+    uword *p;
+
+    VPP_LOCK();
+
+    u32 idx = get_swif_idx(vam, hwif_name);
+    if (idx == (u32) -1) {
+	VPP_UNLOCK();
```
AkeelAli: Indentation to match the rest of the file (here and below)

AkeelAli (approved):
LGTM. Minor formatting comment.

vslib/vpp/vppxlate/SaiVppXlate.c
yue-fred-gao: done

PR #1862, opened 2026-04-20
[202511] Smart Counter Poll to allow counters to work properly on Broadcom platforms.

justin-wong-ce:
(Unable to re-open previous PR https://github.com/sonic-net/sonic-sairedis/pull/1755, copying over description)

#### Summary
This is change is to address https://github.com/sonic-net/sonic-sairedis/issues/1753 and the 202511 implementation of the HLD added in https://github.com/sonic-net/SONiC/pull/2190.

Existing 202511 assumes all ports support the same counter capabilities. This causes issues on most Broadcom platform switches as there are different types of ports on a switch that does not support the same set of counters.

Fix by dynamically discovering what each interface is capable of during initialization of `syncd`.

For more detials please refer to the above issue and HLD.

#### Testing
Testing is done on both a `Arista-7060X6-16PE-384C-B-O128S2` and `Arista-7260CX3-D108C8` on 202511 with the tests:
```
sonic-mgmt/tests/dhcp_relay/test_dhcp_counter_stress.py
sonic-mgmt/tests/drop_packets/test_drop_counters.py
sonic-mgmt/tests/drop_packets/test_configurable_drop_counters.py
sonic-mgmt/tests/gnmi/test_gnmi_countersdb.py
sonic-mgmt/tests/snmp/test_snmp_queue_counters.py
```

Due to ongoing warm reboot test issues in 202511 (https://github.com/sonic-net/sonic-swss/pull/4108), warm reboot tests (`
sonic-mgmt/tests/platform_tests/test_advanced_reboot.py`) are instead conducted on 202505 using `Arista-7260CX3-D108C8` with this change backported to it.

A full `sonic-mgmt` test suite run has also been ran. There are no notable fallout compared to `sonic-mgmt` runs without this change. This is ran on both XGS and DNX Broadcom platforms.

#### Performance Impact:
All logic change is only done in the counter initialisation stage of FlexCounters.cpp - there is no polling logic change at all. 
**Therefore, any performance impact is limited to any new execution of `syncd` - i.e. reboot / `config reload` / `systemctl restart`.**

Tested on several topologies:
image

This impact seems reasonable for what this offers. 

The fastest operation that would cause a `syncd` restart is a `systemctl restart swss`, that is a operation that spans minutes. The worst real life scenario on a HwSKU with older CPU and many interfaces takes additional ~17 seconds. New, high interface-count HwSKUs only takes an extra ~2 seconds.

As for memory usage - the impact is on the order of kilobytes, the impact is negligible as the system has GBs of RAM.

### DIFF
--- a/syncd/FlexCounter.cpp
+++ b/syncd/FlexCounter.cpp
@@ -501,6 +501,38 @@ class CounterContext : public BaseCounterContext
 
 protected:
     sai_object_id_t m_switchId = SAI_NULL_OBJECT_ID;
+    struct CounterGroupRef
+    {
+        size_t idx;
+        size_t size;
+    };
+    CounterGroupRef makeCounterGroupRef(size_t idx, size_t size)
+    {
+        SWSS_LOG_ENTER();
+        return {idx, size};
+    }
+    std::set<StatType>* getSortedCGRef(CounterGroupRef const& cgr)
+    {
+        SWSS_LOG_ENTER();
+        return &m_supportedCounterGroups[cgr.idx];
+    }
+    static bool counterGroupRefDscSorter(CounterGroupRef const& lhs, CounterGroupRef const& rhs)
+    {
+        SWSS_LOG_ENTER();
+        // Must adhere to strict weak ordering
+        if (lhs.idx == rhs.idx)
+        {
+            return false; // The sets are equivalent
+        }
+        else if (lhs.size == rhs.size)
+        {
+            return lhs.idx < rhs.idx; // order by earliest creation when size is same
+        }
+        else
+        {
+            return lhs.size > rhs.size;
+        }
+    }
 
 public:
     typedef CounterIds<StatType> CounterIdsType;
@@ -515,6 +547,125 @@ public:
     BaseCounterContext(name, instance), m_objectType(object_type), m_vendorSai(vendor_sai), m_groupStatsMode(stats_mode)
     {
         SWSS_LOG_ENTER();
+    }
+
+    virtual void addObjectWithCounterGroups(
+            _In_ sai_object_id_t vid,
+            _In_ sai_object_id_t rid,
+            _In_ const std::vector<std::string> &idStrings,
+            _In_ const std::string &per_object_stats_mode) override
+    {
+        SWSS_LOG_ENTER();
+        sai_stats_mode_t instance_stats_mode = SAI_STATS_MODE_READ_AND_CLEAR;
+        sai_stats_mode_t effective_stats_mode;
+        // TODO: use if const expression when c++17 is supported
+        if (HasStatsMode<CounterIdsType>::value)
+        {
+            if (per_object_stats_mode == STATS_MODE_READ_AND_CLEAR)
+            {
+                instance_stats_mode = SAI_STATS_MODE_READ_AND_CLEAR;
+            }
+            else if (per_object_stats_mode == STATS_MODE_READ)
+            {
+                instance_stats_mode = SAI_STATS_MODE_READ;
+            }
+            else
+            {
+                SWSS_LOG_WARN("Stats mode %s not supported for flex counter. Using STATS_MODE_READ_AND_CLEAR", per_object_stats_mode.c_str());
+            }
+
+            effective_stats_mode = (m_groupStatsMode == SAI_STATS_MODE_READ_AND_CLEAR ||
+                                    instance_stats_mode == SAI_STATS_MODE_READ_AND_CLEAR) ? SAI_STATS_MODE_READ_AND_CLEAR : SAI_STATS_MODE_READ;
+        }
+        else
+        {
+            effective_stats_mode = m_groupStatsMode;
+        }
+
+        std::vector<StatType> counter_ids;
+        for (const auto &str : idStrings)
+        {
+            StatType stat;
+            deserializeStat(str.c_str(), &stat);
+            counter_ids.push_back(stat);
+        }
+
+        std::set<StatType> counter_ids_set = setupBaseCounterGroup(rid, counter_ids, effective_stats_mode);
+        counter_ids = std::vector<StatType>(counter_ids_set.begin(), counter_ids_set.end());
+        std::sort(counter_ids.begin(), counter_ids.end());
+        updateSupportedCounterGroups(rid, vid, counter_ids, effective_stats_mode);
+
+        if (m_objectSupportedCountersGroupMap.count(vid) == 0)
+        {
+            return; // This vid has no supported counters
+        }
+        size_t groupIndex = m_objectSupportedCountersGroupMap[vid];
+        std::set<StatType>& supportedIdsSet = m_supportedCounterGroups[groupIndex];
+        if (supportedIdsSet.empty())
+        {
+            return;
+        }
+        std::vector<StatType> supportedIds(supportedIdsSet.begin(), supportedIdsSet.end());
+
+        if (double_confirm_supported_counters)
+        {
+            std::vector<uint64_t> stats(supportedIds.size());
+
+            if (!collectData(rid, supportedIds, effective_stats_mode, false, stats))
+            {
+                SWSS_LOG_ERROR("%s RID %s VID %s can't provide the statistic",  m_name.c_str(),
+                               sai_serialize_object_id(rid).c_str(), sai_serialize_object_id(vid).c_str());
+                throw std::runtime_error("Test counter poll failed on populating m_objectIdsMap");
+            }
+        }
+
+        // Perform a remove and re-add to simplify the logic here
+        removeObject(vid, false);
+
+        bool supportBulk;
+        // TODO: use if const expression when cpp17 is supported
+        if (HasStatsMode<CounterIdsType>::value)
+        {
+            supportBulk = false;
+        }
+        else
+        {
+            supportBulk = no_double_check_bulk_capability || checkBulkCapability(vid, rid, supportedIds);
+        }
+
+        if (!supportBulk)
+        {
+            auto counter_data = std::make_shared<CounterIds<StatType>>(rid, supportedIds);
+            // TODO: use if const expression when cpp17 is supported
+            if (HasStatsMode<CounterIdsType>::value)
+            {
+                counter_data->setStatsMode(instance_stats_mode);
+            }
+            m_objectIdsMap.emplace(vid, counter_data);
+        }
+        else if (m_counterChunkSizeMapFromPrefix.empty())
+        {
+            std::sort(supportedIds.begin(), supportedIds.end());
+            auto bulkContext = getBulkStatsContext(supportedIds, "default", default_bulk_chunk_size);
+            addBulkStatsContext(vid, rid, supportedIds, *bulkContext.get());
+        }
+        else
+        {
+            std::map<std::string, vector<StatType>> counter_prefix_map;
+            std::vector<StatType> default_partition;
+            createCounterBulkChunkSizePerPrefixPartition(supportedIds, counter_prefix_map, default_partition);
+
+            for (auto &counterPrefix : counter_prefix_map)
+            {
+                std::sort(counterPrefix.second.begin(), counterPrefix.second.end());
+                auto bulkContext = getBulkStatsContext(counterPrefix.second, counterPrefix.first, m_counterChunkSizeMapFromPrefix[counterPrefix.first]);
+                addBulkStatsContext(vid, rid, counterPrefix.second, *bulkContext.get());
+            }
+
+            std::sort(default_partition.begin(), default_partition.end());
+            auto bulkContext = getBulkStatsContext(default_partition, "default", default_bulk_chunk_size);
+            addBulkStatsContext(vid, rid, default_partition, *bulkContext.get());
+        }
     }
 
     // For those object type who support per object stats mode, e.g. buffer pool.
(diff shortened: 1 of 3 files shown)


yxieca:
Master branch PR: https://github.com/sonic-net/sonic-sairedis/pull/1774

yxieca:
@justin-wong-ce , please help address the merge conflict in unit test.

justin-wong-ce:
> @justin-wong-ce , please help address the merge conflict in unit test.

Conflicts resolved, please review, thanks.

syncd/FlexCounter.cpp
```diff
@@ -520,6 +520,38 @@ class CounterContext : public BaseCounterContext
...
+    }
+    std::set<StatType>* getSortedCGRef(CounterGroupRef const& cgr)
+    {
+        SWSS_LOG_ENTER();
+        return &m_supportedCounterGroups[cgr.idx];
+    }
+    static bool counterGroupRefDscSorter(CounterGroupRef const& lhs, CounterGroupRef const& rhs)
+    {
+        SWSS_LOG_ENTER();
+        // Must adhere to strict weak ordering
```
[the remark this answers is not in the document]
justin-wong-ce: This is required by the build CI check.

syncd/FlexCounter.cpp
```diff
@@ -1141,9 +1547,23 @@ class CounterContext : public BaseCounterContext
...
+                }
+                else if (m_failedPolls[{rid, vid}] < 3)
+                {
+                    m_failedPolls[{rid, vid}] += 1;
+                    SWSS_LOG_DEBUG("counter read failed %d times on RID 0x%" PRIx64 " on intf 0x%" PRIx64, m_failedPolls[{rid, vid}], rid, vid);
+                }
+                else
+                {
+                    SWSS_LOG_ERROR("counter read failed more than 3 times on RID 0x%" PRIx64 " on intf 0x%" PRIx64, rid, vid);
+                }
```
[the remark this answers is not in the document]
justin-wong-ce: This is by design - poll failures are rare when things are working as intended.

Having to check the map on every single poll is detrimental to performance.
Something is seriously wrong if polling fails with many vid-rid pairs, fails more than 3 times, and persists - the switch should be checked immediately and triaged in that case.

Note any `config reload` or `reboot` would restart `syncd`, which clears the map.

syncd/FlexCounter.cpp
```diff
@@ -536,6 +568,125 @@ class CounterContext : public BaseCounterContext
...
+            if (!collectData(rid, supportedIds, effective_stats_mode, false, stats))
+            {
+                SWSS_LOG_ERROR("%s RID %s VID %s can't provide the statistic",  m_name.c_str(),
+                               sai_serialize_object_id(rid).c_str(), sai_serialize_object_id(vid).c_str());
+                throw std::runtime_error("Test counter poll failed on populating m_objectIdsMap");
+            }
+        }
+
+        // Perform a remove and re-add to simplify the logic here
+        removeObject(vid, false);
```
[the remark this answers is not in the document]
justin-wong-ce: This is leftover logic from the previous implementation - this is no longer needed in this updated version. Removing.

syncd/FlexCounter.cpp
```diff
@@ -1521,6 +1994,118 @@ class CounterContext : public BaseCounterContext
...
+            {
+                // Success - Check support for counters not in counter group (extra counters)
+                std::vector<StatType> extraCounters;
+                std::set<StatType> newCounters;
+
+                // Counter groups are a subset of counter_ids in any situation
+                // Vectors need to be sorted for set_difference for defined behavior - counter_ids are already sorted
+                std::sort(countersToPoll.begin(), countersToPoll.end());
+                std::set_difference(counter_ids.begin(), counter_ids.end(), countersToPoll.begin(), countersToPoll.end(),
+                                    std::inserter(extraCounters, extraCounters.begin()));
```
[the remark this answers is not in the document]
justin-wong-ce: Ack, swapping to `back_inserter`.

unittest/syncd/TestFlexCounter.cpp
```diff
@@ -2316,6 +2316,186 @@ TEST(FlexCounter, noEniDashMeterCounter)
...
+    };
+
+    // Force non-bulk path so per-port counter groups are polled individually.
+    sai->mock_bulkGetStats = [](sai_object_id_t, sai_object_type_t, uint32_t,
+                                const sai_object_key_t *, uint32_t,
+                                const sai_stat_id_t *, sai_stats_mode_t,
+                                sai_status_t *, uint64_t *)
+    {
+        return SAI_STATUS_FAILURE;
+    };
```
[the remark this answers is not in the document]
justin-wong-ce: Added context for the test.

syncd/FlexCounter.cpp
```diff
@@ -536,6 +568,130 @@ class CounterContext : public BaseCounterContext
...
+            }
+        }
+
+        // Perform a remove and re-add to simplify the logic here
+        // This remove function removes the counter from m_objectIdsMap and clears counter group mappings:
+        // - There is special handling logic for m_objectIdsMap below.
+        // - Counter group mappings are no longer needed for this vid since it is already stored in supportedIds.
+        // - This vid will also not have the same counter_ids for discovery later on, except from intentional discoveries
+        //   via addObject, which we would want to rediscover anyways.
+        removeObject(vid, false);
```
lolyu: 🔴 **Major — this defeats the group cache for the single-add path.** `updateSupportedCounterGroups()` just created/assigned the per-vid group above, but `removeObject(vid, false)` now unconditionally calls `cleanupCounterGroupMapping(vid)` (see L1490+). When this vid is the **sole referencer** of the freshly-discovered group — which is exactly the case for the *first* port of each capability class — cleanup erases the mapping, drops the ref from `m_counterGroupsSorted`, and `.clear()`s the set. So the group is destroyed before the next port can match it, forcing every subsequent port back into per-counter probing instead of a single matching poll. The local `supportedIds` is a copy so *this* vid still polls correctly (which is why the unit test passes — it only asserts final values, not group reuse), but the cross-port cache benefit is lost. This is Copilot's earlier C1; the reply said "Removing" but it's still here at HEAD. Suggest splitting removal into "drop polling state" vs "drop group mapping", and not invoking `cleanupCounterGroupMapping` from this internal remove/re-add.

syncd/FlexCounter.cpp
```diff
@@ -1101,6 +1471,45 @@ class CounterContext : public BaseCounterContext
...
+        if (!referenced)
+        {
+            // Remove from sorted list so updateSupportedCounterGroups won't match
+            m_counterGroupsSorted.erase(
+                std::remove_if(m_counterGroupsSorted.begin(), m_counterGroupsSorted.end(),
+                    [removedGroupIdx](const CounterGroupRef& ref) { return ref.idx == removedGroupIdx; }),
+                m_counterGroupsSorted.end());
+
+            // Clear the group data (can't erase from vector without invalidating other indices)
+            m_supportedCounterGroups[removedGroupIdx].clear();
```
lolyu, line 1481: 🟡 **Minor — empty group slots are never reclaimed.** `m_supportedCounterGroups[removedGroupIdx].clear()` empties the set but leaves the slot in place (per the comment, to avoid invalidating other indices). Over syncd's lifetime with port flap / repeated `config reload` churn, `m_supportedCounterGroups` accumulates empty sets and `updateSupportedCounterGroups` still iterates them on every match attempt. Not blocking, but worth either a periodic compaction (rebuild the vector + remap indices) or a free-list of reusable slots. At minimum, the empty slots add work to the match loop.

syncd/FlexCounter.cpp
```diff
@@ -1141,9 +1552,23 @@ class CounterContext : public BaseCounterContext
...
+                    SWSS_LOG_DEBUG("counter read failed 1 time on RID 0x%" PRIx64 " on intf 0x%" PRIx64, rid, vid);
+                }
+                else if (m_failedPolls[{rid, vid}] < 3)
+                {
+                    m_failedPolls[{rid, vid}] += 1;
+                    SWSS_LOG_DEBUG("counter read failed %d times on RID 0x%" PRIx64 " on intf 0x%" PRIx64, m_failedPolls[{rid, vid}], rid, vid);
+                }
+                else
+                {
+                    SWSS_LOG_ERROR("counter read failed more than 3 times on RID 0x%" PRIx64 " on intf 0x%" PRIx64, rid, vid);
```
lolyu, line 1545: 🟡 **Minor — ERROR is logged on *every* poll once the count exceeds 3.** I take the point from the earlier thread that not clearing `m_failedPolls` on success is a deliberate perf choice (avoid a map lookup per poll) — fair. But because the counter never resets, this `else` branch emits `SWSS_LOG_ERROR` every poll interval, indefinitely, for a single stuck (rid,vid). That's a syslog flood for one bad counter. Keeping your design, consider firing the ERROR **once** (e.g. only when `m_failedPolls[{rid,vid}] == 3`, then a final increment so it doesn't re-fire) or rate-limiting it. That preserves the "escalate after 3" signal without per-poll spam.

syncd/FlexCounter.cpp
```diff
@@ -520,6 +520,38 @@ class CounterContext : public BaseCounterContext
...
+    {
+        SWSS_LOG_ENTER();
+        return {idx, size};
+    }
+    std::set<StatType>* getSortedCGRef(CounterGroupRef const& cgr)
+    {
+        SWSS_LOG_ENTER();
+        return &m_supportedCounterGroups[cgr.idx];
+    }
+    static bool counterGroupRefDscSorter(CounterGroupRef const& lhs, CounterGroupRef const& rhs)
```
lolyu: 🟡 **Minor (Copilot C3) — `SWSS_LOG_ENTER()` inside a `std::sort` comparator.** The comparator runs O(n log n) times per sort and `SWSS_LOG_ENTER` records on each call, which adds up during init on high-port-count SKUs. You noted the build lint requires it — if it can't be satisfied another way, a one-line comment explaining *why* it's here (lint requirement, not by choice) would save the next reader; ideally make this a local lambda the lint doesn't flag so the hot comparator stays side-effect-free.

unittest/syncd/TestFlexCounter.cpp
```diff
@@ -2316,6 +2316,188 @@ TEST(FlexCounter, noEniDashMeterCounter)
...
+
+    // Skip HW capability query; counter support is determined by getStats probing.
+    sai->mock_queryStatsCapability = [](sai_object_id_t, sai_object_type_t,
+                                        sai_stat_capability_list_t *)
+    {
+        return SAI_STATUS_FAILURE;
+    };
+
+    // Force non-bulk path so per-port counter groups are polled individually.
```
lolyu, line 2408: 🔴 **Major — the tested path isn't the Broadcom production path.** This forces `mock_bulkGetStats → FAILURE` to exercise the non-bulk path, but Broadcom (the whole motivation for this PR) polls via **bulk** by default, through `bulkAddObjectWithCounterGroups()` — a separate implementation that picks the largest group (`m_counterGroupsSorted[0]`) and falls back per-object via `bulkUnsupportedCounters`. None of that logic is covered here. The realistic production case is ports with **disjoint** capability sets (your Port 1 already supports a set that isn't a subset of Port 3's) under the bulk path. Could you add a bulk-path variant of this test with disjoint per-port capabilities? That's where a group-discovery regression would actually bite on the target platforms.

lolyu (commented):
**Review:** https://github.com/sonic-net/sonic-sairedis/pull/1862

Thanks for this — per-port dynamic counter discovery is the right fix for the heterogeneous-port-capability problem on Broadcom, and the hardware testing across two Arista SKUs plus a full sonic-mgmt run is solid diligence. Comments are non-blocking; none of these corrupt COUNTERS_DB (polled values are correct). They're about cache effectiveness, test coverage of the production path, and log/memory hygiene.

Tracking the earlier Copilot review against current HEAD: the `back_inserter` fix (✅) and the test non-bulk note (✅) are in. Three remain open and I've expanded on the two that matter most:

**Main points**
1. 🔴 The single-add path defeats its own group cache: after `updateSupportedCounterGroups()` assigns the group, `removeObject(vid, false)` → `cleanupCounterGroupMapping(vid)` can erase the just-created group when this vid is its only referencer (exactly the first port of each capability class). Subsequent ports then re-probe individually instead of matching in one poll. (This is Copilot's C1 — you replied "Removing" on Jun 5 but it's still here at HEAD.)
2. 🔴 The new `dynamicCounterGroups` test forces the **non-bulk** path, but Broadcom — the entire motivation — polls via **bulk**, which is a separate code path (`bulkAddObjectWithCounterGroups`) with different discovery logic. The production scenario (disjoint per-port capabilities under bulk) is untested.
3. 🟡 Three hygiene items inline (empty-slot leak, ERROR log flood >3×, comparator logging).

Details inline.

syncd/FlexCounter.cpp
justin-wong-ce: `config reload` would cause `swss` and` syncd` to restart, which means all the counter capabilities would be lost and needs to be re-discovered.  Port flap should resuse the previous vid.

I don't think it is crucial for this to exist but there is value in having this functionality.
Off the top of my head, adding this functionality would mean adding some more logic to `syncd`, adding changes to polling logic (not modified here), and potentially new config for how often this compacting should occur - way out of scope of this review.

This would likely need to be a follow-up or second stage change.

syncd/FlexCounter.cpp
justin-wong-ce: I will be adding some checks to run this logic only in the case where desired counter group is different from what has been passed before. I have some changes that I am running extended testing on (sonic-mgmt).testing.

unittest/syncd/TestFlexCounter.cpp
justin-wong-ce: Test case added.

syncd/FlexCounter.cpp
justin-wong-ce: Moved to using in-line lambda

syncd/FlexCounter.cpp
justin-wong-ce: Changed logic to only log ERR once.

justin-wong-ce:
Running extended testing on physical DUT with sonic-mgmt.

StormLiangMS:
hi @lolyu remind for review.

lolyu (approved):
LGTM

vmittal-msft:
@volodymyrsamotiy @liat-grozovik this is proper fix for counter issue. Can you please check if you still see config reload/cold boot issue with this change ?

vmittal-msft:
@volodymyrsamotiy @liat-grozovik can you please help check this for your reload issue so we can make a call for 202511 ?
@Javier-Tan @abdosi for viz

vmittal-msft:
@volodymyrsamotiy @liat-grozovik please help check above

PR #1870, opened 2026-04-24
vslib/vpp: register hostif info for link-state updates

janpal23:
## Summary
- register VPP hostif info during hostif creation
- allow runtime link events to match the hostif ifindex
- clean up hostif bookkeeping on remove

## Problem
On the VPP path, hostif creation did not populate `m_hostif_info_map`. As a result, runtime `RTM_NEWLINK` events were dropped by `syncOnLinkMsg()` because `hasIfIndex()` could not match the interface, so no SAI port-state-change notification was sent and `oper_status` in APPL_DB could remain stale.

## Fix
Call `hostif_create_tap_veth_forwarding(name, tapfd, obj_id)` from `SwitchVpp::vs_create_hostif_tap_interface()` and remove the matching hostif bookkeeping in `SwitchVpp::vs_remove_hostif_tap_interface()`.

### DIFF
--- a/vslib/vpp/SwitchVppHostif.cpp
+++ b/vslib/vpp/SwitchVppHostif.cpp
@@ -218,6 +218,41 @@ bool SwitchVpp::hostif_create_tap_veth_forwarding(
     return true;
 }
 
+bool SwitchVpp::register_hostif_info(
+        _In_ const std::string &tapname,
+        _In_ int tapfd,
+        _In_ sai_object_id_t port_id)
+{
+    SWSS_LOG_ENTER();
+
+    // VPP uses the TAP directly through Linux CP. No packet socket or veth
+    // forwarding is needed here; syncOnLinkMsg() only needs this ifindex
+    // to accept runtime RTM_NEWLINK events for this hostif.
+    int ifindex = if_nametoindex(tapname.c_str());
+    if (ifindex == 0)
+    {
+        SWSS_LOG_ERROR("failed to get interface index for %s", tapname.c_str());
+
+        return false;
+    }
+
+    m_hostif_info_map[tapname] =
+        std::make_shared<HostInterfaceInfo>(
+                ifindex,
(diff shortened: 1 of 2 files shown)


vslib/vpp/SwitchVppHostif.cpp
```diff
@@ -389,6 +389,13 @@ sai_status_t SwitchVpp::vs_create_hostif_tap_interface(
         return SAI_STATUS_FAILURE;
     }
 
+    if (!hostif_create_tap_veth_forwarding(name, tapfd, obj_id))
```
yue-fred-gao: hostif_create_tap_veth_forwarding does a lot of things that are not needed for vpp. What you need is only register the interface in m_hostif_info_map. Can you remove the socket related code and rename the function pertaining to registering host interface in map?

vslib/vpp/SwitchVppHostif.cpp
```diff
@@ -462,32 +469,23 @@ sai_status_t SwitchVpp::vs_remove_hostif_tap_interface(
...
-    // m_hostif_info_map.erase(it);
-
-    // remove interface mapping
+    sai_object_id_t port_id = getPortIdFromIfName(name);
+    auto it = m_hostif_info_map.find(name);
 
-    // std::string vname = vpp_get_veth_name(name, info->m_portId);
+    if (it != m_hostif_info_map.end())
+    {
+        SWSS_LOG_NOTICE("attempting to remove host info entry for tap device: %s", name.c_str());
```
yue-fred-gao: Please lower it to INFO

yue-fred-gao (commented):
Please create a sonic-buildimage PR with sonic-sairedis submodule pointing to this PR branch. That will trigger sonic-vpp t1-lag sanity. Please share the result to make sure it doesn't break sanity

yue-fred-gao (approved):

PR #1871, opened 2026-04-24
[meta] Skip validation for mandatory deprecated attributes in Meta.

ksravani-hcl:
What i did:
Updated the generic object creation validation logic in Meta.cpp to check the isdeprecated flag of SAI attributes. If an attribute is marked as both mandatory and deprecated, skips the requirement check instead of potentially failing the operation.

Why i did:
In SAI, some legacy attributes are marked as mandatory for backward compatibility but are simultaneously deprecated in favor of newer mechanisms. Enforcing mandatory checks on deprecated attributes can cause unnecessary validation failures in orchestration agents that have already migrated to modern attributes, hindering smooth transitions and system stability.

How i did:
Verified the persistence mechanism through the local build and UT.

### DIFF
--- a/meta/Meta.cpp
+++ b/meta/Meta.cpp
@@ -3896,6 +3896,13 @@ sai_status_t Meta::meta_generic_validation_create(
             continue;
         }
 
+        if (md.isdeprecated)
+        {
+            // skip deprecated attributes
+            META_LOG_NOTICE(md, "attribute is mandatory but deprecated, skipping validation");
+            continue;
+        }
+
         const auto &it = attrs.find(md.attrid);
 
         if (it == attrs.end())


kishanps:
@ksravani-hcl @sreeiyer-1 Please fix the verification part in the description notes, there is no Bazel env here.

mholankar (approved):

mholankar:
@kcudnik can you please approve this PR?

kcudnik:
in description you said that if attribute is depreacated then LOG will be added, but no log message exists in this PR, also was this discussed further ? did all vendors agreed to skip this or all depreacated attributes from specifisc SAI version ?

ksravani-hcl:
> in description you said that if attribute is depreacated then LOG will be added, but no log message exists in this PR, also was this discussed further ? did all vendors agreed to skip this or all depreacated attributes from specifisc SAI version ?

Corrected the log. This is discussed only internally but the change is applicable only to SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_VR_ID in saitunnel.h (https://github.com/opencomputeproject/SAI/blob/c72635e6ca35371ba0bd82d36a5b2c500c021e16/inc/saitunnel.h#L995) which is the only attribute in the SAI/inc that is both mandatory and deprecated.

kcudnik:
i would add warning line or at least NOTICE in this case taht any attribute as depreacated, as i sspect not all vendors will initially support that right away

meta/Meta.cpp
```diff
@@ -3896,6 +3896,11 @@ sai_status_t Meta::meta_generic_validation_create(
             continue;
         }
 
+        if (md.isdeprecated) {
+            // skip deprecated attributes
```
kcudnik, line 3901: add log here warning or notice that specific attribute was mandatory on create and depreacated

kcudnik (dismissed):

meta/Meta.cpp
```diff
@@ -3896,6 +3896,12 @@ sai_status_t Meta::meta_generic_validation_create(
             continue;
         }
 
+        if (md.isdeprecated) {
```
kcudnik: move { to new line

ksravani-hcl: moved!

meta/Meta.cpp
ksravani-hcl: added a log notice.

kcudnik (approved):

ksravani-hcl:
@kcudnik
as all the checks are successful, Could you please merge this PR.

kcudnik (approved):

PR #1881, opened 2026-05-01
zmq: reset ZMQ_REQ socket when zmq_poll times out

### LINKED ISSUE #26300: Bug: orchagent CrmOrch hits zmq timeout and crashes during reboots longer than 1m
### Is it platform specific

generic

### Importance or Severity

Medium

## Errors/crash summary

During reboot, the `reboot` script ( [source](https://github.com/sonic-net/sonic-utilities/blob/6b47fed6979ffa52f369e7d5d2f88a183fe63169/scripts/reboot#L94-L120) ) gracefully shuts down syncd/SAI, but does not shut down swss/orchagent.  There is a problem in this scenario, involving a zmq message queue between orchagent and SAI. The SAI side of the queue gets shut down. However, orchagent, specifically CrmOrch, continues to send requests for counters to SAI over the queue, and these requests will eventually time out. (A description of that program flow is given further below in this issue ticket.)

The zmq request timeout is 60s.  [  [channel creation with timeout](https://github.com/sonic-net/sonic-sairedis/blob/9a87a71c80a74f7dc997da9cee7aec8bb9fd8b18/lib/Channel.cpp#L12) ] [ [timeout value](https://github.com/sonic-net/sonic-sairedis/blob/9a87a71c80a74f7dc997da9cee7aec8bb9fd8b18/lib/sairedis.h#L46) ]

The CrmOrch polling interval is 300s.  [  [timer creation](https://github.com/sonic-net/sonic-swss/blob/c20ded7bcef07eb68995aa2347f39c4bc22c9191/orchagent/crmorch.cpp#L402) ] [ [timer interval default value](https://github.com/sonic-net/sonic-swss/blob/c20ded7bcef07eb68995aa2347f39c4bc22c9191/orchagent/crmorch.cpp#L12) ] (Also available in CLI command `crm show summary`.)

So if reboot extends past 1 minute, orchagent will log errors and crash some time in the next 5 minutes after that.  The syslog will contain messages like the following:

```
2026 Mar 19 20:06:17.973495 sonic ERR swss#orchagent: :- wait: zmq_poll timed out for: object_type_get_availability_response
2026 Mar 19 20:06:17.973495 sonic ERR swss#orchagent: :- getResAvailability: Failed to get availability counter for DASH_VNET CRM resource
2026 Mar 19 20:06:17.973495 sonic ERR swss#orchagent: :- set: zmq_send failed, on endpoint tcp://127.0.0.1:5555, zmqerrno: 156384763: Operation cannot be accomplished in current state
2026 Mar 19 20:06:17.976515 sonic INFO swss#supervisord: orchagent terminate called after throwing an instance of 'std::runtime_error'
2026 Mar 19 20:06:17.976916 sonic INFO swss#supervisord: orchagent   what():  :- set: zmq_send failed, on endpoint tcp://127.0.0.1:5555, zmqerrno: 156384763: Operation cannot be accomplished in current state
```
We see 2 different kinds of unwanted ERR level messages: the first few messages are directly from the timeout, and the last messages are from a crash which is a secondary effect triggered by the timeout.

 This program control flow that leads to these messages is described in more detail, below:

## Program flow details

The orchagent is running CrmOrch, which periodically (300 seconds) calls the `getResAvailableCounters()` function  ( [code](https://github.com/sonic-net/sonic-swss/blob/c20ded7bcef07eb68995aa2347f39c4bc22c9191/orchagent/crmorch.cpp#L878C15-L878C38) ). This function sequentially fetches many counters from SAI via the zmq. In the log above, this function is called after syncd/SAI is shut down, and the first request for counters to be put on the queue is for CRM_DASH_VNET, and it times out.  (Note: The other CrmResourceTypes that appear earlier in the list in `getResAvailableCounters()` get disabled at the beginning of the program execution when SAI replies with SAI_STATUS_NOT_SUPPORTED or similar.)

After the timeout, the state machine for the queue gets into a bad state. When `getResAvailableCounters()` continues on and attempts to send the next request for the next counter, the internal zmq state machine is still expecting to receive a message, and the program crashes due to this conflict - emitting the lines at the bottom of the log snippet. That error comes from the ZeroMQ implementation, similar to this line: 

https://github.com/zeromq/libzmq/blob/7d95ac02ae1af9a2f6896794d12b9de2ae30a3a8/src/req.cpp#L34

### Steps to Reproduce

Once syncd has been terminated by reboot, the issue should reproduce within 6 minutes, as long as the switch does not shut down.

Suggestions for how to do this:

* NVIDIA/Mellanox SmartSwitches - I am running "reboot -p" on the dpu, which only triggers the pre-reboot steps. Then wait.  A more realistic scenario is to perform a sonic-to-sonic upgrade, and reboot. Sometimes the firmware update process that is triggered during the reboot will be long enough for the timeout to occur, and the messages will appear in the DPU's syslog.

* Mellanox Switches (not SmartSwitches) - I don't think it's possible to reproduce without hacking because there is a special case in reboot within the `stop_sonic_services()` function to skip the graceful shutdown of syncd. Shutting down syncd might cause other issues.

* Other Platforms - Assuming the switch uses ZeroMQ for orchagent->syncd communication ("synchronous mode"?), you may be able to reproduce by editing `reboot` to exit or sleep after `stop_sonic_services()` is called.

Actual behavior:

1) orchagent CrmOrch continues to send messages to SAI, which is no longer running.
2) orchagent times out, emmitting unwanted ERR level messages to syslog. 
3) orchagent crashes, and logs more bad messages.

Expected behvior.

1) No ERR level logs.
2) The requests from orchagent to SAI stop before stopping the syncd/SAI container, so that timeouts do not occur, and
3) orchagent shouldn't crash.

### Relevant log output

```shell
2026 Mar 19 20:06:17.973495 sonic ERR swss#orchagent: :- wait: zmq_poll timed out for: object_type_get_availability_response
2026 Mar 19 20:06:17.973495 sonic ERR swss#orchagent: :- getResAvailability: Failed to get availability counter for DASH_VNET CRM resource
2026 Mar 19 20:06:17.973495 sonic ERR swss#orchagent: :- set: zmq_send failed, on endpoint tcp://127.0.0.1:5555, zmqerrno: 156384763: Operation cannot be accomplished in current state
2026 Mar 19 20:06:17.976515 sonic INFO swss#supervisord: orchagent terminate called after throwing an instance of 'std::runtime_error'
2026 Mar 19 20:06:17.976916 sonic INFO swss#supervisord: orchagent   what():  :- set: zmq_send failed, on endpoint tcp://127.0.0.1:5555, zmqerrno: 156384763: Operation cannot be accomplished in current state
```

```shell
This occurs on the 202511 release candidate branch.
```

### Attach files (if any)

_No response_

prabhataravind:
### Description of PR

Summary:
Fixes https://github.com/sonic-net/sonic-buildimage/issues/26300

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

##### Work item tracking

#### How did you do it?
Reset ZMQ_REQ socket when zmq_poll times out.

Leaving the socket in a bad state can cause can orchagent to crash with runtime_error when crmorch (or any orch) calls zmq_send() on it.

Impacts only platforms that use zmq for orchagent -> syncd communication

### Documentation

``` 
- zmq_poll times out in wait() → logs the error and returns SAI_STATUS_FAILURE without receiving the response. This leaves the ZMQ REQ socket stuck in the "waiting for recv" state.
  
- The caller (orchagent CRM) logs the DASH_VNET failure, then continues to the next SAI operation.
  
- The next set() call tries zmq_send on the same REQ socket, but ZMQ's REQ/REP pattern enforces strict send→recv→send→recv ordering. Since the recv was never done after the previous send, zmq_send fails with EFSM (156384763 = ZMQ_EFSM — "Operation cannot be accomplished in current state").

- SWSS_LOG_THROW fires → std::runtime_error → orchagent crashes.
```

### DIFF
--- a/lib/ZeroMQChannel.cpp
+++ b/lib/ZeroMQChannel.cpp
@@ -280,10 +280,24 @@ sai_status_t ZeroMQChannel::wait(
 
         if (rc == 0)
         {
-            SWSS_LOG_ERROR("zmq_poll timed out for: %s", command.c_str());
+            // REQ socket is now stuck in WAITING FOR REPLY state because
+            // zmq_send was done but zmq_recv did not complete. Close and
+            // recreate the socket to reset the state machine (Issue #26300).
 
-            // notice, at this point we could throw, since in REP/REQ pattern
-            // we are forced to use send/recv in that specific order
+            SWSS_LOG_ERROR("zmq_poll timed out for: %s, resetting REQ socket", command.c_str());
+
+            zmq_close(m_socket);
+
+            m_socket = zmq_socket(m_context, ZMQ_REQ);
+
+            int connect_rc = zmq_connect(m_socket, m_endpoint.c_str());
+
+            if (connect_rc != 0)
+            {
+                SWSS_LOG_THROW("failed to reconnect zmq endpoint %s after timeout, zmqerrno: %d",
+                        m_endpoint.c_str(),
+                        zmq_errno());
+            }
 
             return SAI_STATUS_FAILURE;
         }


prsunny (approved):

prabhataravind:
hi @saiarcot895 could you help merge?

lolyu (approved):

PR #1860, opened 2026-04-19
[vpp] add ipinip tunnel encap/decap support

lolyu:
Add VPP ipinip tunnel encap/decap support.
Please refer to HLD to design details: https://github.com/sonic-net/sonic-platform-vpp/pull/222.

```
 tests/decap/test_decap.py::test_decap ✓                                                                                                                                                                         100%
```

* Microsoft ADO (number only): 37902564

Signed-off-by: Longxiang Lyu <lolv@microsoft.com>

PR #1883, opened 2026-05-05
[vpp] Fix LAG subinterface hwif name resolution

### LINKED ISSUE #227: PortChannel/LAG subinterfaces are not working on SONiC-VPP
PortChannel/LAG subinterfaces are not working on SONiC-VPP.

This was reproduced by enabling the sonic-mgmt `port_in_lag` parameter for the sub-port forwarding test:

```text
sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_packet_routed_with_valid_vlan[port_in_lag]
```

Environment from that run:

```text
Topology: t1-lag-vpp
Platform: kvm
ASIC type: vpp
DUT: vlab-vpp-01
```

Observed result:

```text
test_packet_routed_with_valid_vlan[port_in_lag] failed on both the original attempt and retry.

AssertionError: Did not receive expected packet on any of ports [16] for device 0.
```

Expected behavior:

`PortChannelX.<vlan>` / LAG subinterfaces should be programmed in VPP and should route the valid VLAN packet, the same way the test expects for regular sub-port interfaces.

Actual behavior:

The valid VLAN packet is not routed out to the expected PTF port for `port_in_lag`.

Source-level suspicion:

The RIF SUB_PORT create path appears to accept LAG parents, but still calls `create_sub_interface(tap_to_hwif_name(dev), vlan_id, vlan_id)` where `dev` is `PortChannelX`. `tap_to_hwif_name()` is backed by `sonic_vpp_ifmap.ini`, which is generated with physical `Ethernet*` mappings only, so `PortChannelX` does not resolve to the VPP `BondEthernetX` hwif and the VPP subinterface is not created on the correct parent.

This issue is specifically for `PortChannelX.<vlan>` / LAG subinterfaces not being programmed/routed correctly in SONiC-VPP.

Bojun-Feng:
### Description of PR
LAG subinterfaces (PortChannelX.vlan) are non-functional on SONiC-VPP due to two related bugs:

1. The RIF SUB_PORT create/delete paths call `create_sub_interface(tap_to_hwif_name(dev), vlan_id, vlan_id)` where `dev` is "PortChannelX". Since `sonic_vpp_ifmap.ini` only maps physical Ethernet interfaces, `tap_to_hwif_name("PortChannelX")` returns "Unknown" and the VPP subinterface is never created on the correct BondEthernet parent.

2. `vpp_get_hwif_name()` returns early for LAG objects without appending the `.vlan_id` suffix, causing `vpp_set_interface_mtu` and `vpp_set_interface_state` to target the parent BondEthernet instead of the subinterface.

Changes made:
* In the create/delete paths, check `ot == SAI_OBJECT_TYPE_LAG` and construct the VPP hwif name as `BondEthernetX` from `bond_info.id` instead of calling `tap_to_hwif_name()`
* Fix `vpp_get_hwif_name()` to let the LAG branch fall through to the vlan_id append logic so MTU/state operations target the correct subinterface

Summary:
Fixes https://github.com/sonic-net/sonic-platform-vpp/issues/227
### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

LAG subinterfaces are broken on VPP platform. The `test_packet_routed_with_valid_vlan[port_in_lag]` test fails because the VPP subinterface is never created on the correct BondEthernet parent. Additionally, post-creation operations (MTU, admin state) would target the wrong interface due to `vpp_get_hwif_name()` returning the parent bond name without the `.vlan_id` suffix.

##### Work item tracking

#### How did you do it?

Applied the same LAG-aware hwif resolution pattern already used in the [VRF assignment section](https://github.com/sonic-net/sonic-sairedis/blob/d3ba80722ef9e2f907f748702912d8d6b0dceebd/vslib/vpp/SwitchVppRif.cpp#L1614-L1620) and in [SwitchVppFdb.cpp](https://github.com/sonic-net/sonic-sairedis/blob/d3ba80722ef9e2f907f748702912d8d6b0dceebd/vslib/vpp/SwitchVppFdb.cpp#L137-L145):

1. In the SUB_PORT create/delete paths, check `ot == SAI_OBJECT_TYPE_LAG` and construct `"BondEthernetX"` directly from `bond_info.id` (already populated via `get_lag_bond_info()`) instead of calling `tap_to_hwif_name()`.

2. In `vpp_get_hwif_name()`, restructured the LAG branch to set `hwifname` and fall through to the shared `vlan_id` append logic (the `if (vlan_id) { snprintf(..., "%s.%u", ...) }` block) rather than returning early with just the parent name.

Todo: Run `test_packet_routed_with_valid_vlan[port_in_lag]` in the t1-lag-vpp topology (Platform: kvm, ASIC type: vpp).

VPP platform only. Physical port subinterfaces and non-subinterface LAG operations are unaffected (the `else` branches preserve original behavior).

### Documentation

N/A

### DIFF
--- a/vslib/vpp/SwitchVppRif.cpp
+++ b/vslib/vpp/SwitchVppRif.cpp
@@ -1583,7 +1583,15 @@ sai_status_t SwitchVpp::vpp_create_router_interface(
         snprintf(host_subifname, sizeof(host_subifname), "%s.%u", dev, vlan_id);
 
         /* The host(tap) subinterface is also created as part of the vpp subinterface creation */
-        create_sub_interface(tap_to_hwif_name(dev), vlan_id, vlan_id);
+        const char *parent_hwif;
+        char hw_subif_parent[32];
+        if (ot == SAI_OBJECT_TYPE_LAG) {
+            snprintf(hw_subif_parent, sizeof(hw_subif_parent), "%s%d", BONDETHERNET_PREFIX, bond_info.id);
+            parent_hwif = hw_subif_parent;
+        } else {
+            parent_hwif = tap_to_hwif_name(dev);
+        }
+        create_sub_interface(parent_hwif, vlan_id, vlan_id);
 
         /* Get new list of physical interfaces from VS */
         refresh_interfaces_list();
@@ -1869,7 +1877,18 @@ sai_status_t SwitchVpp::vpp_remove_router_interface(sai_object_id_t rif_id)
     uint16_t vlan_id = attr.value.u16;
 
     std::string if_name;
-    bool found = getTapNameFromPortId(obj_id, if_name);
+    platform_bond_info_t bond_info;
+    bool found;
+    if (ot == SAI_OBJECT_TYPE_LAG) {
+        status = get_lag_bond_info(obj_id, bond_info);
+        if (status != SAI_STATUS_SUCCESS) {
+            return status;
+        }
+        if_name = std::string(PORTCHANNEL_PREFIX) + std::to_string(bond_info.id);
+        found = true;
+    } else {
+        found = getTapNameFromPortId(obj_id, if_name);
+    }
     if (found == false)
     {
         SWSS_LOG_ERROR("host interface for port id %s not found", sai_serialize_object_id(obj_id).c_str());
@@ -1878,7 +1897,15 @@ sai_status_t SwitchVpp::vpp_remove_router_interface(sai_object_id_t rif_id)
 
     const char *dev = if_name.c_str();
 
-    delete_sub_interface(tap_to_hwif_name(dev), vlan_id);
+    const char *parent_hwif;
+    char hw_del_parent[32];
+    if (ot == SAI_OBJECT_TYPE_LAG) {
+        snprintf(hw_del_parent, sizeof(hw_del_parent), "%s%d", BONDETHERNET_PREFIX, bond_info.id);
+        parent_hwif = hw_del_parent;
+    } else {
+        parent_hwif = tap_to_hwif_name(dev);
+    }
+    delete_sub_interface(parent_hwif, vlan_id);
     /* Get new list of physical interfaces from VS */
     refresh_interfaces_list();
 
--- a/vslib/vpp/SwitchVppRif.cpp
+++ b/vslib/vpp/SwitchVppRif.cpp
@@ -343,6 +343,9 @@ bool SwitchVpp::vpp_get_hwif_name (
 {
     SWSS_LOG_ENTER();
 
+    const char *hwifname;
+    char hw_bondifname[32];
+
     if (objectTypeQuery(object_id) == SAI_OBJECT_TYPE_LAG) {
         platform_bond_info_t bond_info;
         sai_status_t status = get_lag_bond_info(object_id, bond_info);
@@ -350,21 +353,22 @@ bool SwitchVpp::vpp_get_hwif_name (
         {
             return false;
         }
-        ifname = std::string(BONDETHERNET_PREFIX) + std::to_string(bond_info.id);
-        return true;
+        snprintf(hw_bondifname, sizeof(hw_bondifname), "%s%d", BONDETHERNET_PREFIX, bond_info.id);
+        hwifname = hw_bondifname;
+    } else {
+        std::string if_name;
+        bool found = getTapNameFromPortId(object_id, if_name);
+
+        if (found == false)
+        {
+            SWSS_LOG_ERROR("host interface for port id %s not found", sai_serialize_object_id(object_id).c_str());
+            return false;
+        }
+        hwifname = tap_to_hwif_name(if_name.c_str());
     }
 
-    std::string if_name;
-    bool found = getTapNameFromPortId(object_id, if_name);
 
-    if (found == false)
-    {
-        SWSS_LOG_NOTICE("host interface for port id %s not found", sai_serialize_object_id(object_id).c_str());
-        return false;
-    }
-
-    const char *hwifname = tap_to_hwif_name(if_name.c_str());
-    char hw_subifname[32];
+    char hw_subifname[64];
     const char *hw_ifname;
 
     if (vlan_id) {
--- a/vslib/vpp/SwitchVppRif.cpp
+++ b/vslib/vpp/SwitchVppRif.cpp
@@ -353,7 +353,7 @@ bool SwitchVpp::vpp_get_hwif_name (
         {
             return false;
         }
-        snprintf(hw_bondifname, sizeof(hw_bondifname), "%s%d", BONDETHERNET_PREFIX, bond_info.id);
+        snprintf(hw_bondifname, sizeof(hw_bondifname), "%s%u", BONDETHERNET_PREFIX, bond_info.id);
         hwifname = hw_bondifname;
     } else {
         std::string if_name;
@@ -1590,7 +1590,7 @@ sai_status_t SwitchVpp::vpp_create_router_interface(
         const char *parent_hwif;
         char hw_subif_parent[32];
         if (ot == SAI_OBJECT_TYPE_LAG) {
-            snprintf(hw_subif_parent, sizeof(hw_subif_parent), "%s%d", BONDETHERNET_PREFIX, bond_info.id);
+            snprintf(hw_subif_parent, sizeof(hw_subif_parent), "%s%u", BONDETHERNET_PREFIX, bond_info.id);
             parent_hwif = hw_subif_parent;
         } else {
             parent_hwif = tap_to_hwif_name(dev);
@@ -1904,7 +1904,7 @@ sai_status_t SwitchVpp::vpp_remove_router_interface(sai_object_id_t rif_id)
     const char *parent_hwif;
     char hw_del_parent[32];
     if (ot == SAI_OBJECT_TYPE_LAG) {
-        snprintf(hw_del_parent, sizeof(hw_del_parent), "%s%d", BONDETHERNET_PREFIX, bond_info.id);
+        snprintf(hw_del_parent, sizeof(hw_del_parent), "%s%u", BONDETHERNET_PREFIX, bond_info.id);
         parent_hwif = hw_del_parent;
     } else {
         parent_hwif = tap_to_hwif_name(dev);
--- a/vslib/vpp/SwitchVppRif.cpp
+++ b/vslib/vpp/SwitchVppRif.cpp
@@ -343,7 +343,7 @@ bool SwitchVpp::vpp_get_hwif_name (
 {
     SWSS_LOG_ENTER();
 
-    const char *hwifname;
+    const char *hwifname = nullptr;
     char hw_bondifname[32];
 
     if (objectTypeQuery(object_id) == SAI_OBJECT_TYPE_LAG) {
@@ -367,6 +367,7 @@ bool SwitchVpp::vpp_get_hwif_name (
         hwifname = tap_to_hwif_name(if_name.c_str());
     }
 
+    if (!hwifname) return false;
 
     char hw_subifname[64];
     const char *hw_ifname;
@@ -1888,7 +1889,7 @@ sai_status_t SwitchVpp::vpp_remove_router_interface(sai_object_id_t rif_id)
         if (status != SAI_STATUS_SUCCESS) {
             return status;
         }
-        if_name = std::string(PORTCHANNEL_PREFIX) + std::to_string(bond_info.id);
+        if_name = std::string(BONDETHERNET_PREFIX) + std::to_string(bond_info.id);
         found = true;
     } else {
         found = getTapNameFromPortId(obj_id, if_name);
--- a/vslib/vpp/SwitchVppRif.cpp
+++ b/vslib/vpp/SwitchVppRif.cpp
@@ -336,6 +336,22 @@ void SwitchVpp::vpp_intf_remove_prefix_entry (const std::string& intf_name)
     m_intf_prefix_map.erase(it);
 }
 
+const char* SwitchVpp::vpp_resolve_parent_hwif(
+      _In_ sai_object_type_t ot,
+      _In_ uint32_t bond_id,
+      _In_ const char *tap_name,
+      _Out_ char *buf,
+      _In_ size_t buflen)
+{
+    SWSS_LOG_ENTER();
+
+    if (ot == SAI_OBJECT_TYPE_LAG) {
+        snprintf(buf, buflen, "%s%d", BONDETHERNET_PREFIX, bond_id);
+        return buf;
+    }
+    return tap_to_hwif_name(tap_name);
+}
+
 bool SwitchVpp::vpp_get_hwif_name (
       _In_ sai_object_id_t object_id,
       _In_ uint32_t vlan_id,
@@ -1587,14 +1603,8 @@ sai_status_t SwitchVpp::vpp_create_router_interface(
         snprintf(host_subifname, sizeof(host_subifname), "%s.%u", dev, vlan_id);
 
         /* The host(tap) subinterface is also created as part of the vpp subinterface creation */
-        const char *parent_hwif;
         char hw_subif_parent[32];
-        if (ot == SAI_OBJECT_TYPE_LAG) {
-            snprintf(hw_subif_parent, sizeof(hw_subif_parent), "%s%d", BONDETHERNET_PREFIX, bond_info.id);
-            parent_hwif = hw_subif_parent;
-        } else {
-            parent_hwif = tap_to_hwif_name(dev);
-        }
+        const char *parent_hwif = vpp_resolve_parent_hwif(ot, bond_info.id, dev, hw_subif_parent, sizeof(hw_subif_parent));
         create_sub_interface(parent_hwif, vlan_id, vlan_id);
 
         /* Get new list of physical interfaces from VS */
@@ -1623,14 +1633,8 @@ sai_status_t SwitchVpp::vpp_create_router_interface(
 
     vpp_add_ip_vrf(vrf_obj_id, vrf_id);
     if (ret == 0 && vrf_id != 0) {
-	const char *hwif_name;
 	char hw_bondifname[32];
-	if (ot == SAI_OBJECT_TYPE_LAG) {
-	    snprintf(hw_bondifname, sizeof(hw_bondifname), "%s%d", BONDETHERNET_PREFIX, bond_info.id);
-	    hwif_name = hw_bondifname;
-	} else {
-	    hwif_name = tap_to_hwif_name(dev);
-	}
+	const char *hwif_name = vpp_resolve_parent_hwif(ot, bond_info.id, dev, hw_bondifname, sizeof(hw_bondifname));
 	SWSS_LOG_NOTICE("Setting interface vrf on hwif_name %s", hwif_name);
 	set_interface_vrf(hwif_name, vlan_id, vrf_id, false);
     }
@@ -1786,14 +1790,8 @@ sai_status_t SwitchVpp::vpp_router_interface_remove_vrf(
 
     linux_ifname = if_name.c_str();
 
-    const char *hwif_name;
     char hw_bondifname[32];
-    if (objectTypeQuery(obj_id) == SAI_OBJECT_TYPE_LAG) {
-        snprintf(hw_bondifname, sizeof(hw_bondifname), "%s%d", BONDETHERNET_PREFIX, bond_info.id);
-        hwif_name = hw_bondifname;
-    } else {
-        hwif_name = tap_to_hwif_name(if_name.c_str());
-    }
+    const char *hwif_name = vpp_resolve_parent_hwif(objectTypeQuery(obj_id), bond_info.id, if_name.c_str(), hw_bondifname, sizeof(hw_bondifname));
 
     SWSS_LOG_NOTICE("Resetting to default vrf for interface %s, %s", linux_ifname, hwif_name);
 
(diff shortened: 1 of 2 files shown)


Bojun-Feng:
Note to self: `addRemoveDashMeterCounter` and `bulkChunksize` unit tests are flaky due to race conditions, both tests keep failing probabilistically on different builds. I had to rerun tests multiple times.

Luckily we have the option to only rerun failed tests, else it would have taken a large number of runs to get both tests passing on all builds simutaneously.

Need to open an issue (or find existing issue) to track this.

Bojun-Feng:
Hi @cyw233, CI is all green. Would appreciate your review when you have time, thanks!

Additionally, I have not yet verified the fix by enabling the sonic-mgmt `port_in_lag` parameter for the sub-port forwarding test.

I can try to set up the environment on my end and run the test but that would take a few days. Would it be possible for you to help verify it?

cyw233:
> Hi @cyw233, CI is all green. Would appreciate your review when you have time, thanks!
> 
> Additionally, I have not yet verified the fix by enabling the sonic-mgmt `port_in_lag` parameter for the sub-port forwarding test.
> 
> I can try to set up the environment on my end and run the test but that would take a few days. Would it be possible for you to help verify it?

@Bojun-Feng , great to see the PR checkers are passing!

We are currently having some internal limitations/restrictions on setting up the environment and still working a solution... Maybe let's both set up our environment and run the test in parallel. Will keep you updated here

lunyue-ms:
Hi @Bojun-Feng

It seems that your patch can fix the issue! Today I applied your commits and used the new image for a test. It seems that the LAG sub-interface can be displayed via vppctl. Following is the test steps.
Next week, we will verify this fix on our CI testbed in the t1-lag-vpp topology (Platform: kvm, ASIC type: vpp). Hopefully the related tests can get passed.

(BondEthernet1.10 will not be there when I test the image without your patch)

```
admin@sonic:~$ 
admin@sonic:~$ 
admin@sonic:~$ sudo config portchannel add PortChannel1
admin@sonic:~$ sudo config portchannel member add PortChannel1 Ethernet0
admin@sonic:~$ sudo config subinterface add PortChannel1.10 10
admin@sonic:~$ sudo config interface ip add PortChannel1.10 10.1.0.1/24
admin@sonic:~$ 
admin@sonic:~$ 
admin@sonic:~$ 
admin@sonic:~$ 
admin@sonic:~$ redis-cli -n 0 keys "*PortChannel*"
1) "LAG_MEMBER_TABLE:PortChannel1:Ethernet0"
2) "INTF_TABLE:PortChannel1.10:10.1.0.1/24"
3) "LAG_TABLE:PortChannel1"
4) "INTF_TABLE:PortChannel1.10"
admin@sonic:~$ 
admin@sonic:~$ docker exec -it syncd vppctl show interface
              Name               Idx    State  MTU (L3/IP4/IP6/MPLS)     Counter          Count     
BondEthernet1                     5      up          9000/0/0/0     tx packets                     4
                                                                    tx bytes                     520
BondEthernet1.10                  7      up          9100/0/0/0     
bobm0                             1      up          9122/0/0/0     rx packets                     4
                                                                    rx bytes                     560
                                                                    tx packets                   279
                                                                    tx bytes                   37262
                                                                    drops                          4
                                                                    ip6                            4
                                                                    rx-miss                     1135
                                                                    tx-error                       2
bobm1                             2      up          9122/0/0/0     rx packets                    24
                                                                    rx bytes                    3160
                                                                    tx packets                    45
                                                                    tx bytes                   10715
                                                                    drops                         24
                                                                    ip6                           24
                                                                    rx-miss                     1135
                                                                    tx-error                       2
local0                            0     down          0/0/0/0       drops                          2
tap4096                           3      up          9000/0/0/0     rx packets                   272
                                                                    rx bytes                   36352
                                                                    drops                          1
                                                                    ip6                           17
tap4097                           4      up          9000/0/0/0     rx packets                    46
                                                                    rx bytes                   10845
                                                                    drops                          1
                                                                    ip6                            5
tap4098                           6      up          9000/0/0/0     rx packets                     4
                                                                    rx bytes                     520
                                                                    ip6                            4
tap4098.10                        8      up           0/0/0/0       
admin@sonic:~$ 
admin@sonic:~$ show subinterfaces status
  Sub port interface    Speed    MTU    Vlan    Admin                  Type
--------------------  -------  -----  ------  -------  --------------------
     PortChannel1.10       1G   9100      10       up  802.1q-encapsulation
admin@sonic:~$ 
```

Bojun-Feng:
Thanks for the update @lunyue-ms! Looking forward to the CI testbed results next week. Let me know if anything comes up during the t1-lag-vpp topology run, happy to help debug if needed.

lunyue-ms:
Hi @Bojun-Feng 

I setup my local test environment and did some tests. The VLAN related sub interface tests were failed.
[test_packet_routed_with_valid_vlan.log](https://github.com/user-attachments/files/27691699/test_packet_routed_with_valid_vlan.log)

I'm trying to figure out what's the possible reason now and will update the ticket after.

Bojun-Feng:
> Hi @Bojun-Feng
> 
> I setup my local test environment and did some tests. The VLAN related sub interface tests were failed. [test_packet_routed_with_valid_vlan.log](https://github.com/user-attachments/files/27691699/test_packet_routed_with_valid_vlan.log)
> 
> I'm trying to figure out what's the possible reason now and will update the ticket after.

Hi @lunyue-ms, thanks for sharing! I wanted to provide some updates — I did some research on the log to look into why the `[port]` case (Ethernet sub-interface) is also failing, not just `[port_in_lag]`.

Seems like the test expects ICMP replies from the DUT's sub-interfaces (e.g., `Ethernet64.20`, `PortChannel1.20`), but VPP never responds.

Looking at the code, the root cause is in `SwitchVppRoute.cpp` — when a connected subnet route is installed with `next_hop_type = ROUTER_INTERFACE` (which is the normal case for directly-connected IPs on sub-ports), the IP programming call is commented out:

https://github.com/sonic-net/sonic-sairedis/blob/88bc51ae95df66977601957515e5527119ffd4c5/vslib/vpp/SwitchVppRoute.cpp#L148-L151

This was commented out in the upstream `sonic-platform-vpp` repo during the Debian Bullseye migration (https://github.com/sonic-net/sonic-platform-vpp/pull/22, see [here](https://github.com/sonic-net/sonic-platform-vpp/pull/22/changes#diff-cbfb8019729fef861ce1c0660547a346f35b41a768d91c62c0c7ac6cf4a71f6bL171) for the exact line of change).

I believe the the original `vpp_add_del_intf_ip_addr` function couldn't handle non-PORT interface types (LAG, loopback) and hard-rejects anything that isn't `SAI_OBJECT_TYPE_PORT`. The assumption was that VPP's LCP plugin would handle IP sync from Linux → VPP for all interfaces, which works for regular ports but doesn't appear to work for sub-interfaces.

There are several directions out of this:

1. Wire the existing `vpp_add_del_intf_ip_addr_norif` into the `ROUTER_INTERFACE` case (this function already resolves interface names from prefixes and handles both PORT and LAG sub-ports).
2. Fix the original `vpp_add_del_intf_ip_addr` to handle LAG/loopback types, then uncomment.
3. Fix LCP sub-interface pairing so IP sync works implicitly (longer-term, likely a WG discussion topic).

A targeted hot-fix would be option (1) scoped to only `SUB_PORT` RIF types, so it doesn't risk regressing the existing LCP-based flow for regular ports. LCP improvements could be addressed separately as a longer-term item.

I think this is out of scope for this PR (which addresses LAG hwif name resolution) and would also live in a different repository. Happy to look into a follow-up hot-fix PR for this if it would be helpful.

lunyue-ms:
Hi @Bojun-Feng ,
Thanks for your finding!  It makes sense to use the option1 in the short term. I'm looking forward to your hot-fix in another PR on this issue  :)

Btw, I ran three VPP L3 sub-interface tests on the DUT (vlab-vpp-01) with tap1.20 (172.16.20.1/30) and tap2.30(172.16.30.1/30), and set up ns1/ns2 as traffic endpoints (172.16.20.2/30, 172.16.30.2/30).

 1. Test 1: ns1 -> DUT gateway (172.16.20.1)
Failed (0/4 replies, 100% loss).
 2. Test 2: ns2 -> DUT gateway (172.16.30.1)
Failed (0/4 replies, 100% loss).
 3. Test 3: Inter-VLAN routing through VPP (ns1 <-> ns2)
Passed in both directions (4/4 replies each way), with TTL=63, confirming packets were routed through VPP L3 forwarding.

Additional observation: the same test logic was validated on a separate non-SONiC Ubuntu machine, and it passed there. This indicates the test method is sound, and the failure appears specific to the SONiC-VPP DUT environment (gateway/local interface reply path). 
Atteched the test steps:
[dut_vpp_l3_subinterface_test_report.md](https://github.com/user-attachments/files/27756222/dut_vpp_l3_subinterface_test_report.md)

Bojun-Feng:
Hi @lunyue-ms, I implemented the hot-fix for the sub-interface IP programming issue we discussed at https://github.com/sonic-net/sonic-sairedis/pull/1895. Since it touches a different file (SwitchVppRoute.cpp) and addresses a separate bug, I opened it as a new PR. Also happy to consolidate into this PR if you'd prefer them together.

I also added a small refactor commit on the current PR, adding a helper to deduplicate the LAG→BondEthernet name resolution that was copy-pasted in 4 places.

Seems like the current CI testing server is out of memory, which resulted in tests passing but log collection failing. See the [Azure Pipeline page](https://dev.azure.com/mssonic/build/_build/results?buildId=1117403&view=logs&j=4b2e7278-6a28-5761-e11d-502b1a3033c1&t=b82dde59-e7cc-5983-155e-9c3bfa7dc0ae&s=9f373adc-b8ce-5e80-aa3e-0c43d50b8b1e) ([Snapshot](https://web.archive.org/web/20260519164232/https://dev.azure.com/mssonic/build/_build/results?buildId=1117403&view=logs&j=4b2e7278-6a28-5761-e11d-502b1a3033c1&t=b82dde59-e7cc-5983-155e-9c3bfa7dc0ae&s=9f373adc-b8ce-5e80-aa3e-0c43d50b8b1e))for more details.

Would appreciate if you can help verify on the testbed when you get a chance. Thanks!

lunyue-ms:
Hi @Bojun-Feng ,

Thanks for your update. I was debugging tests related to sub_port_interfaces with your previous changes in this PR. Currently, some tests got passed but somes are still failed.

 
Passed tests:
sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_packet_routed_with_valid_vlan[port]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_packet_routed_with_valid_vlan[port_in_lag]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_untagged_packet_not_routed[port]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_untagged_packet_not_routed[port_in_lag]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_admin_status_down_disables_forwarding[port]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_admin_status_down_disables_forwarding[port_in_lag]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_mtu_inherited_from_parent_port[port]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_mtu_inherited_from_parent_port[port_in_lag]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_vlan_config_impact[port]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_vlan_config_impact[port_in_lag]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_routing_between_sub_ports[port-same-TCP-UDP-ICMP]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_routing_between_sub_ports[port-different-TCP-UDP-ICMP]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_routing_between_sub_ports[port_in_lag-same-TCP-UDP-ICMP]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_routing_between_sub_ports[port_in_lag-different-TCP-UDP-ICMP]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_routing_between_sub_ports_unaffected_by_sub_ports_removal[port-same-TCP-UDP-ICMP]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_routing_between_sub_ports_unaffected_by_sub_ports_removal[port_in_lag-same-TCP-UDP-ICMP]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPortsNegative::test_packet_routed_with_invalid_vlan[port]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPortsNegative::test_packet_routed_with_invalid_vlan[port_in_lag]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPortStress::test_max_numbers_of_sub_ports[port]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPortStress::test_max_numbers_of_sub_ports[port_in_lag]

Failed tests:
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_routing_between_sub_ports_unaffected_by_sub_ports_removal[port-different-TCP-UDP-ICMP]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_routing_between_sub_ports_unaffected_by_sub_ports_removal[port_in_lag-different-TCP-UDP-ICMP]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_routing_between_sub_ports_and_port[port-svi-TCP-UDP-ICMP]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_routing_between_sub_ports_and_port[port-l3-TCP-UDP-ICMP]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_routing_between_sub_ports_and_port[port_in_lag-svi-TCP-UDP-ICMP]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_routing_between_sub_ports_and_port[port_in_lag-l3-TCP-UDP-ICMP]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_tunneling_between_sub_ports[port-same]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_tunneling_between_sub_ports[port-different]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_tunneling_between_sub_ports[port_in_lag-same]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_tunneling_between_sub_ports[port_in_lag-different]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_balancing_sub_ports[port]
 sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_balancing_sub_ports[port_in_lag]

I'm still debugging on this and didn't submit PRs so  far.

I think that you can go ahead with the changes in this PR. Tomorrow I will update use the latest changeset in this PR. 
For your another PR, I havn't  used it so far. It looks that the change was covered by commits in recent days when debugging sub_port_interfaces tests. I will update you when I submit the related PR after. Thanks.

Bojun-Feng:
Seems like an unrelated linker error in SWSS `fpmsyncd` is blocking the CI. Ignoring it for now.

@lunyue-ms I have pushed some additional commits that should help with the `routing_between_sub_ports` and `balancing_sub_ports` tests.

For the `tunneling_between_sub_ports` tests, the tunneling feature is simply not implemented. Since its a relatively large feature (likely several hundred lines C++), I believe it deserves its own dedicated feature PR.

Please let me know if the new changes helped, thanks!

**Note:** The previous subport removal implementation called `delete_sub_interface()` followed by `refresh_interfaces_list()`. The new approach strips IPs and admin-downs instead, preserving the VPP sub-interface object.

The motivation was to avoid `refresh_interfaces_list()` during removal, which is suspected as a contributor to the test failure in `routing_between_sub_ports_unaffected_by_sub_ports_removal`, since it destroys and rebuilds the entire local cache. Admin-down + preserve is a lighter-weight approach that sidesteps the issue. The trade-off is that removed sub-interfaces accumulate as admin-down stubs in VPP, bounded by unique VLAN IDs per parent (max 4094). 

If this approach still doesn't pass the test, the commit should be reverted. If the test is passing, then the proper long-term fix would be `delete_sub_interface()` + a targeted single-entry cache removal so we avoid the full rebuild.

lunyue-ms:
Hi @Bojun-Feng , quick status:

I opened two PRs for this work, both currently open:

- https://github.com/sonic-net/sonic-sairedis/pull/1907 — includes your two
  earliest LAG hwif commits (you're co-author), plus the rest of the sub-port /
  LAG sub-port fixes needed for traffic (LCP host on `be<id>.<vlan>` + tc
  bridge, RIF IP, idempotency, etc.).
- https://github.com/sonic-net/sonic-platform-vpp/pull/237 — companion
  changes: `lcp-sync` on, `lcp-auto-subint` off, DPDK promisc reconciler.

Full `test_sub_port_interfaces.py` on `vlab-vpp-01` (`vms-kvm-vpp-t1-lag-vpp`)
with both PRs applied: **22 passed / 10 failed**. The 10 are all out of scope:
- 2 SVI — sonic-buildimage #26936
- 4 IPinIP — sairedis #1860 (lolyu)
- 4 LAG sub-port follow-up — separate DUT-originated ARP gap; `[port]`
  counterparts PASS, tracked for a follow-up PR

On your SVI bits (`ca4ccc6f` → `_norif`): #1907 changed the SUB_PORT path to
OID-based, so picking it as-is conflicts. Happy to help rebase onto
post-#1907 master in a follow-up — with you as co-author.

Bojun-Feng:
Hi @lunyue-ms, thanks for the detailed update and for opening #1907! Great to see the test results improving.

I've reset this PR to the scoped LAG hwif fix only (commits 1-2). The later commits (sub-port IP programming, admin-down preservation, etc.) are superseded by your approach, so I've dropped them to avoid conflicts.

Since this PR is now a clean, self-contained fix for the LAG hwif name resolution bug (which #1907 also depends on), would it make sense to merge this one first? That way #1907 can rebase cleanly on top and we avoid duplicate commits. Happy to coordinate on timing.

For the SVI work, I'm happy to pick that up as a follow-up once #1907 lands. Thanks for the offer to collaborate on it!

@cyw233 The PR is now in its final form, just the two LAG hwif fixes you helped triage in https://github.com/sonic-net/sonic-platform-vpp/issues/227. Would appreciate your review when you have a moment.

lunyue-ms:
Hi @Bojun-Feng, thanks for the update. I'm happy to wait for thie PR be merged and I can rebase https://github.com/sonic-net/sonic-sairedis/pull/1907  then.

vslib/vpp/SwitchVppRif.cpp
```diff
@@ -343,28 +343,32 @@ bool SwitchVpp::vpp_get_hwif_name (
...
         {
             return false;
         }
-        ifname = std::string(BONDETHERNET_PREFIX) + std::to_string(bond_info.id);
-        return true;
-    }
-
-    std::string if_name;
-    bool found = getTapNameFromPortId(object_id, if_name);
+        snprintf(hw_bondifname, sizeof(hw_bondifname), "%s%d", BONDETHERNET_PREFIX, bond_info.id);
```
lunyue-ms: %d → %u for bond_info.id (it's uint32_t)  Other two locations in line1593 and 1907.

lunyue-ms (changes requested):

vslib/vpp/SwitchVppRif.cpp
Bojun-Feng: fixed, thanks for the catch!

lunyue-ms (dismissed):
LGTM now.

Bojun-Feng:
Hi @cyw233 @lolyu — just checking in on this PR. CI is all green, @lunyue-ms already approved it, #1907 is waiting to rebase on top. Please let me know if there's anything else needed from my side.

vslib/vpp/SwitchVppRif.cpp
```diff
@@ -1869,7 +1881,18 @@ sai_status_t SwitchVpp::vpp_remove_router_interface(sai_object_id_t rif_id)
...
     std::string if_name;
-    bool found = getTapNameFromPortId(obj_id, if_name);
+    platform_bond_info_t bond_info;
+    bool found;
+    if (ot == SAI_OBJECT_TYPE_LAG) {
+        status = get_lag_bond_info(obj_id, bond_info);
+        if (status != SAI_STATUS_SUCCESS) {
+            return status;
+        }
+        if_name = std::string(PORTCHANNEL_PREFIX) + std::to_string(bond_info.id);
```
lolyu: ⚠️ **Major — `PORTCHANNEL_PREFIX` vs `BONDETHERNET_PREFIX` asymmetry**

`if_name` is set to `PORTCHANNEL_PREFIX + bond_info.id` here (e.g. `"PortChannel1"`), but the very next block that actually calls `delete_sub_interface` builds `hw_del_parent` from `BONDETHERNET_PREFIX + bond_info.id` — so the VPP delete itself is fine. However `dev = if_name.c_str()` is derived from this value and used in `host_subifname` logging (and any future caller expecting the Linux/tap name). This creates a latent asymmetry with the create path, which never sets `if_name` at all for the LAG branch.

If `if_name`/`dev` is intentionally the Linux tap-side name (i.e. `PortChannel1`), please add a comment explaining that. If it should be the VPP hwif name for consistency with the create path, change to `BONDETHERNET_PREFIX`.

```suggestion
        if_name = std::string(BONDETHERNET_PREFIX) + std::to_string(bond_info.id);
```

vslib/vpp/SwitchVppRif.cpp
```diff
@@ -343,28 +343,32 @@ bool SwitchVpp::vpp_get_hwif_name (
 {
     SWSS_LOG_ENTER();
 
+    const char *hwifname;
```
lolyu: ⚠️ **Major — `hwifname` declared uninitialized; dangling pointer risk if `objectTypeQuery` returns an unexpected type**

`hwifname` is assigned in both the LAG and else branches, so the normal paths are safe. But if `objectTypeQuery(object_id)` returns neither `SAI_OBJECT_TYPE_LAG` nor a port type (e.g. a trunk member or some future object type), the `else` branch runs `getTapNameFromPortId`, which may return `found == false` → `return false` — still safe. The concern is a hypothetical third SAI type that reaches neither `return false` early exit: `hwifname` would be unread past the `if/else`, so practically no UB today.

Still, initialising defensively eliminates the compiler warning and makes intent explicit:

```suggestion
    const char *hwifname = nullptr;
    char hw_bondifname[32];
```
And add a null-check before the `vlan_id` block below:
```cpp
    if (!hwifname) return false;
```

lolyu (commented):
Two observations on the remove path and `vpp_get_hwif_name` worth addressing before merge.

yue-fred-gao (approved):

PR #1861, opened 2026-04-20
[syncd] Fix PORT_PHY_ATTR::collectData to check initAttrData return value

ZhaohuiS:
#### Why I did it
PORT_PHY_ATTR::collectData in FlexCounter.cpp calls `initAttrData` without checking its return value. When Broadcom SAI does not support attributes like `SAI_PORT_ATTR_RX_SIGNAL_DETECT` (attribute 149), the subsequent `sai_get` call fails with `Feature unavailable (0xfffffff0)`, generating ERR-level syslog entries every ~10 seconds (FlexCounter polling interval).

This causes **all tests using loganalyzer** to fail at teardown on Arista-7260CX3 (Broadcom) platforms. On 202511 image `20251110.20`, this produced **8,700+ test failures**.

The regression is triggered by the combination of:
- sonic-sairedis #1674 / #1785 (adds PORT_PHY_ATTR FlexCounter polling support)
- sonic-swss #4223 / #4407 (activates PORT_PHY_ATTR polling via FLEX_COUNTER_DB entries)

A partial fix was applied in #1799 / #1815 for PORT_PHY_SERDES_ATTR::collectData, but the same fix was **not** applied to PORT_PHY_ATTR::collectData.

#### How I did it
- Changed `initAttrData` return type from `void` to `bool`
- Each switch case now returns `true` on success, error paths return `false`
- In `collectData`, check the return value of `initAttrData` in the init loop
- If any attribute initialization fails, log a WARN and skip the entire object (`continue`)
- This matches the pattern already used in PORT_PHY_SERDES_ATTR::collectData (fixed by #1799)

#### How to verify it
1. Deploy on an Arista-7260CX3 (Broadcom) testbed with 202511 image
2. Verify no more `RX signal detect` ERR syslog entries from syncd
3. Run any test with loganalyzer enabled — teardown should pass

#### Related PRs
- #1674 — PORT_PHY_ATTR flex counter support (master)
- #1785 — cherry-pick to 202511
- #1799 — Partial fix for PORT_PHY_SERDES_ATTR (master)
- #1815 — cherry-pick of #1799 to 202511
- #1855 — Same fix targeting 202511 branch
- sonic-swss #4223 / #4407 — Activation trigger

#### ADO Reference
- Microsoft ADO PBI: 37325902
- Arista issue: https://github.com/aristanetworks/sonic-qual.msft/issues/1232
- Public tracking: https://github.com/sonic-net/sonic-mgmt/issues/24023

PR #1884, opened 2026-05-08
[syncd] Cache "permanently unsupported" PORT_PHY_ATTR / PORT_PHY_SERD…

Xichen96:
…ES_ATTR per RID

#### Why I did it
Building on PR #1861 (`Fix PORT_PHY_ATTR::collectData to check initAttrData return value`), there is a remaining gap that still triggers per-poll ERR syslog spam from `PortPhyAttrContext::collectData` and `PortPhySerdesAttrContext::collectData` on Arista 720dt M0/MX testbeds.

PR #1861 short-circuits when `m_portLaneCountMap` has *no* entry for the (rid, attr_id) pair (i.e. when the lane-count probe in `updatePortLaneCountMap` itself fails — Arista 7260CX3 case). For 720dt ports without an optical module, the lane-count probe **succeeds** so `initAttrData()` returns true, then `sai_get_port_attribute` is invoked and BCM SAI returns either:

  * `Feature unavailable (0xfffffff0)` -> SAI_STATUS_NOT_SUPPORTED (-2), or
  * `Invalid port (0xffffffee)` -> SAI_STATUS_INVALID_PORT_NUMBER (-9)

Both BCM SAI driver layer and the syncd FlexCounter wrapper layer emit ERR-level syslog every poll cycle (~10s). On a fresh 720dt with ~24 unplugged front ports this generates 25,000+ ERR lines per nightly run and breaks loganalyzer at teardown — same downstream impact as the bug PR #1861 set out to fix.

#### How I did it
Add a per-context `m_unsupportedAttrs` cache mapping rid -> set of attrs the SAI driver permanently does not support for that port:

  * `PortPhyAttrContext::collectData` and `PortPhySerdesAttrContext::collectData`:
    - Before the `initAttrData` loop: if every attr in this batch is in the cache for this rid, silently `continue`.
    - After `m_vendorSai->get()` returns NOT_SUPPORTED or INVALID_PORT_NUMBER: insert every attr in the batch into the cache, log NOTICE (not ERROR) once, and `continue`.
    - Other (transient/unexpected) failures keep the existing SWSS_LOG_ERROR + `continue`.
  * `removeObject`: clean up the cache for the removed RID, mirroring the existing `m_portLaneCountMap` / `m_portSerdesIdToLaneCountMap` cleanup.

The first poll per rid still emits one BCM SAI driver ERR (we can't suppress the vendor binary), but subsequent polls are completely silent. For 720dt this drops syslog ERR volume from ~25,000 per nightly to ~50 per syncd start (~500x reduction) and removes the FlexCounter wrapper ERRs entirely (demoted to NOTICE on first occurrence).

#### How to verify it
1. Deploy on an Arista 720dt (or any platform with unplugged front ports on a Broadcom ASIC) with the 202511 image.
2. Run `show logging | grep -E "(brcm_sai_get_port_attribute_cmn|collectData: Failed to get port attr)"`.
3. Before the fix: thousands of ERR lines, growing every 10s.
4. After the fix: at most a few NOTICE lines per port at syncd start, then silent.
5. Run any test with loganalyzer enabled — teardown should pass without the temporary `loganalyzer_common_ignore.txt` patches now needed for `collectData ... -2` and `collectData ... -9`.

Unit test added: `TestPortPhyAttr.CollectDataSkipsAfterUnsupportedStatus`
- Mocks SAI to succeed lane-count probe but return NOT_SUPPORTED on collectData's actual sai_get.
- Asserts `multiAttrGets == 1` after multiple poll cycles, proving the cache prevents repeat calls.
- Asserts COUNTERS_DB is empty for the failing RID.

#### Related PRs
- #1861 -- Fix PORT_PHY_ATTR::collectData to check initAttrData return value (extends the same defensive pattern; this PR closes the remaining gap)
- #1799 / #1815 -- Partial fix for PORT_PHY_SERDES_ATTR (same shape)
- sonic-net/sonic-mgmt #21564 -- Temporary loganalyzer ignore for `-9` collectData failures (can be removed once this lands)

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

##### Work item tracking

### DIFF
--- a/syncd/FlexCounter.cpp
+++ b/syncd/FlexCounter.cpp
@@ -1,6 +1,7 @@
 #include <inttypes.h>
 #include <unordered_map>
 #include <map>
+#include <set>
 #include <vector>
 #include <string>
 #include <mutex>
@@ -1967,6 +1968,13 @@ public:
                 SWSS_LOG_DEBUG("PORT_PHY_ATTR: Removing RID 0x%" PRIx64 " from m_portLaneCountMap", it->second->rid);
                 m_portLaneCountMap.erase(lane_it);
             }
+
+            auto unsupp_it = m_unsupportedAttrs.find(it->second->rid);
+            if (unsupp_it != m_unsupportedAttrs.end())
+            {
+                SWSS_LOG_DEBUG("PORT_PHY_ATTR: Removing RID 0x%" PRIx64 " from m_unsupportedAttrs", it->second->rid);
+                m_unsupportedAttrs.erase(unsupp_it);
+            }
         }
 
         // Clean up lane metadata for this VID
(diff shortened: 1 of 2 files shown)


Xichen96:
Closing in favor of a cleaner approach.

Reviewer feedback (offline) made the right point: this PR caches failures **after** they happen, which papers over an inconsistency between the lane-count probe and the actual sai_get poll. A cleaner approach is to extend the probe in updatePortLaneCountMap to also do a test poll, so probe and poll behavior match by construction. That way #1861's existing skip mechanism (m_portLaneCountMap absence -> initAttrData returns false -> collectData skips) handles the 720dt case naturally with no new data structure.

A follow-up PR with the probe-augmentation approach will be opened shortly.

Note: the underlying root cause is a Broadcom SAI driver behavior — sai_get_port_attribute(RX_SIGNAL_DETECT) returns NOT_SUPPORTED / INVALID_PORT_NUMBER for ports without an inserted module, when per the SAI spec it should return success with current_status=false on all lanes. A vendor-side issue will be filed separately.

PR #1885, opened 2026-05-08
[syncd] PORT_PHY_ATTR: verify the actual data read at probe time

Xichen96:
#### Why I did it
After PR #1861 (`Fix PORT_PHY_ATTR::collectData to check initAttrData return value`) there is a remaining gap that still triggers per-poll ERR-level syslog spam from `PortPhyAttrContext::collectData` on Arista 720dt M0/MX testbeds.

PR #1861 short-circuits when `m_portLaneCountMap` has no entry for a given (rid, attr) pair (i.e. when the lane-count probe in `updatePortLaneCountMap` itself fails -- the Arista 7260CX3 case). For 720dt's unplugged front ports the lane-count probe **succeeds** so the map gets populated and `initAttrData()` returns true, then `sai_get_port_attribute` is invoked and BCM SAI returns either:

* `Feature unavailable (0xfffffff0)` -> `SAI_STATUS_NOT_SUPPORTED` (-2), or
* `Invalid port (0xffffffee)` -> `SAI_STATUS_INVALID_PORT_NUMBER` (-9)

Both BCM SAI driver layer and the syncd FlexCounter wrapper layer emit ERR-level syslog every poll cycle (~10s). On a fresh 720dt with ~24 unplugged front ports this generates 25,000+ ERR lines per nightly run and breaks loganalyzer at teardown -- the same downstream impact PR #1861 set out to fix, just on a different platform.

The root cause is a vendor-side inconsistency: per the SAI spec, `SAI_PORT_ATTR_RX_SIGNAL_DETECT` returns `sai_port_lane_latch_status_t` whose `current_status` is a boolean -- "no signal" is a valid value (`current_status = false` on all lanes), not an error. BCM SAI's mapping of "no module / no live signal" to NOT_SUPPORTED / INVALID_PORT_NUMBER is therefore non-spec-compliant. A vendor issue is being filed separately to correct this at the source.

Until that lands, this PR closes the syncd-side gap by aligning the probe and the actual read so they agree at addCounter time.

#### How I did it
In `PortPhyAttrContext::updatePortLaneCountMap`, after the existing lane-count probe (`get` with `count=1, list=nullptr` expecting BUFFER_OVERFLOW), perform a second `get` with a properly sized buffer allocated from the just-discovered lane count. Only register the (rid, attr) pair into `m_portLaneCountMap` when **both** phases succeed.

If the verify-poll fails:
* NOT_SUPPORTED (-2) / INVALID_PORT_NUMBER (-9): log NOTICE once explaining the attr will not be polled for this port until syncd restart, then continue without registering.
* Any other status: log ERROR (preserving existing behavior for unexpected failures), then continue without registering.

Skipped (rid, attr) pairs naturally fall through PR #1861's existing `initAttrData()` false-return path, and `collectData` continues to silently skip them every poll. No new data structure, no per-poll cost.

Cost: one extra SAI `get` per attr per port at addCounter time (syncd startup); for 64 ports x 3 attrs that is 192 additional calls total, all during initialization. Trivial.

#### How to verify it
1. Deploy on an Arista 720dt (or any Broadcom platform with unplugged front ports configured for PORT_PHY_ATTR polling) with the 202511 image.
2. Before the fix:
   ```
   show logging | grep -E "(brcm_sai_get_port_attribute_cmn|collectData: Failed to get port attr)" | wc -l
   ```
   produces thousands of ERR lines, growing every 10s.
3. After the fix: the same command produces at most ~3 NOTICE lines per unplugged port (one per attr) at syncd start, then silent.
4. Run any test with loganalyzer enabled -- teardown passes without the temporary `loganalyzer_common_ignore.txt` patches now needed for `collectData ... -2` and `collectData ... -9`.

Unit test added: `TestPortPhyAttr.UpdatePortLaneCountMapSkipsWhenVerifyPollFails`
- Mocks SAI to return BUFFER_OVERFLOW + count for the lane-count probe (`attr_count==1` with `list==nullptr`) but NOT_SUPPORTED for the verify-poll (`attr_count==1` with `list!=nullptr`).
- Asserts the verify-poll runs exactly once per attr at addCounter.
- Asserts `collectData` never reaches `sai_get` for the rejected (rid, attr) pairs across multiple poll cycles.
- Asserts COUNTERS_DB has no entries for the failing RID.

#### Why this approach (vs caching after the failure)
An earlier draft (PR #1884, now closed) cached failures in a new `m_unsupportedAttrs` map populated inside `collectData` after the first failed read. That worked but added a new data structure and required one wasted poll per (rid, attr) per syncd start before the silence kicked in. The verify-poll approach eliminates the wasted poll, reuses PR #1861's existing skip mechanism, and aligns probe/poll behavior so they cannot disagree at runtime.

#### Related PRs / issues
- #1861 -- Fix PORT_PHY_ATTR::collectData to check initAttrData return value (this PR extends the same defensive pattern by feeding it earlier)
- #1884 -- earlier (now closed) draft of this fix using a per-poll cache approach
- #1799 / #1815 -- Partial fix for PORT_PHY_SERDES_ATTR
- sonic-net/sonic-mgmt#21564 -- Temporary loganalyzer ignore for `-9` collectData failures (can be removed once this lands)
- sonic-net/sonic-mgmt#24023 -- Tracking issue for the loganalyzer impact

### DIFF
--- a/syncd/FlexCounter.cpp
+++ b/syncd/FlexCounter.cpp
@@ -1846,7 +1846,7 @@ public:
                 continue;
             }
 
-            // Query SAI for lane count (expecting BUFFER_OVERFLOW)
+            // Phase 1: probe SAI for the lane count (expecting BUFFER_OVERFLOW)
             sai_status_t status = Base::m_vendorSai->get(
                     Base::m_objectType,
                     rid,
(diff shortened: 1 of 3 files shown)


Xichen96:
Closing — work on this is being handed off to others, and on review our verify-poll approach turns out to be the wrong fix. The platform-side regression should be addressed elsewhere; this PR is no longer the right path forward.

If anything from the analysis is still useful, the discussion thread above remains in the record.

PR #1886, opened 2026-05-08
[DO NOT MERGE] [202511 backport] [syncd] PORT_PHY_ATTR: verify the actual data read at probe time

Xichen96:
## ⚠️ DO NOT MERGE — Artifact build only ⚠️

This PR exists **only to trigger Azure Pipelines** to build a `syncd` `.deb`
artifact for testing on physical Arista 720dt M0/MX testbeds. It is not for
review or merge.

The actual upstream review is happening on the master PR:
**sonic-net/sonic-sairedis#1885**

Once #1885 is merged, the proper 202511 cherry-pick will be requested via
the `Request for 202511 Branch` label on #1885 and this PR will be closed.

---

#### What it does
Adds a verify-poll phase to `PortPhyAttrContext::updatePortLaneCountMap` so
the lane-count probe and the actual `sai_get_port_attribute` agree at
addCounter time. Suppresses ~25,000 ERR-level syslog lines per nightly run
on Arista 720dt M0/MX testbeds where unplugged front ports cause BCM SAI to
return `NOT_SUPPORTED`/`INVALID_PORT_NUMBER` for `RX_SIGNAL_DETECT`,
`FEC_ALIGNMENT_LOCK`, and `RX_SNR`. Extends PR #1861's defensive pattern.

See the master PR (#1885) for full rationale, design discussion, alternative
approaches considered, and unit test details.

#### Cherry-picked commits
- `acf49fd3` `[syncd] PORT_PHY_ATTR: verify the actual data read at probe time` (cherry-pick of `045e21e7` from master branch)
- `de24757d` `[tests] aspellcheck: add Arista, dt, loganalyzer, MX, PMD, teardown` (cherry-pick of `ef3a2e52` from master branch)

#### Cross references
- Upstream master PR: sonic-net/sonic-sairedis#1885
- Predecessor: sonic-net/sonic-sairedis#1861
- Tracking issue: sonic-net/sonic-mgmt#24023
- Vendor bug filed: aristanetworks/sonic-qual.msft#1262

PR #1887, opened 2026-05-10
[test] FlexCounter.bulkChunksize: skip re-probe polls in per-counter assertions

Xichen96:
## Description

The `FlexCounter.bulkChunksize` unit test occasionally fails on PRs targeting either `master` or `202511` with messages like:

```
TestFlexCounter.cpp:1574: Failure
Expected equality of these values:
  object_count
    Which is: 1
  3
```

PR #1836 added a guard for the **unified** path (`if (unifiedBulkChunkSize > 0)`) so that re-probe polls of bulk capability with `object_count == 1` do not trigger the unified-chunk-size assertion. The comment in the same block also documents the intent to skip re-probe polls for the per-counter switch below:

```cpp
// After the merge-back, FlexCounter also re-probes
// bulk capability with single-object calls (object_count=1)
// that have all counters. Skip the assertion for both
// per-prefix polls and re-probe polls.
```

However the **implementation** only handles the per-prefix-poll case in the unified branch — the per-counter switch still fires `EXPECT_EQ` on re-probe polls, occasionally failing with `object_count=1` vs. expected 3 / 6 / 2.

Recently observed in CI on the 202511 backport build for #1885 (PR #1886) — the same flake fired on the **same identical code** that passed on master.

## Changes

`unittest/syncd/TestFlexCounter.cpp`: add the same `object_count == 1` early-skip before the `switch (counter_ids[j])` block so the implementation matches the documented intent. **No production code change.** Same fix is needed on master and 202511.

## How to verify

Local: `make check && tests/unittest/syncd/syncdtests --gtest_filter=FlexCounter.bulkChunksize --gtest_repeat=20`. Without the patch the test occasionally fails; with the patch it consistently passes.

## Which release branch to backport

- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [x] 202511

cc @theasianpianist — follow-up to #1836.

### DIFF
--- a/unittest/syncd/TestFlexCounter.cpp
+++ b/unittest/syncd/TestFlexCounter.cpp
@@ -1586,6 +1586,16 @@ TEST(FlexCounter, bulkChunksize)
                     }
                     continue;
                 }
+                // Skip re-probe polls (single-object capability probes
+                // that happen after the merge-back, with object_count==1).
+                // The per-counter assertions below check steady-state
+                // per-prefix chunk sizes, which only apply when
+                // object_count > 1. This matches the documented intent
+                // in the comment above and the unified-path guard.
+                if (object_count == 1)
+                {
+                    continue;
+                }
                 switch (counter_ids[j])
                 {
                 case SAI_PORT_STAT_IF_IN_OCTETS:


lolyu (approved):
### 📊 Overview
**Files Changed**: 1 ()
**Lines**: +10 / -0

This is a targeted test-only fix that completes the re-probe poll guard introduced in #1836. No production code is touched.

### ✅ Strengths

- **Correctly identifies the root cause**: PR #1836 added the `object_count == 1` guard for the unified path but left the per-counter `switch` block exposed to re-probe polls, causing the intermittent `object_count: 1 vs. 3/6/2` failures.
- **Minimal and focused**: The fix is exactly one guard clause, placed at the right boundary — after the per-prefix `continue` block but before the `switch` — so it applies only to re-probe polls without masking legitimate per-prefix assertions.
- **Comment is clear**: The added comment explains *why* the skip is valid (re-probe polls are single-object capability probes that precede steady-state polling) and explicitly ties the intent back to the documented comment above, making it easy for future readers to understand.
- **Consistent with existing pattern**: Mirrors the guard already in the unified-path branch (`if (unifiedBulkChunkSize > 0)`) in both structure and rationale.
- **No functional regression risk**: Test-only change; CI flakiness is eliminated without weakening coverage of the actual steady-state per-counter chunk size assertions.

### 📝 Review Findings

No issues found.

### 🧪 Testing

The fix directly addresses a known intermittent CI failure on both  and  targets. The existing  test already provides the coverage; this change makes it reliable rather than flaky.

### Status
✅ Approved — clean fix, well-documented, no concerns.

theasianpianist (approved):

lolyu:
Thanks @Xichen96, nice fix.

justin-wong-ce:
FYI @lolyu this addresses the `bulkChunkSize` flaky fail mentioned on https://github.com/sonic-net/sonic-sairedis/issues/1978 and needs a backport to 202605.

justin-wong-ce:
Oh the fix is already cherry-picked to 202605 - didn't see a reference. All good - we shouldn't see `bulkChunkSize` being flaky anymore.

PR #1894, opened 2026-05-19
feat: add VPP route flow counter support

cyw233:
### Description of PR

Summary:
Add VPP-backed route flow counter support in `vslib/vpp`.

This change makes SAI route counters attached through `SAI_ROUTE_ENTRY_ATTR_COUNTER_ID` read from VPP route statistics instead of only existing as logical SAI counter objects. When a route is programmed into VPP, the VPP route `stats_index` is captured and cached. When flex counter polling reads a SAI counter bound to a route, the VPP SAI layer resolves the counter back to the route, reads VPP route packet/byte stats, subtracts the saved base value, and returns SAI counter stats.

Main changes:
- Capture VPP route `stats_index` from route add replies.
- Track route-to-counter and counter-to-route bindings for `SAI_ROUTE_ENTRY_ATTR_COUNTER_ID`.
- Add a VPP route stats query helper for `/net/route/to`.
- Dispatch route-bound SAI counter reads to VPP route stats.
- Implement clear/read-and-clear behavior with saved base values.
- Refresh route stats index and base values on route updates and next-hop group member add/remove.
- Clean up route counter state when routes or counters are removed.

Fixes # (issue): https://github.com/sonic-net/sonic-platform-vpp/issues/229

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

VPP route flow counters need to reflect dataplane traffic for matched routes. Before this change, the VPP SAI route counter path did not keep the VPP route stats index, did not maintain route/counter bindings, and did not redirect SAI counter reads to VPP route statistics. As a result, route flow counters could be created without packet and byte counters being backed by VPP route stats.

Reviewers may want to start with:
- `vslib/vpp/SwitchVppRoute.cpp` for route create/update/remove and counter binding lifecycle.
- `vslib/vpp/SwitchVpp.cpp` for route-bound counter stats reads and clear behavior.
- `vslib/vpp/vppxlate/SaiVppXlate.c` for VPP route `stats_index` capture.
- `vslib/vpp/vppxlate/SaiRouteStats.c` for VPP route stats query logic.

##### Work item tracking

#### How did you do it?

The VPP SAI layer now treats a route-bound SAI counter as a view over the matching VPP route stats entry.

Implementation details:
- Extended route add/delete plumbing to optionally return a VPP route `stats_index`.
- Stored the route stats index in `m_routeStatsIndexMap`, keyed by serialized route entry.
- Added `m_counterToRouteMap` and `m_routeToCounterMap` so stats reads can resolve counter object ID to route and route updates/removes can resolve route to counter.
- Added `m_routeCounterStatsBaseMap` to store base packet/byte values used for clear and read-and-clear modes.
- Added `getRouteStatsExt()` and `getRouteCounterStats()` so `getStatsExt()` can handle route-bound SAI counter objects by querying VPP route stats.
- Added `vpp_route_stats_query()` to read VPP route packet and byte counters from `/net/route/to` by `stats_index`.
- Rejected non-null counter binding when a route has no valid VPP stats index, instead of silently accepting an unreadable route counter.
- Refreshed stats index and base counter values when route programming changes, including next-hop group member add/remove.
- Removed route counter bindings and cached stats state when the route or counter is deleted.

Ran the changes with the "flow counters increment with traffic" test gap fix in https://github.com/sonic-net/sonic-mgmt/pull/24553 and can confirm they all passed
image

This change is specific to the VPP virtual switch implementation in `vslib/vpp`. It does not change non-VPP vendor SAI behavior.

### Documentation

Documentation is tracked in the SONiC VPP platform HLD: https://github.com/sonic-net/sonic-platform-vpp/pull/236


vslib/vpp/vppxlate/SaiRouteStats.c
```diff
@@ -0,0 +1,67 @@
...
+  char pathbuf[] = "/net/route/to";
+  vpp_route_stats_query_t query;
+
+  memset(stats, 0, sizeof(*stats));
+
+  query.stats_index = stats_index;
+  query.stats = stats;
+  query.found = false;
+
+  if (vpp_stats_dump(pathbuf, NULL, handle_stat_two, &query) != 0)
```
lolyu, line 61: For a device with N monitored routes, R total routes, this function will be called N times, right? And each time this dump function will loop all R routes, the total complexity is `R * N`
Little bit concerned about the performance here.

vslib/vpp/SwitchVppNexthop.cpp
```diff
@@ -317,11 +317,45 @@ SwitchVpp::createNexthopGroupMember(
 
     // Add the specific path to each route using this NHG
     for (auto route : *routes) {
```
lolyu: createNexthopGroupMember/removeNexthopGroupMember have common codes, better to wrap into a helper function

vslib/vpp/SwitchVpp.cpp
```diff
@@ -1198,6 +1391,21 @@ sai_status_t SwitchVpp::remove(
...
+        sai_object_id_t objectId;
+        sai_deserialize_object_id(serializedObjectId, objectId);
+
+        auto routeIt = m_counterToRouteMap.find(objectId);
+        if (routeIt != m_counterToRouteMap.end())
+        {
+            std::string route = routeIt->second;
+            removeRouteCounterBinding(route);
+        }
+        m_routeCounterStatsBaseMap.erase(objectId);
```
lolyu, line 1560: LINE 1405/1406 are redundant erase when bound

vslib/vpp/vppxlate/SaiVppXlate.c
```diff
@@ -462,25 +465,53 @@ void vpp_mutex_unlock ()
...
-    return 1;
+    for (uint32_t idx = 1; idx < VPP_MAX_CTX; idx++) {
+        uint32_t mask = (uint32_t)(1 << idx);
+        if ((idx_map.index_map & mask) == 0) {
+            idx_map.index_map |= mask;
+            return idx;
+        }
+    }
+
+    return 0;
```
yue-fred-gao: Can you define a constant like INVALID_INDEX or some meaningful one with value 0? And a one line comment to explain index 0 is reserved

vslib/vpp/vppxlate/SaiVppXlate.c
```diff
@@ -383,10 +383,13 @@ do {                                                            \
...
 
-#define VPP_MAX_CTX 2
+#define VPP_MAX_CTX 16
+#define VPP_CTX_INDEX_MASK 0xff
+#define VPP_CTX_GENERATION_SHIFT 8
 typedef struct _vpp_index_map_ {
-    uint8_t index_map;
+    uint32_t index_map;
     uintptr_t ptr[VPP_MAX_CTX];
+    uint32_t generation[VPP_MAX_CTX];
```
yue-fred-gao, line 414: nice enhancement

vslib/vpp/vppxlate/SaiRouteStats.c
yue-fred-gao: Agreed. Maybe you can implement a bulk stats dump with multiple indexes. In one dump, you can retrieve counters of all the supplied indexes. Routes have much higher scale than interface and acl. We need to be careful here.

vslib/vpp/SwitchVpp.h
```diff
@@ -677,6 +708,11 @@ namespace saivs
             std::map<sai_object_id_t, std::list<sai_object_id_t>> m_acl_tbl_grp_mbr_map;
             std::map<sai_object_id_t, std::list<sai_object_id_t>> m_acl_tbl_grp_ports_map;
             std::map<sai_object_id_t, vpp_ace_cntr_info_t> m_ace_cntr_info_map;
+            std::map<sai_object_id_t, std::string> m_counterToRouteMap;
```
yue-fred-gao: I hope we don't maintain the map between objects individually  but use SaiObjectDB. You can add below mapping to sai_child_relation_defs in SaiObjectDB.cpp (second entry is new)
{SAI_OBJECT_TYPE_ROUTE_ENTRY, {{SAI_OBJECT_TYPE_NEXT_HOP_GROUP, SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID, SAI_ATTR_VALUE_TYPE_OBJECT_ID},
                               {SAI_OBJECT_TYPE_COUNTER, SAI_ROUTE_ENTRY_ATTR_COUNTER_ID, SAI_ATTR_VALUE_TYPE_OBJECT_ID}}},
For m_routetoCounterMap lookup, you can do
auto route_obj = m_object_db.get(SAI_OBJECT_TYPE_ROUTE_ENTRY, serializedRouteId);
auto linked_counter = route_obj->get_linked_object(SAI_OBJECT_TYPE_COUNTER, SAI_ROUTE_ENTRY_ATTR_COUNTER_ID);

for the reverse look up, you need to use get_child_objs.
auto counter_obj = m_object_db.get(SAI_OBJECT_TYPE_COUNTER, sai_serialize_object_id(counter_oid));
auto routes = counter_obj->get_child_objs(SAI_OBJECT_TYPE_ROUTE_ENTRY);

vslib/vpp/vppxlate/SaiRouteStats.c
yue-fred-gao: I see the difficulty after I read more of the code. Currently the design is one route index per call so it is hard to convert to a batch update. Maybe we can maintain a cache with 5 second expiration or even shorter? The cache is populated with vpp_stats_dump with all the route stats. When the cache is valid, the get will use it. Otherwise, it can trigger a cache refresh. This might be more efficient and less change.

vslib/vpp/SwitchVppNexthop.cpp
```diff
@@ -386,11 +420,45 @@ SwitchVpp::removeNexthopGroupMember(
...
+            }
+        }
+
         SWSS_LOG_INFO("NHG member removed. Removing path from route %s", route.first.c_str());
-        status = IpRoutePathAddRemove(route.second.get(), &member, false);
+        status = IpRoutePathAddRemove(route.second.get(), &member, false, &stats_index);
         if (status != SAI_STATUS_SUCCESS) {
             SWSS_LOG_ERROR("Failed to remove path from route %s, status %d", route.first.c_str(), status);
             // Continue with other routes
+        } else if (stats_index != UINT32_MAX) {
```
yue-fred-gao: This is a good candidate to make it a function. It is to improve reusability also more readable with clear intention of the function.

vslib/vpp/SwitchVppRoute.cpp
```diff
@@ -204,30 +206,129 @@ sai_status_t SwitchVpp::IpRouteAddRemove(
...
+        removeRouteCounterBinding(serializedObjectId);
+        return SAI_STATUS_SUCCESS;
+    }
+
+    sai_object_id_t old_counter_oid = SAI_NULL_OBJECT_ID;
+    std::map<sai_stat_id_t, uint64_t> old_counter_base;
+    std::map<sai_stat_id_t, uint64_t> old_counter_carry;
+    bool has_old_counter_base = false;
+    bool has_old_counter_carry = false;
+    auto oldCounterIt = m_routeToCounterMap.find(serializedObjectId);
```
yue-fred-gao: can this really happen that a route is bound with another counter id?

vslib/vpp/SwitchVppRoute.cpp
```diff
@@ -352,22 +534,217 @@ sai_status_t SwitchVpp::updateIpRoute(
...
+            {
+                return status;
+            }
+            m_routeStatsIndexMap.erase(serializedObjectId);
+        }
+
+        auto status = IpRouteAddRemove(&route_mod_obj, true, &stats_index);
+        if (status != SAI_STATUS_SUCCESS)
+        {
+            auto rollback_status = rollbackRoute(route_db_obj.get(), NULL, old_stats_index);
```
yue-fred-gao, line 690: There are lots rollback code which is the current implementation lacks. But I feel we need a more systematic way. For example, we can use transaction log. For each vpp api call, it should have a matching call of reverse operation. For each such success call, we enque the reverse operation with all the parameters needed for the call. We can do the same for the internal mappings. For the SaiObjectDB map, we typically leave it at the end so it won't be written to the db until everything succeeds. This is something to consider for the future.

vslib/vpp/vppxlate/SaiRouteStats.c
cyw233: Nice catch. Done: added a single-pass `vpp_route_stats_dump_all` API and a short-lived full-dump cache so each polling cycle does one dump plus O(1) lookups.

vslib/vpp/SwitchVppNexthop.cpp
cyw233: Good catch, done.

vslib/vpp/vppxlate/SaiVppXlate.c
cyw233: Good suggestion, I introduced a new variable called `VPP_INVALID_CTX_INDEX`

vslib/vpp/SwitchVpp.h
cyw233: Done. Added the `{COUNTER, SAI_ROUTE_ENTRY_ATTR_COUNTER_ID, OBJECT_ID}` child relation in `sai_child_relation_defs` and removed `m_routeToCounterMap` / `m_counterToRouteMap`, replacing them with three helpers: `getRouteBoundCounter`, `getCounterBoundRoute`, and `readRouteStatsByIndex`.

cyw233: The three stats/index maps stay as accounting state.

vslib/vpp/SwitchVppNexthop.cpp
cyw233: Done

vslib/vpp/SwitchVppRoute.cpp
cyw233: Good question. It cannot happen from the create path (fresh route, no prior binding). It can only happen via the `SAI_ROUTE_ENTRY_ATTR_COUNTER_ID` set path, and a different counter (A -> B) only if a client issues a direct rebind without setting NULL first, which is SAI-legal but not expected from current SONiC `FlowCounterRouteOrch`. I kept the capture, added a clarifying comment, and log a warning if an unexpected A -> B rebind is ever seen.

vslib/vpp/SwitchVppRoute.cpp
cyw233: Agree, and the current rollback is indeed scattered. Maybe we can we track it as a separate future design item rather than in this PR

vslib/vpp/SwitchVpp.cpp
```diff
@@ -746,6 +769,296 @@ void SwitchVpp::setPortStats(
...
+        return SAI_STATUS_FAILURE;
+    }
+
+    stats[SAI_COUNTER_STAT_PACKETS] = route_stats.packets;
+    stats[SAI_COUNTER_STAT_BYTES] = route_stats.bytes;
+
+    return SAI_STATUS_SUCCESS;
+}
+
+sai_object_id_t SwitchVpp::getRouteBoundCounter(
```
yue-fred-gao, line 828: Is it better to pass route SaiObject and return counter SaiObject? I see some cases the caller already has route SaiObject, then convert to id to call this function. This function then read the SaiObject by the ID. It is not very efficient. 
For the case where the caller doesn't have SaiObject, it is easy to load the object from the db before calling this function.

Please consider the same for the other 2 functions

cyw233: Good point on the redundant object -> id ->object round-trip. I looked at all the call sites and applied it where it's a true case, but kept the ID-based contract for the rest. 

Added a `getRouteBoundCounter(const SaiObject*)` overload and used it in `removeIpRoute`.

The other callers of `getRouteBoundCounter` / `getCounterBoundRoute` are in the create/set/rollback paths, where these helpers must read the committed binding from  SaiObjectDB  before `create_internal` / `set_internal` persists the new one. The object the caller holds there is the in-flight `SaiCachedObject` carrying the new attributes (a different object). Passing it in would silently read the new binding and break the old-counter capture /  rollback logic, so the re-fetch by ID is deliberate.

`getCounterBoundRoute`  /  `getRouteCounterStats`: their callers only hold a counter `sai_object_id_t`, never a counter `SaiObject`, so switching those to take/return objects would just push the same `SaiObjectDB` lookup to each caller plus boilerplate, with no net saving.

yue-fred-gao (approved):

PR #1895, opened 2026-05-19
[vpp]: Program sub-interface IP addresses directly into VPP

### LINKED ISSUE #227: PortChannel/LAG subinterfaces are not working on SONiC-VPP

Bojun-Feng:
### Description of PR

Summary:
Sub-interface IP addresses are not programmed into VPP, causing VPP to be unable to respond to packets destined to its own sub-interface IPs (e.g., ICMP echo-request to a gateway address on Ethernet64.20 or PortChannel1.20). Inter-VLAN L3 forwarding through VPP works, but local-delivery is broken.

The root cause is that `vpp_add_del_intf_ip_addr` was commented out in the `ROUTER_INTERFACE` route case during the Debian Bullseye migration because the original function hard-rejected non-PORT interface types. VPP's LCP plugin handles IP sync for regular ports, but does not handle sub-interfaces. Additionally, the PortChannel branch in `vpp_add_del_intf_ip_addr_norif` never appended the `.vlan_id` suffix, so LAG sub-interfaces would get the IP on the parent bond instead of the sub-interface.

Fixes https://github.com/sonic-net/sonic-platform-vpp/issues/227

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

`test_packet_routed_with_valid_vlan[port]` and `test_packet_routed_with_valid_vlan[port_in_lag]` both fail on VPP because VPP sub-interfaces have no IP programmed — they can forward traffic between sub-ports but cannot reply to packets destined to themselves.

##### Work item tracking

#### How did you do it?

* Wire `vpp_add_del_intf_ip_addr_norif` into the ROUTER_INTERFACE route case for SUB_PORT RIFs
* Fix PortChannel branch in `vpp_add_del_intf_ip_addr_norif` to append `.vlan_id` for sub-interfaces
* Bump all `hw_subifname` buffers from [32] to [64] to prevent truncation on long hwif names

Todo: Run `test_packet_routed_with_valid_vlan[port]` and `test_packet_routed_with_valid_vlan[port_in_lag]` in the t1-lag-vpp topology (Platform: kvm, ASIC type: vpp).

VPP platform only. Regular PORT RIFs are unaffected (gated by `SAI_ROUTER_INTERFACE_TYPE_SUB_PORT` check). LCP-based IP sync for non-sub-interface ports remains unchanged.

### Documentation

N/A

### DIFF
--- a/vslib/vpp/SwitchVppRif.cpp
+++ b/vslib/vpp/SwitchVppRif.cpp
@@ -364,7 +364,7 @@ bool SwitchVpp::vpp_get_hwif_name (
     }
 
     const char *hwifname = tap_to_hwif_name(if_name.c_str());
-    char hw_subifname[32];
+    char hw_subifname[64];
     const char *hw_ifname;
 
     if (vlan_id) {
@@ -785,7 +785,7 @@ sai_status_t SwitchVpp::vpp_add_del_intf_ip_addr (
     }
 
     const char *hwifname = tap_to_hwif_name(if_name.c_str());
-    char hw_subifname[32];
+    char hw_subifname[64];
     const char *hw_ifname;
 
     if (vlan_id) {
@@ -1001,7 +1001,7 @@ sai_status_t SwitchVpp::vpp_add_del_intf_ip_addr_norif (
     }
 
     const char *hwifname;
-    char hw_subifname[32];
+    char hw_subifname[64];
     char hw_bviifname[32];
     char hw_bondifname[32];
     const char *hw_ifname;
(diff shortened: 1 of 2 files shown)


Bojun-Feng:
Closing — changes consolidated into #1883

PR #1909, opened 2026-05-26
[syncd] Survive transient redis disconnect in FlexCounter and notification threads

augusdn:
### Description of PR

`syncd-vpp` does not survive a transient redis (database container) disconnect: it dies with `SIGABRT` from `std::terminate`, originating in two `std::thread` bodies that have no top-level `try/catch`.

Reproducer (sonic-platform-vpp, vlab-vpp-01):
```bash
# inside syncd container, before patch:
docker exec database kill -SIGKILL $(docker exec database pgrep -f /usr/bin/redis-server)
# supervisord log inside syncd container:
WARN exited: syncd (terminated by SIGABRT (core dumped); not expected)
```

The exception that escapes is `swss::RedisError("RedisError: Failed to redisGetReply in RedisPipeline::pop, err=1: errstr=Broken pipe")` thrown by `RedisPipeline::pop()` ([sonic-swss-common/common/redispipeline.h:124](https://github.com/sonic-net/sonic-swss-common/blob/master/common/redispipeline.h#L124)). The two unprotected thread bodies are:

1. `FlexCounter::flexCounterThreadRunFunction` — calls `collectCounters(...)` / `runPlugins(...)`, both of which write to the FlexCounter `RedisPipeline` and then `flush()`. If redis is gone mid-flush, `pop()` throws.
2. `NotificationProcessor::ntf_process_function` — calls `processNotification(item)`, which in turn writes to the notification redis channel.

On Broadcom platforms this isn't visible because `dummy_syncd` doesn't drive these code paths the same way; on VPP, `saivpp` does, so the bug surfaces immediately.

This PR catches `std::exception` at the top of both `std::thread` bodies (the same `catch(std::exception)` pattern already used elsewhere in syncd — `Syncd.cpp:4896, 5527, 5562, 5860, 6004`). For the FlexCounter thread, the catch also **recreates** `DBConnector` + `RedisPipeline` + `Table`, because a `pop()` throw leaves `m_remaining > 0` in the pipeline and the next loop iteration would just throw the same error forever once redis is back.

This unblocks the `test_critical_process_monitoring` test on VPP (sonic-mgmt#25821).

Summary:
Fixes the SIGABRT half of https://github.com/sonic-net/sonic-buildimage/issues/25821 (the `process_monitoring` test failure on VPP). The remaining `database`-container-restart concerns (rsyslog/script restart) are addressed in sonic-mgmt / sonic-platform-vpp follow-ups, not here.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

`syncd` should not abort when its redis connection is briefly lost. The current behavior makes any redis hiccup a fatal crash, which is overly aggressive: redis losing a connection (e.g. supervisord-restarting redis-server inside the `database` container) is a *transient* condition that the rest of SONiC is expected to ride through.

#### How did you do it?

`syncd/FlexCounter.cpp` — `flexCounterThreadRunFunction`:
- Convert raw `db`, `pipeline`, `countersTable` pointers/locals to `std::unique_ptr` so they can be reset and recreated.
- Wrap the per-iteration `collectCounters(...)` + `runPlugins(...)` calls in `try/catch (const std::exception &e)`.
- On catch, log the error and **recreate** the unique_ptrs. This drops the broken pipeline (and its non-zero `m_remaining`) so the next iteration starts clean. A nested `try/catch` around the recreate avoids tight-looping if redis is still unreachable.

`syncd/NotificationProcessor.cpp` — `ntf_process_function`:
- Wrap `processNotification(item)` in `try/catch (const std::exception &e)`.
- On catch, log and continue. No state recreation needed — `processNotification` uses short-lived redis writers per call.

On a `sonic-platform-vpp` testbed (`vlab-vpp-01`, KVM, t1-lag):

1. Built `syncd-vs_1.0.0_amd64.deb` from this branch via `sonic-buildimage` (`BLDENV=bookworm SONIC_BUILD_JOBS=1 make -f Makefile.work target/debs/bookworm/syncd-vs_1.0.0_amd64.deb`).
2. Extracted the patched `/usr/bin/syncd` and hot-swapped it into the running `syncd` container on the DUT.
3. Reproduced the kill: `docker exec database kill -SIGKILL <redis-server pids>` (both 6379 and 6400).
4. Observed: **no SIGABRT**, syncd's PID stayed unchanged through redis death and through supervisord respawning `redis-server`.

Before patch:
```
WARN exited: syncd (terminated by SIGABRT (core dumped); not expected)
```

After patch:
```
syncd  RUNNING  pid 71, uptime 0:02:31    # same PID across redis kill+restart cycle
>>> SUCCESS: syncd (pid 71) survived redis SIGKILL <<<
>>> OVERALL SUCCESS: syncd (pid 71) survived redis-server restart cycle <<<
```

No new tests added — existing `test_critical_process_monitoring` in sonic-mgmt is the integration coverage; this PR unblocks the syncd half of it.

The bug is visible on `sonic-platform-vpp` because `saivpp` implements real counter collection and the FlexCounter pipeline has pending writes when redis dies. On platforms whose SAI doesn't drive the same FlexCounter / notification paths under load, the throw site can go unhit, which is why this hasn't been noticed before.

The fix is platform-neutral — it hardens the `std::thread` bodies for all syncd consumers.

### DIFF
--- a/syncd/FlexCounter.cpp
+++ b/syncd/FlexCounter.cpp
@@ -3535,9 +3535,32 @@ void FlexCounter::flexCounterThreadRunFunction()
     // (a partially-drained pipeline leaves m_remaining nonzero and would loop
     // on the same error forever), and let the thread keep polling so it
     // recovers naturally when redis comes back.
-    auto db = std::make_unique<swss::DBConnector>(m_dbCounters, 0, m_isTcpConn);
-    auto pipeline = std::make_unique<swss::RedisPipeline>(db.get());
-    auto countersTable = std::make_unique<swss::Table>(pipeline.get(), COUNTERS_TABLE, true);
+    std::unique_ptr<swss::DBConnector> db;
+    std::unique_ptr<swss::RedisPipeline> pipeline;
+    std::unique_ptr<swss::Table> countersTable;
+
+    auto recreateRedisObjects = [&]() -> bool {
+        try
+        {
+            countersTable.reset();
+            pipeline.reset();
+            db.reset();
+            db = std::make_unique<swss::DBConnector>(m_dbCounters, 0, m_isTcpConn);
+            pipeline = std::make_unique<swss::RedisPipeline>(db.get());
+            countersTable = std::make_unique<swss::Table>(pipeline.get(), COUNTERS_TABLE, true);
+            return true;
+        }
+        catch (const std::exception &re)
+        {
+            SWSS_LOG_ERROR("FlexCounter %s: failed to (re)create redis pipeline: %s",
+                           m_instanceId.c_str(), re.what());
+            return false;
+        }
+    };
+
+    // Initial creation. If it fails (e.g. database not yet up), the loop will
+    // keep retrying via recreateRedisObjects() before each collection cycle.
+    recreateRedisObjects();
 
     while (m_runFlexCounterThread)
     {
@@ -3547,6 +3570,24 @@ void FlexCounter::flexCounterThreadRunFunction()
         {
             auto start = std::chrono::steady_clock::now();
 
+            // Ensure redis objects are live before use. A previous iteration's
+            // exception handler tore them down; if recreate still fails (redis
+            // is still down), skip this poll and try again next iteration.
+            // Never call collectCounters/runPlugins with a null countersTable
+            // or null db - dereferencing a null unique_ptr is undefined
+            // behaviour and segfaults inside the callee.
+            if ((!db || !pipeline || !countersTable) && !recreateRedisObjects())
+            {
+                // Snapshot poll interval under the mutex before sleeping; the
+                // setter is guarded by m_mtx, so reading it here keeps the
+                // wait synchronized.
+                uint32_t sleepMs = m_pollInterval;
+                MUTEX_UNLOCK; // explicit unlock before sleeping
+                std::unique_lock<std::mutex> lk(m_mtxSleep);
+                m_cvSleep.wait_for(lk, std::chrono::milliseconds(sleepMs));
+                continue;
+            }
+
             try
             {
                 collectCounters(*countersTable);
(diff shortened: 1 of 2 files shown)


augusdn:
### v2 update: avoid SEGV when redis recreate fails after exception

Pushed `68db2a9d` to address a regression introduced by the first commit on this branch.

#### What went wrong with v1
The original `try`/`catch` in `flexCounterThreadRunFunction` tore down the redis
`DBConnector`/`RedisPipeline`/`Table` inside the catch handler and recreated them in
the same block. If the recreate **itself** threw (redis still unreachable when the next
`swss::DBConnector` constructor ran), the nested catch logged and fell through, leaving
`countersTable` as a reset (null) `unique_ptr`. The next loop iteration evaluated
`*countersTable` to pass `swss::Table&` into `collectCounters()` →
`CounterContext::collectData()` → `Table::set()`, dereferencing the null pointer and
faulting with **SIGSEGV at FlexCounter.cpp:1174**. So v1 traded one crash signal (SIGABRT)
for another (SIGSEGV) on the exact same code path.

Verified end-to-end on `vlab-vpp-01` (sonic-vpp KVM testbed) with the v1 binary — 3-for-3
reproduction of the SIGSEGV.

#### v2 fix
- Move `db`/`pipeline`/`countersTable` declarations outside the loop as null
  `unique_ptr`s.
- Extract recreate logic into a `recreateRedisObjects()` lambda that returns `bool` and
  never throws across its boundary.
- At the **top** of each iteration, if any of the three pointers is null, call
  `recreateRedisObjects()`. If recreate fails, snapshot `m_pollInterval` under the mutex
  (matches the existing `correction` pattern), `MUTEX_UNLOCK`, sleep on `m_cvSleep`
  for that interval, and `continue` — never proceed to `collectCounters` with a null
  `Table`.
- Catch handler is now **teardown-only** (reset the three pointers). Recreate happens at
  the next iteration's top under the null-check guard.

#### Verification
Built `syncd-vs` from this branch, hot-swapped onto `vlab-vpp-01`, ran:

\\\
tests/process_monitoring/test_critical_process_monitoring.py::test_monitoring_critical_processes
  --skip_vendor_specific_container=database
\\\

(the `--skip_vendor_specific_container=database` flag matches physical-testbed behavior —
on Broadcom physical, `FEATURE|database` is absent, so the test never kills redis; on VPP
`FEATURE|database = always_enabled`, so without the flag the test also kills redis in the
`database` container, which is out of scope for this sairedis PR. A companion sonic-mgmt
PR adds this skip permanently for sonic-vpp.)

Result:
\\\
=================== 1 passed, 279 warnings in 501.93s (0:08:21) ===================
\\\

syncd PID 68 stayed RUNNING through the entire `swss` kill burst (all 14 programs:
`orchagent`, `portsyncd`, `neighsyncd`, `fdbsyncd`, `vlanmgrd`, `intfmgrd`,
`portmgrd`, `fabricmgrd`, `buffermgrd`, `vrfmgrd`, `nbrmgrd`, `vxlanmgrd`,
`coppmgrd`, `tunnelmgrd`) and was cleanly SIGKILLed by the test's syncd-container loop
iteration. Behavior mirrors a successful physical-testbed reference run
(`7260cx3.t1.202511_NightlyTest_20260525.1`, plan id `6a141da095e3f84fd1e9a37e`).

No new cores. /cc @sonic-net/sonic-fundamentals

PR #1912, opened 2026-05-28
[meta]: Add SAI_OBJECT_TYPE_METER_BUCKET_ENTRY to MetaKeyHasher

lolyu:
### What I did

* Microsoft ADO (number only): 38615565

Added handling for `SAI_OBJECT_TYPE_METER_BUCKET_ENTRY` in [meta/MetaKeyHasher.cpp](meta/MetaKeyHasher.cpp):
- `operator==` overload for `sai_meter_bucket_entry_t` comparing `(switch_id, eni_id, meter_class)`.
- `sai_get_hash` for the same struct.
- `case` in the extension switch of both `MetaKeyHasher::operator()` (hash) and `MetaKeyHasher::operator()` (equality).
- Unit test `MetaKeyHasher.operator_eq_meter_bucket_entry` in [unittest/meta/TestMetaKeyHasher.cpp](unittest/meta/TestMetaKeyHasher.cpp).

### Why I did it

`MetaKeyHasher::operator()` / `operator==` are the hasher/comparator used by the unordered containers keyed by `sai_object_meta_key_t` inside [meta/Meta.cpp](meta/Meta.cpp) (e.g. `m_saiObjectCollection`). Any meta-key whose object type isn't enumerated in the two switch statements falls through to `SWSS_LOG_THROW("not handled: ...")`.

`SAI_OBJECT_TYPE_METER_BUCKET_ENTRY` is a DASH extension entry type already supported by `sai_serialize_meter_bucket_entry` / `sai_deserialize_meter_bucket_entry` in [meta/SaiSerialize.cpp](meta/SaiSerialize.cpp), but it was never added to the hasher. As a result, unit tests `Meta.quad_mcast_fdb_entry` and `Meta.quad_impc_entry` started failing with:

```
:- operator(): not handled: SAI_OBJECT_TYPE_METER_BUCKET_ENTRY
```

when the meta layer iterated extension entry metadata while validating create/get/remove on mcast_fdb / ipmc entries. The companion test `Meta.remove_meter_bucket_entry` passes because its code path does not exercise the hasher branch.

### How I verified it

- New unit test `MetaKeyHasher.operator_eq_meter_bucket_entry` exercises both `operator()` overloads (hash + equality) with `SAI_OBJECT_TYPE_METER_BUCKET_ENTRY` keys.
- Existing meta unit tests `Meta.quad_mcast_fdb_entry` / `Meta.quad_impc_entry` should now pass on CI.

### Details if related

Hash is computed from `(eni_id, meter_class)`, matching the pattern used for other extension entries such as `SAI_OBJECT_TYPE_ENI_TRUSTED_VNI_ENTRY`. `switch_id` is intentionally not folded into the hash (consistent with other entry hashers) but is included in the equality check.

### DIFF
--- a/meta/MetaKeyHasher.cpp
+++ b/meta/MetaKeyHasher.cpp
@@ -357,6 +357,15 @@ static bool operator==(
     return a.switch_id == b.switch_id && a.eni_id == b.eni_id && a.vni_range == b.vni_range;
 }
 
+static bool operator==(
+        _In_ const sai_meter_bucket_entry_t& a,
+        _In_ const sai_meter_bucket_entry_t& b)
+{
+    // SWSS_LOG_ENTER(); // disabled for performance reasons
+
+    return a.switch_id == b.switch_id && a.eni_id == b.eni_id && a.meter_class == b.meter_class;
+}
+
 bool MetaKeyHasher::operator()(
         _In_ const sai_object_meta_key_t& a,
         _In_ const sai_object_meta_key_t& b) const
@@ -427,6 +436,9 @@ bool MetaKeyHasher::operator()(
 
     if ((sai_object_type_extensions_t)a.objecttype == SAI_OBJECT_TYPE_ENI_TRUSTED_VNI_ENTRY)
         return a.objectkey.key.eni_trusted_vni_entry == b.objectkey.key.eni_trusted_vni_entry;
+
+    if ((sai_object_type_extensions_t)a.objecttype == SAI_OBJECT_TYPE_METER_BUCKET_ENTRY)
+        return a.objectkey.key.meter_bucket_entry == b.objectkey.key.meter_bucket_entry;
 
     SWSS_LOG_THROW("not implemented: %s",
             sai_serialize_object_meta_key(a).c_str());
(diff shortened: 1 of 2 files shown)


yxieca (approved):
Nice, clean tail-fix — pattern-matches the existing `eni_trusted_vni_entry` shape exactly, all 3 fields of `_sai_meter_bucket_entry_t` covered in equality, hash deliberately omits `switch_id` consistent with the rest of the DASH extension entries. Verified against the SAI header (`saiexperimentaldashmeter.h:50-71`).

### Grounded nit (non-blocking)

**Test inequality coverage is asymmetric** — `operator_eq_meter_bucket_entry` only flips `meter_class`. The current assertion would still pass if `operator==` accidentally read `a.eni_id == b.eni_id && a.eni_id == b.eni_id && a.meter_class == b.meter_class` (eni_id clause repeated). Consider adding a parallel `EXPECT_FALSE` that flips `eni_id` (with `meter_class` fixed), and ideally one that flips `switch_id` too, so each of the three compared fields is independently exercised.

### Forward-looking suggestion (separate PR)

Every new DASH extension entry triggers the same 4-piece patch (`operator==`, `sai_get_hash`, equality-dispatch case, hash-dispatch case) plus a near-identical unit test — and forgetting any of the four pieces silently throws at runtime, which is how this bug landed. Worth considering a follow-up to collapse the flat-POD entry hashers into a macro (or template + traits) so that adding a new extension entry becomes a one-line declaration that generates all four pieces *and* the dispatch entry together. The IPv4/IPv6-family entries (`route_entry`, `neighbor_entry`, etc.) would stay custom; this would only touch the ~15 flat DASH-style ones. Not in scope for this PR — flagging for follow-up.

Reviewed by AI agent on behalf of Ying.

PR #1913, opened 2026-05-28
[sonic-sairedis] vslib: extend native hash field list with RDMA BTH fields

jithenderkondam:
This change is a new enhancement to support the two RDMA native hash fields required by the Generic Hash feature. SONiC HLD https://github.com/sonic-net/SONiC/pull/2100 adds per-packet-type ECMP/LAG hashing (IPv4, IPv6, IPNIP, IPV4_RDMA, IPV6_RDMA) and introduces RDMA_BTH_OPCODE and RDMA_BTH_DEST_QP as configurable hash fields for RDMA traffic. vslib should advertise these fields through queryHashNativeHashFieldListCapability so SAI capability discovery on the virtual switch reflects the updated native hash field set used by the Generic Hash stack.

#### How did you do it?
- Extended SwitchStateBase::queryHashNativeHashFieldListCapability in vslib/SwitchStateBase.cpp to return 21 native hash fields (was 19), adding SAI_NATIVE_HASH_FIELD_RDMA_BTH_OPCODE and SAI_NATIVE_HASH_FIELD_RDMA_BTH_DEST_QP.
- Updated the switchHashCapabilitiesGet unit test in unittest/vslib/TestSwitchStateBase.cpp to include the two new RDMA fields in the expected capability list.

Run the sonic-swss unit tests.

No

### Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

Which release branch to backport (provide reason below if selected):
- [x] 202511

Tested branch (Please provide the tested image version):
- [x] 202511


yxieca:
Thanks for the change — the production side looks correct: `RDMA_BTH_OPCODE` (0x22) and `RDMA_BTH_DEST_QP` (0x23) both exist at the pinned SAI submodule SHA, and the `19 → 21` count + index bookkeeping in `queryHashNativeHashFieldListCapability` is consistent.

One blocking item, though — it's why CI (`Build amd64` / `BuildAsan amd64`) is currently red:

**There's a second test that pins this same VS capability set and still expects 19 fields.**
`syncd/tests/TestSyncdMlnx.cpp::queryAttrEnumValuesCapability` builds its expected set `hfSet1` and asserts `ASSERT_EQ(hfSet1, hfSet2)`, where `hfSet2` comes from the capability query. That test links `-lsaivs` with profile `SAI_VS_SWITCH_TYPE_MLNX2700`, so it routes into the very function this PR changes — which now returns 21. Since `hfSet1` there still lists only the original 19 (ending at `SAI_NATIVE_HASH_FIELD_IPV6_FLOW_LABEL`), the assertion fails.

Fix is the same 2-line addition you already made in `TestSwitchStateBase.cpp` — add these two to `hfSet1` in `TestSyncdMlnx.cpp` (around L159-179):
```cpp
SAI_NATIVE_HASH_FIELD_RDMA_BTH_OPCODE,
SAI_NATIVE_HASH_FIELD_RDMA_BTH_DEST_QP
```

Non-blocking nit: the field count `21` is now hard-coded in three spots in `queryHashNativeHashFieldListCapability` (the `< 21` guard, the overflow-path assignment, and the success-path assignment), and the expected set is duplicated across two test files. Not for this PR, but a named constant / array-derived count would make the next field addition a single-site change and avoid exactly this "updated one copy, missed the other" situation.

@lolyu could you take a further look as well? Thanks!

PR #1918, opened 2026-05-29
[smartswitch][meta] add cache policy for specific DASH entry types

nhegde-microsoft:
#### Summary
Addresses [#26683 ](https://github.com/sonic-net/sonic-buildimage/issues/26683)

#### Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

 Issue #26683 exposed a memory scaling issue in smartswitch when large number
 of Dash objects are created.

#### How did you do it?
 All SAI objects and attributes are cached in SaiObjectCollection for validation
 although vendors typically do the same as well. This PR introduces a framework
 to control the cache-policy of an allow-list of SmartSwitch objects. The initial
 default policy is conservative caching of a small set of objects identified in
 Issue #26683 as particularly problematic.

 Verified with Virtual DVS testbeds, as well as real Pensando-based SmartSwitch
 testbeds. Orchagent RSS memory measured for all cases during test runs using
 same scale as originating issue - up to 16 ENIs, 125000 CA-PA mappings, and
 100000 outbound routes.

  Impacts only SmartSwitch platforms

#### Virtual Testbed
32 ENIS, 100k outbound routes, and 12k CA-PA mappings
virtual

#### Real Testbed
8 ENIS, 100k outbound routes, and 12k CA-PA mappings
real

### DIFF
--- a/meta/DashMeta.cpp
+++ b/meta/DashMeta.cpp
@@ -1,0 +1,107 @@
+#include "DashMeta.h"
+
+#include "swss/logger.h"
+#include "sai_serialize.h"
+
+extern "C" {
+#include "saiextensions.h"
+}
+
+using namespace saimeta;
+
+namespace
+{
+    /*
+     * Meta-cache policy for DASH SAI entry types only. Non-DASH object
+     * types are unaffected and always behave as FULL.
+     *
+     *   FULL                cache every attribute (upstream default)
+     *   EXISTENCE_REFCOUNT  cache key + OID attrs only (OidRefCounter
+     *                       still works; non-OID payloads not deep-copied)
+     *   NONE                pass-through, skip all meta bookkeeping
+     */
+    enum class DashCacheMode
+    {
+        FULL,
+        EXISTENCE_REFCOUNT,
+        NONE,
+    };
+
+    constexpr DashCacheMode kDashCacheMode = DashCacheMode::EXISTENCE_REFCOUNT;
+
+    /*
+     * Hard-coded allow-list of DASH object types this policy applies to.
+     *
+     * Scoped to the three high-volume DASH entry types that dominate
+     * orchagent's meta-cache footprint at smart-switch scale (issue #26683):
+     *
+     *   - OUTBOUND_CA_TO_PA_ENTRY
+     *   - OUTBOUND_ROUTING_ENTRY
+     *   - INBOUND_ROUTING_ENTRY
+     */
+    constexpr sai_object_type_t kDashTypes[] =
+    {
+        (sai_object_type_t) SAI_OBJECT_TYPE_OUTBOUND_CA_TO_PA_ENTRY,
+        (sai_object_type_t) SAI_OBJECT_TYPE_OUTBOUND_ROUTING_ENTRY,
+        (sai_object_type_t) SAI_OBJECT_TYPE_INBOUND_ROUTING_ENTRY,
+    };
+}
+
+bool saimeta::isDashObjectType(sai_object_type_t ot)
+{
+    // SWSS_LOG_ENTER() is disabled for performance reasons
+    for (auto t : kDashTypes)
+    {
+        if (t == ot)
+        {
+            return true;
+        }
+    }
+    return false;
+}
+
+bool saimeta::bypassValidation(sai_object_type_t ot)
+{
+    // SWSS_LOG_ENTER() is disabled for performance reasons
+    if (!isDashObjectType(ot))
+    {
+        return false;
+    }
+
+    SWSS_LOG_DEBUG("DASH policy active for object type %s (mode=%s)",
+            sai_serialize_object_type(ot).c_str(),
+            dashCacheModeName());
+
+    return kDashCacheMode == DashCacheMode::NONE;
+}
+
+bool saimeta::shouldCacheAttribute(
+        sai_object_type_t ot,
+        const sai_attr_metadata_t& md)
+{
+    // SWSS_LOG_ENTER() is disabled for performance reasons
+    if (!isDashObjectType(ot))
+    {
+        return true;
+    }
+
+    switch (kDashCacheMode)
+    {
+        case DashCacheMode::FULL:               return true;
+        case DashCacheMode::EXISTENCE_REFCOUNT: return md.isoidattribute;
+        case DashCacheMode::NONE:               return false;
+    }
+    return true;
+}
+
+const char* saimeta::dashCacheModeName()
+{
+    // SWSS_LOG_ENTER() is disabled for performance reasons
+    switch (kDashCacheMode)
+    {
+        case DashCacheMode::FULL:               return "FULL";
+        case DashCacheMode::EXISTENCE_REFCOUNT: return "EXISTENCE_REFCOUNT";
+        case DashCacheMode::NONE:               return "NONE";
+    }
+    return "UNKNOWN";
+}
--- a/meta/DashMeta.cpp
+++ b/meta/DashMeta.cpp
@@ -68,10 +68,6 @@ bool saimeta::bypassValidation(sai_object_type_t ot)
         return false;
     }
 
-    SWSS_LOG_DEBUG("DASH policy active for object type %s (mode=%s)",
-            sai_serialize_object_type(ot).c_str(),
-            dashCacheModeName());
-
     return kDashCacheMode == DashCacheMode::NONE;
 }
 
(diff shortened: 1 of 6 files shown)


vivekkarri:
@judsonwilson-nvidia please help review

meta/DashMeta.cpp
```diff
@@ -0,0 +1,107 @@
...
+    if (!isDashObjectType(ot))
+    {
+        return false;
+    }
+
+    SWSS_LOG_DEBUG("DASH policy active for object type %s (mode=%s)",
+            sai_serialize_object_type(ot).c_str(),
+            dashCacheModeName());
+
+    return kDashCacheMode == DashCacheMode::NONE;
```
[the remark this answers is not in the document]
nhegde-microsoft: Changed repeated debug log to one-time notice log at initialization.

vivekkarri:
Please also validate by running existing DASH tests on nvidia-bluefield platform

securely1g (commented):
## Review: [smartswitch][meta] add cache policy for specific DASH entry types

Well-scoped PR that addresses a real memory scaling problem (issue #26683) conservatively. The three-mode cache policy framework is clean, and the default `EXISTENCE_REFCOUNT` mode is a sensible middle ground — preserving OID ref counting while skipping non-OID payload deep copies for the three highest-volume DASH entry types.

### Strengths
- Conservative default — `EXISTENCE_REFCOUNT` keeps ref counting intact
- Solid unit tests — 5 new test cases covering object caching, ref count behavior, missing/duplicate entry handling, and vnet remove protection across all policy modes
- Clean separation in new `DashMeta.cpp/h` files
- Tightly scoped to the three proven problematic types

### Issues / Suggestions

**1. File permissions (minor fix)**
`DashMeta.cpp` and `DashMeta.h` are `100755` (executable). Source files should be `100644`.

**2. `bypassValidation` in entry-specific validators is dead code today**
Bypass checks are added in both the generic paths (`meta_generic_validation_create/remove/set/get`) and the type-specific validators (`meta_sai_validate_inbound_routing_entry`, etc.). Since `EXISTENCE_REFCOUNT` returns `false` from `bypassValidation()`, the entry-specific checks only activate in `NONE` mode. This is fine for future-proofing but worth a comment clarifying intent.

**3. `NONE` mode deserves a warning comment**
`NONE` mode allows creating duplicates, removing non-existent entries without error, and bypasses all attribute validation. If someone flips the `constexpr` without understanding the implications, meta becomes a complete pass-through for these types. A prominent warning near the enum definition would help.

**4. Hardcoded policy — runtime configurability?**
`kDashCacheMode` is `constexpr`, requiring recompilation to change. For a V1 this is fine, but is runtime configurability (e.g. via SAI switch attribute or config DB) on the roadmap?

**5. Subtle behavioral note on `post_set` + `shouldCacheAttribute`**
In `EXISTENCE_REFCOUNT` mode, if a non-OID attribute is set after creation, it won't be cached. A subsequent meta-layer `get` for that attribute would return stale/missing data from cache. The vendor handles the real get so this is likely fine in practice, but worth documenting as a known behavioral change.

meta/Meta.cpp
```diff
@@ -6314,7 +6375,10 @@ void Meta::meta_generic_validation_post_set(
     // only on create we need to increase entry object types members
     // save actual attributes and values to local db
 
-    m_saiObjectCollection.setObjectAttr(meta_key, md, attr);
+    if (saimeta::shouldCacheAttribute(meta_key.objecttype, md))
```
vivekrnv, line 6378: So, this would only be called for type SAI_ATTR_VALUE_TYPE_OBJECT_ID correct and to track refrerence counting. 

Would it work for SAI_ATTR_VALUE_TYPE_OBJECT_LIST?

prsunny:
@nhegde-microsoft , would you check the copilot and other comments?

meta/Meta.cpp
nhegde-microsoft: Good question! I think it will work for object list as well because the `isoidattribute` field is set if there is at least 1 OID type (see [this code](https://github.com/opencomputeproject/SAI/blob/c67f1152309ca08a94de0be8634a733c5cb25c35/meta/parse.pl#L2444)). Its worth pointing out that the 3 object types in this PR are not lists, but you are right that if we added list objects such as ACLs this would be relevant, and it should work as intended, I think.

meta/DashMeta.cpp
```diff
@@ -0,0 +1,118 @@
...
+     * Scoped to the three high-volume DASH entry types that dominate
+     * orchagent's meta-cache footprint at smart-switch scale (issue #26683):
+     *
+     *   - OUTBOUND_CA_TO_PA_ENTRY
+     *   - OUTBOUND_ROUTING_ENTRY
+     *   - INBOUND_ROUTING_ENTRY
+     */
+    constexpr sai_object_type_t kDashTypes[] =
+    {
+        (sai_object_type_t) SAI_OBJECT_TYPE_OUTBOUND_CA_TO_PA_ENTRY,
```
prabhataravind, line 44: @nhegde-microsoft could we add SAI_OBJECT_TYPE_PA_VALIDATION_ENTRY as well?

theasianpianist (commented):
Can we add unit tests for OUTBOUND_CA_TO_PA and OUTBOUND_ROUTING entries as well?

unittest/meta/TestMetaDash.cpp
```diff
@@ -1377,3 +1377,217 @@ TEST(Meta, bulk_dash_outbound_ca_to_pa_entry)
...
+    attr.value.s32 = SAI_INBOUND_ROUTING_ENTRY_ACTION_VXLAN_DECAP;
+
+    EXPECT_EQ(SAI_STATUS_SUCCESS, m.create(&entry, 1, &attr));
+
+    sai_object_meta_key_t meta_key = {
+        .objecttype = kDashEntryType,
+        .objectkey = { .key = { .inbound_routing_entry = entry } }
+    };
+
+    EXPECT_EQ(!dashBypassActive(), m.objectExists(meta_key))
```
theasianpianist, line 1430: Would be good to also add some validation that the attribute caching is behaving as intended. May need to add a new helper method in `Meta.cpp` similar to the existing `objectExists` helper method.

meta/Meta.cpp
```diff
@@ -1869,6 +1877,11 @@ void Meta::meta_generic_validation_post_remove(
 {
     SWSS_LOG_ENTER();
 
+    if (saimeta::bypassValidation(meta_key.objecttype))
+    {
+        return;
```
prabhataravind, line 1882: The early return in post_create/post_remove when skipCollection is true also skips the attribute OID reference counting loop. This means references like DASH_TUNNEL_ID never get refcounts incremented/decremented, so meta could allow removing a tunnel that's still in use.

Suggest to only guard the collection storage calls (createObject, removeObject, setObjectAttr) with !skipCollection, but keep the OID refcount logic unconditional. No memory impact — refcounts are stored on the parent objects which are already tracked.

This can be added as a second PR later. Not a blocker.

prabhataravind: On further checking, it looks like we are ok. This issue might be applicable only for NONE mode, not the default  EXISTENCE_REFCOUNT mode.

prabhataravind:
@nhegde-microsoft could you also add the memory optimization gains with this change to the PR description, leaving out any platform specifics?

theasianpianist (approved):
lgtm. Please plan to expand test coverage in future PRs as per comments

meta/DashMeta.cpp
nhegde-microsoft: Sure, adding it in next drop.

meta/DashMeta.cpp
```diff
@@ -0,0 +1,118 @@
...
+        (sai_object_type_t) SAI_OBJECT_TYPE_OUTBOUND_ROUTING_ENTRY,
+        (sai_object_type_t) SAI_OBJECT_TYPE_INBOUND_ROUTING_ENTRY,
```
saiarcot895, line 46: Minor code style comment: are the casts necessary?

nhegde-microsoft: I did that because the values are defined as type `_sai_object_type_extensions_t`, but probably not necessary.

saiarcot895: Never mind, that seems to be the convention based on other files.

saiarcot895 (approved):

prabhataravind (approved):

vivekkarri (approved):

PR #1919, opened 2026-06-01
Make ZMQ fields optional in context_config.json

inder-nexthop:
### Description of PR
Currently the runtime knob for southbound ZMQ is authoritative at file granularity instead of field granularity. The ContextConfig data model uses one boolean for "a context_config.json was loaded" (`m_loadedFromJson`) and a plain bool `m_zmqEnable` for the value, so a file that loads successfully but omits zmq_enable is indistinguishable from one that explicitly sets `zmq_enable`: false: both land on the bool's false default. A plain bool only has two inhabitants, so it cannot represent the third case, "operator said nothing, defer to the command line."

The ZMQ fields are read with unchecked nlohmann::json::operator[], which throws when a key is absent. That throw is caught by a single outer try/catch that returns the hardcoded default container, so omitting `zmq_enable` (or an endpoint) does not just lose the ZMQ opinion: it discards the entire context definition (guid, DB names, switch list), silently degrading a partial config to no config.

With this change `zmq_enable` becomes truly optional. When the field is omitted from context_config.json, the command-line argument (-z zmq_sync) decides the transport between orchagent and syncd. When the field is present, the file still wins over the command line, in both directions.

Summary:
Fixes # (issue)

### Type of change
- [ ] Bug fix
- [ ] New feature
- [x] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

The `m_loadedFromJson` flag conflates two different questions, "did we load a file?" and "did the operator express a ZMQ opinion?", into one signal gated on file presence. The system needs a three-state model gated on field presence. This matters concretely for gearbox: a gearbox box has two contexts (main ASIC and gbsyncd), but the command line is process-wide and cannot say "ZMQ to the ASIC, Redis to gbsyncd." Only a per-context field can, and that field must be able to mean enabled, disabled, or unset.

##### Work item tracking

#### How did you do it?
- Three-state enum replaces the boolean pair. bool m_zmqEnable + bool m_loadedFromJson in ContextConfig become a single context_config_zmq_state_t with three named inhabitants: CONTEXT_CONFIG_ZMQ_EMPTY (key absent, defer to command line), CONTEXT_CONFIG_ZMQ_ENABLED (lock to ZMQ), CONTEXT_CONFIG_ZMQ_DISABLED (lock to Redis). EMPTY is the zero value so a default-constructed ContextConfig lands on the defer-to-cmdline state. Backed by uint8_t, kept as a C-style typedef enum to match the SAI/sairedis convention used by sai_redis_communication_mode_t.
 
- Parser becomes tolerant of optional fields. lib/ContextConfigContainer.cpp switches from item["zmq_enable"] to item.contains("zmq_enable") (matching the form orchagent's saihelper.cpp already uses) for all three ZMQ fields. Absent zmq_enable leaves the state at EMPTY instead of throwing; absent endpoints leave the constructor defaults instead of throwing. Required structure (guid, db names, switches) stays strict.
 
- syncd/Syncd.cpp (six sites) and lib/RedisRemoteSaiInterface.cpp (four sites) replace the old m_loadedFromJson && m_zmqEnable / m_loadedFromJson && !m_zmqEnable patterns with m_zmqEnable == ENABLED / m_zmqEnable == DISABLED. The deprecated -s reconcile in Syncd correspondingly maps EMPTY and DISABLED to the demote path while ENABLED keeps the file's lock.
 
- Fixed the runtime-promotion bug in RedisRemoteSaiInterface.cpp. Two changes: the SYNC_MODE attr handler now reads m_redisCommunicationMode == SAI_REDIS_COMMUNICATION_MODE_ZMQ_SYNC (resolved channel state) instead of m_zmqEnable == ENABLED (file opinion), and the m_zmqEnable = ENABLED write inside the COMMUNICATION_MODE = ZMQ_SYNC case is deleted. m_zmqEnable is now immutable after parse and represents only the file's opinion, the resolved channel state lives in m_redisCommunicationMode, which is already maintained correctly. With this change, a context promoted to ZMQ via the SAI attr can be downgraded back to Redis via the same attr.

Added unit tests covering the new states.

### DIFF
--- a/lib/ContextConfig.cpp
+++ b/lib/ContextConfig.cpp
@@ -17,8 +17,7 @@ ContextConfig::ContextConfig(
     m_dbCounters(dbCounters),
     m_dbFlex(dbFlex),
     m_dbState(dbState),
-    m_zmqEnable(false),
-    m_loadedFromJson(false),
+    m_zmqEnable(CONTEXT_CONFIG_ZMQ_EMPTY),
     m_zmqEndpoint("tcp://127.0.0.1:5555"),
     m_zmqNtfEndpoint("tcp://127.0.0.1:5556")
 {
@@ -79,16 +78,46 @@ bool ContextConfig::hasConflict(
 
     // state database can be shared
 
+    // When both contexts have CONTEXT_CONFIG_ZMQ_EMPTY, neither one has
+    // expressed an opinion about ZMQ in the config file. Matching
+    // endpoints are usually the shared constructor defaults from an
+    // ordinary non-ZMQ multi-context layout that omits the optional
+    // zmq_endpoint fields, so the loader must not reject the file. The
+    // match is still worth flagging in the log: if a cmdline override
+    // later promotes both contexts to ZMQ they will collide at bind
+    // time, and an operator scanning startup logs gets a hint at why.
+    bool bothZmqEmpty =
+            m_zmqEnable == CONTEXT_CONFIG_ZMQ_EMPTY
+            && ctx->m_zmqEnable == CONTEXT_CONFIG_ZMQ_EMPTY;
+
     if (m_zmqEndpoint == ctx->m_zmqEndpoint)
     {
-        SWSS_LOG_ERROR("zmqEndpoint %s conflict", m_zmqEndpoint.c_str());
-        return true;
+        if (bothZmqEmpty)
+        {
+            SWSS_LOG_WARN("zmqEndpoint %s matches between EMPTY contexts; "
+                    "cmdline ZMQ promotion would collide at bind",
+                    m_zmqEndpoint.c_str());
+        }
+        else
+        {
+            SWSS_LOG_ERROR("zmqEndpoint %s conflict", m_zmqEndpoint.c_str());
+            return true;
+        }
     }
 
     if (m_zmqNtfEndpoint == ctx->m_zmqNtfEndpoint)
     {
-        SWSS_LOG_ERROR("zmqNtfEndpoint %s conflict", m_zmqNtfEndpoint.c_str());
-        return true;
+        if (bothZmqEmpty)
+        {
+            SWSS_LOG_WARN("zmqNtfEndpoint %s matches between EMPTY contexts; "
+                    "cmdline ZMQ promotion would collide at bind",
+                    m_zmqNtfEndpoint.c_str());
+        }
+        else
+        {
+            SWSS_LOG_ERROR("zmqNtfEndpoint %s conflict", m_zmqNtfEndpoint.c_str());
+            return true;
+        }
     }
 
     return false;
--- a/lib/ContextConfig.cpp
+++ b/lib/ContextConfig.cpp
@@ -90,7 +90,17 @@ bool ContextConfig::hasConflict(
             m_zmqEnable == CONTEXT_CONFIG_ZMQ_EMPTY
             && ctx->m_zmqEnable == CONTEXT_CONFIG_ZMQ_EMPTY;
 
-    if (m_zmqEndpoint == ctx->m_zmqEndpoint)
+    // A context locked to Redis (CONTEXT_CONFIG_ZMQ_DISABLED) never binds or
+    // connects ZMQ sockets, so a shared endpoint with such a context can
+    // never collide at bind time. Flagging it as a conflict would make
+    // ContextConfigContainer::insert() throw, which loadFromFile() catches by
+    // discarding the entire parsed context_config.json and falling back to the
+    // default container, so a DISABLED context must not trigger that path.
+    bool eitherZmqDisabled =
+            m_zmqEnable == CONTEXT_CONFIG_ZMQ_DISABLED
+            || ctx->m_zmqEnable == CONTEXT_CONFIG_ZMQ_DISABLED;
+
+    if (m_zmqEndpoint == ctx->m_zmqEndpoint && !eitherZmqDisabled)
     {
         if (bothZmqEmpty)
         {
@@ -105,7 +115,7 @@ bool ContextConfig::hasConflict(
         }
     }
 
-    if (m_zmqNtfEndpoint == ctx->m_zmqNtfEndpoint)
+    if (m_zmqNtfEndpoint == ctx->m_zmqNtfEndpoint && !eitherZmqDisabled)
     {
         if (bothZmqEmpty)
         {
--- a/lib/ContextConfig.cpp
+++ b/lib/ContextConfig.cpp
@@ -78,57 +78,65 @@ bool ContextConfig::hasConflict(
 
     // state database can be shared
 
-    // When both contexts have CONTEXT_CONFIG_ZMQ_EMPTY, neither one has
-    // expressed an opinion about ZMQ in the config file. Matching
-    // endpoints are usually the shared constructor defaults from an
-    // ordinary non-ZMQ multi-context layout that omits the optional
-    // zmq_endpoint fields, so the loader must not reject the file. The
-    // match is still worth flagging in the log: if a cmdline override
-    // later promotes both contexts to ZMQ they will collide at bind
-    // time, and an operator scanning startup logs gets a hint at why.
-    bool bothZmqEmpty =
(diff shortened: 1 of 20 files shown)


inder-nexthop:
CC @deepak-singhal0408, @venkit-nexthop

deepak-singhal0408:
@inder-nexthop, we discussed about this change and the current consensus is, while this change is good for mid/long term, for current 202605 release, for the platforms needing zmq support(and has context_config.json) should keep this enabled/hardcoded via contest_config.json itself.

lets keep this PR open, and raise another one to enable it by default on NH Platforms.

cc: @prsunny @abdosi @prabhataravind

deepak-singhal0408 (dismissed):
Clean architectural improvement. The three-state enum (`EMPTY/ENABLED/DISABLED`) correctly models field-level precedence that the old `bool + m_loadedFromJson` pair could not express. The immutability fix for `m_zmqEnable` (no longer mutated by COMM_MODE attr handler) is a genuine bug fix enabling bidirectional transport toggling. Test coverage is comprehensive — parser, reconciliation, gearbox multi-context, and toggle scenarios all covered.

LGTM

deepak-singhal0408:
@prabhataravind @prsunny , please review as well. if looks ok, we can merge in master.  thanks,

lib/ContextConfig.cpp
```diff
@@ -79,16 +78,46 @@ bool ContextConfig::hasConflict(
...
+        {
+            SWSS_LOG_WARN("zmqEndpoint %s matches between EMPTY contexts; "
+                    "cmdline ZMQ promotion would collide at bind",
+                    m_zmqEndpoint.c_str());
+        }
+        else
+        {
+            SWSS_LOG_ERROR("zmqEndpoint %s conflict", m_zmqEndpoint.c_str());
+            return true;
+        }
```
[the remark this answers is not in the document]
deepak-singhal0408: @inder-nexthop this seems like a valid comment. we need to handle both explicitly zmq "disabled" or zmq config "ommited" case in context_config. please fix this.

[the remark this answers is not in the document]
inder-nexthop: Okay I will take a look @deepak-singhal0408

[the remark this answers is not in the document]
inder-nexthop: @deepak-singhal0408 addressed the review comments and added corresponding unit test coverage for this as well.

lib/ContextConfig.cpp
```diff
@@ -79,16 +78,56 @@ bool ContextConfig::hasConflict(
...
+    // connects ZMQ sockets, so a shared endpoint with such a context can
+    // never collide at bind time. Flagging it as a conflict would make
+    // ContextConfigContainer::insert() throw, which loadFromFile() catches by
+    // discarding the entire parsed context_config.json and falling back to the
+    // default container, so a DISABLED context must not trigger that path.
+    bool eitherZmqDisabled =
+            m_zmqEnable == CONTEXT_CONFIG_ZMQ_DISABLED
+            || ctx->m_zmqEnable == CONTEXT_CONFIG_ZMQ_DISABLED;
+
+    if (m_zmqEndpoint == ctx->m_zmqEndpoint && !eitherZmqDisabled)
```
prabhataravind: The logic seems reasonable, but I think this can be simplified for readability.

Right now the condition mixes endpoint equality with `!eitherZmqDisabled`, and then handles `bothZmqEmpty` inside the block. I think it would be easier to follow if the flow were written as:

- if endpoints differ: no issue
- if either context is `CONTEXT_CONFIG_ZMQ_DISABLED`: no conflict
- if both are `CONTEXT_CONFIG_ZMQ_EMPTY`: warn only
- otherwise: hard conflict

Also, the `zmqEndpoint` and `zmqNtfEndpoint` sections are structurally identical, so this feels like a good candidate for a small utility/helper function to avoid duplicating the same decision logic twice.

unittest/syncd/Makefile.am
```diff
@@ -22,13 +22,12 @@ tests_SOURCES = main.cpp \
...
 				TestMdioIpcServer.cpp \
 				TestPortStateChangeHandler.cpp \
-				TestSelectablesTracker.cpp \
 				TestWorkaround.cpp \
 				TestSyncd.cpp \
 				TestVendorSai.cpp \
 				TestFlowDump.cpp
 
-tests_CXXFLAGS = $(DBGFLAGS) $(AM_CXXFLAGS) $(CXXFLAGS_COMMON)
+tests_CXXFLAGS = $(DBGFLAGS) $(AM_CXXFLAGS) $(CXXFLAGS_COMMON) -fno-access-control
```
prabhataravind, line 30: This seems consistent with other unit test targets in the repo, since -fno-access-control is already used elsewhere. Still, because it disables private/protected checks for the whole unittest/syncd target, I wonder if there’s a narrower option here, such as testing via observable behavior or a more targeted test hook.

syncd/Syncd.cpp
```diff
@@ -174,20 +178,17 @@ Syncd::Syncd(
 
     bool isDpuSwitch = switchType == "dpu";
 
-    if (m_contextConfig->m_zmqEnable && isDpuSwitch && !isVirtualSwitch)
+    if (zmqActive && isDpuSwitch && !isVirtualSwitch)
```
prabhataravind, line 181: This will conflict with #1819

deepak-singhal0408: Agreed — this will conflict with #1819, and that's intentional. We need this PR (#1919) to merge first. #1819 needs to be restructured to use the resolved `m_redisCommunicationMode` (post-reconciliation) as its guard condition instead of the raw `m_zmqEnable` bool. The three-state enum from this PR is what makes that possible for platforms where `zmq_enable` is omitted from context_config.json.

I'll post a separate detailed comment on #1819 with the restructuring guidance.

@inder-nexthop — could you address prabhat's comments on this PR (hasConflict simplification + Makefile question) so we can get this merged?

unittest/syncd/Makefile.am
inder-nexthop: It's a fair point but I'd lean toward keeping it as-is: it's already the established pattern for the other unittest targets in this repo, and the tri-state tests need to poke m_zmqEnable directly (the test at TestContextConfig.cpp:18 sets it). A narrower hook would mean adding test-only accessors to production headers, unless there is another approach you had in mind for this?

lib/ContextConfig.cpp
inder-nexthop: Done.

syncd/Syncd.cpp
inder-nexthop: Addressed above comments, will rebase the PR #1819 once this has merged.

deepak-singhal0408:
@inder-nexthop fix the DCO sign-off? and see if CI retriggers. if not, we may need to close/reopen PR.

deepak-singhal0408:
## CI Failure Analysis — VS Test Stage

The P4RT/VS test failures on this PR are **not PR-induced**. They are caused by a systemic issue affecting all sonic-sairedis PR builds since ~Jun 22.

### Root Cause

The `_INCLUDE_DOCKER` restructuring of `docker-sonic-vs` (merged to sonic-buildimage master Jun 17) renamed the supervisor program from `[program:start.sh]` to `[program:start]`. The DVS `check_services_ready()` in `sonic-swss/tests/conftest.py:588` still looks up `"start.sh"`, which no longer exists — causing the readiness check to time out after 60 seconds and all DVS-based tests to fail with `AttributeError: NoneType has no attribute get_logs`.

### Fix

[sonic-swss PR #4696](https://github.com/sonic-net/sonic-swss/pull/4696) addresses this by accepting either `"start"` or `"start.sh"`. Once that merges, the sairedis CI pipeline will pick up the updated sonic-swss and DVS tests should pass again.

### This PR

Unit tests and lint checks pass. The only failure is the VS test stage, which is blocked on the fix above. A CI retry after #4696 merges should go green.

deepak-singhal0408 (approved):

prabhataravind (approved):

PR #1920, opened 2026-06-02
[vslib] Add support for SPAN mirror sessions

taran9999:
### Description of PR
This PR implements support for SPAN mirror sessions. Defines translation layer with vpp api, provides support for creating and removing mirror session entries from CLI. Addresses Phase I of this HLD: https://github.com/sonic-net/sonic-platform-vpp/blob/master/docs/HLD/vpp-mirror.md

Summary:
Addresses issue sonic-net/sonic-buildimage#25786.

### Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

To enable span data plane testing on KVM testbed. Integration with sonic is needed, and is addressed in this PR.

#### How did you do it?
Changes are located in `vslib`, `vslib/vpp`, and `vslib/vpp/vppxlate`.
- Defines `vpp_span_enable_disable` in `SaiVppXlate.c/h` as a wrapper to enable/disable a port for span mirror session through vpp.
- Defines structures and constants in `SwitchVpp.h` to track mirror session entries.
- New file `SwitchVppMirror.cpp` to handle creating/removing mirror session entries from CLI.
- Modify `setPort`in `SwitchVpp.cpp`to track ports that are bound to mirror sessions.

Passed the following sonic-mgmt tests on t0_azure.
- span/conftest.py
- span/test_port_mirroring.py


lolyu (commented):
the build is failing due to your change

taran9999:
Looks like maybe just a syntax error from looking at the build log. Let me investigate

vslib/vpp/vppxlate/SaiVppXlate.h
```diff
@@ -354,6 +354,12 @@ typedef enum {
     extern int vpp_sidlist_del(vpp_ip_addr_t *bsid);
     extern int vpp_sr_steer_add_del(vpp_sr_steer_t *sr_steer, bool is_del);
     extern int vpp_sr_set_encap_source(vpp_ip_addr_t *encap_src);
+
+/* SPAN (port mirroring) */
+extern int vpp_span_enable_disable(uint32_t sw_if_index_from,
```
yue-fred-gao: please fix indentation

vslib/vpp/SwitchVpp.cpp
```diff
@@ -1414,6 +1432,64 @@ sai_status_t SwitchVpp::setPort(
...
+
+                    // 1 = RX, 2 = TX, 3 = both
+                    uint8_t state = (pmb.rx ? 1 : 0) | (pmb.tx ? 2 : 0);
+
+                    SWSS_LOG_INFO("Port mirror binding info for port %s: session_oid=%s, rx=%d, tx=%d, dst_sw_if_idx=%u", sid.c_str(), sai_serialize_object_id(session_oid).c_str(), pmb.rx, pmb.tx, pmb.dst_sw_if_idx);
+                    SWSS_LOG_INFO("VPP span enable: src_sw_if=%u, src_hwif_name=%s, dst_sw_if=%u, state=%u", src_sw_if, src_hwif.c_str(), it->second.sw_if_index, state);
+                    vpp_span_enable_disable(src_sw_if, it->second.sw_if_index, state, false);
+                }
+            } else {
+                auto pmb_it = m_port_mirror_bindings.find(portId);
```
yue-fred-gao: should we remove span only for the corresponding direction? If a port has both ingress and egress mirror and mirror of one of the directions got removed, we should keep the other direction

vslib/vpp/SwitchVpp.cpp
```diff
@@ -1414,6 +1432,64 @@ sai_status_t SwitchVpp::setPort(
 
     auto sid = sai_serialize_object_id(portId);
 
+    if(attr->id == SAI_PORT_ATTR_INGRESS_MIRROR_SESSION ||
```
yue-fred-gao: Can we move this to a function in SwitchVppMirror.cpp?

vslib/vpp/SwitchVpp.cpp
```diff
@@ -1414,6 +1432,64 @@ sai_status_t SwitchVpp::setPort(
...
+    {
+        std::string src_hwif;
+        if(!vpp_get_hwif_name(portId, 0, src_hwif)) {
+            SWSS_LOG_WARN("Failed to get hwif name for port %s; skipping VPP SPAN programming", sid.c_str());
+        } else {
+            uint32_t src_sw_if = get_sw_if_idx(src_hwif.c_str());
+
+            if(attr->value.objlist.count > 0) {
+                // bind
+                sai_object_id_t session_oid = attr->value.objlist.list[0];
```
yue-fred-gao: if it is a objlist, it means the same port can have multiple egress or ingress mirror sessions. Does vpp support that? If not, we can pick the first one but log a warning for the others.

taran9999: I believe the ports can be mapped to only one destination interface in vpp, I assume this means one session per port. I can add a warn log for the others.

vslib/vpp/SwitchVpp.cpp
yue-fred-gao: And move the call of the function to updatePort?

vslib/vpp/SwitchVppMirror.cpp
```diff
@@ -0,0 +1,148 @@
...
+            }
+        } else {
+            auto pmb_it = m_port_mirror_bindings.find(portId);
+            if(pmb_it == m_port_mirror_bindings.end()) {
+                SWSS_LOG_WARN("No existing mirror session binding found for port %s; skipping VPP SPAN unprogramming", sid.c_str());
+            } else {
+                m_port_mirror_bindings.erase(pmb_it);
+            }
+
+            // unbind: state = 0
```
lolyu: **Blocking: unbind tears down both directions.** The unbind path (`objlist.count == 0`) always erases the whole `PortMirrorBinding` and calls `vpp_span_enable_disable(src_sw_if, ~0, 0, false)`, killing both RX and TX. As yue-fred-gao already flagged, if a port has both ingress+egress mirror and only one direction is removed, the other should survive.

The dispatch also doesn't distinguish which direction is being cleared. You need to clear only `pmb.rx` or `pmb.tx` based on `attr->id`, recompute `state`, and only fully disable/erase when both are false. Something like:

```cpp
(attr->id == SAI_PORT_ATTR_INGRESS_MIRROR_SESSION) ? pmb.rx = false : pmb.tx = false;
uint8_t state = (pmb.rx ? 1 : 0) | (pmb.tx ? 2 : 0);
if (state == 0) {
    m_port_mirror_bindings.erase(pmb_it);
    vpp_span_enable_disable(src_sw_if, ~0, 0, false);
} else {
    vpp_span_enable_disable(src_sw_if, pmb.dst_sw_if_idx, state, false);
}
```

vslib/vpp/SwitchVppMirror.cpp
```diff
@@ -0,0 +1,148 @@
...
+            SWSS_LOG_INFO("VPP span disable: src_sw_if=%u, src_hwif_name=%s", src_sw_if, src_hwif.c_str());
+            vpp_span_enable_disable(src_sw_if, ~0, 0, false);
+        }
+    }
+    return SAI_STATUS_SUCCESS;
+}
```
lolyu: **Nit: missing newline at end of file** (`\ No newline at end of file`). Please add a trailing newline.

vslib/vpp/SwitchVppMirror.cpp
```diff
@@ -0,0 +1,148 @@
...
+    auto sid = sai_serialize_object_id(portId);
+
+    std::string src_hwif;
+    if(!vpp_get_hwif_name(portId, 0, src_hwif)) {
+        SWSS_LOG_WARN("Failed to get hwif name for port %s; skipping VPP SPAN programming", sid.c_str());
+        return SAI_STATUS_FAILURE;
+    } else {
+        uint32_t src_sw_if = get_sw_if_idx(src_hwif.c_str());
+
+        if(attr->value.objlist.count > 0) {
```
lolyu, line 145: `get_sw_if_idx` returns `int` and can yield `~0` when the hwif isn't found. Here you assign straight into `uint32_t src_sw_if` with no not-found check (unlike `createMirrorSession`, which checks `sw_idx < 0`). Add a guard so a lookup miss doesn't silently program `~0` as the source.

vslib/vpp/SwitchVppMirror.cpp
```diff
@@ -0,0 +1,148 @@
...
+            // bind
+            sai_object_id_t session_oid = attr->value.objlist.list[0];
+            auto it = m_mirror_sessions.find(session_oid);
+            if(it == m_mirror_sessions.end()) {
+                SWSS_LOG_WARN("Mirror session %s not found for port %s; skipping VPP SPAN programming",
+                    sai_serialize_object_id(session_oid).c_str(), sid.c_str());
+                return SAI_STATUS_FAILURE;
+            } else {
+                auto pmb_it = m_port_mirror_bindings.find(portId);
```
lolyu, line 157: You take `list[0]` unconditionally. You agreed in the earlier thread to add a warn log for the additional entries when `count > 1` (VPP maps a port to a single destination interface). That log isn't in this diff yet — please add it.

vslib/vpp/SwitchVppMirror.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+sai_status_t SwitchVpp::bindMirrorPort(
+    _In_ sai_object_id_t portId,
+    _In_ const sai_attribute_t* attr)
+{
+    SWSS_LOG_ENTER();
+
+    auto sid = sai_serialize_object_id(portId);
+
+    std::string src_hwif;
```
lolyu, line 134: This returns `SAI_STATUS_FAILURE` on "hwif not found" but only logs `SWSS_LOG_WARN` with wording "skipping VPP SPAN programming". Since `setPort` propagates this return, it fails the entire port attr-set rather than skipping. Either return `SAI_STATUS_SUCCESS` (if skipping is truly intended) or raise the log to error — the current warn-but-hard-fail is inconsistent.

vslib/vpp/vppxlate/SaiVppXlate.c
```diff
@@ -5054,6 +5110,33 @@ int vpp_sr_set_encap_source(vpp_ip_addr_t *encap_src)
...
+
+    VPP_LOCK();
+
+    __plugin_msg_base = span_msg_id_base;
+
+    M (SW_INTERFACE_SPAN_ENABLE_DISABLE, mp);
+
+    mp->sw_if_index_from = htonl(sw_if_index_from);
+    mp->sw_if_index_to = htonl(sw_if_index_to);
+    mp->state = htonl((uint32_t)state);
```
lolyu: **Blocking: wrong wire encoding of `state`.** `state` is a `u8` field in `vl_api_sw_interface_span_enable_disable_t`, not a u32. `htonl((uint32_t)state)` writes 4 bytes into a 1-byte field (clobbering adjacent struct memory) and sends the wrong value on little-endian hosts. It should be:

```c
mp->state = state;
```

For reference, `sw_if_index_from/to` are genuinely u32 in the .api so their `htonl` is correct, and `is_l2` is a u8/bool assigned directly — both fine.

vslib/vpp/SwitchVpp.cpp
```diff
@@ -1932,6 +1950,18 @@ sai_status_t SwitchVpp::setPort(
 
     auto sid = sai_serialize_object_id(portId);
 
+    if(attr->id == SAI_PORT_ATTR_INGRESS_MIRROR_SESSION ||
```
lolyu: yue-fred-gao asked (twice, still unresolved) to move this mirror-binding dispatch into a function in `SwitchVppMirror.cpp` and call it from `updatePort` rather than inline in `setPort`. The function body already lives in `SwitchVppMirror.cpp`, but the wiring is still through `setPort`. Please confirm with the reviewer whether the `updatePort` relocation is still required and resolve/reply to that thread either way.

vslib/vpp/SwitchVpp.h
```diff
@@ -1280,5 +1280,43 @@ namespace saivs
...
+                sai_object_id_t session_oid;
+                bool rx;
+                bool tx;
+                uint32_t dst_sw_if_idx;
+            };
+
+            // port id to PortMirrorBinding
+            std::map<sai_object_id_t, PortMirrorBinding> m_port_mirror_bindings;
+
+        protected:
```
lolyu, line 1309: Indentation in this `protected:` block is over-indented (16 spaces on the method decls vs. the surrounding 12-space style). Please align with the rest of the class — same class of nit yue-fred-gao raised on `SaiVppXlate.h`.

vslib/vpp/SwitchVpp.h
```diff
@@ -1280,5 +1280,43 @@ namespace saivs
...
                 _Inout_ sai_s32_list_t *enum_values_capability) override;
 
+        private: // VPP mirror
+            constexpr static const int m_maxMirrorSessions = 10;
+            uint32_t m_mirror_session_count = 0;
+            uint16_t m_next_erspan_session_id = 1;
+
+            struct MirrorSessionInfo {
+                uint32_t sw_if_index;
+                bool is_erspan;
```
lolyu: ERSPAN fields (`is_erspan`, `src_ip`, `dst_ip`, `session_id`, `m_next_erspan_session_id`) are defined but unused since only `SAI_MIRROR_SESSION_TYPE_LOCAL` is handled this phase. Fine to keep for Phase II, but a `// Phase II (ERSPAN)` comment would clarify intent. Also `m_maxMirrorSessions = 10` here duplicates the `SAI_SWITCH_ATTR_MAX_MIRROR_SESSION = 10` set in `SwitchStateBase.cpp` — worth a single shared constant.

lolyu (changes requested):
Nice feature work — the VPP SPAN plumbing follows the existing sflow/bfd/sr pattern cleanly and it passes the SPAN sonic-mgmt tests. A few correctness issues to resolve before merge (inline):

**Blocking**
- Unbind tears down both RX and TX directions; single-direction removal should preserve the other (also flagged earlier by @yue-fred-gao).
- `mp->state = htonl((uint32_t)state)` byte-swaps a `u8` API field — wrong wire value + adjacent-memory clobber. Should be a plain `mp->state = state;`.

**Should-fix**
- Missing warn log for extra objlist entries (`count > 1`) — you agreed to add this earlier.
- Unresolved reviewer thread on moving the dispatch to `updatePort` — please resolve/reply.
- No not-found guard on `get_sw_if_idx` in `bindMirrorPort`; warn-but-hard-fail return is inconsistent.

**Nits:** trailing newline, indentation in `SwitchVpp.h`/`SaiVppXlate.h`, `%d` vs `%u` on the unsigned session count, duplicated `10` constant, unused ERSPAN fields (Phase II — a comment would help).

Structure is good and it's close.

vslib/vpp/vppxlate/SaiVppXlate.c
```diff
@@ -1648,6 +1687,8 @@ static void vpp_ext_vpe_init(void)
 
     foreach_vpe_ext_api_reply_msg;
 #undef _
+
+
```
dypet: Remove extra whitespace

vslib/vpp/SwitchVppMirror.cpp
```diff
@@ -0,0 +1,148 @@
...
+            auto pmb_it = m_port_mirror_bindings.find(portId);
+            if(pmb_it == m_port_mirror_bindings.end()) {
+                SWSS_LOG_WARN("No existing mirror session binding found for port %s; skipping VPP SPAN unprogramming", sid.c_str());
+            } else {
+                m_port_mirror_bindings.erase(pmb_it);
+            }
+
+            // unbind: state = 0
+            SWSS_LOG_INFO("VPP span disable: src_sw_if=%u, src_hwif_name=%s", src_sw_if, src_hwif.c_str());
+            vpp_span_enable_disable(src_sw_if, ~0, 0, false);
```
dypet: Both calls to vpp_span_enable_disable in this function discard the return code, so bindMirrorPort returns SAI_STATUS_SUCCESS even if VPP rejected the request. Consider checking/propagating, or add a comment that programming is best-effort.

vslib/vpp/SwitchVppMirror.cpp
```diff
@@ -0,0 +1,148 @@
...
+        info.is_erspan = false;
+    } else {
+        SWSS_LOG_ERROR("Unsupported mirror session type %d", mirror_type);
+        return SAI_STATUS_FAILURE;
+    }
+
+    CHECK_STATUS(create_internal(SAI_OBJECT_TYPE_MIRROR_SESSION, sid, switch_id, attr_count, attr_list));
+
+    m_mirror_sessions[object_id] = info;
+    m_mirror_session_count++;
```
dypet, line 60: m_mirror_session_count is incremented without checking it against m_maxMirrorSessions, but getObjectTypeAvailability returns m_maxMirrorSessions - m_mirror_session_count on unsigned operands. If the count ever exceeds 10 the subtraction underflows to a huge value. Enforce the cap here with SAI_STATUS_INSUFFICIENT_RESOURCES or clamp the availability computation. Consider also checking the bounds when doing m_mirror_session_count-- in removeMirrorSession

vslib/vpp/SwitchVppMirror.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+    auto it = m_mirror_sessions.find(object_id);
+    if(it == m_mirror_sessions.end()) {
+        SWSS_LOG_ERROR("Mirror session %s not found", sai_serialize_object_id(object_id).c_str());
+        return SAI_STATUS_ITEM_NOT_FOUND;
+    }
+
+    CHECK_STATUS(remove_internal(SAI_OBJECT_TYPE_MIRROR_SESSION, sai_serialize_object_id(object_id)));
+
+    m_mirror_sessions.erase(it);
```
dypet, line 118: removeMirrorSession erases the session but leaves any m_port_mirror_bindings entries that reference it. If a port is still bound when the session is removed (out-of-order teardown), the binding is orphaned, VPP SPAN is never disabled, and a later session re-created with the same OID could re-bind to a stale dst_sw_if_idx. Please clean up bindings still referencing this OID before erasing.

dypet (changes requested):

vslib/vpp/vppxlate/SaiVppXlate.c
taran9999: Hi Longxiang, thanks for the review and comments.

I am a bit unsure of this. On the vpp side, `vl_api_sw_interface_span_enable_disable_t_handler` (found in `vpp/src/vnet/span/span_api.c`) calls `ntohl(mp->state)`. I believe `ntohl` expects a `u32`, so outside of the case of an implicit conversion, `mp->state` should be a `u32`. Also, I believe calling `ntohl` with a `u8` parameter would serve no purpose, as `u8` should not be affected by endianness.

I checked the definition of the message struct in `span.api`. The `state` parameter seems to be a generated version of the `span_state` enum found above, which does not explicitly show the type. It seems the API parser will set enum type to `u32` by default (found here: https://github.com/FDio/vpp/blob/246c72f5743e9910703f2de1b067d9611447d68f/src/tools/vppapigen/vppapigen.py#L360).

`mp->state = state` is what I had before, but it seemed to result in the `state` field being discarded (tracing the call with vppctl would show the `state` field being left empty), which is why I changed it to this

dypet (dismissed):

vslib/vpp/vppxlate/SaiVppXlate.c
lolyu: You're right, and my original comment was wrong — apologies. I confirmed against `src/vnet/span/span.api`: the field is `vl_api_span_state_t state`, i.e. the `span_state` enum. vppapigen emits enums as `u32` by default, so the generated C struct field is a 4-byte `u32`, not a `u8`, and the handler does `ntohl(mp->state)`. That means `htonl((uint32_t)state)` is the correct encoding, and plain `mp->state = state` (host-endian) is what breaks on little-endian — matching what you observed with the empty `state` in the vppctl trace.

One optional readability nit, not blocking: since the wire field is genuinely a `u32`, consider making `vpp_span_enable_disable`'s `state` parameter a `uint32_t` (and dropping the `(uint32_t)` cast) so the type matches the API field and there's no implicit narrow-then-widen. Purely cosmetic though.

Resolving this thread — good to go on the encoding.

lolyu (approved):
LGTM

PR #1923, opened 2026-06-02
[smartswitch][meta] dash cache policy tests

nhegde-microsoft:
#### Description of PR
 Followon from previous PR #1918 

#### Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [X] Test improvement

 Addressed pending comments from previous PR. Changes:
 1. Added comments about behaviour change due to new cache modes
 2. Added missing PA_VALIDATION_ENTRY to list of object types subject to cache policy
 3. Unit tests to confirm behaviour change
 4. Unit test to specifically check that attributes are cached (or not) as expected

Ran unit tests and virtual testbed tests and confirmed correctness.

 Impacts only SmartSwitch platforms

### DIFF
--- a/meta/Meta.h
+++ b/meta/Meta.h
@@ -429,6 +429,10 @@ namespace saimeta
             bool objectExists(
                     _In_ const sai_object_meta_key_t& mk) const;
 
+            bool objectHasAttribute(
+                    _In_ const sai_object_meta_key_t& mk,
+                    _In_ sai_attr_id_t attr_id) const;
+
         private: // port helpers
 
             sai_status_t meta_port_remove_validation(
--- a/unittest/meta/TestMetaDash.cpp
+++ b/unittest/meta/TestMetaDash.cpp
@@ -1379,215 +1379,481 @@ TEST(Meta, bulk_dash_outbound_ca_to_pa_entry)
 }
 
 
-// ---------------------------------------------------------------------------
-// DashMeta behavior tests.
-// ---------------------------------------------------------------------------
-
 #include "DashMeta.h"
+#include "sai_serialize.h"
+
+// ---------------------------------------------------------------------------
+// Typed DASH-entry tests.
+//
+// One traits struct per SAI DASH entry type. The same suite of TYPED_TESTs
+// then runs against every entry listed in DashEntryTypes below. To cover a
+// new DASH entry type, add a Traits struct exposing the static members used
+// here and append it to DashEntryTypes.
+// ---------------------------------------------------------------------------
 
 namespace
 {
-    constexpr sai_object_type_t kDashEntryType =
-        (sai_object_type_t)SAI_OBJECT_TYPE_INBOUND_ROUTING_ENTRY;
-
-    bool dashBypassActive()
+    struct InboundRoutingTraits
     {
-        // SWSS_LOG_ENTER() is disabled for performance reasons
-        return bypassValidation(kDashEntryType);
-    }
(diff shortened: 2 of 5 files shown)


meta/Meta.h
```diff
@@ -429,6 +429,10 @@ namespace saimeta
             bool objectExists(
                     _In_ const sai_object_meta_key_t& mk) const;
 
+            bool objectHasAttribute(
+                    _In_ const sai_object_meta_key_t& mk,
+                    _In_ sai_attr_id_t attr_id) const;
```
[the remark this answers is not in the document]
nhegde-microsoft: Done

unittest/meta/TestMetaDash.cpp
```diff
@@ -1379,215 +1379,481 @@ TEST(Meta, bulk_dash_outbound_ca_to_pa_entry)
...
+        static std::vector<sai_attribute_t> createAttrs(const Ctx&)
+        {
+            SWSS_LOG_ENTER();
+            sai_attribute_t a{};
+            a.id = SAI_PA_VALIDATION_ENTRY_ATTR_ACTION;
+            a.value.s32 = SAI_PA_VALIDATION_ENTRY_ACTION_PERMIT;
+            return { a };
+        }
+
+        // PA validation entry has no OID-typed attribute; OID-attr-based
```
theasianpianist, line 1521: Since PA validation entries don't reference other OIDs, it doesn't seem like they will benefit from the new caching policy. Do we need to include them in the caching scheme?

nhegde-microsoft: This is correct that it does not reference other OIDs and the vnet and switch ID it uses are in the key which get counted any way. This was [requested to be added](https://github.com/sonic-net/sonic-sairedis/pull/1918#discussion_r3342945997) in the previous PR by @prabhataravind, I assume for future proofing. It will probably provide reasonable savings at high scale if the policy is changed to `NONE` (no caching for listed object types).

prabhataravind (approved):

unittest/meta/TestMetaDash.cpp
prabhataravind: My thinkinig is that including PA_VALIDATION_ENTRY is not just future-proofing. If I understand correctly, under EXISTENCE_REFCOUNT, all non-OID attribute deep-copies are skipped from m_saiObjectCollection storage - this means the ACTION and IP_ADDR_FAMILY attribute maps are never stored. At millions of entries, avoiding those deep-copied attribute maps is a meaningful memory saving (possibly a couple hundred MBs?).

theasianpianist: Sorry you're right, I had things flipped around in my head.

theasianpianist (approved):

saiarcot895 (approved):

rookie-who:
@nhegde-microsoft — This PR is requested for the **202605 branch** but is missing test evidence. To move forward with cherry-pick approval, please:

1. **Test on the 202605 branch** and add results (image version + test output) to the PR description
2. **Check the `202605` box** in the "Tested Branches" section of the PR description
3. **Add the "Tested for 202605 branch" label** once testing is complete

Also, to help us prioritize correctly, could you clarify: **Is this PR a bug fix / regression fix, a new feature, a platform config change, or a dependency/infra update?** Understanding the nature of the change helps us triage the cherry-pick queue.

Thank you!

*Review by @vaibhavhd via automated tooling*

PR #1928, opened 2026-06-05
feat: handle LAG egress disable updates

cyw233:
### Description of PR

Summary:
Add VPP vslib handling for `SAI_LAG_MEMBER_ATTR_EGRESS_DISABLE` so LAG members can be dynamically detached from, or re-added to, the VPP bond when egress is disabled or enabled.

This also defers VPP bond member creation when a LAG member is created with egress disabled. LCP/tap creation remains delayed until the first enabled member is added, which avoids creating an empty bond tap with a stale generated MAC.

Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [x] Test improvement

VPP vslib already creates and removes LAG members in the VPP bond, but it did not handle updates to `SAI_LAG_MEMBER_ATTR_EGRESS_DISABLE`. As a result, a member whose egress state changed after creation would not be reflected in VPP bond membership.

##### Work item tracking

#### How did you do it?

- Route `SAI_OBJECT_TYPE_LAG_MEMBER` set operations through a new `setLagMember` handler.
- Handle `SAI_LAG_MEMBER_ATTR_EGRESS_DISABLE` transitions by removing the member from the VPP bond when egress is disabled and adding it back when egress is enabled.
- Treat missing `EGRESS_DISABLE` state as the SAI default value, `false`, when deciding whether a VPP bond membership update is needed.
- Skip VPP bond member creation when a LAG member is initially created with egress disabled.
- Extract LCP/tap setup into `vpp_ensure_lag_lcp()` so LCP creation is still performed when the first enabled member is added.

Added VPP-enabled vslib unit coverage for `getLagMemberEgressDisableAction()` to verify the expected no-op, detach, and add decisions for current and requested egress-disable states.

Also ran the `pc/test_lag_member_forwarding.py` sonic-mgmt test to validate this feature (PR https://github.com/sonic-net/sonic-mgmt/pull/25135 to enable it):

image

This change is specific to the VPP vslib path.

### Documentation

HLD: https://github.com/sonic-net/sonic-platform-vpp/pull/250

### DIFF
--- a/vslib/vpp/SwitchVpp.h
+++ b/vslib/vpp/SwitchVpp.h
@@ -286,6 +286,21 @@ namespace saivs
             virtual sai_status_t setLag(
                     _In_ sai_object_id_t lagId,
                     _In_ const sai_attribute_t* attr);
+            virtual sai_status_t setLagMember(
+                    _In_ sai_object_id_t lagMemberId,
+                    _In_ const sai_attribute_t* attr);
+
+            enum class LagMemberEgressDisableAction
+            {
+                NONE,
+                DETACH,
+                ADD,
+            };
+
+            static LagMemberEgressDisableAction getLagMemberEgressDisableAction(
+                    _In_ bool requested_egress_disable,
+                    _In_ bool current_attr_found,
+                    _In_ bool current_egress_disable);
 
             sai_status_t vpp_create_lag(
                     _In_ sai_object_id_t lag_id,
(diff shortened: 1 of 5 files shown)


vslib/vpp/SwitchVpp.h
```diff
@@ -307,6 +322,11 @@ namespace saivs
                     _In_ sai_object_id_t lag_member_oid);
 	    sai_status_t vpp_remove_lag_member(
                     _In_ sai_object_id_t lag_member_oid);
+            sai_status_t vpp_ensure_lag_lcp(
```
lolyu, line 325: wrong indentation

lolyu (commented):
if a member is created with egress disabled - that is never added to VPP bond. when it is later removed, what will happen with your current implementation?

PR #1934, opened 2026-06-08
[vpp][fdb] Add fdb support

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

The VPP SAI virtual switch (saivpp) currently has no FDB support — MAC learning, aging, and move events are not propagated to SONiC's FDB_TABLE. This means L2 forwarding state is invisible to the SONiC control plane on VPP-based platforms.

This PR adds FDB support to saivpp using VPP's push-based `WANT_L2_MACS_EVENTS2` API, which delivers MAC learn/age/move notifications within ~10 ms of the hardware event directly to the SAI layer without requiring periodic polling.

##### Work item tracking
- Microsoft ADO **(number only)**: 38333755

#### How did you do it?

Please refer to HLD: https://github.com/sonic-net/SONiC/pull/2368

Ran the existing sonic-mgmt `tests/fdb/test_fdb.py` suite on a VPP KVM t0 testbed (vms-kvm-vpp-t0) in a 10-iteration soak:
* 40/40 test cases passed, 0 errors, 0 failures.
Tests exercise: FDB learning via Ethernet frames, ARP requests, and ARP replies; self-MAC non-learning; and FDB aging/cleanup.

VPP platform only (`saivpp` / `docker-syncd-vpp`). No changes to other SAI implementations.

Requires VPP with the l2 plugin supporting `WANT_L2_MACS_EVENTS2` (available since VPP 22.06).

### DIFF
--- a/vslib/VirtualSwitchSaiInterfaceFdb.cpp
+++ b/vslib/VirtualSwitchSaiInterfaceFdb.cpp
@@ -1,8 +1,5 @@
 #include "VirtualSwitchSaiInterface.h"
 #include "SwitchStateBase.h"
-#ifdef USE_VPP
-#include "SwitchVpp.h"
-#endif
 
 #include "meta/sai_serialize.h"
 
@@ -325,4 +322,10 @@ void VirtualSwitchSaiInterface::ageFdbs()
     {
         it.second->processFdbEntriesForAging();
     }
+}
+
+void VirtualSwitchSaiInterface::setFdbAgingWakeFunction(std::function<void()> fn)
+{
+    for (auto &kv : m_switchStateMap)
+        kv.second->setFdbAgingWakeFunction(fn);
 }
--- a/vslib/VirtualSwitchSaiInterfaceFdb.cpp
+++ b/vslib/VirtualSwitchSaiInterfaceFdb.cpp
@@ -1,6 +1,8 @@
 #include "VirtualSwitchSaiInterface.h"
 #include "SwitchStateBase.h"
+#ifdef USE_VPP
 #include "vpp/SwitchVpp.h"
+#endif
 
 #include "meta/sai_serialize.h"
 
--- a/vslib/VirtualSwitchSaiInterfaceFdb.cpp
+++ b/vslib/VirtualSwitchSaiInterfaceFdb.cpp
@@ -327,14 +327,18 @@ void VirtualSwitchSaiInterface::ageFdbs()
 
 void VirtualSwitchSaiInterface::initFdbEventHandling(std::function<void()> fn)
 {
+    SWSS_LOG_ENTER();
+
     for (auto &kv : m_switchStateMap)
     {
         kv.second->initFdbEventHandling(fn);
-    }       
+    }
 }
 
 void VirtualSwitchSaiInterface::deinitFdbEventHandling()
 {
+    SWSS_LOG_ENTER();
+
     for (auto &kv : m_switchStateMap)
     {
         kv.second->deinitFdbEventHandling();
--- a/vslib/VirtualSwitchSaiInterfaceFdb.cpp
+++ b/vslib/VirtualSwitchSaiInterfaceFdb.cpp
@@ -329,6 +329,8 @@ void VirtualSwitchSaiInterface::initFdbEventHandling(std::function<void()> fn)
 {
     SWSS_LOG_ENTER();
 
+    m_fdbEventFn = fn;
+
     for (auto &kv : m_switchStateMap)
     {
         kv.second->initFdbEventHandling(fn);
--- a/vslib/VirtualSwitchSaiInterfaceFdb.cpp
+++ b/vslib/VirtualSwitchSaiInterfaceFdb.cpp
@@ -1,5 +1,6 @@
 #include "VirtualSwitchSaiInterface.h"
 #include "SwitchStateBase.h"
+#include "vpp/SwitchVpp.h"
 
 #include "meta/sai_serialize.h"
 
@@ -324,8 +325,18 @@ void VirtualSwitchSaiInterface::ageFdbs()
     }
 }
 
-void VirtualSwitchSaiInterface::setFdbAgingWakeFunction(std::function<void()> fn)
+void VirtualSwitchSaiInterface::initFdbEventHandling(std::function<void()> fn)
 {
     for (auto &kv : m_switchStateMap)
-        kv.second->setFdbAgingWakeFunction(fn);
+    {
+        kv.second->initFdbEventHandling(fn);
+    }       
+}
+
+void VirtualSwitchSaiInterface::deinitFdbEventHandling()
+{
+    for (auto &kv : m_switchStateMap)
+    {
+        kv.second->deinitFdbEventHandling();
+    }
 }
--- a/vslib/vpp/SwitchVppFdb.cpp
+++ b/vslib/vpp/SwitchVppFdb.cpp
@@ -1520,6 +1520,12 @@ void SwitchVpp::swif_bdid_untrack(const char *hwif_name)
     if (swif != (uint32_t)~0u)
     {
         m_swif_to_bdid.erase(swif);
+
+        // Invalidate the sw_if_index -> SAI port OID memoization as well.
+        // VPP recycles sw_if_index values after an interface is deleted, so a
+        // stale entry would mis-attribute FDB learn/move/age notifications (and
+        // vpp_fdb_entries_invalidate_by_port matches) to the previous port OID.
+        m_swif_to_port_id.erase(swif);
     }
 }
 
--- a/vslib/vpp/SwitchVppFdb.cpp
+++ b/vslib/vpp/SwitchVppFdb.cpp
@@ -1379,7 +1379,7 @@ sai_object_id_t SwitchVpp::getPortIdFromSwIfIndex(uint32_t sw_if_index)
     return getPortIdFromIfName(std::string(tapname));
 }
 
-void SwitchVpp::generateFdbLearnedOrMoveEvent(const VppFdbKey &key, uint32_t sw_if_index, sai_fdb_event_t event_type)
+bool SwitchVpp::generateFdbLearnedOrMoveEvent(const VppFdbKey &key, uint32_t sw_if_index, sai_fdb_event_t event_type)
 {
     SWSS_LOG_ENTER();
 
@@ -1389,7 +1389,7 @@ void SwitchVpp::generateFdbLearnedOrMoveEvent(const VppFdbKey &key, uint32_t sw_
     if (port_id == SAI_NULL_OBJECT_ID)
     {
         SWSS_LOG_ERROR("FDB: cannot resolve port OID for sw_if_index %u", sw_if_index);
-        return;
+        return false;
     }
 
     sai_object_id_t bv_id = SAI_NULL_OBJECT_ID;
@@ -1401,7 +1401,7 @@ void SwitchVpp::generateFdbLearnedOrMoveEvent(const VppFdbKey &key, uint32_t sw_
     {
         SWSS_LOG_ERROR("FDB: bv_id or bridge_port not found for sw_if_index %u bd %u",
                        sw_if_index, key.bd_id);
-        return;
+        return false;
     }
 
     std::set<FdbInfo>::iterator existing_it = m_fdb_info_set.end();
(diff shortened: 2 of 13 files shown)


yxieca:
## Multi-model review — [vpp][fdb] Add fdb support

Thanks for this — the core VPP FDB push-notification design is solid: the producer (VPP events thread) only touches the MAC-event queue mutex, the consumer drains under `m_apimutex`, the single FIFO preserves event ordering, and the FFI boundary is POD-only with `memcpy`/`memcmp` for the MAC key. A few things to address before merge.

### Should-fix

**1. `l2_macs_event` batch is truncated at 128, dropping FDB deltas** (`vslib/vpp/vppxlate/SaiVppXlate.c`, `vl_api_l2_macs_event_t_handler`)
`l2_macs_event` carries `mac[n_macs]` with a `u32` count. When `n_macs > VPP_MAC_EVENT_BATCH_MAX (128)`, the handler logs an error and drops entries `[128..n)`. FDB events are state deltas, not telemetry: a dropped age leaves a stale MAC, a dropped learn means orchagent never learns it. `max_macs_in_event=10` makes >128 unlikely in practice, but please consume the full batch (e.g. process in chunks and invoke the callback repeatedly) rather than truncating.

### Questions

**2. Teardown ordering of `deinitFdbEventHandling()`**
The MAC-event callback is deregistered at the end of `~SwitchVpp`, after the aging thread is already joined. This is safe as long as `staticMacEventCb` is only ever dispatched on `m_vpp_thread` (which `~SwitchVpp` joins first). Can you confirm VPP's socket client doesn't deliver unsolicited `l2_macs_event` on a separate internal rx thread? If it can, there's a teardown window where the callback could run during/after destruction. Hoisting the deregistration to the start of `Sai::uninitialize` would remove the ambiguity.

**3. Warm-boot reconciliation of `m_vpp_fdb_entries`**
The snapshot map is in-memory only and isn't persisted to the warm-boot file. After a warm reboot it starts empty while VPP's L2FIB still holds entries, so an age event for a pre-reboot MAC won't be in the snapshot and no aged notification is generated (stale MAC). How is the snapshot reconciled with VPP's L2FIB across warm boot?

### Scope — unrelated change bundled here

**4. `syncd/scripts/syncd_init_common.sh`** contains two changes unrelated to VPP FDB and not mentioned in the description:
- The CONFIG_DB key read by the init script is renamed `orch_southbound_zmq_enabled` → `swss_zmq` (that key was added recently in #1835). If the producer side isn't renamed in lockstep this would silently disable southbound ZMQ on every syncd platform. Please confirm the coordinated change and call it out.
- nvidia-bluefield hugepages bumped `11700` → `16000`.

### Nits
- Process-global FFI state `g_mac_event_cb` / `g_mac_event_ctx` / `g_mac_event_batch` vs per-switch `initFdbEventHandling`: the last registration wins and one switch's teardown nulls the callback for all. Latent today (single-switch), but worth an assert/comment documenting the single-switch assumption.
- ADD path records `m_vpp_fdb_entries[key]` even when `generateFdbLearnedOrMoveEvent` early-returns on a port-resolve failure; a later DELETE then calls `generateFdbAgedEvent`, which won't find it in `m_fdb_info_set` and logs an ERROR on a benign race. Consider only recording the snapshot on a confirmed-generated event, or downgrading that log.
- `Meta::objectHasAttribute()` (new in `meta/Meta.{cpp,h}`) has no production caller — it's only used by `TestMetaDash.cpp`.
- The `Sai.fdbAgingWakeEvent` unit test is effectively a start/stop "doesn't deadlock" smoke test (100ms sleep, no MAC events injected); the comment over-states that it exercises the wake path.
- The action codes `0`/`1`/`2` (ADD/DELETE/MOVE) are duplicated across the C struct comment, the trampoline, and the C++ switch — an enum would prevent drift.

vslib/vpp/vppxlate/SaiVppXlate.c
```diff
@@ -952,6 +952,62 @@ vl_api_l2fib_flush_bd_reply_t_handler (vl_api_l2fib_flush_bd_reply_t *msg)
...
+vl_api_l2_macs_event_t_handler (vl_api_l2_macs_event_t *mp)
+{
+    if (!g_mac_event_cb)
+        return;
+
+    u32 n = ntohl(mp->n_macs);
+    if (n == 0) return;
+
+    // Process the full batch in chunks to avoid dropping FDB state deltas.
+    for (u32 off = 0; off < n; off += VPP_MAC_EVENT_BATCH_MAX)
```
yue-fred-gao, line 1046: Is there a concern of causing the caller's WR time out if there are too many events? The timeout is 5 seconds. PR https://github.com/sonic-net/sonic-sairedis/pull/1933 introduces a separated socket and polling thread for event notification. Probably we can ignore this concern with the assumption the other PR will be merged

vslib/vpp/SwitchVppFdb.cpp
```diff
@@ -1304,6 +1311,10 @@ sai_status_t SwitchVpp::vpp_fdbentry_flush(
                     const char *hwif_name = ifname.c_str();
                     auto ret = l2fib_flush_int(hwif_name);
                     SWSS_LOG_NOTICE(" Flush by interface on hwif_name %s  Successful ret_val: %d", hwif_name, ret);
+                    if (is_tunnel_bridge_port(br_port_id))
```
yue-fred-gao: is this already handled at line 1291?

vslib/vpp/SwitchVppFdb.cpp
```diff
@@ -1339,4 +1352,227 @@ sai_status_t SwitchVpp::vpp_fdbentry_flush(
...
-}
\ No newline at end of file
+}
+
+/*
+ * Resolve a VPP sw_if_index to a SAI port OID via the 3-step lookup chain:
+ * sw_if_index -> VPP hw interface name -> Linux tap/SONiC port name -> SAI port OID.
+ * Returns SAI_NULL_OBJECT_ID on any lookup failure.
+ */
+sai_object_id_t SwitchVpp::getPortIdFromSwIfIndex(uint32_t sw_if_index)
```
yue-fred-gao: This is a useful function. Can you move it to SwitchVpp.cpp?

vslib/vpp/SwitchVppFdb.cpp
```diff
@@ -1339,4 +1352,227 @@ sai_status_t SwitchVpp::vpp_fdbentry_flush(
...
+    SWSS_LOG_INFO("FDB: invalidated by bd_id=%u, removed %zu entries, %zu remain",
+                  bd_id, before - m_vpp_fdb_entries.size(), m_vpp_fdb_entries.size());
+}
+
+void SwitchVpp::vpp_fdb_entries_invalidate_by_port(sai_object_id_t port_id)
+{
+    SWSS_LOG_ENTER();
+    /* O(N) scan: resolves sw_if_index -> port OID via VPP API for each entry.
+     * This is acceptable since flush-by-interface is infrequent.
+     * TODO: Improve this by storing the sw_if_index -> port OID mapping somewhere. */
```
yue-fred-gao: Agree. This lookup is costly. Please consider to add the mapping.

vslib/VirtualSwitchSaiInterfaceFdb.cpp
```diff
@@ -1,8 +1,6 @@
 #include "VirtualSwitchSaiInterface.h"
 #include "SwitchStateBase.h"
-#ifdef USE_VPP
```
yue-fred-gao, line 3: is this intentional?

vslib/vpp/vppxlate/SaiVppXlate.c
lolyu: Yes, this can be a concern if we have a large MAC event workload and the the mac event callback called in this handler is a heavy operation.

We can **safely ignore this for now** as we don't have flooded MAC event test scenarios in the sonic-mgmt test suite and the mac event handler called here `SwitchVpp::staticMacEventCb` is lightweight:

1. Acquires a mutex, pushes events to a queue
2. Calls `m_fdbAgingWakeFn()` — which just signals an event/semaphore to wake the aging thread
So the impact is minimal in practice.

The work done inside the WR loop for each MAC event batch is:
A mutex lock + N queue.push() ops
One wake signal

This is O(microseconds) per batch, well within the 5-second WR timeout.

Let me add a **TODO** comment here to move the events into a separated socket.

vslib/vpp/SwitchVppFdb.cpp
lolyu: Good catch

vslib/vpp/SwitchVppFdb.cpp
lolyu: Moved.

vslib/VirtualSwitchSaiInterfaceFdb.cpp
lolyu: reverted.

vslib/vpp/SwitchVppFdb.cpp
lolyu: Yes, add a map to store the sw if -> port OID mapping.

vslib/vpp/vppxlate/SaiVppXlate.c
yue-fred-gao: yes, we can leave it for now. When PR #1933 is merged, we should move Mac event to the new event socket and thread for consistency. staticMacEventCb is detached from sending notifications, which is asynchronously handled by the fdb aging thread.

yue-fred-gao (dismissed):

yxieca:
## Follow-up review — [vpp][fdb] Add fdb support (re-reviewed at `94d2e66`)

Thanks for the rework — re-reviewed the updated branch. The items from the previous round are addressed:

- ✅ **Batch handling** — the `l2_macs_event` handler now consumes the full batch in chunks and invokes the callback per chunk, so events beyond 128 are no longer dropped.
- ✅ **Scope** — the unrelated `syncd_init_common.sh` and `meta/Meta` changes are gone; the PR is now a focused saivpp/vslib diff.
- ✅ **ADD snapshot** — `m_vpp_fdb_entries[key]` is now recorded only on a confirmed-generated event.
- ✅ **Teardown ordering** — `deinitFdbEventHandling()` now runs first in `~SwitchVpp`, before the events thread is joined.
- ✅ **Single-switch global cb** — documented + `assert`.

A few live items remain:

### Should-fix
**`m_swif_to_port_id` cache is never invalidated.** `resolvePortIdFromSwIfIndex` memoizes `sw_if_index → SAI port OID` permanently — there's no `erase`/`clear` on the map. VPP recycles `sw_if_index` values after an interface is deleted, so the next interface reusing that index resolves to the *previous* port's OID. That mis-attributes FDB learn/move/age notifications, and `vpp_fdb_entries_invalidate_by_port` matches against the same stale cache. The companion `m_swif_to_bdid` is already untracked on port-leave via `swif_bdid_untrack` — suggest invalidating `m_swif_to_port_id[sw_if_index]` in the same place.

### Questions
1. **Teardown thread model.** Deregistering before the join is safe as long as `l2_macs_event` is only ever dispatched on the WR() polling loop (per the code comment). Can you confirm VPP's socket client never delivers it on a separate internal rx thread? If it can, there's still a window where `staticMacEventCb` could touch a half-destroyed switch.
2. **Warm boot.** `m_vpp_fdb_entries` / `m_fdb_info_set` are in-memory only. After a warm reboot they start empty while VPP's L2FIB still holds entries, so an age for a pre-reboot MAC hits the "unknown MAC, skipping" path → no aged notification → stale FDB_TABLE in SONiC. How is the snapshot reconciled with VPP's L2FIB across warm boot, or is FDB warm-boot explicitly out of scope for saivpp?

### Nits
- The single-switch `assert(g_mac_event_cb == NULL ...)` is compiled out under `NDEBUG`; a second registration would silently clobber the global in a release build. Consider a runtime log + early-return alongside the assert.
- `SwitchVpp.h` comment references `onVppMacEvent()`; the actual producer is `staticMacEventCb`.
- The unit test exercises the aging thread's start/stop lifecycle; the queue-drain path in `processFdbEntriesForAging` isn't directly covered.

yue-fred-gao (approved):

PR #1937, opened 2026-06-11
PR  1/2 - SaiVppSflow skeleton in phase one

ritvikiscool9:
### Description of PR

This PR contains an implementation of the SaiVppSflow skeleton, specifically for the four tasks:
- create/remove/set for SAI_OBJECT_TYPE_SAMPLEPACKET.
- SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE set handler that calls sflow_enable_disable and sflow_sampling_rate_set for the global rate (per-port rate honored only if all ports use the same rate; otherwise last-writer-wins, log a WARNING).
- SAI_HOSTIF_TRAP_TYPE_SAMPLEPACKET create/remove as bookkeeping
- SAI_HOSTIF of type GENETLINK + table entry as bookkeeping

This is PR 1/2 for phase one. The second one can be found here: https://github.com/sonic-net/sonic-platform-vpp/pull/254 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

The majority of changes needed for phase one of sFlow. The main change is the implementation of the SaiVppSflow translator. 

##### Work item tracking

#### How did you do it?

Used past code as a reference to develop the new sFlow features

The HLD specifies 3 conditions for the exit criteria:
- config sflow enable; config sflow interface enable Ethernet0; config sflow collector add c1 1.1.1.1 6343 results in flow samples reaching sflowtool -p 6343.
- show sflow interface reflects configured state.
- COUNTERS_DB-based counter samples appear at the configured polling interval.

I tested my code changes in a test environment and verified that their results matched and everything worked.

For conditions 1 & 3:

Screenshot 2026-06-18 at 11 22 35 AM

This screenshot shows `sflowtool -p 6343` output on the collector switch, with both record types:
- `Flow` record: Condition 1 (sample packets). These are sampled packets captured on ingress to a sampled port. The fields show `10.0.0.1 -> 10.0.0.2`, ethertype `0x0800` (IPv4), protocol 1 (ICMP), the ping traffic sent from the peer switch (r1) into the sampled interface on the collector switch (r2). Their appearance confirms VPP's sFlow plugin is sampling live dataplane traffic and the samples are reaching the collector via hsflowd. 
- `CNTR` records: Condition 3 (counter samples). These are the periodic traffic independent interface counters read from COUNTERS_DB and exported at the configured polling interval (20s). 

I used a two-switch topology (r1 <-> r2, connected on Ethernet 0), so that real cross-switch traffic could ingress on a sampled port. r1 was given `10.0.0.1/24` and  r2 `10.0.0.2/24` on the connected interface. A ping from r1 -> r2 produced ingress packets on r2's port. sFlow was configured and the collector run on r2:

```
sudo config sflow enable
sudo config sflow interface enable Ethernet0
sudo config sflow collector add c1 127.0.0.1 --port 6343
sudo config sflow interface sample-rate Ethernet0 256
sudo config save -y
docker exec sflow sflowtool -p 6343 -l -k
```

I used `127.0.0.1` instead of the HLD's example `1.1.1.1` so the collector and sflowtool run on the same switch via looback. 

For condition 2:

Screenshot 2026-06-18 at 11 32 19 AM

- Admin state: `up` on all sampled interfaces - sFlow is enabled per port
- Sampling rate: Ethernet0 shows 256, which was set by `sudo config sFlow interface sample-rate Ethernet0 256`, while Ethernet4/Ethernet8 show the default 40000
- Sampling direction: `rx`, ingress sampling. This is the default set in `sflowmgr`

### DIFF
--- a/vslib/vpp/SwitchVpp.cpp
+++ b/vslib/vpp/SwitchVpp.cpp
@@ -1431,6 +1431,13 @@ sai_status_t SwitchVpp::create(
         return sflow_hostif_trap_samplepacket_create(object_id, switch_id, attr_count, attr_list);
     }
 
+    if(object_type == SAI_OBJECT_TYPE_HOSTIF_TABLE_ENTRY)
+    {
+        sai_object_id_t object_id;
+        sai_deserialize_object_id(serializedObjectId, object_id);
+        return sflow_hostif_table_entry_create(object_id, switch_id, attr_count, attr_list);
+    }
+
     if (object_type == SAI_OBJECT_TYPE_MACSEC_PORT)
     {
         sai_object_id_t object_id;
@@ -1780,6 +1787,11 @@ sai_status_t SwitchVpp::remove(
     if(object_type == SAI_OBJECT_TYPE_HOSTIF_TRAP)
     {
         return sflow_hostif_trap_samplepacket_remove(serializedObjectId);
+    }
+
+    if(object_type == SAI_OBJECT_TYPE_HOSTIF_TABLE_ENTRY)
+    {
+        return sflow_hostif_table_entry_remove(serializedObjectId);
     }
 
     if (object_type == SAI_OBJECT_TYPE_ACL_ENTRY)
--- a/vslib/vpp/SwitchVpp.cpp
+++ b/vslib/vpp/SwitchVpp.cpp
@@ -1906,6 +1906,45 @@ sai_status_t SwitchVpp::setPort(
 
     UpdatePort(portId, 1, attr);
 
+    if (attr->id == SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE)
+    {
+        sai_object_id_t sp_oid = attr->value.oid;
+
+        if(sp_oid == SAI_NULL_OBJECT_ID)
+        {
+            m_sflow_port_to_samplepacket.erase(portId);
+            sflow_enable_disable(portId, false);
+        }
+        else 
+        {
+            sai_attribute_t rate_attr;
+            rate_attr.id = SAI_SAMPLEPACKET_ATTR_SAMPLE_RATE;
+            uint32_t rate = 0;
+
+            auto serialized_id = sai_serialize_object_id(sp_oid);
+
+            if(get(SAI_OBJECT_TYPE_SAMPLEPACKET, serialized_id, 1, &rate_attr) == SAI_STATUS_SUCCESS)
+            {
+                rate = rate_attr.value.u32;
+            }
+
+            if(m_sflow_sample_rate != 0 && m_sflow_sample_rate != rate)
+            {
+                SWSS_LOG_WARN("sFlow sample rate mismatch: global=%u port %s requesting=%u (last-writer-wins)",
+                    m_sflow_sample_rate,
+                    sai_serialize_object_id(portId).c_str(),
+                    rate);
+            }
+
+            m_sflow_port_to_samplepacket[portId] = sp_oid;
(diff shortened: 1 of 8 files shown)


vslib/vpp/SwitchVpp.cpp
```diff
@@ -1759,6 +1795,45 @@ sai_status_t SwitchVpp::setPort(
 
     UpdatePort(portId, 1, attr);
 
+    if (attr->id == SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE)
```
yue-fred-gao: can we move this to a function in SwitchVppSflow.cpp?

vslib/vpp/SwitchVpp.cpp
```diff
@@ -1759,6 +1795,45 @@ sai_status_t SwitchVpp::setPort(
...
+
+            if(m_sflow_sample_rate != 0 && m_sflow_sample_rate != rate)
+            {
+                SWSS_LOG_WARN("sFlow sample rate mismatch: global=%u port %s requesting=%u (last-writer-wins)",
+                    m_sflow_sample_rate,
+                    sai_serialize_object_id(portId).c_str(),
+                    rate);
+            }
+
+            m_sflow_port_to_samplepacket[portId] = sp_oid;
```
yue-fred-gao: we don't need to maintain the mapping? The information is available in the Port SAI object, which is already stored

vslib/vpp/SwitchVpp.h
```diff
@@ -1022,7 +1024,46 @@ namespace saivs
...
+                    _In_ const sai_attribute_t *attr_list);
 
+            sai_status_t samplePacketRemove(
+                    _In_ const std::string &serializedObjectId);
+
+            sai_status_t samplePacketSet(
+                    _In_ sai_object_id_t entry_id,
+                    _In_ const sai_attribute_t *attr);
+
+            sai_status_t sflow_enable_disable(
```
yue-fred-gao: let's try to use the same case (camel case) for the functions. I know we are not very consistent in the past.

vslib/vpp/SwitchVppSflow.cpp
```diff
@@ -0,0 +1,174 @@
...
+    _In_ const sai_attribute_t *attr_list)
+{
+    SWSS_LOG_ENTER();
+
+    auto sid = sai_serialize_object_id(object_id);
+
+    const sai_attribute_value_t *trap_type_attr = nullptr;
+    uint32_t attr_index = 0;
+
+    sai_status_t status = find_attrib_in_list(
```
yue-fred-gao: Please consider to use SaiCachedObject like https://github.com/sonic-net/sonic-sairedis/blob/9fe84d3c819939ac410ff717f9a858653f5b5461/vslib/vpp/SwitchVppNexthop.cpp#L277 and get_mandatory_attr

vslib/vpp/SwitchVpp.cpp
yue-fred-gao: sorry, one more request. Can we move the call to UpdatePort? We have similar handling there like INGRESS_ACL

yue-fred-gao (commented):
Please resolve conflict, fix DCO and trigger t1-lag-vpp sanity via sonic-buildimage PR (create a PR in sonic-buildimage with src/sonic-sairedis submodule updated to this PR branch)

yue-fred-gao (approved):

PR #1945, opened 2026-06-18
Address racey SAI PHY Serdes object causing bad syncd setup

### LINKED ISSUE #1946: SAI Port PHY Serdes object causing syslog ERRs when polling
Occasionally during long-running testing (`sonic-mgmt` full test suite), there will be continous polling of the ERR syslogs:
```
30979 2026 Jun  7 20:56:03.499522 qspf201 ERR syncd#syncd: :- initAttrData: PORT_PHY_SERDES_ATTR: port_serdes_rid:0xb005700000000 has no serdes count attribute information
30980 2026 Jun  7 20:56:03.499522 qspf201 WARNING syncd#syncd: :- collectData: PORT_PHY_SERDES_ATTR: Failed to initialize attribute SAI_PORT_SERDES_ATTR_TX_FIR_TAPS_LIST for       RID:0xb005700000000, skipping object
```

This is caused by some SAI read failure earlier during `syncd`'s setup:
```
26929 2026 Jun  7 20:55:59.007758 qspf201 ERR syncd#syncd: [none] SAI_API_PORT:brcm_sai_get_port_serdes_attribute_cmn:18248 Port 11 hardware lane count get failed with error       -17.

26930 2026 Jun  7 20:55:59.007758 qspf201 ERR syncd#syncd: :- updatePortSerdesTapsCountMap: PORT_PHY_SERDES_ATTR: Failed to get port serdes count attr SAI_PORT_SERDES_ATTR_T      X_FIR_COUNT for port_serdes RID:0xb005700000000, status:-17
```

The reason this happens is because the Port PHY Serdes object only supports `WRITE` in SAI. Sometime during `syncd`'s initialization, the configuration for this object changes, and requires the object to be re-created for the configuration change to take effect.
There is likely not enough time difference between the creation of the object, and the object being ready for SAI interactions.

This issue occurs around ~2.5% chance on a `config reload` or `reboot`.

justin-wong-ce:
### Description of PR

Summary:
Mitigate a race condition in Port PHY Serdes initialization (`syncd`) where the SAI is still busy with the object from its creation when `syncd`'s intialization attempts to read attributes from the object for syncd's own setup.

This race condition will cause a failing SAI call that is needed for initializing `syncd`, causing polls to the Port PHY Serdes object to fail continuously later on. The chances of this happening is around ~2.5% per `config reload` / `reboot`.

Do this by allowing 5 retries with 10ms wait intervals in between tries. The retry will only trigger when the failure reason is `SAI_STATUS_OBJECT_IN_USE`.

Fixes #1946

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

Fix syslog ERRs that logs continuously when this scenario occurs.

##### Work item tracking

#### How did you do it?
Added multiple attempts for interacing with Port PHY Serdes object over SAI during setup.

Continuous syslog ERRs no longer occur. The retry logic can be seen working in the syslogs as well. The most retry attempts observed is 1.

Broadcom.

### DIFF
--- a/syncd/FlexCounter.cpp
+++ b/syncd/FlexCounter.cpp
@@ -2231,28 +2231,29 @@ public:
         sai_attribute_t attr;
         attr.id = SAI_PORT_SERDES_ATTR_PORT_ID;
 
-        sai_status_t status = SAI_STATUS_OBJECT_IN_USE;
-        while (status == SAI_STATUS_OBJECT_IN_USE)
+        for (uint32_t tries = 0; tries < 5; tries++)
         {
-            status = Base::m_vendorSai->get(Base::m_objectType, port_serdes_rid, 1, &attr);
-            if (status != SAI_STATUS_OBJECT_IN_USE)
+            sai_status_t status = Base::m_vendorSai->get(Base::m_objectType, port_serdes_rid, 1, &attr);
+
+            if (status == SAI_STATUS_SUCCESS)
             {
+                port_rid = attr.value.oid;
+                return true;
+            }
+            else if (status == SAI_STATUS_OBJECT_IN_USE and tries < 2)
+            {
+                // SAI object is busy - retry in 10ms
+                SWSS_LOG_WARN("PORT_PHY_SERDES_ATTR: SAI object in use, retry getting port RID for port serdes RID:0x%" PRIx64 "...",
+                               port_serdes_rid);
+                std::this_thread::sleep_for(chrono::milliseconds(10));
+            }
+            else
+            {
+                SWSS_LOG_ERROR("PORT_PHY_SERDES_ATTR: Failed to get port RID for port serdes RID:0x%" PRIx64 ", status:%d",
+                               port_serdes_rid, status);
                 break;
             }
-            // SAI object is busy - retry in 10ms
-            SWSS_LOG_WARN("PORT_PHY_SERDES_ATTR: SAI object in use, retry getting port RID for port serdes RID:0x%" PRIx64 "...",
-                           port_serdes_rid);
-            std::this_thread::sleep_for(chrono::milliseconds(10));
         }
-
-        if (status == SAI_STATUS_SUCCESS)
-        {
-            port_rid = attr.value.oid;
-            return true;
-        }
-
-        SWSS_LOG_ERROR("PORT_PHY_SERDES_ATTR: Failed to get port RID for port serdes RID:0x%" PRIx64 ", status:%d",
-                       port_serdes_rid, status);
         return false;
     }
 
@@ -2326,28 +2327,27 @@ public:
         attr.value.u32list.count = 0;        // Query with count=0 to get the actual lane count
         attr.value.u32list.list = nullptr;
 
-        sai_status_t status = SAI_STATUS_OBJECT_IN_USE;
-        while (status == SAI_STATUS_OBJECT_IN_USE)
+        for (uint32_t tries = 0; tries < 5; tries++)
         {
-            status = Base::m_vendorSai->get(SAI_OBJECT_TYPE_PORT, port_rid, 1, &attr);
-            if (status != SAI_STATUS_OBJECT_IN_USE)
-            {
+            sai_status_t status = Base::m_vendorSai->get(SAI_OBJECT_TYPE_PORT, port_rid, 1, &attr);
+
+            // The SAI status expected is SAI_STATUS_BUFFER_OVERFLOW since we pass in a nullptr
+            // This is the agreed method with Broadcom for retrieving the actual lane count
+            if (status == SAI_STATUS_BUFFER_OVERFLOW)
                 break;
+            else if (status == SAI_STATUS_OBJECT_IN_USE && tries < 2)
+            {
+                // SAI object is busy - retry in 10ms
+                SWSS_LOG_WARN("PORT_PHY_SERDES_ATTR: SAI object in use, retry getting hardware lane count for port RID:0x%" PRIx64 "...",
+                              port_rid);
+                std::this_thread::sleep_for(chrono::milliseconds(10));
             }
-
-            // SAI object is busy - retry in 10ms
-            SWSS_LOG_WARN("PORT_PHY_SERDES_ATTR: SAI object in use, retry getting hardware lane count for port RID:0x%" PRIx64 "...",
-                          port_rid);
-            std::this_thread::sleep_for(chrono::milliseconds(10));
-        }
-
-        // The SAI status expected is SAI_STATUS_BUFFER_OVERFLOW since we pass in a nullptr
-        // This is the agreed method with Broadcom for retrieving the actual lane count
-        if (status != SAI_STATUS_BUFFER_OVERFLOW)
-        {
-            SWSS_LOG_ERROR("PORT_PHY_SERDES_ATTR: Failed to get hardware lane count for port RID:0x%" PRIx64 ", status:%d",
-                           port_rid, status);
-            return;
+            else
+            {
+                SWSS_LOG_ERROR("PORT_PHY_SERDES_ATTR: Failed to get hardware lane count for port RID:0x%" PRIx64 ", status:%d",
+                              port_rid, status);
+                return;
+            }
         }
 
         laneCount = attr.value.u32list.count;
@@ -2375,32 +2375,34 @@ public:
             sai_attribute_t attr;
             attr.id = attrId;
 
-            sai_status_t status = SAI_STATUS_OBJECT_IN_USE;
-            while (status == SAI_STATUS_OBJECT_IN_USE)
+            bool failed = false;
+            for (uint32_t tries = 0; tries < 5; tries++)
             {
-                status = Base::m_vendorSai->get(
+                sai_status_t status = Base::m_vendorSai->get(
                     Base::m_objectType,
                     port_serdes_rid,
                     1,
                     &attr);
 
-                if (status != SAI_STATUS_OBJECT_IN_USE)
-                {
+                if (status == SAI_STATUS_SUCCESS)
                     break;
+                else if (status == SAI_STATUS_OBJECT_IN_USE && tries < 2)
+                {
+                    // SAI object is busy - retry in 10ms
+                    SWSS_LOG_WARN("PORT_PHY_SERDES_ATTR: SAI object in use, retry getting port serdes count attr %s for port_serdes RID:0x%" PRIx64 "...",
+                                  sai_serialize_port_serdes_attr(attrId).c_str(), port_serdes_rid);
+                    std::this_thread::sleep_for(chrono::milliseconds(10));
                 }
-
-                // SAI object is busy - retry in 10ms
-                SWSS_LOG_WARN("PORT_PHY_SERDES_ATTR: SAI object in use, retry getting port serdes count attr %s for port_serdes RID:0x%" PRIx64 "...",
-                              sai_serialize_port_serdes_attr(attrId).c_str(), port_serdes_rid);
-                std::this_thread::sleep_for(chrono::milliseconds(10));
+                else
+                {
+                    SWSS_LOG_ERROR("PORT_PHY_SERDES_ATTR: Failed to get port serdes count attr %s for port_serdes RID:0x%" PRIx64 ", status:%d",
+                                  sai_serialize_port_serdes_attr(attrId).c_str(), port_serdes_rid, status);
+                    failed = true;
+                    continue;
+                }
             }
-
-            if (status != SAI_STATUS_SUCCESS)
-            {
-                SWSS_LOG_ERROR("PORT_PHY_SERDES_ATTR: Failed to get port serdes count attr %s for port_serdes RID:0x%" PRIx64 ", status:%d",
-                               sai_serialize_port_serdes_attr(attrId).c_str(), port_serdes_rid, status);
+            if (failed)
                 continue;
-            }
 
             count = attr.value.u32;
 
--- a/syncd/FlexCounter.cpp
+++ b/syncd/FlexCounter.cpp
@@ -2231,7 +2231,7 @@ public:
         sai_attribute_t attr;
         attr.id = SAI_PORT_SERDES_ATTR_PORT_ID;
 
-        for (uint32_t tries = 0; tries < 5; tries++)
+        for (uint32_t tries = 1; tries <= 5; tries++)
         {
             sai_status_t status = Base::m_vendorSai->get(Base::m_objectType, port_serdes_rid, 1, &attr);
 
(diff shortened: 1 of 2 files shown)


syncd/FlexCounter.cpp
```diff
@@ -2229,16 +2230,30 @@ class PortPhySerdesAttrContext : public AttrContext<sai_port_serdes_attr_t, Port
...
         attr.id = SAI_PORT_SERDES_ATTR_PORT_ID;
-        sai_status_t status = Base::m_vendorSai->get(Base::m_objectType, port_serdes_rid, 1, &attr);
 
-        if (status == SAI_STATUS_SUCCESS)
+        for (uint32_t tries = 0; tries < 3; tries++)
         {
-            port_rid = attr.value.oid;
-            return true;
-        }
+            sai_status_t status = Base::m_vendorSai->get(Base::m_objectType, port_serdes_rid, 1, &attr);
```
r12f, line 2236: hey Justin - the fix is a bit of "by-chance". 

Also supposedly,
1) flex counter should not fail syncd init
2) flex counter should work after the objects gets created

so I wonder if we have looked deeper into the failure/race and see what might be happening?

justin-wong-ce: As mentioned above, the race condition is from the SAI not being able to handle the Port PHY Serdes object creation and accessing of its attribute in extremely quick succession.

Either the SAI needs to be able to delay a request and not return `SAI_STATUS_OBJECT_IN_USE`, or Sonic (`syncd` initialization for this case) needs to be able to handle `SAI_STATUS_OBJECT_IN_USE` properly.


> the fix is a bit of "by-chance"

I can change the fix such that it will retry until SAI does not return the `SAI_STATUS_OBJECT_IN_USE`, rather than just 3 times - this will remove the chance element.

I think it is reasonable to retry when the SAI returns `SAI_STATUS_OBJECT_IN_USE` since the object is busy and not available for read.

justin-wong-ce: Please let me know what you think of the latest changes

r12f: what if we just simply skip the failed round and retry in the next pulling interval?

r12f: retrying 3 times is not going to guarantee the fix anyway.

justin-wong-ce: This error is occuring during the initialization stage and this code that is being edited is for the initialization of Port PHY Serdes.
This will print the following ERR syslog once - during `syncd`'s intialization:
```
30979 2026 Jun  7 20:56:03.499522 qspf201 ERR syncd#syncd: :- initAttrData: PORT_PHY_SERDES_ATTR: port_serdes_rid:0xb005700000000 has no serdes count attribute information
30980 2026 Jun  7 20:56:03.499522 qspf201 WARNING syncd#syncd: :- collectData: PORT_PHY_SERDES_ATTR: Failed to initialize attribute SAI_PORT_SERDES_ATTR_TX_FIR_TAPS_LIST for RID:0xb005700000000, skipping object
```

If `SAI_STATUS_OBJECT_IN_USE` is returned here and we don't retry, this will result missing info (maps) that is used to do other initalization steps later on, as it is treated as a failure.

If the info is missing, it will cause polling to fail all the time.
The following ERR syslog will print at every single poll interval:
```
26929 2026 Jun  7 20:55:59.007758 qspf201 ERR syncd#syncd: [none] SAI_API_PORT:brcm_sai_get_port_serdes_attribute_cmn:18248 Port 11 hardware lane count get failed with error -17.

26930 2026 Jun  7 20:55:59.007758 qspf201 ERR syncd#syncd: :- updatePortSerdesTapsCountMap: PORT_PHY_SERDES_ATTR: Failed to get port serdes count attr SAI_PORT_SERDES_ATTR_TX_FIR_COUNT for port_serdes RID:0xb005700000000, status:-17
```

dhanasekar-arista: @justin-wong-ce  i see you have removed the retry loop, I dont think this is a right approach.
We should not retry without an upper bound. If we believe a retry of 3 times with a 10ms is not enough we can increase this to 5 retries but we should not leave it unbounded. 
@r12f fyi.

syncd/FlexCounter.cpp
```diff
@@ -2311,13 +2326,28 @@
...
-        sai_status_t status = Base::m_vendorSai->get(SAI_OBJECT_TYPE_PORT, port_rid, 1, &attr);
 
-        if (status != SAI_STATUS_BUFFER_OVERFLOW)
+        for (uint32_t tries = 0; tries < 5; tries++)
         {
-            SWSS_LOG_ERROR("PORT_PHY_SERDES_ATTR: Failed to get hardware lane count for port RID:0x%" PRIx64 ", status:%d",
-                          port_rid, status);
-            return;
+            sai_status_t status = Base::m_vendorSai->get(SAI_OBJECT_TYPE_PORT, port_rid, 1, &attr);
+
```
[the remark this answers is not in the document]
justin-wong-ce: This seems to be a stale check and I do not have the permissions to dismiss/resolve it.

thisptr-sh:
Hi @justin-wong-ce , can we get this PR merged? Are there any remaining blockers?

This issue is the biggest source of loganalyzer teardown failures across our SONiC 202511 hardware qualification on Broadcom TH5 (Arista 7060X6-512). We're seeing the same periodic ERR from initAttrData (port_serdes_rid:0x1005700000000 has no lane count information) every 10 second.

syncd/FlexCounter.cpp
justin-wong-ce: @r12f please advise whether if we can go forward with 5 retries

justin-wong-ce:
> Hi @justin-wong-ce , can we get this PR merged? Are there any remaining blockers?
> 
> This issue is the biggest source of loganalyzer teardown failures across our SONiC 202511 hardware qualification on Broadcom TH5 (Arista 7060X6-512). We're seeing the same periodic ERR from initAttrData (port_serdes_rid:0x1005700000000 has no lane count information) every 10 second.

I am awaiting feedback in this thread: https://github.com/sonic-net/sonic-sairedis/pull/1945#discussion_r3574755429

I do not have merge permissions. To pre-emptively resolve any potential code coerage issues I am working on a unit test at the moment.

justin-wong-ce:
Note: I am aware of the cherry-pick conflict to 202511. I have a local branch that addresses this conflict ready and will make a direct backport when this PR is approved.

syncd/FlexCounter.cpp
```diff
@@ -2946,18 +2375,34 @@ class PortPhySerdesAttrContext : public AttrContext<sai_port_serdes_attr_t, Port
...
+                    SWSS_LOG_WARN("PORT_PHY_SERDES_ATTR: SAI object in use, retry getting port serdes count attr %s for port_serdes RID:0x%" PRIx64 "...",
+                                  sai_serialize_port_serdes_attr(attrId).c_str(), port_serdes_rid);
+                    std::this_thread::sleep_for(chrono::milliseconds(10));
+                }
+                else
+                {
+                    SWSS_LOG_ERROR("PORT_PHY_SERDES_ATTR: Failed to get port serdes count attr %s for port_serdes RID:0x%" PRIx64 ", status:%d",
+                                  sai_serialize_port_serdes_attr(attrId).c_str(), port_serdes_rid, status);
+                    failed = true;
+                    continue;
```
tjchadaga: should this be 'break'?

tjchadaga:
@justin-wong-ce - are these failure logs only seen on a specific platform?

justin-wong-ce:
> @justin-wong-ce - are these failure logs only seen on a specific platform?

I do not have access to other platforms - I can only test on Broadcom.

I cannot say for sure if it happens on other platforms so I only noted `Broadcom` in the PR description.

justin-wong-ce:
Restarting CI due to failure from:
```
The process '/usr/bin/python' failed with exit code 1
BuildTrixie • amd64 • [Auto] Generate build coverage properties

ENOENT: no such file or directory, open '/__w/1/s/.diff-coverage/codecoverageProperties.json'
BuildTrixie • amd64 • [Auto] Update build coverage properties
Bash exited with code '100'.
Test • vstest • Install dependencies
```

r12f:
hi @tjchadaga - do you mind to help check this PR?

rookie-who:

syncd/FlexCounter.cpp
saiarcot895: Just to check, is it that because flexcounter is running in its own thread and starts as part of syncd initialization, it could be that flexcounter is starting to get port stats before SAI initialization (triggered from the sync main thread) completes? If so, should the flexcounter thread start instead be delayed until after SAI initialization?

justin-wong-ce: Yes and no - the situaiton with the Port PHY Serdes objects are a bit special.

The Port PHY Serdes objects are first initialized like any other counter during `syncd`'s initialization. There are no problems up to this part.

After a port configuration update, the configuration for these objects also change.

Since these objects only support `CREATE`, the objects have to be re-created to be reconfigured. What happens in the code is it will re-create these objects, then proceed to read from the objects from SAI so it can instantiate some variables in `FlexCounter.cpp` for polling - these reads usually happens on the next poll on these objects (which may or may not be the 1st poll).

The time between is not sufficient in some cases since the reconfiguration happens late into initialization or sometimes when polling has already started.
Althought its a rare occurence, it still needs to be accounted for.

syncd/FlexCounter.cpp
```diff
@@ -2229,16 +2230,30 @@ class PortPhySerdesAttrContext : public AttrContext<sai_port_serdes_attr_t, Port
...
         attr.id = SAI_PORT_SERDES_ATTR_PORT_ID;
-        sai_status_t status = Base::m_vendorSai->get(Base::m_objectType, port_serdes_rid, 1, &attr);
 
-        if (status == SAI_STATUS_SUCCESS)
+        for (uint32_t tries = 1; tries <= 5; tries++)
         {
-            port_rid = attr.value.oid;
-            return true;
-        }
+            sai_status_t status = Base::m_vendorSai->get(Base::m_objectType, port_serdes_rid, 1, &attr);
```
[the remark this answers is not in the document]
prgeor: @justin-wong-ce ?

syncd/FlexCounter.cpp
```diff
@@ -2830,16 +2230,30 @@ class PortPhySerdesAttrContext : public AttrContext<sai_port_serdes_attr_t, Port
...
+            {
+                port_rid = attr.value.oid;
+                return true;
+            }
+            else if (status == SAI_STATUS_OBJECT_IN_USE and tries < 5)
+            {
+                // SAI object is busy - retry in 10ms
+                SWSS_LOG_WARN("PORT_PHY_SERDES_ATTR: SAI object in use, retry getting port RID for port serdes RID:0x%" PRIx64 "...",
+                               port_serdes_rid);
+                std::this_thread::sleep_for(chrono::milliseconds(10));
```
prgeor, line 2248: @justin-wong-ce this is risky to sleep assuming we have say 256 ports which will result in 256 * 10 * 5 = 12.8 secs.

syncd/FlexCounter.cpp
```diff
@@ -2946,18 +2375,34 @@ class PortPhySerdesAttrContext : public AttrContext<sai_port_serdes_attr_t, Port
...
-                continue;
+                sai_status_t status = Base::m_vendorSai->get(
+                    Base::m_objectType,
+                    port_serdes_rid,
+                    1,
+                    &attr);
+
+                if (status == SAI_STATUS_SUCCESS)
+                    break;
+                else if (status == SAI_STATUS_OBJECT_IN_USE && tries < 5)
```
prgeor, line 2389: @justin-wong-ce I think we need to understand who else is modifying the serdes object to keep it busy in BRCM SDK. Adding a busy retry only mitigates the problem to some extent but there could be other race condition in future where these retries may not be sufficient. Do you know how else is modifying the serdes object?

syncd/FlexCounter.cpp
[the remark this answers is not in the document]
justin-wong-ce: Forgot to update the description after applying changes from https://github.com/sonic-net/sonic-sairedis/pull/1945#discussion_r3440257781

syncd/FlexCounter.cpp
justin-wong-ce: The 2.5% chance is the probability of it happening to 1 port on a config-reload or reboot. Of course yes, there is the chance of it happening to all ports at the same time and I agree this is not an optimal solution.

syncd/FlexCounter.cpp
justin-wong-ce: To my knowledge, there is just not enough time between creating the object and reading from it.

justin-wong-ce: Port phy serdes objects are first created in intialization.

Then some time later, a reconfiguration can happen. This can be during intialization or during polling.
To apply the new config, the object has to be recreated.

There is a chance a poll will happen too quickly after the object is created.

justin-wong-ce: AFAIK, there is no mechanism to partially delay the polling only on the Port PHY Serdes objects, or on object creation. However, I am also not familiar with the `orchagent` code.

If there is a way to:
1. stop polling on certain objects when destroy object
2. when an object isc reated, check the object is ready for polling before turning on polling

^then we should use this approach, this will be the ideal fix.

thisptr-sh:
Hi @justin-wong-ce thanks for moving this forward! Wanted to check in and see if we have any progress on this as this is the last pending blocker on TH5 202511

justin-wong-ce:
> Hi @justin-wong-ce thanks for moving this forward! Wanted to check in and see if we have any progress on this as this is the last pending blocker on TH5 202511

Still in the progress of finding the deeper level root cause. I think this change sufficies as a temporarily workaround, but could be better for `master` / `202605`. For now, it seems like it may still take some time.

You might want to consider backporting the current PR changes as a temp fix to unblock your projects.

justin-wong-ce:
Attempt full fix:
https://github.com/sonic-net/sonic-sairedis/pull/2030

rookie-who:
@justin-wong-ce — This PR is requested for the **202605 branch** but is missing test evidence. To move forward with cherry-pick approval, please:

1. **Test on the 202605 branch** and add results (image version + test output) to the PR description
2. **Check the `202605` box** in the "Tested Branches" section of the PR description
3. **Add the "Tested for 202605 branch" label** once testing is complete

Also, to help us prioritize correctly, could you clarify: **Is this PR a bug fix / regression fix, a new feature, a platform config change, or a dependency/infra update?** Understanding the nature of the change helps us triage the cherry-pick queue.

Thank you!

*Review by @vaibhavhd via automated tooling*

rookie-who:
Hi — this PR has been pending cherry-pick to the **202605 branch** for 72 days with no activity in the last 11 days. Are you still planning to have this cherry-picked to 202605?

If yes, please respond here within **7 days** and:
1. **Test on the 202605 branch** and add results to the PR description
2. **Add the `Tested for 202605 branch` label**

If there is no response within 7 days, we will remove the `Request for 202605 Branch` label automatically.

PR #1948, opened 2026-06-22
Select failed errors in Syncd due to PR#1906

sivat6:
Description of PR
Noticed the following error logs in syncd.

2026 Jun 15 06:51:58.579634 vlab-01 ERR syncd#syncd: :- run: select failed: 2

Summary:
Fixes #
processPendingDampingSync() was directly sending notifications via m_notifications->send() which might affect the fd that select() is managing resulted in inconsistent state.

Introduced a queue with mutex protection. processPendingDampingSync() enqueues the notifications and flushPendingDampingNotifications() dequeues and sends in main thread.

There were few review comments given after the marge of PR#1906. Addressed those review comments.

Type of change
 Bug fix
Approach
What is the motivation for this PR?
Build issue in sonic-buildimage

Work item tracking
None

How did you do it?
The processPendingDampingSync() just enqueues the notifications. Later flushPendingDampingNotifications() dequeues and sends in main thread.

How did you verify/test it?
Unit Testing done. Without the fix, I can see the run:select_failed messages every 1 sec. After the fix, I am not seeing this error logs.

Any platform specific information?
None

Documentation
None

### DIFF
--- a/syncd/Syncd.cpp
+++ b/syncd/Syncd.cpp
@@ -1227,7 +1227,7 @@ bool Syncd::applyAiedAlgorithm(
         // Penalty decays based on last_decay_time tracking
         else if (state.current_penalty < state.aied_config.reuse_threshold)
         {
-            // Penalty decayed below resue threshold. Exit damping.
+            // Penalty decayed below reuse threshold. Exit damping.
             std::string physicalStatusStr = sai_serialize_port_oper_status(state.physical_status);
             std::string advertisedStatusStr = sai_serialize_port_oper_status(state.advertised_status);
             SWSS_LOG_NOTICE("Port VID %s exiting damped state: penalty (%u) < "
--- a/syncd/Syncd.cpp
+++ b/syncd/Syncd.cpp
@@ -877,7 +877,6 @@ sai_status_t Syncd::processLinkEventDampingConfigSet(
     // Extract object type and object ID
     std::string strObjectType = key.substr(0, colon_pos);
     std::string strObjectId = key.substr(colon_pos + 1);
-
     sai_object_type_t objectType;
     sai_deserialize_object_type(strObjectType, objectType);
 
@@ -885,7 +884,7 @@ sai_status_t Syncd::processLinkEventDampingConfigSet(
     if (objectType != SAI_OBJECT_TYPE_PORT)
     {
         SWSS_LOG_ERROR("invalid object type for link event damping config: %s",
-		strObjectType.c_str());
+                strObjectType.c_str());
         sendLinkEventDampingConfigResponse(SAI_STATUS_INVALID_PARAMETER);
         return SAI_STATUS_INVALID_PARAMETER;
     }
@@ -911,10 +910,11 @@ sai_status_t Syncd::processLinkEventDampingConfigSet(
         return SAI_STATUS_INVALID_PARAMETER;
     }
 
-    // Link event damping is a software-based feature implemented in syncd.
-    // Store the configuration parameters on the port object so that
-    // OnPortStateChange can apply the damping algorithm before forwarding notifications.
-    // The damping parameters will be used to decide whether to suppress link state changes.
+    /* Link event damping is a software-based feature implemented in syncd.
+     * Store the configuration parameters on the port object so that OnPortStateChange
+     * can apply the damping algorithm before forwarding notifications. The damping
+     * parameters will be used to decide whether to suppress link state changes.
+     */
     sai_status_t status = SAI_STATUS_SUCCESS;
 
     // Acquire lock to protect damping state
@@ -930,7 +930,7 @@ sai_status_t Syncd::processLinkEventDampingConfigSet(
         std::string strAttrValue = fvValue(v);
 
         SWSS_LOG_DEBUG("processing link event damping attribute: %s = %s",
-		strAttrId.c_str(), strAttrValue.c_str());
+                strAttrId.c_str(), strAttrValue.c_str());
 
         // Deserialize attribute ID
         sai_redis_port_attr_t attrId;
@@ -947,11 +947,8 @@ sai_status_t Syncd::processLinkEventDampingConfigSet(
                 SWSS_LOG_INFO("setting link event damping algorithm on port %s: %d",
                         strObjectId.c_str(), algo);
 
-                // Link event damping is a software-only feature as of now
-                // Store the configuration locally for use in notification
-                // processing.
+                // Store the configuration locally for use in notification processing
                 dampingState.algorithm = algo;
-
                 status = SAI_STATUS_SUCCESS;
                 break;
             }
@@ -960,7 +957,6 @@ sai_status_t Syncd::processLinkEventDampingConfigSet(
             {
                 // Allocate temporary memory for the config structure
                 sai_redis_link_event_damping_algo_aied_config_t config{};
-
                 sai_deserialize_redis_link_event_damping_aied_config(strAttrValue, config);
 
                 SWSS_LOG_INFO("setting link event damping AIED config on port %s: "
@@ -970,11 +966,7 @@ sai_status_t Syncd::processLinkEventDampingConfigSet(
                         config.suppress_threshold, config.reuse_threshold,
                         config.decay_half_life, config.flap_penalty);
 
-                // Link event damping is a software-only feature as of now
-                // Store the configuration locally for use in notification
-                // processing.
                 dampingState.aied_config = config;
-
                 status = SAI_STATUS_SUCCESS;
                 break;
             }
@@ -997,7 +989,6 @@ sai_status_t Syncd::processLinkEventDampingConfigSet(
     }
 
     sendLinkEventDampingConfigResponse(status);
-
     return status;
 }
 
@@ -1013,7 +1004,6 @@ void Syncd::sendLinkEventDampingConfigResponse(
     }
 
     std::string strStatus = sai_serialize_status(status);
-
     std::vector<swss::FieldValueTuple> entry;
 
     SWSS_LOG_INFO("sending link event damping config response: %s", strStatus.c_str());
@@ -1064,7 +1054,6 @@ void Syncd::decayPenalty(
     }
 
     // Penalty decay formula: P(t) = P0 * (0.5 ^ (t / half_life))
-    // We use floating point for the calculation
     double half_lives = (double)elapsed_ms / state.aied_config.decay_half_life;
     double decay_factor = std::pow(0.5, half_lives);
     uint32_t decayed_penalty = (uint32_t)(state.current_penalty * decay_factor);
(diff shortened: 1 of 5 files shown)


prsunny:
Can you check build failure?

syncd/Syncd.cpp
```diff
@@ -1533,7 +1553,6 @@ void Syncd::processPendingDampingSync()
                 state.advertised_status != state.physical_status)
             {
                 state.pending_state_sync = false;
-
                 state.advertised_status = state.physical_status;
```
lolyu: 🟠 Correctness: this commits the in-memory state transition (`pending_state_sync = false`, `advertised_status = physical_status`) at **enqueue** time, but the actual `m_notifications->send` happens later in `flushPendingDampingNotifications`. If the queued batch is subsequently **dropped by the overflow guard** (line ~1591, `pop()` oldest) or **stranded** (when `hasAnyValidDampingConfig()` is false at flush time — the gating the Copilot reviewer flagged), the notification is never sent, and because `pending_state_sync` was already cleared here, nothing ever retries it. The port's advertised state stays mismatched indefinitely with no self-heal. Note the overflow-drop case is strictly worse than the gating case: gating can still recover if damping config returns, but a dropped batch cannot recover at all. Suggest updating `advertised_status` (and clearing `pending_state_sync`) only **after** a successful send in flush, so a dropped or stranded batch is naturally retried on the next pass.

syncd/Syncd.cpp
```diff
@@ -6830,11 +6952,18 @@ void Syncd::run()
...
             else
             {
-                SWSS_LOG_ERROR("select failed: %d", result);
+                SWSS_LOG_ERROR("Select returned unknown selectable: %p", sel);
             }
 
-            // Process if any pending state sync due to link event damping
-            processPendingDampingSync();
+            // Only process damping functions if at least one port has valid damping config
+            if (hasAnyValidDampingConfig())
```
lolyu: 🟡 Two back-to-back O(ports) scans under `m_linkEventDampingMutex` on the hot path: `hasAnyValidDampingConfig()` here, then `processPendingDampingSync()` re-scans the same map immediately. The gate buys little — `flushPendingDampingNotifications()` already swaps the queue under lock and early-returns when empty, so it's nearly free when there's nothing pending. Making `flush` **unconditional** (drop the `hasAnyValidDampingConfig()` gate around the flush specifically) would both eliminate the redundant per-iteration scan AND fix the notification-stranding half of issue #1 — a stranded queue would always drain on the next loop iteration regardless of current config. The same applies to the TIMEOUT-path copy of this block above (~line 6821).

unittest/syncd/Makefile.am
```diff
@@ -28,7 +28,7 @@ tests_SOURCES = main.cpp \
 				TestVendorSai.cpp \
 				TestFlowDump.cpp
 
-tests_CXXFLAGS = $(DBGFLAGS) $(AM_CXXFLAGS) $(CXXFLAGS_COMMON)
+tests_CXXFLAGS = $(DBGFLAGS) $(AM_CXXFLAGS) $(CXXFLAGS_COMMON) -fno-access-control
```
lolyu: 🟡 Question: `-fno-access-control` disables access control for the entire test translation unit, letting any test reach private members directly (the new tests use `m_syncd->m_pendingNotifications`, `m_linkEventDampingMutex`, `m_portLinkEventDampingStates`). It works, but it couples tests to private layout and is a broad instrument compared with a `friend` test fixture or narrow test-only accessors. Is global `-fno-access-control` the preferred white-box pattern for this repo's syncd tests, or would a friend declaration be more contained? Not blocking — just flagging the precedent.

syncd/Syncd.cpp
```diff
@@ -1546,14 +1565,90 @@ void Syncd::processPendingDampingSync()
...
+        portVids.reserve(notifications.size() * 24);
+
+        for (size_t i = 0; i < notifications.size(); ++i)
+        {
+            if (i > 0) portVids += ", ";
+            portVids += sai_serialize_object_id(notifications[i].port_id);
+        }
+
+        SWSS_LOG_NOTICE("processPendingDampingSync: enqueuing %zu port state notifications for ports: %s",
+                        notifications.size(), portVids.c_str());
```
lolyu, line 1556: 🟢 Minor DRY: this port-VID logging loop is duplicated almost verbatim in `flushPendingDampingNotifications` (the `portVids` reserve + join). Could extract a small `serializePortVids(notifications)` helper used by both. Cosmetic.

lolyu (changes requested):
**Review:** https://github.com/sonic-net/sonic-sairedis/pull/1948

Reviewed the full `Syncd.cpp` (not just the diff) to trace the threading model. The core fix is **correct**: `m_notifications->send` now has exactly two call sites — `flushPendingDampingNotifications` and the shutdown path — both on the main thread, while the timer thread (`checkDampedPortsTimeout`) only sets `pending_state_sync` and writes STATE_DB counters under `m_linkEventDampingMutex`. That's precisely the `select failed: 2` race, correctly resolved. Nice work decoupling the send.

Requesting changes on one correctness issue (inline #1): because the in-memory `advertised_status` is committed at *enqueue* time but the send happens later, a batch that is dropped by the overflow guard — or stranded by the `hasAnyValidDampingConfig()` gate the Copilot reviewer already flagged — leaves the port's advertised/physical state permanently mismatched with **no retry**, since `pending_state_sync` was already cleared. Inline #2 proposes a single change (make `flush` unconditional) that resolves both the stranding and a redundant hot-path scan. #3 (`-fno-access-control`) is a question; #4 a minor DRY nit. The central fix is sound — these are about hardening the failure paths around it.

syncd/Syncd.cpp
sivat6: Removed hasAnyValidDampingConfig() checks.

yijingyan2:
Hi @sivat6, could you please help check the test failure? if it takes some time to fix, can I revert your change to unblock the submodule advance PR? Thanks!

syncd/Syncd.cpp
sivat6: Modified. The clearing of pending_state_sync and updating advertised_status is moved to flush notifications function.

sivat6:
> Hi @sivat6, could you please help check the test failure? if it takes some time to fix, can I revert your change to unblock the submodule advance PR? Thanks!

Hi, This PR is pending due to baseline issue. It will be merged once the baseline issue resolved.

prsunny:
@lolyu , could you rereview and merge?

@sivat6 , please change the title of PR to reflect what is done for the PR instead of "fix elastic failure"

sivat6:
> @lolyu , could you rereview and merge?
> 
> @sivat6 , please change the title of PR to reflect what is done for the PR instead of "fix elastic failure"

Changed the title.

lolyu (approved):
**Review:** https://github.com/sonic-net/sonic-sairedis/pull/1948 (re-review at `6fcb8c8`)

Thanks for the revision — verified all the changes against the full source, not just the diff. Everything I flagged is genuinely resolved, and the fixes are correct (not just relocated). Dismissing my earlier CHANGES_REQUESTED and approving.

**#1 (blocking correctness) — fixed correctly.** `processPendingDampingSync` now *only* enqueues; it no longer touches `pending_state_sync` or `advertised_status`. The in-memory commit moved into `flushPendingDampingNotifications` (lines 1616-1617) and — crucially — sits **inside the success path, after `m_notifications->send()` returns**. So a batch dropped by the overflow guard or failed in the `try/catch` leaves `pending_state_sync = true`, and the next pass re-enqueues it. The permanent advertised/physical mismatch (and the strictly-worse overflow-drop case) is now self-healing. Exactly the retry semantics I was asking for.

**#2 / the gating issue — fixed.** `processPendingDampingSync()` + `flushPendingDampingNotifications()` are now called unconditionally on both the TIMEOUT and unknown-selectable paths; the `hasAnyValidDampingConfig()` gate is gone, so a stranded queue always drains on the next loop iteration. This also removes the redundant double O(ports) scan.

**#4 (DRY) — fixed** via the new `serializePortVids()` helper, used by both paths. The Copilot swap-vs-copy nit is also addressed (`emplace()` + `swap` at enqueue).

**New surface checked:** moving the commit into the send path means `flush` now touches both `m_pendingNotificationsMutex` and `m_linkEventDampingMutex`. Confirmed there's **no lock-ordering inversion** — both `processPendingDampingSync` and `flush` acquire one mutex, release it, then acquire the other; neither ever holds both simultaneously. The `try/catch` around the send (one bad batch can't strand the rest) is a nice touch.

The core decoupling remains correct: `m_notifications->send` is main-thread-only, the timer thread only sets flags under the damping mutex. This is a sound fix for the `select failed: 2` race. LGTM.

sivat6:
> **Review:** #1948 (re-review at `6fcb8c8`)
> 
> Thanks for the revision — verified all the changes against the full source, not just the diff. Everything I flagged is genuinely resolved, and the fixes are correct (not just relocated). Dismissing my earlier CHANGES_REQUESTED and approving.
> 
> **#1 (blocking correctness) — fixed correctly.** `processPendingDampingSync` now _only_ enqueues; it no longer touches `pending_state_sync` or `advertised_status`. The in-memory commit moved into `flushPendingDampingNotifications` (lines 1616-1617) and — crucially — sits **inside the success path, after `m_notifications->send()` returns**. So a batch dropped by the overflow guard or failed in the `try/catch` leaves `pending_state_sync = true`, and the next pass re-enqueues it. The permanent advertised/physical mismatch (and the strictly-worse overflow-drop case) is now self-healing. Exactly the retry semantics I was asking for.
> 
> **#2 / the gating issue — fixed.** `processPendingDampingSync()` + `flushPendingDampingNotifications()` are now called unconditionally on both the TIMEOUT and unknown-selectable paths; the `hasAnyValidDampingConfig()` gate is gone, so a stranded queue always drains on the next loop iteration. This also removes the redundant double O(ports) scan.
> 
> **#4 (DRY) — fixed** via the new `serializePortVids()` helper, used by both paths. The Copilot swap-vs-copy nit is also addressed (`emplace()` + `swap` at enqueue).
> 
> **New surface checked:** moving the commit into the send path means `flush` now touches both `m_pendingNotificationsMutex` and `m_linkEventDampingMutex`. Confirmed there's **no lock-ordering inversion** — both `processPendingDampingSync` and `flush` acquire one mutex, release it, then acquire the other; neither ever holds both simultaneously. The `try/catch` around the send (one bad batch can't strand the rest) is a nice touch.
> 
> The core decoupling remains correct: `m_notifications->send` is main-thread-only, the timer thread only sets flags under the damping mutex. This is a sound fix for the `select failed: 2` race. LGTM.

Thanks [lolyu](https://github.com/lolyu)

prsunny:
Original PR - https://github.com/sonic-net/sonic-sairedis/pull/1906

PR #1950, opened 2026-06-23
[vpp] SAIVPP unit-test harness (docker-sai-test-vpp) + devdocs

nicholasching:
### Description of PR

Summary:
Adds the self-contained Docker harness (`docker-sai-test-vpp`) that runs the upstream OCP `sai_test` PTF suite against the real VPP SAI backend (`libsaivs`) in one container — VPP + `saiserver` + PTF + veth/AF_PACKET topology — and produces a per-test compatibility matrix. This is test scaffolding only (everything under `.azure-pipelines/docker-sai-test-vpp/`); it does not touch any production/runtime code, so it cannot affect existing `sonic-sairedis` functionality. Reviewers can start at `.azure-pipelines/docker-sai-test-vpp/README.md` and `run_test.sh`.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [x] Test improvement

There was no in-tree way to run the OCP `sai_test` suite against the VPP SAI backend and track per-test coverage. This harness provides that, plus the development docs and compatibility matrices used to drive the backend fixes.

#### How did you do it?
Added under `.azure-pipelines/docker-sai-test-vpp/`:
- `Dockerfile` and `run_test.sh` (orchestrates Redis → veth/PortChannel topology → VPP → saiserver → PTF), plus T0 config (`sai.profile`, `lanemap.ini`, `port-map.ini`, `ptf-port-map.ini`) and `gen_compatibility_matrix.py`.
- Config-signature grouping: because the VPP SAI backend builds the switch once per `saiserver` process, `run_test.sh` groups tests by their `setUp` config signature and restarts the backend per group, so one container can run any mix of tests.
- `ISOLATE_EACH_TEST` (default on): run each test in its own group with a fresh backend + own config to eliminate cross-test config-reuse contamination; set `0` for faster grouped runs.
- L3-over-LAG forwarding (harness portion): `run_test.sh` assigns the DUT-side connected IPs to each LAG (`be<N>`) and SVI (`bvi<vlan>`) so routed-to-LAG/SVI traffic forwards in the standalone PTF environment (the backend portion is in the backend-fixes PR).
- Debug hold to keep VPP/saiserver/veths alive after a `--debug` run for `vppctl` inspection.
- `devdocs/`, `demodocs/`, and `results/` compatibility matrices; `.gitignore` for generated artifacts.

Running the harness against `sai_route_test`/`sai_rif_test`/`sai_neighbor_test`/`sai_ecmp_test` produces a compatibility matrix (current state: 34 PASS, captured in `devdocs/`/`results/`). The harness is self-contained and does not build or run any production code.

VPP platform test infrastructure only; isolated under `.azure-pipelines/docker-sai-test-vpp/`. No impact on the `sonic-sairedis` build or runtime behavior.

### Documentation
This PR adds the harness `README.md` plus rolling/dated progress notes and deep-dive debug logs under `.azure-pipelines/docker-sai-test-vpp/devdocs/` and demo docs under `demodocs/`.


nicholasching:
It appears the checks are failing due to an upstream issue which should be resolved when #4696 is merged into sonic-swss.

.azure-pipelines/docker-sai-test-vpp/demodocs/demo-6-16.md
```diff
@@ -0,0 +1,212 @@
+# VPP SAI Unit-Test Framework — Stakeholder Demo
```
yue-fred-gao: I feel the docs are too much, especially the progress mds. People may get lost if there is too much information. Let's consolidate the information with README.md.

.azure-pipelines/docker-sai-test-vpp/results/compatibility-matrix-6-18.md
```diff
@@ -0,0 +1,147 @@
+# VPP SAI Compatibility Matrix
```
yue-fred-gao: Keep one with the latest result

.azure-pipelines/docker-sai-test-vpp/Dockerfile
```diff
@@ -0,0 +1,124 @@
+FROM debian:bookworm
```
yue-fred-gao: can we move to trixie? Latest sonic-vpp has moved to the new debian. I am not sure if there is any compatibility issues with vpp debian packages

.azure-pipelines/docker-sai-test-vpp/README.md
```diff
@@ -0,0 +1,245 @@
...
+      target/debs/bookworm/libsaivs_*.deb     target/debs/bookworm/libsaivs-dev_*.deb
+
+# Build libsaivs / libsairedis
+NOTRIXIE=1 make target/debs/bookworm/libsairedis_1.0.0_amd64.deb
+
+# Build saiserver + saithrift client (if those changed)
+NOTRIXIE=1 BLDENV=bookworm make -f Makefile.work target/debs/bookworm/libsaithrift-dev_0.9.4_amd64.deb
+
+# Stage the fresh packages into the harness build context
+cp target/debs/bookworm/libsaivs_*.deb     target/debs/bookworm/libsaivs-dev_*.deb \
```
yue-fred-gao: I think we have 3 use cases.
1. run UT locally in workspace during development. This may include changes in sairedis/vslib/vpp, new test cases in sairedis/SAI and new vpp deb packages. We can assume the workspace is sonic-buildimage. The dependent packages can be downloaded from target/debs/trxie.
2. run UT in sonic-sairedis PR pipeline. This may include changes in sairedis/vslib/vpp and new test cases. The PR pipeline build stage will build libsairedis and libsaivs. libswsscommon can be downloaded from other pipeline artifacts. Are all the dependent debs available either locally or from other pipeline?
3. run UT in sonic-platform-vpp PR pipeline. The pipeline build stage will build new vpp packages. Can we download docker-sai-test-vpp docker image from sairedis pipeline artifact? And replace vpp packages with the locally built ones?

.azure-pipelines/docker-sai-test-vpp/README.md
```diff
@@ -0,0 +1,245 @@
...
+   target/debs/bookworm/libsairedis_*.deb  target/debs/bookworm/libsairedis-dev_*.deb \
+   target/debs/bookworm/saiserver_*.deb    target/debs/bookworm/python-saithrift_*.deb \
+   src/sonic-sairedis/.azure-pipelines/docker-sai-test-vpp/debs/
+```
+> Behind a corporate proxy, prefix `make` with your Docker build credentials,
+> e.g. `DOCKER_CONFIG=<dir>`. VPP `.deb`s come from the `sonic-platform-vpp`
+> pipeline; drop new ones into `debs/` to upgrade VPP.
+
+### Building the image
```
yue-fred-gao: Can we add a script to automate the build process?

.azure-pipelines/docker-sai-test-vpp/README.md
```diff
@@ -0,0 +1,245 @@
...
+### Environment knobs
+| Variable | Default | Meaning |
+|---|---|---|
+| `PORT_COUNT` | 32 | number of `OEthernetX`/`OEthX_peer` veth pairs |
+| `COMMON_CONFIGURED_REUSE` | 1 | 1 = config-signature grouping + reuse; 0 = legacy single ptf invocation |
+| `KEEP_VETHS_UP_SECONDS` | 120 | how long the per-group watchdog keeps VPP host-interfaces/veths up |
+| `TEST_FILTER` | — | alternative way to pass a single selector via env |
+
+### Logs inside the container
+- `/var/log/saiserver.log` — high-level SAI RPC trace (saiserver runs with SWSS
```
yue-fred-gao: Is this still available after you remove SWSS logger from saiserver? It is important we keep SAI logs. Can we redefine SWSS_LOG_INFO etc with simple file logger?

.azure-pipelines/docker-sai-test-vpp/run_test.sh
```diff
@@ -0,0 +1,1190 @@
...
+    for ((lag_index = 0; lag_index < LAG_COUNT; lag_index++)); do
+        local pc_if="PortChannel${lag_index}"
+        if ip link show "$pc_if" >/dev/null 2>&1; then
+            ip link delete "$pc_if" >/dev/null 2>&1
+        fi
+    done
+    set -e
+}
+
+create_portchannels()
```
yue-fred-gao: Is this test case specific? Could this be moved to setup()/teardown() of the test case?

.azure-pipelines/docker-sai-test-vpp/run_test.sh
```diff
@@ -0,0 +1,1190 @@
...
+# LAG: the IP goes on the kernel "be<N>" netdev (the BondEthernet LCP host-if),
+# which linux_nl mirrors onto BondEthernet<N>. SVI: a BVI has no LCP host-if, so
+# the IP is set directly in VPP on "bvi<vlan_id>" via vppctl.
+#
+# Unlike keep_vpp_veths_up, this loops for the whole backend lifetime (until
+# stop_backend/cleanup kills it) rather than a fixed window: a single config group
+# can run many tests for well over KEEP_VETHS_UP_SECONDS, and later tests (e.g. the
+# SVI MAC-learning/aging suite) churn interface/FDB state, so the connected IPs
+# must be continuously re-asserted, not just during the initial config build.
+keep_lag_rif_ips_up()
```
yue-fred-gao: I don't quite understand why this is needed. when do we need IP on the BE interface in the kernel?

.azure-pipelines/docker-sai-test-vpp/run_test.sh
```diff
@@ -0,0 +1,1190 @@
...
+    for ((port_index = 0; port_index < PORT_COUNT; port_index++)); do
+        echo "Ethernet$((port_index * 4)) host-OEthernet${port_index}" >> "$SONIC_VPP_IFMAP"
+    done
+}
+
+generate_vpp_config()
+{
+    VPP_CONF="$(mktemp /tmp/vpp-sai-test.XXXXXX.conf)"
+
+    cat > "$VPP_CONF" <<EOF
```
yue-fred-gao: Can we move this to a template, from which you can generate the .conf? It is more friendly to separate data from code

.azure-pipelines/docker-sai-test-vpp/run_test.sh
```diff
@@ -0,0 +1,1190 @@
...
+
+    log "Creating VPP host interfaces"
+
+    for ((port_index = 0; port_index < PORT_COUNT; port_index++)); do
+        if [[ "$batch_count" -eq 0 ]]; then
+            batch_cmds="$(mktemp /tmp/vpp-sai-test-create.XXXXXX.cmds)"
+            batch_start="$port_index"
+        fi
+
+        echo "create host-interface name OEthernet${port_index}" >> "$batch_cmds"
```
yue-fred-gao: I think from vpp startup.conf, you can specify the initial config. vpp will execute the commands there after boot up. You can create a Jinja2 template with initial config then generate the initial config from the template.

.azure-pipelines/docker-sai-test-vpp/run_test.sh
```diff
@@ -0,0 +1,1190 @@
...
+
+set_vpp_rx_mode_interrupt()
+{
+    local rx_mode_cmds
+
+    log "Setting VPP AF_PACKET RX mode to interrupt"
+    rx_mode_cmds="$(mktemp /tmp/vpp-sai-test-rxmode.XXXXXX.cmds)"
+
+    for ((port_index = 0; port_index < PORT_COUNT; port_index++)); do
+        echo "set interface rx-mode host-OEthernet${port_index} interrupt" >> "$rx_mode_cmds"
```
yue-fred-gao: move to initial config template?

.azure-pipelines/docker-sai-test-vpp/run_test.sh
```diff
@@ -0,0 +1,1190 @@
...
+    terminate_process saiserver "$SAISERVER_PID"; SAISERVER_PID=""
+    terminate_process vpp "$VPP_PID"; VPP_PID=""
+    terminate_process redis "$REDIS_PID"; REDIS_PID=""
+    # Drop persisted SAI object IDs and the link-up marker so the next group's
+    # first test rebuilds (and re-persists) its own config and re-asserts veths.
+    rm -rf /tmp/sai_model 2>/dev/null || true
+    rm -f "$LINKS_UP_MARKER" 2>/dev/null || true
+}
+
+# Emit a run plan: tab-separated "<group_id>\t<module.Class>\t<signature>" lines,
```
yue-fred-gao, line 829: Ideally each test case should restore the tb to initial state with teardown. Does this not work?

.azure-pipelines/docker-sai-test-vpp/demodocs/demo-6-16.md
nicholasching: I agree, was mainly maintaining those documents for context, no need to include as much depth for documentation. All done, removed and integrated into README now.

.azure-pipelines/docker-sai-test-vpp/results/compatibility-matrix-6-18.md
nicholasching: Since only one is left, I decided to remove `/results/` directory entirely and include the supported tests in README instead.

.azure-pipelines/docker-sai-test-vpp/README.md
nicholasching: Good idea! I added in commit [cd93fc5](https://github.com/sonic-net/sonic-sairedis/pull/1950/commits/cd93fc552354899583c34f66905eb66cfbff4dc1); can now run script to automate staging/build. It will also auto-build sonic-sairedis if its `.debs` are missing and show more informative error messages than `docker build` before if any third-party `.debs` dependencies (eg. vpp, libswsscommon, etc.) are missing.

.azure-pipelines/docker-sai-test-vpp/README.md
nicholasching: With SWSS logger removed, logs will go back to /var/log/syslog by default. I just pushed new change [b9319d8](https://github.com/sonic-net/sonic-sairedis/pull/1950/commits/b9319d86c8a7dcaf04f42d19c386ee5624fca6b6) adding a preload shim to continue routing logs to `var/log/saiserver.log`.

.azure-pipelines/docker-sai-test-vpp/README.md
nicholasching: Use case 1 is fully implemented and automated by `build_harness.sh`. I plan to implement use cases 2 and 3 during phase 3 (will be in a separate PR) and will mimic the build pattern used in `.azure-pipelines/build-docker-sonic-vs-template.yml`. For now, I have documented these use cases in `README.md` as under development.

> Are all the dependent debs available either locally or from other pipeline?

Yes, all `.deb` packages are available from the existing pipelines. The `sonic-sairedis` PR pipeline's build stage will compile fresh `libsaivs`,`libsairedis`, `libsaimetadata`, `saiserver/saiserverv2`, and `python-saithrift/python-saithriftv2` artifacts and `libswsscommon`, `libyang`, and `vpp` artifacts can be retrieved from their upstream pipelines.

> Can we download docker-sai-test-vpp docker image from sairedis pipeline artifact and replace vpp packages with the locally built ones?

Yes, after the `docker-sai-test-vpp` image is built and published as a pipeline artifact in `sonic-sairedis` (use case 2 implementation), we intend to have `sonic-platform-vpp` pull the image and layer VPP changes on top.

yue-fred-gao: sounds good

.azure-pipelines/docker-sai-test-vpp/run_test.sh
nicholasching: `create_portchannels()` is not test case specific; I do not think we can move it to the test case's`setUp()`/`tearDown()`.

VPP backend, `libsaivs`, relies on `PortChannel<N>` netdevs at startup to derive bond IDs and configure traffic redirection (via `tc-mirred` redirects from the host tap to the `PortChannel` interface). Because In production SONiC, `teamd` creates these netdevs in the kernel. 

Since our standalone UT harness does not run `teamd`, we use `create_portchannels()` to emulate these netdevs. This has to happen before setup() of any test case because the interfaces need to exist in the kernel namespace before `saiserver` starts up and initializes `livsaivs`, after which the test cases can run.

.azure-pipelines/docker-sai-test-vpp/run_test.sh
nicholasching: In production SONiC, `IntfMgr` configures IP addresses on `PortChannel` and SVI interfaces. Netlink listener, `linux_nl`, mirrors these IP assignments into VPP's FIB.

Since our standalone UT harness does not run the SONiC control plane, where `IntfMgr` lives, the kernel TAP interfaces are created without any IP addresses and VPP cannot build the connected route or resolve ARP/NDP neighbours on that subnet and fails L3-over-LAG forwarding tests.

Because these `be<N>` TAP interfaces are created lazily by the VPP SAI backend only when the test's `setUp()` requests a LAG RIF, `run_test.sh` does not know when the python test will create RIF.  `keep_lag_rif_ips_up()` starts a background watchdog to detect when they appear and assign a connected IP to mimic `IntfMgr`'s behaviour and makes sure VPP has connected routes for routing tests.

.azure-pipelines/docker-sai-test-vpp/run_test.sh
nicholasching: Because`libsaivs` is designed to initialize the switch and host-interfaces only once per process lifetime, `teardown()` will not fully restore the tb to initial state.

`teardown()` in `sai_test` will set `common_configured=false` and delete SAI objects, but `saiserver` will fail to rebind w/VPP host-interfaces and crash on full teardown/recreation within the same `saiserver` process due to `libsaivs` limitation.

To make sure each test starts w/same initial state, we perform full restart on backend processes (redis, vpp, and saiserver) and wipe `/tmp/sai_model` between tests.

.azure-pipelines/docker-sai-test-vpp/run_test.sh
nicholasching: I agree. I moved the startup configuration from `run_test.sh` to the dedicated template file in `vpp_startup.conf.template`.

.azure-pipelines/docker-sai-test-vpp/run_test.sh
nicholasching: All done! I refactored the initialization sequence to perform interface setup atomically at boot time using the `unix { exec <path> }` directive in `startup.conf` with a dynamically generated VPP CLI script rather than running post-boot `vppctl` commands.

.azure-pipelines/docker-sai-test-vpp/run_test.sh
nicholasching: All done; please see thread above.

.azure-pipelines/docker-sai-test-vpp/Dockerfile
nicholasching: All done! I just validated the migration to Trixie.

There were several platform compatibility issues with Trixie's newer environment; I implemented the fixes below:
1. **Python 3.12+ Compatibility:** Trixie uses Python 3.13, which removed the legacy `imp` module. This broke the PTF test runner. Resolved this by adding an `importlib` compatibility fallback in `SAI/test/ptf/ptf`.
2. **Pip Package Conflicts:** The PTF installation failed due to conflicts with the Debian-managed `six` package. Resolved this by adding `--ignore-installed` to the `pip install /opt/ptf` command in the Dockerfile.
3. **`libyang` Dependency on `libpcre3`:** The `libyang` v1 package used by `libswsscommon` depends on `libpcre3`, which is not natively installable in Trixie. We updated `build_harness.sh` to explicitly stage `libpcre3_*.deb` from `target/debs/trixie/` and install it during the Docker build.

After the migration, I ran the full test in isolation mode again and matched the coverage from our Bookworm baseline (34 passing) to verify the migration to Trixie was successful.

nicholasching:
Hi @yue-fred-gao, thank you for the PR review. I just finished code edits to address all comments and requested 2nd round of review. Thanks!

nicholasching:
Hi Fred (@yue-fred-gao), the `t1-lag-vpp` sanity test has passed on the associated `sonic-buildimage` PR: https://github.com/sonic-net/sonic-buildimage/pull/28145

Github Run: https://github.com/sonic-net/sonic-buildimage/runs/84906281352
Azure Build: https://dev.azure.com/mssonic/build/_build/results?buildId=1154935&view=logs&s=859b8d9a-8fd6-5a5c-6f5e-f84f1990894e&j=314bc92a-d36a-5eac-5c45-43e8f1288876

.azure-pipelines/docker-sai-test-vpp/run_test.sh
yue-fred-gao: The portchannel netdev is needed when vpp sai receives SAI call to create portchannel. Because the sai call is from a test case, I feel these netdev can be created in the setup

.azure-pipelines/docker-sai-test-vpp/run_test.sh
yue-fred-gao: I see. Can this be done in the setup? This is needed by specific test cases, shouldn't in run_test

.azure-pipelines/docker-sai-test-vpp/run_test.sh
yue-fred-gao: I see. It sounds like saivpp doesn't handle host-interface removal properly. What is implemented here is a workaround. Please investigate if we can fix the bug.

.azure-pipelines/docker-sai-test-vpp/Dockerfile
yue-fred-gao: great

.azure-pipelines/docker-sai-test-vpp/run_test.sh
yue-fred-gao: great!

.azure-pipelines/docker-sai-test-vpp/run_test.sh
nicholasching: Thanks Fred; I agree, it can be done that way! I removed `create_portchannels()`; instead, the `PortChannel<N>` netdev is now created in `LagConfiger.create_lag()` right before `sai_thrift_create_lag`, so it exists right before when it is needed by the VPP SAI backend to derive the bond id. This is done through a harness-side helper, `vpp_ut_support.ensure_portchannel` wired into the config code and gated by `SAI_VPP_UT_HARNESS`, so upstream `sai_test` behavior is unchanged when run outside this harness.

.azure-pipelines/docker-sai-test-vpp/run_test.sh
nicholasching: Yes. Along with the change in the last comment, I removed the `keep_lag_rif_ips_up` watchdog from `run_test.sh`. The DUT-side connected IPs are now assigned in `RouteConfiger.create_router_interface()` at the moment a LAG or VLAN(SVI) RIF is created (vpp_ut_support.assign_lag_rif_ips / assign_svi_rif_ips), so only the test cases that actually build those RIFs trigger it, and only for the interface being configured.

.azure-pipelines/docker-sai-test-vpp/Dockerfile
```diff
@@ -0,0 +1,141 @@
...
+
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+
+# Build from the sonic-sairedis repository root:
+# docker build -f .azure-pipelines/docker-sai-test-vpp/Dockerfile -t docker-sai-test-vpp .
+COPY .azure-pipelines/docker-sai-test-vpp /opt/docker-sai-test-vpp
+COPY SAI/test/ptf /opt/ptf
+COPY SAI/test/sai_test /sai_test
+COPY .azure-pipelines/docker-sai-test-vpp/vpp_ut_support.py /sai_test/config/vpp_ut_support.py
+COPY .azure-pipelines/docker-sai-test-vpp/harness_sai_test/overrides/lag_configer.py /sai_test/config/lag_configer.py
```
yue-fred-gao: This creates maintenance debt. If the one in sai_test is modified and creates a new dependency from certain test case, the "overrides" may become incompatible. One option is to modify sai_test lag_configer.py. It can be based on some env or config option like SIMULATE_SONIC. It is not necessary to be vpp specific.

.azure-pipelines/docker-sai-test-vpp/run_test.sh
yue-fred-gao: This is good. I am only concerned with the way harness overrides dropped in.

.azure-pipelines/docker-sai-test-vpp/run_test.sh
nicholasching: Hi @yue-fred-gao, I found the host-interface removal bug, but it was not enough to remove the workaround; I found some other issues.

In terms of the host-interface bug,`sai_remove_hostif` (vs_remove_hostif_tap_interface) only cleared its software maps and never tore down the VPP linux-cp pair or the enabled IPv6 on the hwif. As a result, re-creating a host interface in the same `saiserver` process failed.`config_lcp_hostif` and `sw_interface_ip6_enable_disable` returned `VALUE_EXIST(-81)`, the ip6-enable failed, and `create_hostif` returned a null OID and the RPC socket dropped. I implemented a fix for this in the backend fixes PR (#1952). Remove now disables IPv6 and deletes the linux-cp pair, and create is idempotent to `VALUE_EXIST`. With this fix, I have verified a host interface can now be removed and recreated within the same `saiserver` process.

Unfortunately, this was not the only issue blocking a per-test teardown in place of the backend restart between tests. When the whole T0 config is rebuilt in the same process, the same "already exists" issue appears in other object types (VLAN, bridge port next, and the L3 objects, RIF/nexthop/neighbor/route, have no teardown helpers). As a result, the teardown behaviour of each object will need to be investigated/resolved individually before replacing the backend restart with teardown between tests.

Since this fix would require large modifications to the VPP SAI backend (to implement per-object-type remove teardown + create idempotency) and the OCP `sai_test` framework (to add teardown helpers to the L3 objects), would it be better to merge the current PRs and defer this task to a future PR? Once we implement this fix in the future, it will be easy to adapt our UT harness.

yue-fred-gao: sure. Please add a comment so we remember this.

nicholasching: Thank you! Added comment in [4da0942](https://github.com/sonic-net/sonic-sairedis/pull/1950/commits/4da09423a2b63f9f79b98a51ba1049b13800d93e).

.azure-pipelines/docker-sai-test-vpp/Dockerfile
nicholasching: This reply consolidates this thread with https://github.com/sonic-net/sonic-sairedis/pull/1950#discussion_r3521912941

I agree, this creates two sources of truth. I've moved the PortChannel and LAG/SVI-IP setup out of the harness overrides and into `sai_test`, gated by a SIMULATE_SONIC option, disabled by default. As a result, the harness no longer needs to overlay `sai_test`, and `run_test.sh` just configures `SIMULATE_SONIC=1` instead. 

The `sai_test` change is a separate SAI PR (https://github.com/opencomputeproject/SAI/pull/2316). The changes in SAI were made to be platform agnostic to respect the shared framework, therefore, VPP operations are passed in from the harness as a command template rather than hardcoding `vppctl` in `sai_test`.

I've re-validated the full integration of the changes made across (#1950, #1951, #1952, and the 4 SAI PRs) and verified there are no regressions against the 34-PASS baseline.

.azure-pipelines/docker-sai-test-vpp/run_test.sh
```diff
@@ -0,0 +1,1018 @@
...
+    PTF_ARGS+=(--xunit --xunit-dir "$xunit_dir")
+
+    # The T0 sanity flood test sends one unknown-unicast frame and expects it to
+    # be flooded to every other member port of the VLAN. PTF's flood verifier
+    # (verify_each_packet_on_multiple_port_lists) consumes only one copy from the
+    # expected port list and then asserts via verify_no_other_packets() that no
+    # further packets remain - which is impossible for a real flood that delivers
+    # a copy to every member port. --relax makes verify_no_other_packets() a
+    # no-op, which is the intended/standard mode for flooding tests.
+    PTF_ARGS+=(--relax)
```
dypet: PTF_ARGS+=(--relax) is added unconditionally to all targets, not just the T0 cases mentioned above. Please scope --relax to only the flood test cases rather than the entire run.

.azure-pipelines/docker-sai-test-vpp/run_test.sh
```diff
@@ -0,0 +1,1018 @@
...
+                ;;
+        esac
+    done
+    test_rc="${PIPESTATUS[0]}"
+    set -e
+
+    # Collect this invocation's JUnit XML into the shared results dir.
+    if [[ -n "$TEST_RESULTS_DIR" ]]; then
+        mkdir -p "$TEST_RESULTS_DIR"
+        cp -f "$xunit_dir"/*.xml "$TEST_RESULTS_DIR"/ 2>/dev/null || true
```
dypet: With ISOLATE_EACH_TEST=1 (default) or multi-group runs, two invocations that emit the same TEST-<module>.xml filename will clobber each other, and gen_compatibility_matrix.py (which globs TEST-*.xml) would then undercount. Suggest namespacing the copied files per invocation (e.g. prefix with group/idx or test target) so no results are lost.

dypet (changes requested):

.azure-pipelines/docker-sai-test-vpp/run_test.sh
nicholasching: Good catch, thank you @dypet. `--relax` disabled PTF's `verify_no_other_packets()` check, and applying it globally masked unexpected forwarding in non-flood tests. I changed the harness to add it only for an explicit allowlist of positive flood classes that use `verify_each_packet_on_multiple_port_lists()`. All other selectors now run with strict unexpected-packet verification. The change is in `67b5980b`.

After more testing, we found this incorrect implementation to be hiding some background IPv6 lifecycle/control packets that strict PTF correctly treated as unexpected traffic, bringing down # of successful tests from 34 -> 19. The README.md at this PR stage has been updated to reflect that.

Upon implementing an opt-in, exact PTF filter for simulated-router Router Solicitation and MLDv2 startup frames, plus Linux IPv6 setup ordering that prevents DAD/router-discovery packets, the # of successful tests was restored 19 -> 30. These fixes involve both `SAI` and `sonic-sairedis` components. To prevent re-opening `vpp_backend_fixes` these fixes will be bundled with the others in the Phase 3 PRs. The associated SAI fix is here: https://github.com/opencomputeproject/SAI/pull/2321

.azure-pipelines/docker-sai-test-vpp/run_test.sh
nicholasching: Thank you, I agree. I added deterministic invocation namespaces when collecting each report, for example `TEST-group-3-test-8-sai_route_test.xml`, while retaining the `TEST-` prefix expected by the matrix generator. The README also clears stale XML before a new run. This change is in `2ce71f7b`. A collision test was performed, confirming both same-named source reports and testcase rows were retained.

dypet (dismissed):

vslib/vpp/SwitchVppRif.cpp
```diff
@@ -339,39 +341,160 @@ void SwitchVpp::vpp_intf_remove_prefix_entry (const std::string& intf_name)
...
+    else
     {
+        sai_attribute_t attr = {};
+        attr.id = SAI_PORT_ATTR_HW_LANE_LIST;
+        attr.value.u32list.count = sizeof(lanes) / sizeof(lanes[0]);
+        attr.value.u32list.list = lanes;
+
+        if (get(SAI_OBJECT_TYPE_PORT, port_id, 1, &attr) != SAI_STATUS_SUCCESS)
+        {
+            SWSS_LOG_DEBUG("lane list unavailable for port %s",
```
yue-fred-gao: LANE_LIST is mandatory in PORT. This should be an error message. Same for all the DEBUG message in the returning false case in this function.

vslib/vpp/SwitchVppRif.cpp
```diff
@@ -339,39 +341,160 @@ void SwitchVpp::vpp_intf_remove_prefix_entry (const std::string& intf_name)
...
+
+    if (vpp_get_swif_idx_by_name(mapped_hwifname) == static_cast<uint32_t>(-1))
+    {
+        SWSS_LOG_DEBUG("VPP interface %s is not present for port %s",
+                mapped_hwifname, port_name.c_str());
+        return false;
+    }
+
+    if_name = mapped_hwifname;
+    SWSS_LOG_DEBUG("resolved port %s lane set to %s/%s",
```
yue-fred-gao: use INFO. DEBUG level is overwhelmed in syncd due to SWSS_LOG_ENTER in every function.

nicholasching:
Hi Fred (@yue-fred-gao), good news, the t1-lag-vpp sanity test has passed on the associated sonic-buildimage PR: https://github.com/sonic-net/sonic-buildimage/pull/28145

Github Run: https://github.com/sonic-net/sonic-buildimage/runs/93065072611
Azure Build: https://dev.azure.com/mssonic/build/_build/results?buildId=1187086&view=logs&s=859b8d9a-8fd6-5a5c-6f5e-f84f1990894e&j=314bc92a-d36a-5eac-5c45-43e8f1288876

yue-fred-gao (approved):

PR #1953, opened 2026-06-24
feat: handle LAG egress disable updates

cyw233:
### Description of PR

Summary:
Add VPP vslib handling for `SAI_LAG_MEMBER_ATTR_EGRESS_DISABLE` so LAG members can be dynamically detached from, or re-added to, the VPP bond when egress is disabled or enabled.

This also defers VPP bond member creation when a LAG member is created with egress disabled. LCP/tap creation remains delayed until the first enabled member is added, which avoids creating an empty bond tap with a stale generated MAC.

Fixes # (issue) https://github.com/sonic-net/sonic-buildimage/issues/25771

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [x] Test improvement

VPP vslib already creates and removes LAG members in the VPP bond, but it did not handle updates to `SAI_LAG_MEMBER_ATTR_EGRESS_DISABLE`. As a result, a member whose egress state changed after creation would not be reflected in VPP bond membership.

##### Work item tracking

#### How did you do it?

- Route `SAI_OBJECT_TYPE_LAG_MEMBER` set operations through a new `setLagMember` handler.
- Handle `SAI_LAG_MEMBER_ATTR_EGRESS_DISABLE` transitions by removing the member from the VPP bond when egress is disabled and adding it back when egress is enabled.
- Treat missing `EGRESS_DISABLE` state as the SAI default value, `false`, when deciding whether a VPP bond membership update is needed.
- Skip VPP bond member creation when a LAG member is initially created with egress disabled.
- Extract LCP/tap setup into `vpp_ensure_lag_lcp()` so LCP creation is still performed when the first enabled member is added.

Added VPP-enabled vslib unit coverage for `getLagMemberEgressDisableAction()` to verify the expected no-op, detach, and add decisions for current and requested egress-disable states.

Also ran the `pc/test_lag_member_forwarding.py` sonic-mgmt test to validate this feature (PR https://github.com/sonic-net/sonic-mgmt/pull/25135 to enable it):

image

This change is specific to the VPP vslib path.

### Documentation

HLD: https://github.com/sonic-net/SONiC/pull/2401

### DIFF
--- a/vslib/vpp/SwitchVpp.h
+++ b/vslib/vpp/SwitchVpp.h
@@ -1167,8 +1167,10 @@ namespace saivs
             std::shared_ptr<std::thread> m_vpp_thread;
 
         private: // VPP
+	    // m_lag_bond_map and m_egress_disabled_lag_member_ports are only accessed on
+	    // the LAG create/set/remove path, which the VS layer serializes through a
+	    // single queue, so they require no additional locking.
 	    std::map<sai_object_id_t, platform_bond_info_t> m_lag_bond_map;
-	    std::mutex LagMapMutex;
 	    std::set<sai_object_id_t> m_egress_disabled_lag_member_ports;
 
             static int currentMaxInstance;
--- a/vslib/vpp/SwitchVpp.h
+++ b/vslib/vpp/SwitchVpp.h
@@ -390,6 +390,21 @@ namespace saivs
             virtual sai_status_t setLag(
                     _In_ sai_object_id_t lagId,
                     _In_ const sai_attribute_t* attr);
+            virtual sai_status_t setLagMember(
+                    _In_ sai_object_id_t lagMemberId,
+                    _In_ const sai_attribute_t* attr);
+
+            enum class LagMemberEgressDisableAction
+            {
+                NONE,
+                DISABLE,
+                ENABLE,
+            };
+
+            static LagMemberEgressDisableAction getLagMemberEgressDisableAction(
+                    _In_ bool requested_egress_disable,
+                    _In_ bool current_attr_found,
+                    _In_ bool current_egress_disable);
 
             sai_status_t vpp_create_lag(
                     _In_ sai_object_id_t lag_id,
@@ -411,6 +426,17 @@ namespace saivs
                     _In_ sai_object_id_t lag_member_oid);
 	    sai_status_t vpp_remove_lag_member(
                     _In_ sai_object_id_t lag_member_oid);
+	    sai_status_t vpp_ensure_lag_lcp(
+                    _In_ sai_object_id_t lag_oid);
+	    sai_status_t vpp_set_lag_member_egress_disable(
+                    _In_ sai_object_id_t lag_member_oid,
+                    _In_ bool egress_disable);
+	    sai_status_t get_lag_member_port(
+                    _In_ sai_object_id_t lag_member_oid,
+                    _Out_ sai_object_id_t& port_oid);
+	    sai_status_t get_lag_member_bond_index(
+                    _In_ sai_object_id_t lag_member_oid,
+                    _Out_ uint32_t& bond_sw_if_index);
 
             /* FDB Entry and Flush SAI Objects */
             sai_status_t FdbEntryadd(
@@ -1143,6 +1169,7 @@ namespace saivs
         private: // VPP
 	    std::map<sai_object_id_t, platform_bond_info_t> m_lag_bond_map;
 	    std::mutex LagMapMutex;
+	    std::set<sai_object_id_t> m_egress_disabled_lag_member_ports;
 
             static int currentMaxInstance;
 
--- a/vslib/vpp/SwitchVppFdb.cpp
+++ b/vslib/vpp/SwitchVppFdb.cpp
@@ -772,7 +772,19 @@ sai_status_t SwitchVpp::createLagMember(
     auto sid = sai_serialize_object_id(object_id);
 
     CHECK_STATUS(create_internal(SAI_OBJECT_TYPE_LAG_MEMBER, sid, switch_id, attr_count, attr_list));
-    return vpp_create_lag_member(attr_count, attr_list);
+
+    // Add the member to the VPP bond first so the bond (and its LCP/tap) inherits
+    // the member MAC, then detach it from the bond if it is created egress-disabled.
+    CHECK_STATUS(vpp_create_lag_member(attr_count, attr_list));
+
+    auto egress_disable = sai_metadata_get_attr_by_id(SAI_LAG_MEMBER_ATTR_EGRESS_DISABLE, attr_count, attr_list);
+    if (egress_disable != NULL && egress_disable->value.booldata)
+    {
+        SWSS_LOG_NOTICE("LAG member %s created with egress disabled, detach from VPP bond", sid.c_str());
+        CHECK_STATUS(vpp_set_lag_member_egress_disable(object_id, true));
+    }
+
+    return SAI_STATUS_SUCCESS;
 }
 
 sai_status_t SwitchVpp::vpp_create_lag_member(
@@ -785,7 +797,6 @@ sai_status_t SwitchVpp::vpp_create_lag_member(
     bool is_passive = false;
     int ret;
     uint32_t bond_if_idx;
-    uint32_t bond_id;
     sai_object_id_t lag_oid, lag_port_oid;
 
     //Get the bond interface index from attr SAI_LAG_MEMBER_ATTR_LAG_ID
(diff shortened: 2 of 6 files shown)


vslib/vpp/SwitchVpp.h
```diff
@@ -978,6 +1012,7 @@ namespace saivs
         private: // VPP
 	    std::map<sai_object_id_t, platform_bond_info_t> m_lag_bond_map;
 	    std::mutex LagMapMutex;
+	    std::set<sai_object_id_t> m_egress_disabled_lag_member_ports;
```
lolyu, line 1174: 🟠 `LagMapMutex` (declared just above this set, line ~1014) is **never locked anywhere** — verified by grepping both files touched here: zero `lock_guard`/`unique_lock` on it. So both `m_lag_bond_map` and this new `m_egress_disabled_lag_member_ports` are accessed unsynchronized. This set is written in `vpp_set_lag_member_egress_disable`/`vpp_clear_lag_member_egress_disable` (LAG-member set path) and read in `UpdatePort` via `.count()` (port set path). If syncd/VS ever dispatches port and LAG-member operations on different threads, that's a data race on a `std::set`. It may be safe today if VS processes everything single-threaded through its queue — but then `LagMapMutex` is dead code that *looks* like it provides protection. Suggest either wiring the mutex around both structures, or removing it and adding a comment documenting the single-threaded assumption.

vslib/vpp/SwitchVppFdb.cpp
```diff
@@ -766,7 +766,19 @@ sai_status_t SwitchVpp::createLagMember(
     auto sid = sai_serialize_object_id(object_id);
 
     CHECK_STATUS(create_internal(SAI_OBJECT_TYPE_LAG_MEMBER, sid, switch_id, attr_count, attr_list));
-    return vpp_create_lag_member(attr_count, attr_list);
+
+    // Add the member to the VPP bond first so the bond (and its LCP/tap) inherits
```
lolyu, line 776: 🟡 This inverts the deferred-LCP invariant the earlier design relied on. Previously LCP/tap creation was deferred until the first *enabled* member, with the stated reason that creating the tap on an empty bond leaves it with a stale generated MAC. Now an egress-disabled first member is added to the bond (triggering `vpp_ensure_lag_lcp`) and then immediately set admin-down — so the tap is created while its only member is down. The comment says adding the member first lets the bond/LCP inherit the member MAC, which sounds correct, but it's the opposite ordering from before. Can you confirm the tap MAC is still correctly inherited when the first (and only) member is added-then-immediately-downed? If yes this is cleaner than the defer approach; just want the MAC concern explicitly retested for that case.

vslib/vpp/SwitchVppFdb.cpp
```diff
@@ -843,37 +854,287 @@ sai_status_t SwitchVpp::vpp_create_lag_member(
...
+    }
+
+    auto sid = sai_serialize_object_id(lagMemberId);
+
+    auto &objectHash = m_objectHash.at(SAI_OBJECT_TYPE_LAG_MEMBER);
+    if (objectHash.find(sid) == objectHash.end())
+    {
+        SWSS_LOG_ERROR("not found %s:%s",
+                sai_serialize_object_type(SAI_OBJECT_TYPE_LAG_MEMBER).c_str(),
+                sid.c_str());
```
lolyu, line 946: 🟢 The default-to-`true` on a failed get is sound and well-documented (VS doesn't materialize SAI defaults, and a LAG member is admin-up unless told otherwise). One coverage gap: `getLagMemberEffectiveAdminUp` is unit-tested as a pure function, but its *composition* with `get_port_admin_state`'s default isn't. Worth a small test for the "port explicitly admin-down, then egress toggled" matrix to lock in that effective state never wrongly re-ups a port SONiC deliberately downed.

vslib/vpp/SwitchVppRif.cpp
```diff
@@ -617,7 +617,11 @@ sai_status_t SwitchVpp::UpdatePort(
 
     if (attr_type != NULL)
     {
-        vpp_set_interface_state(object_id, 0, attr_type->value.booldata);
+        // A LAG member port that is egress-disabled must stay down regardless of
+        // the requested admin state. Effective state = admin_state && !egress_disable.
+        bool admin_up = attr_type->value.booldata;
+        bool egress_disabled = m_egress_disabled_lag_member_ports.count(object_id) > 0;
+        vpp_set_interface_state(object_id, 0, getLagMemberEffectiveAdminUp(admin_up, egress_disabled));
```
lolyu: 🟢 Minor: this `.count() > 0` read of `m_egress_disabled_lag_member_ports` is the cross-path reader referenced in the threading note on `SwitchVpp.h`. The logic itself — making admin-state and egress-disable compose via `getLagMemberEffectiveAdminUp` rather than clobbering each other — is the clean heart of this design and reads correctly. No change needed here beyond whatever you decide on the synchronization question.

lolyu (commented):
**Review:** https://github.com/sonic-net/sonic-sairedis/pull/1953

I read the full `SwitchVppFdb.cpp` (not just the diff) and the `UpdatePort` interaction, and compared against the earlier detach-based attempt.

**This redesign is the right call and fixes a real regression.** Modeling egress-disable as a member-port admin-down (`effective_up = admin_state && !egress_disable`) — keeping the member attached to the bond — resolves the remove-path bug the detach approach had: there, a member created egress-disabled was never added to the bond, so `removeLagMember` → `delete_bond_member` operated on an absent member and VPP's `BOND_DETACH_MEMBER` would reject it (`SAI_STATUS_FAILURE`). Here the member is always added then downed, so it's genuinely in the bond at remove time. This also directly adopts the "set the port down instead of detaching" suggestion from the HLD review. Nice.

A few things to confirm before merge (inline). The substantive one is #1: `LagMapMutex` is declared but **never locked anywhere** (verified across both touched files), so the new `m_egress_disabled_lag_member_ports` set — written on the LAG-member set path, read on the port `UpdatePort` path — is unsynchronized. Likely latent if VS is single-threaded, but the unused mutex implies a protection that isn't there. #2 is a confirm-the-MAC question on the changed LCP-creation ordering. Rest are minor.

Also: the **PR description is now stale** — it still says "removing the member from the VPP bond" and "defers VPP bond member creation," which describes the old approach. The code keeps the member attached and uses admin-down. Worth updating the description (and the linked HLD) so reviewers aren't misled.

vslib/vpp/SwitchVpp.h
cyw233: Changed the implementation from the admin-state-down model to a detach-from-bond model

vslib/vpp/SwitchVppFdb.cpp
cyw233: Intentional and still valid. Flow for an egress-disabled first member: add to bond -> bond adopts member MAC -> vpp_ensure_lag_lcp creates the tap with that MAC -> detach member. The link stays admin-up; the tap MAC is taken before the detach and unaffected by it. Validated with pc/test_lag_member_forwarding.py (https://github.com/sonic-net/sonic-mgmt/pull/25135)

vslib/vpp/SwitchVppFdb.cpp
cyw233: Changed the implementation from the admin-state-down model to a detach-from-bond model

vslib/vpp/SwitchVppRif.cpp
cyw233: Changed the implementation from the admin-state-down model to a detach-from-bond model

vslib/vpp/SwitchVppFdb.cpp
```diff
@@ -849,27 +860,245 @@ sai_status_t SwitchVpp::vpp_create_lag_member(
...
+    int ret;
+    if (egress_disable)
+    {
+        ret = delete_bond_member(hwif_name);
+    }
+    else
+    {
+        uint32_t bond_sw_if_index;
+        CHECK_STATUS(get_lag_member_bond_index(lag_member_oid, bond_sw_if_index));
+        ret = create_bond_member(bond_sw_if_index, hwif_name, false, false);
```
dypet, line 1079: What's the behavior of create_bond_member if the bond member has already been created / attached? Does this need to be checked before calling create?

cyw233: Good question. Checked against VPP 26.06: it explicitly guards double-add. If the interface is already a bond member it returns `VNET_API_ERROR_VALUE_EXIST` and no-ops, so there's no state corruption. So a pre-check isn't required for correctness.

The one wrinkle:  `create_bond_member` doesn't normalize that return, so `VALUE_EXIST` surfaces as `SAI_STATUS_FAILURE` with a false "failed to re-attach" log, even though "already a member" is exactly the desired end state on the re-attach path. In the current flow this is unreachable because re-attach is gated by  `getLagMemberEgressDisableAction == ENABLE` (only fires when the member was previously detached), so it's not a functional bug today.

To make it robust against state divergence (e.g. warm-restart/resync) and symmetric with the detach path, I can treat `VALUE_EXIST` as success on the re-attach path. Happy to add that if you'd prefer; otherwise the VPP-side guard already prevents any harm.

dypet (dismissed):

vslib/vpp/SwitchVpp.h
```diff
@@ -1143,6 +1169,7 @@ namespace saivs
...
```
lolyu, line 1174: 🟡 **Downgraded from my earlier 🟠 to cleanup — but still worth resolving.** In the previous (admin-down) revision this set had a cross-path reader in `UpdatePort` (SwitchVppRif.cpp), which made the missing synchronization a real two-path race. At this head that reader is gone: `m_egress_disabled_lag_member_ports` is referenced only in SwitchVppFdb.cpp, all on the LAG-member create/set/remove path, which VS serializes through a single queue — so no live race remains.

What's left is that `LagMapMutex` (line 1171, just above) is **still declared and never locked anywhere** — verified across all three touched files, zero `lock_guard`/`unique_lock` on it. Note `BFD_MUTEX` at SwitchVpp.h:24 shows this codebase does use a lock-guard macro pattern, so an unused LAG mutex reads as an intended-but-missing protection for `m_lag_bond_map` + this set. Suggest either wiring it around both structures, or deleting it and adding a one-line comment documenting the single-threaded-queue assumption these structures rely on. Cleanup, not blocking.

vslib/vpp/SwitchVppFdb.cpp
lolyu, line 1079: 🟡 **Concur with @dypet's thread — normalize `VALUE_EXIST` as success here.** Your analysis is right: VPP 26.06 guards double-add and returns `VNET_API_ERROR_VALUE_EXIST`, but `create_bond_member` doesn't normalize it, so a benign "already a member" surfaces as `SAI_STATUS_FAILURE` with a false "failed to re-attach" ERROR (L1082-1086). It's unreachable in the current gated flow (re-attach only fires from `ENABLE`, i.e. previously detached) — but it becomes reachable on **warm-restart / resync state divergence**, which is exactly when a spurious failure hurts most. Treating `VALUE_EXIST` as success on the re-attach path (a few lines) makes this idempotent and symmetric with the detach side. Non-blocking, but I'd take your offer to add it.

lolyu (commented):
**Re-review:** https://github.com/sonic-net/sonic-sairedis/pull/1953

Re-reviewed at head `6b0f86c` after the approach change from admin-down to detach-from-bond. I read the full `SwitchVppFdb.cpp` / `SwitchVpp.h` / `SwitchVppRif.cpp` at this head, not just the diff.

**Both of my earlier substantive concerns are now resolved — verified from source:**

1. **The remove-path regression that originally made me prefer admin-down is closed.** My worry with any detach approach was that a member created egress-disabled is detached from the bond, so `removeLagMember → delete_bond_member` would later act on a non-member and VPP would reject it. The new `removeLagMember` (L1110-1127) guards exactly this: it checks `m_egress_disabled_lag_member_ports.count(port_oid)` and skips `vpp_remove_lag_member` for an already-detached member, erasing the tracking entry instead. And `createLagMember` adds to the bond first (so the tap inherits the member MAC via `vpp_ensure_lag_lcp`) *then* detaches. Consistent end-to-end. Nice iteration.

2. **My LCP-ordering / tap-MAC question is answered** — add-member → bond adopts MAC → tap created with that MAC → detach; MAC is taken before the detach and unaffected, validated with `pc/test_lag_member_forwarding.py`. Good.

**My original 🟠 mutex finding is correctly downgraded by the approach change.** In the admin-down model, `m_egress_disabled_lag_member_ports` had a genuine cross-path reader (`UpdatePort` in `SwitchVppRif.cpp` read it while the LAG-member path wrote it). At this head that reader is gone — the set is referenced *only* in `SwitchVppFdb.cpp`, entirely on the LAG-member create/set/remove path, which VS serializes through one queue. So it's no longer a live race; what remains is just the dead `LagMapMutex` (cleanup). Details inline.

One new non-blocking item, agreeing with @dypet's thread: normalizing `VALUE_EXIST` as success on the re-attach path would make it idempotent and warm-reboot-safe. Inline.

Net: the design flipped, but the flip is sound and the blocking regression is verifiably fixed. dypet has already approved. The two items below are both non-blocking cleanups. No objection from me to merging once you've decided on the `VALUE_EXIST` nit.

vslib/vpp/SwitchVpp.h
cyw233: Yes, `LagMapMutex` is declared but never locked, and it was added in `552cc66c` with `m_lag_bond_map`). Both  `m_lag_bond_map` and `m_egress_disabled_lag_member_ports` are only touched on the LAG create/set/remove path, which VS serializes through its single queue, so there's no live race and wiring the mutex would just add a no-op lock.

Therefore, I removed the dead `LagMapMutex` and add a one-line comment documenting that these structures rely on the single-threaded VS queue.

vslib/vpp/SwitchVppFdb.cpp
cyw233: Agreed. `create_bond_member` returns the raw VPP status, so `VALUE_EXIST` surfaces as `SAI_STATUS_FAILURE`, whereas the detach side already normalizes `NO_SUCH_ENTRY` through `vpp_normalize_ret`. 

Therefore, I made it symmetric by handling the add case in `vpp_normalize_ret` (is_del=false → treat  VNET_API_ERROR_VALUE_EXIST as success) and calling it from `create_bond_member`, so re-attach is idempotent under warm-restart/resync divergence.

lolyu (approved):
**Approving** at head `b3211579`.

Both items from my last review are resolved — verified from the commit diff:

1. **Dead `LagMapMutex` → removed + documented.** The never-locked mutex is gone from `SwitchVpp.h`, replaced with a comment stating that `m_lag_bond_map` and `m_egress_disabled_lag_member_ports` are only touched on the LAG create/set/remove path, which the VS layer serializes through a single queue, so no locking is needed. Correct call given there's no live race.

2. **`VALUE_EXIST` on re-attach → normalized, and better than I proposed.** Rather than a local special-case, you extended the existing `vpp_normalize_ret` helper to map add + `VALUE_EXIST` → success, symmetric with the delete + `NO_SUCH_ENTRY` case it already handled, and wired it into `create_bond_member` (post-`WR`, pre-error-log). This makes *every* `create_bond_member` caller idempotent — not just the egress re-attach path — and reads as a matched pair with the detach side. The benign "already a member" on warm-restart/resync no longer surfaces as a spurious `SAI_STATUS_FAILURE` + false "failed to re-attach" ERROR.

Every finding across all review rounds is now closed: the original remove-path regression (fixed in the detach redesign), the LCP-ordering/tap-MAC question (answered with a passing `pc/test_lag_member_forwarding.py` run), the dead mutex, and the `VALUE_EXIST` idempotency. The design flipped once (admin-down → detach), but the detach implementation is complete and robust, and the final commit even improves a shared helper. No remaining concerns.

Thanks for the careful iteration on this one.

PR #1963, opened 2026-07-02
enable FORTIFY_SOURCE=3

xq9mend:
#### Why I did it

Upgrade FORTIFY_SOURCE from the dpkg default of =2 to =3 for stronger compile-time buffer overflow detection using `__builtin_dynamic_object_size`.

##### Work item tracking
- Microsoft ADO:

#### How I did it

- `debian/rules`: add `DEB_CPPFLAGS_MAINT_APPEND = -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3`
  - `-U_FORTIFY_SOURCE` clears the existing `=2` before setting `=3` to avoid duplicate-definition warnings
  - Guard: `FORTIFY_SOURCE` is incompatible with AddressSanitizer (both provide overlapping runtime instrumentation); guarded with `ifneq ($(ENABLE_ASAN), y)` — the same variable `debian/rules` already uses to control the ASAN build path (line 31)

#### How to verify it

Build the package and confirm `-D_FORTIFY_SOURCE=3` appears in compiler invocations. Build with `ENABLE_ASAN=y` and confirm `-D_FORTIFY_SOURCE=3` is absent.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511
- [ ] 202605

- [ ] 

#### Description for the changelog

debian/rules: upgrade FORTIFY_SOURCE to =3; guard against ASAN builds

#### Link to config_db schema for YANG module changes

N/A — no YANG changes.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -16,6 +16,12 @@ endif
 export DEB_BUILD_MAINT_OPTIONS = hardening=+all
 
 # see ENVIRONMENT in dpkg-buildflags(1)
+# Upgrade FORTIFY_SOURCE to level 3 (dpkg default is =2).
+# Guard: FORTIFY_SOURCE is incompatible with AddressSanitizer; skip when
+# the ASAN build passes --enable-asan via DEB_CONFIGURE_EXTRA_FLAGS.
+ifeq ($(findstring --enable-asan,$(DEB_CONFIGURE_EXTRA_FLAGS)),)
+export DEB_CPPFLAGS_MAINT_APPEND = -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3
+endif
 # package maintainers to append CFLAGS
 #export DEB_CFLAGS_MAINT_APPEND  = -Wall -pedantic
 # package maintainers to append LDFLAGS
--- a/debian/rules
+++ b/debian/rules
@@ -18,8 +18,8 @@ export DEB_BUILD_MAINT_OPTIONS = hardening=+all
 # see ENVIRONMENT in dpkg-buildflags(1)
 # Upgrade FORTIFY_SOURCE to level 3 (dpkg default is =2).
 # Guard: FORTIFY_SOURCE is incompatible with AddressSanitizer; skip when
-# the ASAN build passes --enable-asan via DEB_CONFIGURE_EXTRA_FLAGS.
-ifeq ($(findstring --enable-asan,$(DEB_CONFIGURE_EXTRA_FLAGS)),)
+# building with ENABLE_ASAN=y (the variable that controls the ASAN build path).
+ifneq ($(ENABLE_ASAN), y)
 export DEB_CPPFLAGS_MAINT_APPEND = -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3
 endif
 # package maintainers to append CFLAGS


debian/rules
```diff
@@ -16,6 +16,12 @@ endif
 export DEB_BUILD_MAINT_OPTIONS = hardening=+all
 
 # see ENVIRONMENT in dpkg-buildflags(1)
+# Upgrade FORTIFY_SOURCE to level 3 (dpkg default is =2).
+# Guard: FORTIFY_SOURCE is incompatible with AddressSanitizer; skip when
+# the ASAN build passes --enable-asan via DEB_CONFIGURE_EXTRA_FLAGS.
+ifeq ($(findstring --enable-asan,$(DEB_CONFIGURE_EXTRA_FLAGS)),)
```
lolyu: 🔴 This guard is keyed off the wrong variable and is effectively a no-op. ASAN in this `rules` file is enabled via the `ENABLE_ASAN` env var, not `DEB_CONFIGURE_EXTRA_FLAGS` — see lines 31–33 just below:

```makefile
ifeq ($(ENABLE_ASAN), y)
    configure_opts += --enable-asan
endif
```

sonic-buildimage's `slave.mk` also drives ASAN through `ENABLE_ASAN` (`export enable_asan="$(ENABLE_ASAN)"`); nothing injects `--enable-asan` into `DEB_CONFIGURE_EXTRA_FLAGS`. So on a real ASAN build (`ENABLE_ASAN=y`), `$(findstring --enable-asan,$(DEB_CONFIGURE_EXTRA_FLAGS))` is empty → the `ifeq` is true → `_FORTIFY_SOURCE=3` is applied during the ASAN build this guard is meant to exclude (the opposite of the intent). It only "works" in a config that never actually occurs.

Suggest guarding on the same variable the file already uses:

```makefile
ifneq ($(ENABLE_ASAN), y)
export DEB_CPPFLAGS_MAINT_APPEND = -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3
endif
```

That matches lines 31–33 and actually excludes ASAN builds.

lolyu (commented):
Reviewed at head `b184bfa`. The intent is good and the FORTIFY mechanics are right (`-U_FORTIFY_SOURCE` before `-D..=3` correctly avoids dpkg's `=2` redefinition warning). One real problem with the ASAN guard, plus a couple of minor notes.

**The ASAN guard keys off the wrong variable and is effectively a no-op.** The guard tests `DEB_CONFIGURE_EXTRA_FLAGS` for `--enable-asan`, but this `debian/rules` enables ASAN from the `ENABLE_ASAN` env var, not that variable (lines 31–33: `ifeq ($(ENABLE_ASAN), y)  configure_opts += --enable-asan`). sonic-buildimage's `slave.mk` drives ASAN via `ENABLE_ASAN` (`export enable_asan="$(ENABLE_ASAN)"`) — nothing puts `--enable-asan` into `DEB_CONFIGURE_EXTRA_FLAGS`. So on a real ASAN build (`ENABLE_ASAN=y`), `findstring` returns empty, the `ifeq` is true, and `_FORTIFY_SOURCE=3` gets applied during exactly the ASAN build the guard was meant to exclude — the opposite of its stated purpose. It would only ever fire if someone manually threaded `--enable-asan` through `DEB_CONFIGURE_EXTRA_FLAGS`, which isn't the sanctioned path. Inline fix suggested below.

**Minor, non-blocking:**
- `_FORTIFY_SOURCE=3` only differs meaningfully from `=2` with glibc ≥ 2.34 + GCC ≥ 12. On bookworm-based builds that's fine; if any active target branch still builds on a bullseye/glibc-2.31 base, `=3` degrades toward `=2` and can emit `-Wp` noise. Worth confirming the target toolchain.
- The PR description ("How I did it") no longer matches the diff — it omits the ASAN guard entirely. Worth updating once the guard is fixed.
- If this FORTIFY_SOURCE=3 change is being rolled across multiple sonic repos, make sure the corrected guard idiom (`ENABLE_ASAN`) is used consistently so the same bug doesn't propagate.

Non-blocking overall — the one-line guard fix makes it correct.

xq9mend:
Fixed in eac68cb — ASAN guard changed to `ifneq $(ENABLE_ASAN), y` to match the variable already used on line 31 of `debian/rules`.

lolyu (approved):
**Approving** at head `eac68cb`. My blocking finding is resolved — verified from source:

- The guard is now `ifneq ($(ENABLE_ASAN), y)` (line 22), keyed off the exact variable that enables ASAN at line 31 (`ifeq ($(ENABLE_ASAN), y) → configure_opts += --enable-asan`). So `_FORTIFY_SOURCE=3` is now correctly excluded from ASAN builds, instead of being applied during exactly the build it was meant to skip.
- The old `DEB_CONFIGURE_EXTRA_FLAGS`/`findstring` variant is fully replaced (not layered over), the code comment now describes the real mechanism, and the PR description was updated with an `ENABLE_ASAN=y` verification step.

The FORTIFY mechanics were always right (`-U_FORTIFY_SOURCE` before `-D_FORTIFY_SOURCE=3` avoids dpkg's `=2` redefinition warning). 

Non-blocking, unchanged from my earlier review:
- `_FORTIFY_SOURCE=3` only differs meaningfully from `=2` on glibc ≥ 2.34 + GCC ≥ 12 (fine on bookworm; worth a glance if any target branch still builds on a bullseye/glibc-2.31 base).
- If this FORTIFY_SOURCE=3 change is being rolled across multiple sonic repos, use the same `ENABLE_ASAN` guard idiom everywhere so the original bug doesn't reappear elsewhere.

Nice fix — clean and correct.

PR #1989, opened 2026-07-12
Harden SAI designated initializers against new SAI struct fields

### LINKED ISSUE #1988: sonic-sairedis fails to build against current SAI master (-Werror=missing-field-initializers on new SAI struct fields)
### Describe the bug

`sonic-sairedis` fails to compile when built against a **current SAI master** (i.e. any SAI newer than the currently pinned `v1.18.1`). SAI has added new members to two structs that `sonic-sairedis` initializes with C++ designated initializers, and because the build uses `-Werror -Wextra -Wmissing-field-initializers` (see `configure.ac`), every such addition is a hard build break.

### Errors

Building `syncd` / the meta unit tests against SAI master:

```
SwitchNotifications.h: error: missing initializer for member
  '_sai_switch_notifications_t::on_next_hop_group_hw_protection_switchover'
  [-Werror=missing-field-initializers]

TestSaiSerialize.cpp: error: missing initializer for member
  '_sai_qos_map_params_t::dei' [-Werror=missing-field-initializers]
TestSaiSerialize.cpp: error: missing initializer for member
  '_sai_qos_map_params_t::vc'  [-Werror=missing-field-initializers]
```

### Root cause

Two upstream SAI changes added struct members that `sonic-sairedis`'s designated initializers do not list:

| New member | SAI struct | Added by (opencomputeproject/SAI) |
|---|---|---|
| `on_next_hop_group_hw_protection_switchover` | `sai_switch_notifications_t` | #2269 "HW FRR switchover notification support for protection groups" (May 2026) |
| `dei`, `vc` | `sai_qos_map_params_t` | #2263 "SAI support for Credit Based Flow Control (UEC 1.0)" (Apr 2026) |

The affected initializers:

- `syncd/SwitchNotifications.h` — `Slot::Slot()` builds a `sai_switch_notifications_t` with a designated initializer listing every `on_*` callback.
- `unittest/meta/TestSaiSerialize.cpp` — `serialize_qos_map` builds a `sai_qos_map_t` whose `.key`/`.value` (`sai_qos_map_params_t`) are designated-initialized.

With `-Wall -Wextra -Werror -Wmissing-field-initializers`, GCC (verified on GCC 12/13) errors on the *missing* members even though they are designated initializers, so the build fails.

### Why this isn't seen on master today

The SAI submodule in `sonic-sairedis` is pinned to `c67f115` ("Update SAI version to v1.18.1", Mar 30 2026), which predates both #2263 and #2269. So the structs currently have no extra members and the initializers happen to be complete. The break is latent and will surface the moment the SAI pin is advanced (which is inevitable). It also breaks any downstream that tracks SAI master (e.g. platform vendors carrying SAI extensions).

### Steps to reproduce

1. Update the `SAI` submodule to current SAI master (or any commit ≥ #2263 / #2269).
2. Build `syncd` and the meta unit tests.
3. Compilation fails with the errors above.

### Expected behavior

`sonic-sairedis` should build against newer SAI. New notification callbacks / QoS-map params that syncd hasn't wired up yet should default to `nullptr`/`0` rather than breaking the build.

### Proposed fix

Wrap these two designated initializers in a scoped `#pragma GCC diagnostic ignored "-Wmissing-field-initializers"` so future SAI struct-field additions no longer break the build (unset members keep their `nullptr`/`0` default). PR to follow.

jjin62:
#### Why I did it

Fixes #1988.

`sonic-sairedis` fails to compile against a current SAI master. Two SAI structs gained new members that our designated initializers don't list, and the build uses `-Werror -Wextra -Wmissing-field-initializers`, so each addition is a hard build break:

- `on_next_hop_group_hw_protection_switchover` added to `sai_switch_notifications_t` (opencomputeproject/SAI#2269)
- `dei`, `vc` added to `sai_qos_map_params_t` (opencomputeproject/SAI#2263)

```
SwitchNotifications.h: error: missing initializer for member
  '_sai_switch_notifications_t::on_next_hop_group_hw_protection_switchover'
TestSaiSerialize.cpp: error: missing initializer for member '_sai_qos_map_params_t::dei'
TestSaiSerialize.cpp: error: missing initializer for member '_sai_qos_map_params_t::vc'
```

This isn't visible on `master` yet only because the `SAI` submodule is still pinned to `v1.18.1`, which predates both changes. It breaks the moment the SAI pin advances (inevitable) and already breaks any downstream tracking SAI master.

#### How I did it

Wrap the two affected designated initializers in a scoped `#pragma GCC diagnostic push` / `ignored "-Wmissing-field-initializers"` / `pop`:

- `syncd/SwitchNotifications.h` — the `sai_switch_notifications_t` init in `Slot::Slot()`
- `unittest/meta/TestSaiSerialize.cpp` — the `sai_qos_map_t` init in `serialize_qos_map`

New/unset members keep their default (`nullptr` for callbacks, `0` for the test vector), which is the intended behavior until syncd explicitly wires a new notification up. No functional change; the warning stays enabled everywhere else.

Rationale for suppressing rather than adding the fields: adding `.on_next_hop_group_hw_protection_switchover = nullptr` etc. would not compile on `master` today, because those members don't exist in the currently pinned SAI. Suppressing the warning at these two sites is version-independent — it builds against both the current SAI pin and newer SAI.

Alternative considered: bump the SAI submodule and add the members explicitly. That's a larger, separate change and only defers the next break; hardening the init sites fixes the class of problem.

#### How to verify it

- Current pin: `make` builds clean as before (the pragma is inert — no fields are missing).
- Point the `SAI` submodule at current SAI master and rebuild: without this change `syncd`/meta tests fail with the errors above; with it they build.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

### DIFF
--- a/syncd/SwitchNotifications.h
+++ b/syncd/SwitchNotifications.h
@@ -147,6 +147,15 @@ namespace syncd
         {
             public:
 
+                // SAI periodically adds new members to sai_switch_notifications_t
+                // (e.g. new on_* notification callbacks). Because this designated
+                // initializer is compiled with -Werror -Wextra
+                // -Wmissing-field-initializers, any such addition breaks the build
+                // until this list is updated. New callbacks default to nullptr
+                // (no handler), which is the desired behavior until syncd wires
+                // them up, so suppress the missing-field warning for this init.
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmissing-field-initializers"
                 Slot():
                     SlotBase({
                             .on_switch_state_change = &Slot<context>::onSwitchStateChange,
@@ -173,6 +182,7 @@ namespace syncd
                             .on_ha_scope_event = &Slot<context>::onHaScopeEvent,
                             .on_flow_bulk_get_session_event = &Slot<context>::onFlowBulkGetSessionEvent,
                             }) { }
+#pragma GCC diagnostic pop
 
                 virtual ~Slot() {}
 
--- a/unittest/meta/TestSaiSerialize.cpp
+++ b/unittest/meta/TestSaiSerialize.cpp
@@ -1298,9 +1298,16 @@ TEST(SaiSerialize, serialize_qos_map)
 
     attr.id = SAI_QOS_MAP_ATTR_MAP_TO_VALUE_LIST;
 
+    // sai_qos_map_params_t gains new members over SAI releases (e.g. dei, vc);
+    // any addition breaks this designated initializer under -Werror
+    // -Wmissing-field-initializers. The unset members default to 0, which is
+    // fine for this test, so suppress the missing-field warning here.
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmissing-field-initializers"
     sai_qos_map_t qm = {
         .key   = { .tc = 1, .dscp = 2, .dot1p = 3, .prio = 4, .pg = 5, .queue_index = 6, .color = SAI_PACKET_COLOR_RED, .mpls_exp = 0, .fc = 2 },
         .value = { .tc = 11, .dscp = 22, .dot1p = 33, .prio = 44, .pg = 55, .queue_index = 66, .color = SAI_PACKET_COLOR_GREEN, .mpls_exp = 0, .fc = 2 } };
+#pragma GCC diagnostic pop
 
     attr.value.qosmap.count = 1;
     attr.value.qosmap.list = &qm;


jjin62:
Closing for now. On reflection we'd rather keep the -Wmissing-field-initializers signal (it usefully flags when SAI adds a new notification/param that syncd should wire up) instead of suppressing it. The proper fix belongs with the SAI submodule bump, adding the new members explicitly at that time. Will revisit when SAI is bumped past opencomputeproject/SAI#2263 / #2269.

PR #1993, opened 2026-07-15
Fix saisdkdump profile file handling.

judsonwilson-nvidia:
#### Why I did it
`saisdkdump` did not parse the SAI profile file passed with `-p`: `profile_get_value` returned the file path for every key, and `profile_get_next_value` always failed. A new feature on the nvidia-bluefield platform, DASH CT dumps from the SDK in techsupport, needs a profile key (`CT_TABLE_DUMP_ENABLE`) to reach SAI to trigger the CT dump in the SDK dump.

##### Work item tracking
- Microsoft ADO **(number only)**: N/A

#### How I did it
- Add `ProfileMap` (based on `saidiscovery`) to parse `key=value` lines, skip comments, and back SAI profile callbacks.
- Load the profile in `saisdkdump` before `sai_api_initialize`; build `libsaisdkdump_profile.a`.
- Add gtest coverage in `unittest/saisdkdump/` (KV parsing, malformed lines, iterator reset, clear).

#### How to verify it
```bash
./autogen.sh && ./configure --disable-static --disable-python2 --with-sai=vs
make -C saisdkdump libsaisdkdump_profile.a
make -C unittest/saisdkdump check
```

- [ ] 201811
- [ ] 201911
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

- [x] master, between 202605–202611

#### Description for the changelog
Fix saisdkdump profile file flag.

### DIFF
--- a/saisdkdump/saisdkdump.cpp
+++ b/saisdkdump/saisdkdump.cpp
@@ -1,11 +1,12 @@
 #include <iostream>
 #include <sstream>
 #include <ctime>
-#include <sstream>
 
 #include <unistd.h>
 #include <getopt.h>
+#include <cstring>
 
+#include "ProfileMap.h"
 #include "swss/logger.h"
 
 extern "C" {
@@ -15,6 +16,8 @@ extern "C" {
 // TODO split to multiple cpp
 
 std::string sai_profile = "/tmp/sai.profile";
+
+static ProfileMap g_profileMap;
 
 void print_usage()
 {
@@ -46,7 +49,7 @@ const char* profile_get_value(
 {
     SWSS_LOG_ENTER();
 
-    return sai_profile.c_str();
+    return g_profileMap.getValue(variable);
 }
 
 int profile_get_next_value(
@@ -55,7 +58,8 @@ int profile_get_next_value(
         _Out_ const char** value)
 {
     SWSS_LOG_ENTER();
-    return -1;
+
+    return g_profileMap.getNextValue(variable, value);
 }
 
 sai_service_method_table_t test_services = {
@@ -118,6 +122,11 @@ int main(int argc, char **argv)
         strStream << "/tmp/saisdkdump_" << now->tm_mday << "_" << now->tm_mon + 1 << "_" << now->tm_year + 1900 << "_" << now->tm_hour << "_" << now->tm_min << "_" << now->tm_sec;
         fileName = strStream.str();
         SWSS_LOG_INFO("The dump file is not specified, generated \"%s\" file name", fileName.c_str());
+    }
+
+    if (!g_profileMap.loadFromFile(sai_profile))
+    {
+        exit(EXIT_FAILURE);
     }
 
     sai_status_t status = sai_api_initialize(0, (sai_service_method_table_t*)&test_services);
--- a/saisdkdump/saisdkdump.cpp
+++ b/saisdkdump/saisdkdump.cpp
@@ -4,6 +4,7 @@
 
 #include <unistd.h>
 #include <getopt.h>
+#include <cerrno>
 #include <cstring>
 
 #include "ProfileMap.h"
@@ -26,7 +27,8 @@ void print_usage()
     std::cerr << "Following SAI dump options can be specified:" << std::endl;
     std::cerr << "-------------------------------------------" << std::endl;
     std::cerr << "--dump_file -f   Full path for dump file" << std::endl;
-    std::cerr << "--profile -p     Full path to SAI profile file [ default is " << sai_profile << " ]" << std::endl;
+    std::cerr << "--profile -p     Full path to SAI profile file [ default is " << sai_profile
+              << "; if missing, an empty profile is used ]" << std::endl;
     std::cerr << "--help  -h       usage" << std::endl;
 }
 
@@ -81,6 +83,7 @@ int main(int argc, char **argv)
     };
 
     bool fileSpecified = false;
+    bool profileSpecified = false;
     std::string fileName;
     int option_index = 0;
     int c = 0;
@@ -100,6 +103,7 @@ int main(int argc, char **argv)
                 if (optarg != NULL)
                 {
                     sai_profile = std::string(optarg);
+                    profileSpecified = true;
                 }
                 break;
 
@@ -122,6 +126,14 @@ int main(int argc, char **argv)
         strStream << "/tmp/saisdkdump_" << now->tm_mday << "_" << now->tm_mon + 1 << "_" << now->tm_year + 1900 << "_" << now->tm_hour << "_" << now->tm_min << "_" << now->tm_sec;
         fileName = strStream.str();
         SWSS_LOG_INFO("The dump file is not specified, generated \"%s\" file name", fileName.c_str());
+    }
+
+    if (!profileSpecified && access(sai_profile.c_str(), F_OK) != 0 && errno == ENOENT)
+    {
+        SWSS_LOG_NOTICE("default SAI profile '%s' does not exist, using empty profile",
+                sai_profile.c_str());
+
+        sai_profile.clear();
     }
 
     if (!g_profileMap.loadFromFile(sai_profile))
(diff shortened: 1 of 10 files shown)


vivekkarri (dismissed):

vivekkarri (dismissed):

saisdkdump/saisdkdump.cpp
```diff
@@ -120,6 +124,11 @@ int main(int argc, char **argv)
         SWSS_LOG_INFO("The dump file is not specified, generated \"%s\" file name", fileName.c_str());
     }
 
+    if (!g_profileMap.loadFromFile(sai_profile))
```
lolyu, line 139: 🟠 Possible regression for the default (no-`-p`) case. `sai_profile` defaults to `/tmp/sai.profile` (not empty), and `-p` is optional. `loadFromFile` returns `false` when the file can't be opened, and here that maps to `exit(EXIT_FAILURE)`. So if a user runs `saisdkdump` without `-p` and `/tmp/sai.profile` doesn't exist, the tool now hard-fails — whereas before this PR it ran fine (the callbacks just returned the path / `-1` and SAI init proceeded). For a general techsupport/diagnostic tool, failing hard when no profile is present at the default path is a meaningful change.

Note the asymmetry that suggests this is unintended: `loadFromFile("")` returns `true` and loads nothing (a deliberate "no profile is fine" path), but the non-empty default defeats it, so that graceful branch is never taken unless the user explicitly passes `-p ""`.

Suggest mirroring the existing `-f` handling: track whether `-p` was explicitly given (like `fileSpecified`) and only hard-fail on a missing file when the user *explicitly* requested one; tolerate absence of the default path (warn + empty map, continue). That keeps prior behavior for the common no-`-p` case while still catching a genuine "pointed at a profile that doesn't exist" mistake.

If the bluefield techsupport flow always passes `-p` and the default-path fail-hard is acceptable/intended, feel free to confirm and disregard — just flagging since `saisdkdump` is invoked as a general diagnostic tool.

lolyu (commented):
Reviewed at head `7363cb5`. The core fix is correct and well-tested — `profile_get_value` previously returned the file *path* for every key and `profile_get_next_value` always returned `-1`, so `-p` was effectively dead; the new `ProfileMap` (modeled on `saidiscovery`) properly parses `key=value`, skips `#`/`;` comments, handles CRLF, and backs both callbacks with the right iterator semantics. Good gtest coverage and clean build wiring (`.gitignore`/`configure.ac`/Makefiles/unittest subdir).

One behavior concern on the default (no-`-p`) invocation — inline. Otherwise LGTM; nice work.

saisdkdump/saisdkdump.cpp
judsonwilson-nvidia: Added a fix in the latest commit, where it defaults to the same file as before, but if the file was not specified and the default file doesn't exist, it uses an empty file.  This seems like the safest approach - least likely to break something. (Recall that before it wasn't actually using the profile file.)

I had AI run a series of tests of the "old" implementation (what you reviewed) and the new one with the change.
# saisdkdump — Old vs New Test Report 


### How "old" vs "new" was confirmed

| Signal | Old | New |
|---|---|---|
| Help text | `default is /tmp/sai.profile` | adds `; if missing, an empty profile is used` |
| Strings | `failed to open profile map file` only | also `default SAI profile '%s' does not exist, using empty profile` |

---

## Profile missing / `-p` behavior

| Case | Setup | Command | Old | New |
|---|---|---|---|---|
| A | default profile present | `saisdkdump -f …` | exit 0, dump OK (~44768 B) | exit 0, dump OK (~44768 B) |
| B | default missing, no `-p` | `saisdkdump -f …` | exit 1, no dump | **exit 0, dump OK** + syslog `using empty profile` |
| C | missing, `-p /tmp/sai.profile` | `saisdkdump -p /tmp/sai.profile -f …` | exit 1, no dump | exit 1, no dump (unchanged) |
| D | `-p /tmp/nope.profile` | `saisdkdump -p /tmp/nope.profile -f …` | exit 1, no dump | exit 1, no dump (unchanged) |

### Notes

- **Case B** is the intended behavior change.
- **Cases C and D** confirm that an explicit `-p` still fails even when the path is the default (`/tmp/sai.profile`); i.e. if the user specifies a path that happens to equal the default path, the program should fail. The empty-profile fallback behavior is triggered by lack of the flag, not the default file name.
- Old failure syslog example:

```text
ERR syncd#syncd#saisdkdump: :- loadFromFile: failed to open profile map file: '/tmp/sai.profile' : No such file or directory
```

- New Case B syslog example:

```text
NOTICE syncd#syncd#saisdkdump: :- main: default SAI profile '/tmp/sai.profile' does not exist, using empty profile
```

Relevant new code:

```cpp
if (!profileSpecified && access(sai_profile.c_str(), F_OK) != 0 && errno == ENOENT)
{
    SWSS_LOG_NOTICE("default SAI profile '%s' does not exist, using empty profile",
            sai_profile.c_str());
    sai_profile.clear();
}
```

This matches techsupport's bare `docker exec syncd saisdkdump -f …` fallback path in `generate_dump` when CT-dump profile prep fails.

---

Also added a fix to ignore whitespace-only lines in the profile, and tested it;

## Blank / whitespace lines

Same profile used for both binaries:

```text
SAI_DUMP_STORE_PATH=/var/log/bluefield/sdk-dumps
SAI_DUMP_STORE_AMOUNT=10
<empty>
DASH_ACL_DEFAULT_RULE_ACTION=permit
<spaces only>
<tabs only>
no_equals_here
PORT_LAYOUT=1x400G
```

| Binary | Exit | Dump | Syslog warnings |
|---|---|---|---|
| Old (`saisdkdump.old_pretest`) | 0 | OK | 4 warnings: empty, spaces, tabs, and `no_equals_here` |
| New (`saisdkdump`) | 0 | OK | 1 warning: `no_equals_here` only |

### Old syslog (excerpt)

```text
WARNING ...#saisdkdump.old_pretest: :- loadFromFile: not found '=' in line
WARNING ...#saisdkdump.old_pretest: :- loadFromFile: not found '=' in line
WARNING ...#saisdkdump.old_pretest: :- loadFromFile: not found '=' in line <tabs>
WARNING ...#saisdkdump.old_pretest: :- loadFromFile: not found '=' in line no_equals_here
```

### New syslog (excerpt)

```text
WARNING ...#saisdkdump: :- loadFromFile: not found '=' in line no_equals_here
```

**Verdict:** whitespace-only lines are skipped; real malformed lines still warn.

---

vivekkarri (approved):

lolyu (approved):
Approving at head `dfa835c`. My earlier finding — the default (no-`-p`) invocation hard-failing when `/tmp/sai.profile` is absent — is resolved, and verified from source:

```cpp
if (!profileSpecified && access(sai_profile.c_str(), F_OK) != 0 && errno == ENOENT)
{
    SWSS_LOG_NOTICE("default SAI profile '%s' does not exist, using empty profile", sai_profile.c_str());
    sai_profile.clear();
}
if (!g_profileMap.loadFromFile(sai_profile)) { exit(EXIT_FAILURE); }
```

This mirrors the existing `-f`/`fileSpecified` pattern exactly, and the semantics are right:
- **No `-p`, default file missing** → `sai_profile.clear()` → `loadFromFile("")` takes the empty-path success branch → runs with an empty profile (prior behavior preserved). This is the regression I flagged, fixed.
- **Explicit `-p <path>` missing** → guard skipped (`profileSpecified`) → still hard-fails, which is correct: a user-named profile that doesn't exist is a real error.
- The `errno == ENOENT` scoping is a nice touch — only "file absent" triggers the fallback, not permission errors.

Your old-vs-new test matrix (Case B flipping `exit 1` → `exit 0 + "using empty profile"`, Cases C/D confirming explicit `-p` still fails, and the fallback keying off the *flag* not the filename) confirms each path, and grounding it in the techsupport `saisdkdump -f …` fallback is the right justification.

Bonus whitespace-only-line fix (`find_first_not_of(" \t") == npos`) is a good cleanup too — kills the spurious `not found '='` warnings on blank lines, with a new gtest covering it.

The core fix (dead `-p` flag now actually parses the profile via `ProfileMap`) remains correct and well-tested. Nice work — thanks for the thorough follow-up.

PR #1996, opened 2026-07-16
[vpp] saivpp VXLAN L3 decap: source-independent decap, inner source MAC, BD teardown fix

aaronber0614:
### Description of PR

Summary:
Enable L3 VXLAN decap on the sonic-vpp platform (saivpp / vslib/vpp) and fix
several pre-existing decap setup/teardown defects found during review.

Part of sonic-net/sonic-buildimage#25777

Depends on the VPP patch series in sonic-net/sonic-platform-vpp#262 being in the
built image first. All changes are confined to vslib/vpp.

Changes:
- Source-independent (secondary-VTEP) VNET tunnel decap (decap_any) on both the
  primary and secondary L3 VNET paths; L2 EVPN tunnels intentionally keep exact
  outer-source validation.
- Set the L3 VXLAN tunnel inner source MAC to the router MAC via a new
  sw_interface_set_mac_by_index binary-API helper, so the inner Ethernet header
  matches what HW ASICs and the VNET decap tests expect (default was VPP's
  auto-generated 02:fe:.. MAC).
- Review hardening of the secondary-VTEP decap setup/teardown: refcount the
  shared VRF0 VTEP local-receive route so tearing down one VNI does not break
  its siblings; check every VPP API return in create_vxlan_decap_term and roll
  back the BD/BVI on failure instead of returning success half-programmed; and
  install the L3 decap terms from the tunnel-create path too, so a tunnel
  created after its map entry still gets programmed.
- Detach the VXLAN tunnel interface from its bridge-domain before deleting the
  BD in remove_vpp_vxlan_decap: the tunnel was still a BD member at delete time
  (it is removed later by remove_vpp_vxlan_encap), so
  vpp_bridge_domain_add_del(is_add=0) failed with -120 (BD in use). Benign (BD
  reclaimed later) but it tripped loganalyzer during teardown.

Files: vslib/vpp/{TunnelManager.cpp, TunnelManager.h, SwitchVpp.cpp,
vppxlate/SaiVppXlate.c, vppxlate/SaiVppXlate.h}.

### Type of change

- [x] Bug fix
- [x] New feature

Enable correct L3 VXLAN decap on sonic-vpp and remove decap setup/teardown
defects: source-dependent decap, wrong inner source MAC, and the BD-in-use
teardown error.

#### How did you do it?
Changes in vslib/vpp (TunnelManager, SwitchVpp, SaiVppXlate) to install
source-independent decap terms, set the tunnel inner source MAC, and fix the BD
member-detach ordering on teardown.

Rebuilt syncd-vpp, hot-swapped onto a t1-lag-vpp testbed, and ran the VXLAN
regression suite: test_vnet_decap 4/4, test_vxlan_multiple_tunnels 16/16, and
vxlan/test_vxlan_ecmp.py with zero -120 bridge-domain-delete errors and zero
leaked bridge-domains after teardown.

sonic-vpp (vslib/vpp) only; no other platform is affected.

### Documentation
No doc/HLD changes.

### DIFF
--- a/vslib/vpp/SwitchVpp.cpp
+++ b/vslib/vpp/SwitchVpp.cpp
@@ -1521,6 +1521,10 @@ sai_status_t SwitchVpp::create(
         sai_status_t status = m_tunnel_mgr.create_l2_vxlan_tunnel(object_id, sw_if_index);
         SWSS_LOG_INFO("L2 VXLAN tunnel create for %s: status=%d sw_if_index=%u",
             serializedObjectId.c_str(), status, sw_if_index);
+
+        // Late-tunnel hook: install any L3 secondary-VTEP decap terms whose
+        // TUNNEL_MAP_ENTRY was created before this tunnel existed (M3).
+        m_tunnel_mgr.handle_l3_vxlan_tunnel_create(object_id);
         return status;
     }
 
--- a/vslib/vpp/SwitchVpp.cpp
+++ b/vslib/vpp/SwitchVpp.cpp
@@ -1879,6 +1879,18 @@ sai_status_t SwitchVpp::remove(
         return status;
     }
 
+    if (object_type == SAI_OBJECT_TYPE_TUNNEL)
+    {
+        sai_object_id_t object_id;
+        sai_deserialize_object_id(serializedObjectId, object_id);
+
+        // Defense in depth: sweep any L3 VNET decap terms this tunnel's VTEP owns
+        // before the SAI object and its DECAP_MAPPERS links are torn down, in case
+        // the TUNNEL is deleted while its TUNNEL_MAP_ENTRYs are still present.
+        m_tunnel_mgr.handle_l3_vxlan_tunnel_removal(object_id);
+        return remove_internal(object_type, serializedObjectId);
+    }
+
     if (object_type == SAI_OBJECT_TYPE_TUNNEL_MAP_ENTRY)
     {
         m_tunnel_mgr.handle_l2_vxlan_tunnel_map_entry_removal(serializedObjectId);
--- a/vslib/vpp/TunnelManager.cpp
+++ b/vslib/vpp/TunnelManager.cpp
@@ -299,6 +299,12 @@ TunnelManager::create_vpp_vxlan_encap(
         } else {
             ip4_nbr_add_del(NULL, sw_if_index, &req.dst_address.addr.ip4, false, true/*no_fib_entry*/, bvi_mac, 1);
         }
+        /* Override the tunnel interface's auto-generated MAC with the router MAC.
+         * When a packet is routed into this L3 VXLAN tunnel, VPP builds the inner
+         * Ethernet header using the tunnel interface's hardware MAC as the source.
+         * Without this, the inner source MAC is VPP's auto MAC (02:fe:..) instead
+         * of the router MAC that HW ASICs (and the VNET decap test) expect. */
+        sw_interface_set_mac_by_index(sw_if_index, bvi_mac);
     }
 
     SWSS_LOG_INFO("successfully created encap for vxlan tunnel %d", sw_if_index);
--- a/vslib/vpp/TunnelManager.cpp
+++ b/vslib/vpp/TunnelManager.cpp
@@ -184,6 +184,13 @@ TunnelManager::tunnel_encap_nexthop_action(
             sai_ip_address_t_to_vpp_ip_addr_t(src_ip, req.src_address);
             sai_ip_address_t_to_vpp_ip_addr_t(dst_ip, req.dst_address);
             req.decap_next_index = ~0;
+            // Primary-VTEP L3 VNET decap is source-independent by design: the
+            // underlay may deliver VXLAN frames to the local VTEP IP from any
+            // outer source. This path only handles VIRTUAL_ROUTER_ID_TO_VNI
+            // (L3 VNET) mappers, so request the source-independent decap term.
+            // L2 EVPN tunnels (create_l2_vxlan_tunnel_for_vni) intentionally
+            // leave decap_any unset to keep exact outer-source validation.
+            req.decap_any = true;
 
             attr.id = SAI_TUNNEL_MAP_ENTRY_ATTR_VNI_ID_VALUE;
             CHECK_STATUS_W_MSG(tunnel_encap_mapper_entry->get_attr(attr),
@@ -623,6 +630,137 @@ TunnelManager::create_l2_vxlan_tunnel(
 }
 
 sai_status_t
+TunnelManager::install_l3_vxlan_decap_terms(
+    _In_ const std::string& map_entry_serialized_oid,
+    _In_ uint32_t vni,
+    _In_ sai_object_id_t tunnel_map_oid,
+    _In_ const SaiObject* map_entry_obj)
+{
+    SWSS_LOG_ENTER();
+
+    if (vni == 0 || tunnel_map_oid == SAI_NULL_OBJECT_ID) {
+        return SAI_STATUS_SUCCESS;
+    }
+
+    sai_object_id_t term_oid;
+    sai_deserialize_object_id(map_entry_serialized_oid, term_oid);
+    if (m_vxlan_decap_term_map.find(term_oid) != m_vxlan_decap_term_map.end()) {
+        return SAI_STATUS_SUCCESS;   // already installed
+    }
+
+    // The map entry VR -> VPP VRF that routes the decapped inner packet.
+    sai_attribute_t attr;
+    attr.id = SAI_TUNNEL_MAP_ENTRY_ATTR_VIRTUAL_ROUTER_ID_VALUE;
+    if (map_entry_obj->get_attr(attr) != SAI_STATUS_SUCCESS) {
+        return SAI_STATUS_SUCCESS;
+    }
+    auto ip_vrf = m_switch_db->vpp_get_ip_vrf(attr.value.oid);
+    if (!ip_vrf) {
+        SWSS_LOG_ERROR("VXLAN decap map entry %s: no VRF for VNI %u",
+                       map_entry_serialized_oid.c_str(), vni);
+        return SAI_STATUS_SUCCESS;
+    }
+
+    // Find the VXLAN tunnel(s) referencing this decap mapper; each tunnel's
+    // ENCAP_SRC_IP is the local VTEP IP to make decappable.
+    auto mapper_obj = m_switch_db->get_sai_object(SAI_OBJECT_TYPE_TUNNEL_MAP,
+        sai_serialize_object_id(tunnel_map_oid));
+    if (!mapper_obj) {
+        return SAI_STATUS_SUCCESS;
+    }
+    auto tunnels = mapper_obj->get_child_objs(SAI_OBJECT_TYPE_TUNNEL);
+    if (!tunnels) {
+        return SAI_STATUS_SUCCESS;
+    }
+
+    std::vector<TunnelVPPData> created;
+    for (auto& tunnel_pair : *tunnels) {
+        auto tunnel_obj = tunnel_pair.second;
+
+        attr.id = SAI_TUNNEL_ATTR_TYPE;
+        if (tunnel_obj->get_attr(attr) != SAI_STATUS_SUCCESS ||
+            attr.value.s32 != SAI_TUNNEL_TYPE_VXLAN) continue;
+
+        attr.id = SAI_TUNNEL_ATTR_ENCAP_SRC_IP;
+        if (tunnel_obj->get_attr(attr) != SAI_STATUS_SUCCESS) continue;
+        sai_ip_address_t vtep_ip = attr.value.ipaddr;
+
+        TunnelVPPData td;
+        bool skipped = false;
+        if (create_vxlan_decap_term(vtep_ip, vni, ip_vrf, td, skipped)
+                != SAI_STATUS_SUCCESS) {
+            SWSS_LOG_ERROR("VXLAN decap map entry %s: decap create failed VNI %u",
+                           map_entry_serialized_oid.c_str(), vni);
+            continue;
+        }
+        if (!skipped) {
+            created.push_back(td);
+        }
+    }
+
+    if (!created.empty()) {
+        m_vxlan_decap_term_map[term_oid] = created;
+    }
+    return SAI_STATUS_SUCCESS;
+}
+
+sai_status_t
+TunnelManager::handle_l3_vxlan_tunnel_create(
+    _In_ sai_object_id_t tunnel_oid)
+{
+    SWSS_LOG_ENTER();
+
+    // Late-tunnel hook (M3): a TUNNEL_MAP_ENTRY of type VNI_TO_VIRTUAL_ROUTER_ID
+    // may be created before the TUNNEL that references its mapper. In that case
+    // handle_l2_vxlan_tunnel_map_entry saw no tunnels and installed nothing. Now
+    // that the tunnel exists, rescan its L3 decap mappers and install any decap
+    // terms still missing. Idempotent via m_vxlan_decap_term_map.
+    auto tunnel_obj = m_switch_db->get_sai_object(SAI_OBJECT_TYPE_TUNNEL,
+        sai_serialize_object_id(tunnel_oid));
+    if (!tunnel_obj) {
+        return SAI_STATUS_SUCCESS;
+    }
+
+    sai_attribute_t attr;
+    attr.id = SAI_TUNNEL_ATTR_TYPE;
+    if (tunnel_obj->get_attr(attr) != SAI_STATUS_SUCCESS ||
+        attr.value.s32 != SAI_TUNNEL_TYPE_VXLAN) {
+        return SAI_STATUS_SUCCESS;
+    }
+
+    auto decap_mappers = tunnel_obj->get_linked_objects(
+        SAI_OBJECT_TYPE_TUNNEL_MAP, SAI_TUNNEL_ATTR_DECAP_MAPPERS);
+
+    for (auto mapper : decap_mappers) {
+        attr.id = SAI_TUNNEL_MAP_ATTR_TYPE;
+        if (mapper->get_attr(attr) != SAI_STATUS_SUCCESS) continue;
+        if (attr.value.s32 != SAI_TUNNEL_MAP_TYPE_VNI_TO_VIRTUAL_ROUTER_ID) continue;
+
+        sai_object_id_t mapper_oid;
+        sai_deserialize_object_id(mapper->get_id(), mapper_oid);
+
+        auto entries = mapper->get_child_objs(SAI_OBJECT_TYPE_TUNNEL_MAP_ENTRY);
+        if (!entries) continue;
+
+        for (auto& entry_pair : *entries) {
+            auto entry = entry_pair.second;
+
+            uint32_t vni = 0;
+            attr.id = SAI_TUNNEL_MAP_ENTRY_ATTR_VNI_ID_KEY;
+            if (entry->get_attr(attr) == SAI_STATUS_SUCCESS) {
+                vni = attr.value.u32;
+            }
+
+            install_l3_vxlan_decap_terms(entry->get_id(), vni, mapper_oid,
+                                         entry.get());
+        }
+    }
+
+    return SAI_STATUS_SUCCESS;
+}
+
+
+sai_status_t
 TunnelManager::handle_l2_vxlan_tunnel_map_entry(
     _In_ const std::string& serializedObjectId,
     _In_ uint32_t attr_count,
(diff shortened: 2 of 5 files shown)


yejianquan:
Reviewed alongside sonic-platform-vpp#262 and sonic-mgmt#26220. The new decap-term code reads well — the rollback ladders in `create_vxlan_decap_term()` and the refcounted `vxlan_secondary_vtep_local_receive()` (so sibling VNIs sharing a secondary VTEP don't tear each other's local-receive down) both look right, and the BD-detach fix in `remove_vpp_vxlan_decap()` for the `-120` "BD in use" failure is a good catch.

One thing worth fixing before merge, plus a couple of nits.

### Stale patch numbers in comments

Two comments reference "patch 0014" as the VPP patch that implements the source-independent decap:

- `vslib/vpp/TunnelManager.cpp` — "decap matches the source-independent (src, vni) wildcard that patch 0014 registers on tunnel add"
- `vslib/vpp/vppxlate/SaiVppXlate.h` — "Must match VXLAN_DECAP_ANY_FLAG in the VPP 0014 patch (src/plugins/vxlan/vxlan.h)"

In the series as it stands in sonic-platform-vpp#262 that patch is **0017** (`0017-sonic-vxlan-vnet-source-independent-decap.patch`); `0014-sflow-per-port-api.patch` is an unrelated sFlow change. This looks like the same numbering drift you already fixed on the 0016 description in 738d1df. Cosmetic, but since these two comments are precisely the pointers a future reader follows to find the other half of the contract, they're worth correcting.

For what it's worth I did check the part that actually matters: `VPP_VXLAN_DECAP_ANY_FLAG` here and `VXLAN_DECAP_ANY_FLAG` in the VPP patch are both `(1u << 31)`, so the two sides agree.

### Encoding the flag in `decap_next_index`

Smuggling `decap_any` in the high bit of the wire `decap_next_index` avoids an API CRC bump, which I understand as the motivation, and `decap_next_index` values are small enough that bit 31 is safe in practice. The cost is a silent version coupling: saivpp built with this change and run against a VPP without patch 0017 would send `decap_next_index = 0x80000001` and get undefined behaviour rather than a clean failure. The `VPP_VERSION` bump in #262 plus in-image version locking covers this in practice — a sentence in the header comment stating that dependency explicitly (rather than only "must match") would make the coupling obvious to anyone tempted to build the two halves separately.

### Nits

`install_l3_vxlan_decap_terms()` returns `SAI_STATUS_SUCCESS` on several paths that have just logged `SWSS_LOG_ERROR` (missing VRF, decap create failure). If that's a deliberate best-effort policy — plausible, given it's also used from the late-tunnel hook where partial state is expected — a short comment saying so would help; otherwise those look like they should propagate.

In `vxlan_decap_term_set_dst()`, the IPv6 branch relies on `s6_addr[8] ^= 0x01` to keep `ip46_address_is_ip4(dst)` false. There is a corner case where that flips the wrong way: a source of the form `::0100:0000:xxxx:xxxx` (bytes 0-7 zero, byte 8 == 0x01, bytes 9-11 zero) produces an all-zero top 12 bytes, i.e. exactly the misclassification the helper exists to prevent. Not a realistic VTEP address, but forcing a fixed non-zero byte and asserting the result would be airtight.

`create_vxlan_decap_term()` defines the `rollback_bd_bvi` lambda and then open-codes the same four steps in the three later failure paths. They're consistent today, which is the main thing, but reusing the lambda would keep them that way.

aaronber0614:
Thanks for the careful read @yejianquan. Pushed `ff8519fe` addressing all of it.

**Stale patch numbers (0014 → 0017).** Fixed both pointers — the `vxlan_decap_term_set_dst` header in `TunnelManager.cpp` and the `VPP_VXLAN_DECAP_ANY_FLAG` comment in `SaiVppXlate.h`. Both said "0014" because that was the patch number when I wrote them on Jul 14; the series was later renumbered to 0015–0017 when the sflow patches (0012–0014) landed on master. Confirmed the two flag values are both `(1u << 31)` as you noted.

> FYI, while there I noticed a *third* reference at `SaiVppXlate.h:253` on the `VPP_BOND_API_LB_ALGO_L34_INNER` comment — "sonic-platform-vpp patch 0010-sonic-inner-aware-flow-hash", where that patch is actually `0011` in the current series. But that line is pre-existing (your commit `06e38d8a` from the inner-aware LAG hash work), not part of this PR's diff, so I left it untouched to keep the PR scoped. Happy to fix it in a separate change if you'd like.

**Version coupling of the high-bit flag.** Added an explicit sentence to the `SaiVppXlate.h` header comment: saivpp built with this flag requires a VPP with patch 0017; against an older VPP the `decap_next_index = 0x80000001` is undefined behaviour rather than a clean failure, so the two must be version-locked (which the `VPP_VERSION` bump + in-image version lock enforces in practice).

**`install_l3_vxlan_decap_terms()` SUCCESS returns.** Confirmed deliberate — it's called from both the map-entry hook and the late-tunnel hook, so a legitimately-incomplete intermediate state (mapper present but no tunnel yet, or vice-versa) has to return SUCCESS or it would fail an otherwise valid create. Added a comment at the top of the function stating that best-effort / idempotent policy; the logged-ERROR paths (no VRF, per-tunnel decap create failure) are skipped and re-attempted on the next hook.

**IPv6 `s6_addr[8]` corner case.** Good catch. Made it correct-by-construction: keep the `s6_addr[8] ^= 0x01` flip for `dst != src`, and additionally force `s6_addr[0] = 0x20` (fixed non-zero, non-multicast) so the top 12 bytes can never collapse to zero regardless of src — the `::0100:0:xxxx:xxxx` case you described is now impossible. I didn't add an `assert()` because it isn't used elsewhere in this file and would be compiled out under `NDEBUG`; forcing the byte makes the invariant hold unconditionally, which is strictly stronger than asserting after the fact.

**`rollback_bd_bvi` lambda.** Reused it in the three post-tunnel failure paths (`create_vpp_vxlan_encap`, bridge, local-receive) instead of open-coding the same four steps.

yejianquan:
Thanks — the updates in `ff8519fe` all look right to me.

The IPv6 fix is better than what I suggested: forcing `s6_addr[0] = 0x20` makes the invariant hold unconditionally instead of only detecting a violation after the fact, and you're right that an `assert()` would have been compiled out under `NDEBUG` anyway. The best-effort policy comment on `install_l3_vxlan_decap_terms()` and the explicit version-coupling sentence on the flag both answer what I was asking for.

On the third stale reference at `SaiVppXlate.h:253` — good catch, and thanks for leaving it out of scope. That one is mine (`06e38d8a`); I'll put up a separate one-line change for it rather than widen this PR.

### One thing I'd like to flag before this merges

I think the Copilot comment on the early return in `install_l3_vxlan_decap_terms()` is a genuine gap, not a false positive, and it happens to sit on the exact scenario this series enables. The guard is keyed on the map-entry OID:

```cpp
sai_object_id_t term_oid;
sai_deserialize_object_id(map_entry_serialized_oid, term_oid);
if (m_vxlan_decap_term_map.find(term_oid) != m_vxlan_decap_term_map.end()) {
    return SAI_STATUS_SUCCESS;   // already installed
}
```

but the work it guards is per-tunnel — the function goes on to loop over every tunnel referencing the mapper and create a decap term for each, storing them as a vector under that one key. So:

1. Map entry `E` is created while tunnel `A` already references mapper `M`. A term is installed for `A` and `m_vxlan_decap_term_map[E]` becomes non-empty.
2. Tunnel `B` is created later against the same mapper `M`. `handle_l3_vxlan_tunnel_create(B)` walks `M`'s entries, reaches `E`, and returns at the guard above.
3. `B`'s secondary-VTEP decap term is never installed, and nothing retries it.

The pure late-tunnel case you designed the hook for is safe, because when no tunnel exists yet `created` stays empty and the `if (!created.empty())` means nothing is recorded, so the next hook re-attempts. The same is true when the only candidate is the primary VTEP and `is_local_skip` fires. The gap needs at least one non-local tunnel installed at map-entry time plus a further tunnel arriving afterwards — which is plausibly what a multi-VTEP ordering in `test_vxlan_multiple_tunnels.py` produces, and that test is being added to the t1-lag-vpp PR set in sonic-mgmt#26220.

Making the guard per (map entry, VTEP) rather than per map entry would close it — e.g. keep the vector but skip only the tunnels whose VTEP IP already has a term recorded under that entry, so a newly-appearing tunnel still gets one. That also makes the function properly idempotent in the sense the comment claims.

Worth confirming against a run of `test_vxlan_multiple_tunnels.py` once the image carries these changes, since that's the case that would expose it.

### On the other three Copilot comments

For what it's worth, I read the one about the `handle_l2_vxlan_tunnel_map_entry` reference in the `handle_l3_vxlan_tunnel_create()` comment as a false positive: the `VNI_TO_VIRTUAL_ROUTER_ID` branch genuinely does live inside `handle_l2_vxlan_tunnel_map_entry`, so the comment is accurate — the function name is just historically misleading. Renaming it is out of scope here.

The two about unchecked return codes (`sw_interface_set_mac_by_index()` in `create_vpp_vxlan_encap()`, and the unbridge in `remove_vpp_vxlan_decap()`) both seem fair to me. The MAC one especially: if it fails the tunnel silently keeps the `02:fe:..` auto MAC, which is the exact condition the change exists to prevent, and the caller still sees success — at minimum worth an error log.

aaronber0614:
Thanks @yejianquan — you're right on all three, and I've pushed the fix as `fc3bb733`.

### Per-(map entry, VTEP) guard

Implemented exactly as you suggested. I kept the vector under the map-entry key but replaced the whole-entry early-return with per-VTEP idempotency. The function now:

- looks up the entry's existing vector (if any) instead of returning on its presence;
- in the tunnel loop, skips only VTEPs whose source IP already has a term recorded — either under this entry from a prior hook (`existing`) or installed earlier in the same pass (`created`) — using the existing `saivs::sai_ip_address_equal` on `TunnelVPPData.src_ip`;
- **appends** the newly-created terms to the entry's vector (`vec.insert(vec.end(), ...)`) instead of overwriting it.

So the step-3 scenario you described — tunnel `B` arriving after entry `E` was processed with tunnel `A` already termed — now installs `B`'s decap term and appends it, while `A` is skipped. That also makes the function properly idempotent in the sense the comment claims. I checked the teardown path (`handle_l2_vxlan_tunnel_map_entry_removal`) iterates the full vector, so appending is safe on removal.

### The two unchecked return codes

Both added as (non-fatal) error logs:

- `sw_interface_set_mac_by_index()` in `create_vpp_vxlan_encap()` — logs if setting the router MAC fails, so the silent `02:fe:..` auto-MAC fallback (the exact condition the change exists to prevent) is now diagnosable instead of returning success.
- the unbridge `set_sw_interface_l2_bridge_by_index(..., is_add=false, ...)` in `remove_vpp_vxlan_decap()` — logs if the detach fails, which is what precedes the `-120` (BD in use) on the subsequent BD delete.

### The `handle_l2_vxlan_tunnel_map_entry` naming comment

Agreed — false positive, left as-is. The `VNI_TO_VIRTUAL_ROUTER_ID` branch does live in `handle_l2_vxlan_tunnel_map_entry`; the name is just historically misleading.

### Verification

Built these changes into a `docker-syncd-vpp` image and ran `test_vxlan_multiple_tunnels.py` on a t1-lag-vpp testbed — i.e. the multi-VTEP ordering you flagged: **16/16 passed**, BGP 24/24 established, no new cores.

And thanks for taking the third stale reference at `SaiVppXlate.h:253` in a separate change — left it out of scope here.

Co-authored with CoPilot

aaronber0614:
@yejianquan quick follow up on your cross note about the map entry granularity early return in `install_l3_vxlan_decap_terms()`.

Fixed in `fc3bb733`. Rather than returning as soon as the map entry has any recorded term, it now skips only the VTEPs that already have a decap term under that entry and installs one for any tunnel that appears later, appending to the entry's vector instead of overwriting it. That also covers the Copilot comment on the same early return, plus the two unchecked return codes (the router MAC in `create_vpp_vxlan_encap` and the unbridge in `remove_vpp_vxlan_decap`).

Worth recording for the merge notes though. I ran `test_vxlan_multiple_tunnels.py` on a t1-lag-vpp testbed with a `docker-syncd-vpp` image carrying the fix and it passes 16 of 16, and it also passes on the pre fix base `ff8519fe`. So the exact ordering that would expose the gap you described isn't actually reproduced by the test as it stands today. I'm treating the change as a defensive correctness fix for the multi VTEP decap path rather than something the current test forces, and the single VTEP path behaves exactly as before.

For the record the run stayed healthy end to end, BGP 24 of 24 and no new cores. If you think it's worth locking down I can add a dedicated multi VTEP ordering case to the sonic-mgmt set as a follow up so this path gets explicit coverage.

Co-authored with CoPilot

vslib/vpp/vppxlate/SaiVppXlate.c
```diff
@@ -3960,6 +3522,38 @@ int sw_interface_set_mac (const char *hwif_name, uint8_t *mac_address)
     return ret;
 }
 
+int sw_interface_set_mac_by_index (uint32_t sw_if_index, uint8_t *mac_address)
```
yue-fred-gao, line 3501: This function is very similar to sw_interface_set_mac. Please refactor to reuse the common code

vslib/vpp/TunnelManager.cpp
```diff
@@ -767,6 +985,334 @@ TunnelManager::handle_l2_vxlan_tunnel_map_entry_removal(
...
+    vxlan_secondary_vtep_local_receive(tunnel_data.src_ip, false);
+
+    // Unbridge + delete the decap VXLAN tunnel.
+    vpp_vxlan_tunnel_t req;
+    memset(&req, 0, sizeof(req));
+    req.dst_port = m_vxlan_port;
+    req.src_port = m_vxlan_port;
+    req.instance = ~0;
+    req.vni = tunnel_data.vni;
+    req.decap_next_index = ~0;
```
yue-fred-gao, line 1419: do we need to set decap_any flag for deletion?

vslib/vpp/SwitchVpp.cpp
```diff
@@ -1573,13 +1521,11 @@ sai_status_t SwitchVpp::create(
...
-        return status;
-    }
 
-    if(object_type == SAI_OBJECT_TYPE_MIRROR_SESSION) {
-        sai_object_id_t object_id;
-        sai_deserialize_object_id(serializedObjectId, object_id);
-        return createMirrorSession(object_id, switch_id, attr_count, attr_list);
+        // Late-tunnel hook: install any L3 secondary-VTEP decap terms whose
+        // TUNNEL_MAP_ENTRY was created before this tunnel existed (M3).
+        m_tunnel_mgr.handle_l3_vxlan_tunnel_create(object_id);
```
yue-fred-gao, line 1527: This is only added to create. Do we need to handle tunnel being deleted when tunnel map entries are kept?

vslib/vpp/TunnelManager.cpp
```diff
@@ -767,6 +985,334 @@ TunnelManager::handle_l2_vxlan_tunnel_map_entry_removal(
...
+    SWSS_LOG_ENTER();
+
+    char            hw_bvi_ifname[32];
+    auto            router_mac = get_router_mac();
+    auto            bvi_mac = router_mac.data();
+    vpp_ip_route_t  bvi_ip_prefix;
+
+    is_local_skip = false;
+
+    // Primary-VTEP guard: if this VTEP IP already belongs to one of our own
```
yue-fred-gao, line 1248: I feel this is very SONiC specific behaviour. Is there stronger hint from other SAI API calls? I think I don't quite understand the sequence of SAI API calls leading to this. Can you please share the SAI calls captured from sairedis.rec? Essentially, how SONiC express primary and secondary vtep? how decap-any is derived from SAI attributes?

vslib/vpp/vppxlate/SaiVppXlate.c
aaronber0614: Resolved. Please re-review @yue-fred-gao

vslib/vpp/TunnelManager.cpp
aaronber0614: No. On delete VPP finds the tunnel by its key (dst, vni, encap_fib, src-port, plus src for v4), which does not include decap_next_index, then uses the flag it stored at create (t->decap_any) to also remove the wildcard entry and clear L3 promiscuous. So the remove request does not need to carry decap_any; setting it would be a no-op. See the delete path of vnet_vxlan_add_del_tunnel in patch 0017.

_Co-authored with CoPilot_

vslib/vpp/SwitchVpp.cpp
aaronber0614: The secondary VTEP decap terms are anchored to the map entries, not the tunnel: handle_l2_vxlan_tunnel_map_entry_removal frees them (via m_vxlan_decap_term_map) on TUNNEL_MAP_ENTRY removal. Only create needs the late tunnel hook, because a map entry can arrive before its tunnel; delete has no symmetric race since removing the entries frees the terms.
You are right about the gap: a TUNNEL deleted while its TUNNEL_MAP_ENTRYs are kept would leak, since nothing sweeps the terms then. Normal VNET teardown removes the entries first, so it does not happen in practice, but I added a symmetric handle_l3_vxlan_tunnel_removal sweep as defense in depth, pushed in 9ccc457a. It runs before remove_internal (while the tunnel and its DECAP_MAPPERS links are still resolvable) and is refcount aware, since terms are deduplicated per VTEP: it frees a term only once no surviving VXLAN tunnel still references that mapper with the same VTEP source IP.

_Co-authored with CoPilot_

vslib/vpp/TunnelManager.cpp
aaronber0614: It is not a heuristic. decap_any is set only on the L3 VNET path, both mapper directions. On the encap side (VIRTUAL_ROUTER_ID_TO_VNI, in tunnel_encap_nexthop_action) every tunnel built toward a remote peer is created with decap_any=true, so the local VTEP decaps VXLAN from any outer source. On the decap side (VNI_TO_VIRTUAL_ROUTER_ID, via install_l3_vxlan_decap_terms) the same flag is set, but that explicit term is only created for a VTEP source IP that is not one of our own interfaces: create_vxlan_decap_term probes with vpp_sw_interface_find_by_ip and skips the local (primary) VTEP, since it is already decapped by the encap path. The L2 EVPN path (VLAN mappers) never sets decap_any, so exact source matching is preserved there. So primary VTEP = the ENCAP_SRC_IP that matches a local interface (Loopback0); a non-local VTEP source is the only case that gets its own explicit term plus a VRF0 local-receive. Remote peer VTEPs, the tunnel destinations, are separate and only appear as tunnel-encap nexthops. The line at 1131 is just the placeholder dst for the decap only term: VPP needs a valid dst different from src, so I set a fixed high byte for v6 and 0.0.0.0 for v4. That is a VPP API constraint, not where decap-any is decided.
On a stronger SAI hint, and to answer how SONiC expresses this: I captured a sairedis.rec from a t1-lag-vpp VNET run. Per VNET, SONiC creates four tunnel maps (L2 VLAN<->VNI and L3 VirtualRouter<->VNI), one VXLAN TUNNEL with ENCAP_SRC_IP set to the local VTEP (10.1.0.32) and PEER_MODE P2MP, and a single TUNNEL_TERM_TABLE_ENTRY of type P2MP with DST_IP equal to that local VTEP and no SRC_IP. Per the SAI spec P2MP matches an exact dst with a masked (any) source, so that one term already expresses source independent (decap-any) decap for the local VTEP. Primary VTEP = ENCAP_SRC_IP (the same address as the term DST_IP); the remote peer VTEPs show up only as SAI_NEXT_HOP_TYPE_TUNNEL_ENCAP nexthops (IPs like 100.0.1.10), with no decap term of their own. So the SAI native decap-any signal is the P2MP term type, and a cleaner v2 would key VXLAN decap-any off it rather than the mapper type plus local-interface check. saivpp already reads the term type on the IPinIP path but the VXLAN term path currently just stores the object. Note this single-loopback topology never emits a non-local VTEP source, so the explicit secondary-term branch above is not exercised here; decap is entirely the primary encap path plus this one P2MP term. 

Excerpt:

c|SAI_OBJECT_TYPE_TUNNEL:...6cc|TYPE=VXLAN|ENCAP_SRC_IP=10.1.0.32
    |ENCAP_MAPPERS=[VLAN_ID_TO_VNI,VIRTUAL_ROUTER_ID_TO_VNI]
    |DECAP_MAPPERS=[VNI_TO_VLAN_ID,VNI_TO_VIRTUAL_ROUTER_ID]|PEER_MODE=P2MP
c|SAI_OBJECT_TYPE_TUNNEL_TERM_TABLE_ENTRY:...6cd|TYPE=P2MP|VR_ID=...002
    |DST_IP=10.1.0.32|TUNNEL_TYPE=VXLAN|ACTION_TUNNEL_ID=...6cc
c|SAI_OBJECT_TYPE_TUNNEL_MAP_ENTRY:...6ce|TYPE=VIRTUAL_ROUTER_ID_TO_VNI|VR=...002|VNI=10000
c|SAI_OBJECT_TYPE_TUNNEL_MAP_ENTRY:...6cf|TYPE=VNI_TO_VIRTUAL_ROUTER_ID|VNI=10000|VR=...002
c|SAI_OBJECT_TYPE_NEXT_HOP:...6d0|TYPE=TUNNEL_ENCAP|IP=100.0.1.10|TUNNEL_ID=...6cc
IPv6 VNET emits the same shape (term TYPE=P2MP, DST_IP=fc00:1::32).

_Co-authored with CoPilot_

yue-fred-gao: Thanks for the explanation.

vslib/vpp/TunnelManager.cpp
```diff
@@ -184,6 +184,13 @@ TunnelManager::tunnel_encap_nexthop_action(
             sai_ip_address_t_to_vpp_ip_addr_t(src_ip, req.src_address);
             sai_ip_address_t_to_vpp_ip_addr_t(dst_ip, req.dst_address);
             req.decap_next_index = ~0;
+            // Primary-VTEP L3 VNET decap is source-independent by design: the
+            // underlay may deliver VXLAN frames to the local VTEP IP from any
+            // outer source. This path only handles VIRTUAL_ROUTER_ID_TO_VNI
+            // (L3 VNET) mappers, so request the source-independent decap term.
+            // L2 EVPN tunnels (create_l2_vxlan_tunnel_for_vni) intentionally
+            // leave decap_any unset to keep exact outer-source validation.
+            req.decap_any = true;
```
yue-fred-gao: should we guard the flag based on SAI_TUNNEL_PEER_MODE_P2MP?

vslib/vpp/TunnelManager.cpp
yue-fred-gao: The vtep_ip is from encap_source_ip in tunnel object and you said "but that explicit term is only created for a VTEP source IP that is not one of our own interfaces". Does this mean the encap_source_ip can be an ip not a local interface address? is this ip advertised by bgp otherwise how remote can reach it?

vslib/vpp/TunnelManager.cpp
aaronber0614: Yes this makes sense. I added guard in cbf310a9

vslib/vpp/TunnelManager.cpp
aaronber0614: Correct. vtep_ip is the tunnel ENCAP_SRC_IP. In the standard VNET model that is always the local Loopback0 VTEP, which BGP advertises into the underlay so remote peers reach us, so it is never non local in a normal config. The primary VTEP guard therefore always fires and the secondary VTEP term branch below is a defensive path a single loopback config never hits

_Co-authored with CoPilot_

yue-fred-gao: Thanks. I am still unclear on where the secondary vtep term is from. Can you highlight the SAI call above? Which field becomes the vtep ip in the decap term?

aaronber0614: @yue-fred-gao 

The secondary VTEP is not a separate SAI object, it is the same TUNNEL,. The one SAI read is in install_l3_vxlan_decap_terms:

```
attr.id = SAI_TUNNEL_ATTR_ENCAP_SRC_IP;
tunnel_obj->get_attr(attr);                     // line 741
sai_ip_address_t vtep_ip = attr.value.ipaddr;   // line 743

```
So the field that becomes the vtep ip in the decap term is SAI_TUNNEL_ATTR_ENCAP_SRC_IP. I read it off each tunnel that is a child of the VNI_TO_VIRTUAL_ROUTER_ID decap mapper (get_child_objs(SAI_OBJECT_TYPE_TUNNEL), line 716) and pass it into create_vxlan_decap_term(vtep_ip, ...) at line 754, where it becomes the VPP tunnel src_address.

There is no primary vs secondary split in SAI. Both are just TUNNELs with an ENCAP_SRC_IP. I make the split locally at line 1261 with vpp_sw_interface_find_by_ip(vtep_ip): if that IP is one of our own interface addresses (the normal single Loopback0 case) VPP already receives it so I skip, and if it is not local I call it a secondary VTEP and add the explicit VRF0 local-receive route so VPP will accept and decap those outer packets.

_Co-authored with CoPilot_

yue-fred-gao: The SAI_TUNNEL_ATTR_ENCAP_SRC_IP in above TUNNEL is 10.1.0.32, which is the normal case with loopback address. Is there a real usecase where the encap source ip is not a local ip? You said it is a defensive path so I am wondering if it is really necessary.

aaronber0614: Hi @yue-fred-gao , I modified the comment. It's not defensive but necessary.  The encap source IP does not have to be a local IP. test_vxlan_multiple_tunnels configures :
1. the loopback0 tunnel (10.1.0.32, local) and 
3. a second tunnel whose SRC is the Loopback0 IP BUT with the last octet set to 100 or 101. (e.g. 10.1.0.100). This address is not assigned to an interface. It then sends the outer packet to this special SRC and expects DUT to decap it.

The test assumes that on ASICs, (not specific to VPP) that the DUT is required to decap the special SRC even if not one of its interface address. VPP is the odd one out in how it gets there. VPP only local-receives an IP that is on an interface or has an explicit local route, so without this branch there is no local-receive for the special src, the outer packet is never punted to vxlan-input, and it is not decapped. 

The change adds a VRF0 /32 or /128 VPP_NEXTHOP_LOCAL receive for exactly those non-interface VTEP IPs, while the primary Loopback0 VTEP is skipped by the local-interface guard. 
In our run that is what takes the special-src subcases from failing to passing.

_Co-authored with CoPilot_

yue-fred-gao: Thanks. Honestly, I don't know how the secondly vtep works in production, especially how the second vtep's source ip is advertised. There is no integration with frr to advertise vtep's source ip via routing protocols. The PR is solid to support the feature. Just how it is intended to be used is unclear. If you can find a document to describe the use case and ways to advertise the source ip, that will be helpful to understand it.

yue-fred-gao (approved):

vslib/vpp/TunnelManager.cpp
aaronber0614: @yue-fred-gao . I looked at this a little bit more. So the "secondary" vtep is a sonic-mgmt test construct, and not really something that's done in production. It's needed to test decap for src that is not the box's interface addess.

yue-fred-gao: Thanks for the clarification. Maybe the sonic-mgmt  goes too far to verify some behavior that is not required in production. It might even be some platform specific behavior. For example, some may implement vtep as an interface, which would naturally use vtep source ip as the pseduo interface ip.  But sonic doesn't model vtep as an interface so the source ip is not added to the control plane and advertised by routing protocols.

PR #1998, opened 2026-07-17
Set fabric port defaults for vs platform

lakshmi-nexthop:
### Description of PR
For UT2 kvm, running test_bgp_facts.py is causing an error due to fabric ports not initialized to default values.
ut2 is the first voq switch in VS platform. And the fabric lanemap is not set, but the default values are not set in the vslib. This caused the API getFabricPortList (in fabricport orch) to return error.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

Support kvm ut2

##### Work item tracking

#### How did you do it?
set the fabric port default values to be 0 (for count) and null for ports.

ran test_bgp_facts.py on vtb and it passed without error

### DIFF
--- a/unittest/vslib/TestSwitchStateBase.cpp
+++ b/unittest/vslib/TestSwitchStateBase.cpp
@@ -324,6 +324,24 @@ TEST_F(SwitchStateBaseTest, switchQoSMaxNumOfTrafficClasses)
     ASSERT_EQ(attr.value.u8, maxTcNum);
 }
 
+TEST_F(SwitchStateBaseTest, switchFabricPortDefaults)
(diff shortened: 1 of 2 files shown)


BYGX-wcr:
Not sure if this is needed. We generate fabric port map for VS platform in build time.

lakshmi-nexthop:
> Not sure if this is needed. We generate fabric port map for VS platform in build time.

Thanks for the info. Could you give us a pointer so we can do the same.

PR #2000, opened 2026-07-20
Add per-port counter discovery toggle argument

justin-wong-ce:
### Description of PR

Summary:
Add an argument to `syncd` to toggle between the new per-port counter discovery feature in `FlexCounter` or to use the legacy method.

This is a part 1 of 2 fix in addressing https://github.com/sonic-net/sonic-buildimage/issues/28460. There are two fix methods under consideration, this part is generic to both methods.

Fixes # (issue)
Partially Fixes https://github.com/sonic-net/sonic-buildimage/issues/28460

Needs backport to 202605.

### Type of change

- [x] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

#### What is the motivation for this PR?
In #1774 , a new counter support discovery method is introduced for being able to read counter support for platforms where different ports can have different capabilities.
The method worked well for some platforms, not so much for some others.

This change adds an argument to `syncd` so there is the capability to toggle the per-port counter discovery. The argument will default to `false` so the legacy code path is used by default.

Since there is now an option to toggle the code path on or off, there is no longer a need for a runtime fallback.

##### Work item tracking
- Microsoft ADO **(number only)**:

#### How did you do it?
Add an argument to `syncd` as well as the logic needed to read the argument in various files in `syncd`.
Unit tests are also updated to account for the new argument, as well as mocking the use of this argument for unit tests that only test for per-port discovery logic.

#### Tested branch

- [x] master
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511
- [ ] 202512
- [x] 202605
- [ ] 202608
- [ ] N/A

#### How did you verify/test it?
Unit tests passes on both master and with the change casted to 202605.
- master: Arista internal build on Jul 20, 2026
- 202605: Arista internal build on Jul 20, 2026
`syncd` is also able to boot on physical hardware and enter the expected code path based on whether the argument was provided for launch or not.

#### Any platform specific information?
Generic

PR #2001, opened 2026-07-20
Auto-use enablePerPortCounterDiscovery syncd arg on Broadcom ASICs

### LINKED ISSUE #28460: sonic-sairedis #1774 (Smart Counter Poll) does O(ports) per-port getStats at counter init, starving orchagent and crashing it on high port-count switches
### Is it platform specific

Generic — affects any switch with a high number of ports. Root cause is in the vendor-agnostic `syncd` FlexCounter counter-initialization code.

### Importance or Severity

High — `orchagent` aborts during initialization; ports are never programmed and all interfaces stay `oper-down` (no data plane). Reproducible on every boot / `config reload`.

### Description of the bug

sonic-sairedis PR **#1774** ("Smart Counter Poll to allow counters to work properly on Broadcom platforms") changed FlexCounter counter initialization from *"assume all ports support the same counters"* to **per-port capability discovery**:
- https://github.com/sonic-net/sonic-sairedis/pull/1774

**What the change does at counter init (in `syncd`):**
- Runs a **capability check** (`queryStatsCapability`) to get the set of counters the object type supports — this is cheap and done once.
- Then, **for every port**, issues **individual `getStats` calls** to empirically determine which of those counters that specific port actually supports (a `getStats` of the candidate counter set per port, and per-counter `getStats` in the fallback). Ports are then bucketed into "counter groups" by what they support.

**Why it breaks on high port-count switches:**
- The per-port `getStats` step turns counter init into an **O(number-of-ports)** loop of live hardware reads.
- A single `getStats` call costs on the order of **~136 ms**, and on a **512-port** switch that is **512 such calls ≈ ~70 seconds** just to discover per-port capability.
- While this loop runs, `syncd` holds vendor SAI/SDK access serialized. `orchagent`'s concurrent **synchronous** SAI calls (e.g. the `sai_query_stats_capability` it issues during counter setup) cannot be serviced — they **starve** behind the discovery loop.
- Those synchronous calls hit the **60 s `getresponse` timeout**, which desynchronizes the request/response channel; `orchagent` then throws an **uncaught exception** and aborts (SIGABRT) in its `FLEX_COUNTER_TABLE` drain. Ports are never programmed.

Note this is the *best case* (uniform ports, one `getStats` per port); heterogeneous ports fall back to per-counter probing and are even more expensive.

**Confirmed on a 512-port switch (instrumented `syncd`/`orchagent`):** the entire per-port discovery runs inside a **single** `syncd` event. `syncd` processes its request queue with **one thread, in order**, so that one event holds the queue at the head for **~74 s** (`elapsed_ms=74220`, 514 ports) and blocks everything behind it. `orchagent`'s next synchronous call therefore waits past its **60 s** `getresponse` timeout → gets 0 values → aborts:

```
discovery START             t=0
orchagent TIMEOUT           t=+60.06 s   (60 s sync-op timeout)
discovery DONE              t=+74.2 s
orchagent -> 0 values -> std::runtime_error -> SIGABRT (FLEX_COUNTER_TABLE drain)
```

The failure condition is simply `ports × per-port getStats > 60 s`, so it is **not platform specific**: invisible at low port counts, hit by any platform carrying #1774 once the count is high enough (empirically > ~256 ports).

### Steps to Reproduce

1. Boot (or run `sudo config reload -y`) an image that includes sonic-sairedis #1774, on a switch with a **high port count**.
2. Watch `syncd` counter init and `orchagent` in the logs, and `show interface status`.

**Actual:** During counter init, `syncd` spends tens of seconds in the per-port `getStats` discovery loop; `orchagent`'s synchronous SAI call times out at 60 s, and `orchagent` aborts (exception in the `FLEX_COUNTER_TABLE` drain) and hangs. All interfaces remain `oper-down`.

**Expected:** Counter init completes within the SAI response-timeout window; `orchagent` stays up and programs ports; interfaces come up.

### Relevant log output

```
# orchagent - synchronous SAI call to syncd times out during counter init, then aborts
ERR  swss#orchagent: wait: SELECT operation result: TIMEOUT on getresponse
ERR  swss#orchagent: waitForGetResponse: logic error, get response returned 0 values!
ERR  swss#orchagent: drain: Exception caught: table=FLEX_COUNTER_TABLE, error=...0 values!
     swss#orchagent: terminate called after throwing an instance of 'std::runtime_error'   # SIGABRT

# syncd - per-port counter-init discovery: one getStats per port, ~136 ms each, ~512 ports
#   bulkAdd START  group=PORT  nports=512
#   ... 512 x [ per-port getStats, ~136 ms each ] ...
#   grouping DONE  group=PORT  nports=512  elapsed_ms=~70000
```

Image includes sonic-sairedis #1774. `show techsupport` available on request / to be attached.

### Attach files (if any)

_(attach `generate_dump` techsupport with the orchagent core + syncd/swss logs)_

justin-wong-ce:
### Description of PR

Summary:
Auto-use the `enablePerPortCounterDiscovery` `syncd` arg for Broadcom ASICs in an easily expandable way such that other ASIC types can be easily added to use this behavior.

This is part 2 of 2 fix in addressing https://github.com/sonic-net/sonic-buildimage/issues/28460. There are two fix methods under consideration, this PR outlines the method of deciding code path via the ASIC type.

This change requires the changes in https://github.com/sonic-net/sonic-sairedis/pull/2000 to work.

Fixes -
Partially Fixes: https://github.com/sonic-net/sonic-buildimage/issues/28460

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

In #1774 , a new counter support discovery method is introduced for being able to read counter support for platforms where different ports can have different capabilities.
The method worked well for some platforms, not so much for some others.

This change auto-uses the new `syncd` arg `enablePerPortCounterDiscovery` (introduced in https://github.com/sonic-net/sonic-sairedis/pull/2000) when the `broadcom` ASIC type is detected in `syncd_init_common.sh`.

##### Work item tracking

#### How did you do it?
Add logic in `syncd_init_common.sh` to auto apply the `syncd` arg when `broadcom` ASIC type is detected.

`syncd` unit tests pass on master and on 202605.
On physical hardware, an image is built with the changes cherry-picked to 202605 and tests are ran in 2 scenarios:
1) With the change as is
2) `syncd_init_common.sh` hotpatched to filter on another asic type.

In both scenarios, `syncd` enters the expected code path.
Additionally, a subset of counters-focused `sonic-mgmt` tests are ran. There is no test result difference compared to tests done without this change.
Tests were ran on `Arista-7260CX3-D108C8` and `Arista-7060X6-64PE-B-C512S2`.

The list of `sonic-mgmt` tests ran:
```
test_pretest.py
dhcp_relay/test_dhcp_counter_stress.py
drop_packets/test_drop_counters.py
drop_packets/test_configurable_drop_counters.py
gnmi/test_gnmi_countersdb.py
snmp/test_snmp_queue_counters.py
test_posttest.py
```

To ensure no warm and fast reboot issues from any potential time increase from `syncd`, some reboot-related tests are also ran. The following `sonic-mgmt` tests pass with this change casted, tested on `Arista-7260CX3-D108C8` and `Arista-7060CX-32S-C32`:
```
platform_tests/test_advanced_reboot.py::test_warm_reboot_sad*
platform_tests/test_advanced_reboot.py::test_warm_reboot_mac_jump
platform_tests/test_advanced_reboot.py::test_warm_reboot
platform_tests/test_advanced_reboot.py::test_fast_reboot
```

Broadcom

### DIFF
--- a/syncd/scripts/syncd_init_common.sh
+++ b/syncd/scripts/syncd_init_common.sh
@@ -66,6 +66,12 @@ if [ "$SUPPORTING_BULK_COUNTER_GROUPS" != "" ]; then
     CMD_ARGS+=" -B $SUPPORTING_BULK_COUNTER_GROUPS"
 fi
 
+case "$SONIC_ASIC_TYPE" in
+    broadcom)
+        CMD_ARGS+=" -G"
+        ;;
+esac
+
 case "$(cat /proc/cmdline)" in
   *SONIC_BOOT_TYPE=fastfast*)
     if [ -e /var/warmboot/warm-starting ]; then
--- a/syncd/scripts/syncd_init_common.sh
+++ b/syncd/scripts/syncd_init_common.sh
@@ -66,6 +66,8 @@ if [ "$SUPPORTING_BULK_COUNTER_GROUPS" != "" ]; then
     CMD_ARGS+=" -B $SUPPORTING_BULK_COUNTER_GROUPS"
 fi
 
+# Enable per-port counter capability discovery (see sonic-buildimage#28460, #1774).
+# Add ASIC types that needs this feature here
 case "$SONIC_ASIC_TYPE" in
     broadcom)
         CMD_ARGS+=" -G"


lolyu:
Overall the approach is sound, but a few points worth addressing:

**1. Placement — `-G` is added outside `config_syncd()`, unlike every other ASIC arg.**
All ASIC-specific arg logic lives in `config_syncd_*()` functions dispatched from `config_syncd()`, and there's already a `config_syncd_bcm()` doing `SONIC_ASIC_SUBTYPE`-gated handling. Adding this new `case` at top level (right after the bulk-counter block) splits Broadcom handling across two places, which a future reader won't expect. Moving `-G` into `config_syncd_bcm()` keeps all broadcom logic together and consistent with the rest of the file.

If the intent is genuinely a multi-ASIC-expandable table via one `case`, that's a fair design goal — but then it shouldn't be broadcom-only sitting outside the dispatch.

**2. No explanatory comment.** Every other non-obvious arg here has at least a hint. A bare `-G` tells the reader nothing. Please add a comment referencing the motivation (per-port counter discovery, #1774 / sonic-buildimage#28460).

Suggested:
```bash
# Enable per-port counter capability discovery (see sonic-buildimage#28460, #1774).
# Broadcom platforms can have per-port counter support differences.
CMD_ARGS+=" -G"
```

**3. Boot-type ordering.** This runs before `config_syncd()` / warm-fast-reboot arg assembly, so `-G` is unconditionally present across all boot types including warm/fast/express reboot. Presumably intended, but worth confirming per-port counter discovery is harmless (and desired) during warm/fast reboot.

**4. Hard dependency on #2000.** `-G` is unknown to syncd until #2000 merges; if merge order is reversed, syncd fails to start on Broadcom. Please ensure #2000 lands first (worth calling out explicitly / gating).

justin-wong-ce:
> If the intent is genuinely a multi-ASIC-expandable table via one `case`, that's a fair design goal — but then it shouldn't be broadcom-only sitting outside the dispatch.

This is the intention, however it is currently unknown which platform the issue is affecting, therefore I do not know what other additional platforms to add here.

> **2. No explanatory comment.** Every other non-obvious arg here has at least a hint. A bare `-G` tells the reader nothing. Please add a comment referencing the motivation (per-port counter discovery, #1774 / sonic-buildimage#28460).

Ack, will add.

> **3. Boot-type ordering.** This runs before `config_syncd()` / warm-fast-reboot arg assembly, so `-G` is unconditionally present across all boot types including warm/fast/express reboot. Presumably intended, but worth confirming per-port counter discovery is harmless (and desired) during warm/fast reboot.

This is intentional. Warm and fast reboot will need to run this in order for counters to be setup properly in those scenarios as well.
The following `sonic-mgmt` tests pass with this change casted, tested on `Arista-7260CX3-D108C8` and `Arista-7060CX-32S-C32`:
```
platform_tests/test_advanced_reboot.py::test_warm_reboot_sad*
platform_tests/test_advanced_reboot.py::test_warm_reboot_mac_jump
platform_tests/test_advanced_reboot.py::test_warm_reboot
platform_tests/test_advanced_reboot.py::test_fast_reboot
```
Will add this testing info to the PR description.

> **4. Hard dependency on #2000.** `-G` is unknown to syncd until #2000 merges; if merge order is reversed, syncd fails to start on Broadcom. Please ensure #2000 lands first (worth calling out explicitly / gating).

Yes.

justin-wong-ce:
Reminder this needs backport to 202605.

lolyu (approved):
LGTM

rookie-who:

justin-wong-ce:
> Hi — this PR is flagged for cherry-pick to the **202605 branch** (`Request for 202605 Branch`). To move forward with the cherry-pick, could you please:
> 
> 1. **Test this change on the 202605 branch** and add the test results to the PR description
> 2. **Add the label `Tested for 202605 branch`** once testing is complete
> 
> This helps us track readiness and ensures we don't cherry-pick untested changes. Thanks!
> 
> _Comment by @vaibhavhd via automated tooling_

The test changes are included in the PR description.

PR #2004, opened 2026-07-23
sFlow Phase Three Sairedis Changes

ritvikiscool9:
### Description of PR

Summary:

This PR contains the changes made to sonic-sairedis regarding phase 3 of sFlow. Specifically, it adds support for per-port sampling rate and per-port direction.

It goes hand-in-hand with this [PR](https://github.com/sonic-net/sonic-buildimage/pull/29080), which loads PSAMPLE for VPP so packets can be sampled.

Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

The motivation is to add support to SONiC-VPP so that every port can have its own sampling rate and direction.

##### Work item tracking

#### How did you do it?

In `SwitchVppSflow.cpp`, `sflowPortSamplePacketSet()` now translates the `INGRESS/EGREE_SAMPLEPACKET_ENABLE` sets into per-port VPP config. It reads the SAMPLEPACKET rate, combines ingress/egress into one direction mask, and programs the port via `sflowInterfaceSamplingRateSet()`, `sflowInterfaceDirectionSet()`, and `sflowEnableDisable()`. In `SaiVppXlate.c` I added the client send functions for those per-interface APIs plus their reply handlers and registrations, without the handlers the calls hung until syncd's 30s watchdog and interfaces never armed. Together these let per-port rate and direction flow from SONiC through SAI into VPP.

My testing environment consisted of two routers who were wired together. r1 is the traffic sender and r2 is the collector that had sFlow running. I configured Ethernet4 and Ethernet8 on both routers:

r1: 
```
sudo config interface ip add Ethernet4 10.0.1.1/24
sudo config interface ip add Ethernet8 10.0.2.1/24
sudo config save -y
```

r2:
```
sudo config interface ip add Ethernet4 10.0.1.2/24
sudo config interface ip add Ethernet8 10.0.2.2/24
sudo config save -y
```  

The HLD has 3 exit criteria conditions that must be met. 

1. Configure two ports with different rates (e.g. 1000 and 5000); collector shows distinct effective rates per port, both reported correctly.

I configured Ethernet4 to have a sampling rate of 1000 and Ethernet8 to have a sampling rate of 5000:
image

I then sent each port 20 000 packets: 
image

Here is how much each port sampled:
image

Ethernet4 sampled 24 of 20,000 packets and Ethernet8 sampled 5 of 20,000 packets, consistent with their configured rates of 1-in-1000 (≈20 expected) and 1-in-5000 (≈4 expected). sFlow sampling is statistical, not a strict every-Nth count, so the per-port totals scatter around the expected value rather than matching it exactly. The rate is calculated through a per-interface skip counter: it is drawn as a random value in the range [1, 2N-1]  and counts down as packets pass through the port. When it reaches zero the packet is sampled and a new skip is calculated. Because the mean skip equals N, the interface samples 1-in-N on average while randomizing which packets are selected.

2. Configure one port ingress-only and another port both-direction; collector shows ingress flow samples on the first and ingress + egress flow samples on the second; the global `sflow direction` setting does not override either.

For this criteria, I configured Ethernet4 to be ingress only and Ethernet8 to be both, I also set the global direction to be egress, to test if the global direction will be over-ridden by per-port direction: 
image

I then sent each port 40 000 packets:
image

The results were:
image

Each port is behaving as expected, Ethernet4 is sampling only packets on ingress and Ethernet8 is sampling packets in both directions.

3. Per-port `samplingN` and `skip` visible in show `sflow per-port output`.

`docker exec -it syncd vppctl show sflow` returns the following:
image

This command prints the global sFlow settings followed by a per-interface block for each enabled port. Each block shows the interface's effective sampling-rate (its per-port `samplingN`, or the global rate if none is set) and, for every VPP thread, that thread's live skip counter — the number of packets remaining until the next sample. In this run Ethernet4 (bobm1) reports sampling-rate 1000 and Ethernet8 (bobm2) reports sampling-rate 5000, each with a per-thread skip in the [1, 2N-1] range, confirming that per-port sampling state is now observable directly from the data plane.


vslib/vpp/vppxlate/SaiVppXlate.c
```diff
@@ -3531,6 +3207,87 @@ int vpp_ip_flow_hash_set (uint32_t vrf_id, uint32_t hash_mask, int addr_family)
...
+        u32 idx = get_swif_idx(vam, hwif_name);
+        if(idx != (u32) - 1){
+            mp->hw_if_index = htonl(idx);
+        } else {
+            SAIVPP_ERROR("Unable to get the sw_index for %s\n", hwif_name);
+            VPP_UNLOCK();
+            return -EINVAL;
+        }
+    } else {
+        SAIVPP_ERROR("No hw_index provided");
```
yue-fred-gao: argument is hwif_name not hw_index

vslib/vpp/SwitchVppSflow.cpp
```diff
@@ -170,34 +170,121 @@ sai_status_t SwitchVpp::sflowPortSamplePacketSet(
...
+        SWSS_LOG_ERROR("Unexpected sFlow port attribute ID %u", attr->id);
+        return SAI_STATUS_INVALID_PARAMETER;
+    }
+
+    bool updating_ingress = attr->id == SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE;
+    sai_object_id_t updated_oid = attr->value.oid;
+
+    sai_attribute_t other_attr{};
+
+    other_attr.id = updating_ingress
```
yue-fred-gao: please consider to rewrite this function using SaiObject. From port SaiObject, you can use get_linked_object to get SAI_OBJECT_TYPE_SAMPLEPACKET, from which you can get attribute like sample rate with get_attr or get_mandatory_attr.
Because the attr argument of this function is from update, which is not written to the object in db yet, you can create a SaiCachedObject that is backed by the object in db. get_attr or get_linked_object for attribute in the update returns the updated value in the update. otherwise, it returns the one from object in db from previous create/update.

vslib/vpp/vppxlate/SaiVppXlate.c
```diff
@@ -3531,6 +3207,87 @@ int vpp_ip_flow_hash_set (uint32_t vrf_id, uint32_t hash_mask, int addr_family)
...
+        SAIVPP_ERROR("%s failed(%d) %s sampling_N %u", __func__, ret, hwif_name, sampling_n);
+    } else {
+        SAIVPP_INFO("%s %s sampling_N %u", __func__, hwif_name, sampling_n);
+    }
+
+    VPP_UNLOCK();
+    return ret;
+}
+
+int vpp_sflow_interface_direction_set(const char *hwif_name, uint32_t direction)
```
yue-fred-gao, line 3905: how is direction both supported?

ritvikiscool9: `direction` is a bitmask, not an enum: 1 = ingress, 2 = egress, 3 = both and `vpp_sflow_interface_direction_set` marshals it.

SAI sets the ingress and egress independently (`SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE/...EGRESS..`), while VOO takes on per-interfaction direction. So `sflowPortSamplePacketSet` ORs the attribute being set with the samplepacket still bound to the opposite direction - both bound gives 3. Neither bound disables sFlow on the port. 

I verified with `sample-direction both` on Eth8, the collector receives 25 ingress and 18 egress samples.

yue-fred-gao (dismissed):

yue-fred-gao (approved):

PR #2008, opened 2026-07-29
[vpp] Add MPLS data-plane support (INSEG + IP-route push)

augusdn:
### Description of PR

Summary:
Part of sonic-net/sonic-buildimage#25782

Adds MPLS data-plane support to the VPP SAI backend so the `tests/mpls` data-plane
tests can run on the sonic-vpp testbed. Today `SAI_OBJECT_TYPE_INSEG_ENTRY` is not
handled at all in `vslib/vpp`, so MPLS traffic is never programmed into VPP.

> [!NOTE]
> Companion PR sonic-net/sonic-mgmt#26619 enables `mpls/test_mpls.py` on sonic-vpp and
> depends on this change being in the image, so it merges after this one.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

sonic-buildimage#25782 asks for MPLS support in sonic-vpp so that `mpls/test_mpls.py`
can be enabled on the VPP KVM testbed. The tests currently can't run because the VPP
SAI backend has no INSEG handling.

##### Work item tracking

#### How did you do it?

10 files, +737 / -5, all under `vslib/vpp`:

| File | Change |
|---|---|
| `vslib/vpp/SwitchVppMpls.cpp` (new) | INSEG create/remove: pop, swap and push disposition into the VPP MPLS FIB. |
| `vslib/vpp/SwitchVpp.{cpp,h}` | Dispatch `SAI_OBJECT_TYPE_INSEG_ENTRY` from `create()`/`remove()`; declare the MPLS helpers and the MPLS-table state. |
| `vslib/vpp/SwitchVppNexthop.{cpp,h}` | Carry the out-label stack on next hops; `n_labels` is explicitly zeroed in `fillNHGrpMember()`. |
| `vslib/vpp/SwitchVppRoute.cpp` | Impose the out-label stack on IP routes (ingress LER), gated on `n_labels > 0`. |
| `vslib/vpp/SwitchVppRif.cpp` | Honour `SAI_ROUTER_INTERFACE_ATTR_ADMIN_MPLS_STATE` (`sw_interface_set_mpls_enable`, ensure the MPLS table exists). |
| `vslib/vpp/vppxlate/SaiVppXlate.{c,h}` | `mpls_table_add_del()`, `mpls_route_add_del()`, and the label stack on fib paths for both MPLS and IP routes. |
| `vslib/Makefile.am` | Build the new source file. |

Details worth a reviewer's attention:

- *pop*: inject an implicit-null out-label in UNIFORM LSP mode so the disposition
  derives the inner IP TTL from the popped MPLS TTL. Without it VPP defaults the
  disposition to PIPE and the inner TTL is only decremented once by the IP stage.
- *swap/push*: read the SAI next hop's out-label stack. `SAI_NEXT_HOP_ATTR_LABELSTACK`
  is a list attribute, so the output buffer has to be supplied before the get,
  otherwise the read fails and the labels are silently dropped - which turns a swap
  into a bare pop.
- Resolve the next hop's router interface to its VPP egress hwif so the path is
  programmed *attached* rather than recursive. VPP only inserts the MPLS disposition
  for an attached next hop, and a recursive labelled path fails to resolve and is
  dropped at the MPLS DROP DPO.
- The SAI INSEG key has no end-of-stack qualifier but the VPP MPLS FIB is keyed by
  `{label, eos}`, so eos=1 is always programmed and eos=0 as well when out-labels are
  present, so a non-bottom label in a stack is handled.
- `mpls_route_add_del()` validates every path (address family and label count) *before*
  allocating the API message, matching what `ip_route_add_del_get_stats()` already does,
  so the error path can't leak the message.

Status is propagated the same way the neighbouring `addIpRoute()`/`removeIpRoute()`
do it: `addMplsRoute()` rolls back the VPP programming if `create_internal()` fails,
and `ensureMplsTable()` only latches its "created" flag once the table really exists.

**With this change in the image.** A throwaway sonic-buildimage build carrying this PR
produced image `SONiC.master-28652.1177040-4e2ccc26c`. Run on a `vms-kvm-vpp-t1-lag` KVM
testbed together with the companion sonic-mgmt PR, ElasticTest plan
`6a69da68f481df03c4e59c5e` - **SUCCESS, 22 tests, 17 passed / 5 skipped / 0 failed /
0 errors**:

```
mpls/test_mpls.py::TestBasicMpls::test_pop_label        PASSED
mpls/test_mpls.py::TestBasicMpls::test_swap_label       PASSED
mpls/test_mpls.py::TestBasicMpls::test_push_label       SKIPPED
mpls/test_mpls.py::TestBasicMpls::test_swap_labelstack  PASSED

3 passed, 1 skipped
```

**Without this change in the image (negative control).** The dependency was verified
rather than assumed. The identical test code, run by the `t1-lag-vpp` PR checker on a
stock sonic-vpp image that does *not* carry this PR, fails:

```
mpls/test_mpls.py|||2 failed for RUN_TEST_CASE_FAILED
```

(ElasticTest plan `6a70603deb2c1b23503d2e2f`, build 1182699; same signature on the
earlier build 1178163.) `mpls/test_mpls.py` is the only failure in that run - the job
executes the whole `t1-lag-vpp` list. That is exactly the gap this PR closes.

**Regression check.** The same image also ran the full `t1-lag-vpp` suite
(1762 tests, 1222 passed) to check this doesn't regress anything else on VPP.

`test_push_label` is skipped for a reason outside this change: it injects the push
route directly into `ROUTE_TABLE` and orchagent does not install that route into
ASIC_DB, so it never reaches the SAI backend.

sonic-vpp only - everything is under `vslib/vpp`. Non-MPLS traffic is unaffected: the
INSEG path only runs for INSEG objects, the next-hop label handling is gated on
`SAI_NEXT_HOP_TYPE_MPLS`, and label imposition is gated on `n_labels > 0`.

Known limitations, called out deliberately rather than half-implemented - happy to
follow up on any of these if reviewers would prefer them in scope:

- IPv6 payloads: `eos_proto_af` is IPv4.
- ECMP / next-hop-group MPLS next hops are not programmed.
- INSEG `set()` does not reprogram VPP (create/remove only).
- `NUM_OF_POP` > 1, and non-default POP TTL/QoS modes, are not implemented.

### Documentation

No user-facing documentation change.

### DIFF
--- a/vslib/vpp/SwitchVpp.h
+++ b/vslib/vpp/SwitchVpp.h
@@ -788,6 +788,11 @@ namespace saivs
             sai_status_t fillMplsNexthop(
                     _In_ const SaiObject *nh_obj,
                     _Out_ vpp_mpls_nexthop_t *vnh);
+            void getOutsegTtl(
+                    _In_ const SaiObject *nh_obj,
+                    _Out_ uint8_t *ttl,
+                    _Out_ uint8_t *exp,
+                    _Out_ uint8_t *is_uniform);
             sai_status_t ensureMplsTable();
 
             sai_status_t IpRouteNexthopEntry(
--- a/vslib/vpp/SwitchVpp.h
+++ b/vslib/vpp/SwitchVpp.h
@@ -781,7 +781,7 @@ namespace saivs
                     _In_ const sai_attribute_t *attr_list);
             sai_status_t removeMplsRoute(
                     _In_ const std::string &serializedObjectId);
-            sai_status_t MplsRouteAddRemove(
+            sai_status_t mplsRouteAddRemove(
                     _In_ const SaiObject *inseg_obj,
                     _In_ const std::string &serializedObjectId,
                     _In_ bool is_add);
--- a/vslib/vpp/SwitchVpp.h
+++ b/vslib/vpp/SwitchVpp.h
@@ -774,6 +774,22 @@ namespace saivs
             sai_status_t removeIpRoute(
                     _In_ const std::string &serializedObjectId);
 
+            sai_status_t addMplsRoute(
+                    _In_ const std::string &serializedObjectId,
+                    _In_ sai_object_id_t switch_id,
+                    _In_ uint32_t attr_count,
+                    _In_ const sai_attribute_t *attr_list);
+            sai_status_t removeMplsRoute(
+                    _In_ const std::string &serializedObjectId);
+            sai_status_t MplsRouteAddRemove(
+                    _In_ const SaiObject *inseg_obj,
+                    _In_ const std::string &serializedObjectId,
+                    _In_ bool is_add);
+            sai_status_t fillMplsNexthop(
+                    _In_ const SaiObject *nh_obj,
+                    _Out_ vpp_mpls_nexthop_t *vnh);
+            sai_status_t ensureMplsTable();
+
             sai_status_t IpRouteNexthopEntry(
                     _In_ uint32_t attr_count,
                     _In_ const sai_attribute_t *attr_list,
@@ -1325,6 +1341,7 @@ namespace saivs
             // SRv6 object tracking for CRM
             constexpr static const int m_maxMySidEntries = 1000;
             uint32_t m_srv6_my_sid_count = 0;
+            bool m_mpls_table_created = false;
 
             std::shared_ptr<RealObjectIdManager> m_realObjectIdManager;
 
--- a/vslib/vpp/SwitchVppMpls.cpp
+++ b/vslib/vpp/SwitchVppMpls.cpp
@@ -207,13 +207,20 @@ sai_status_t SwitchVpp::MplsRouteAddRemove(
     route->label = inseg_entry.label;
     route->is_multipath = false;
     route->nexthop_cnt = 1;
-    route->eos_proto_af = AF_INET;
 
     sai_status_t status = fillMplsNexthop(nh_obj.get(), &route->nexthop[0]);
     if (status != SAI_STATUS_SUCCESS) {
         free(route);
         return status;
     }
+
+    /*
+     * Protocol of the payload exposed once the end-of-stack label is popped.
+     * The SAI INSEG entry does not carry it, so derive it from the resolved
+     * next hop's address family. Must run after fillMplsNexthop() has populated
+     * the address.
+     */
+    route->eos_proto_af = route->nexthop[0].addr.sa_family;
 
     bool has_outlabels = (route->nexthop[0].n_labels > 0);
 
--- a/vslib/vpp/SwitchVppMpls.cpp
+++ b/vslib/vpp/SwitchVppMpls.cpp
@@ -1,0 +1,279 @@
+/*
+ * VPP MPLS backend for SAI INSEG (MPLS in-segment / local label) entries.
+ * Issue: sonic-buildimage#25782 - enable MPLS data plane testing on VPP.
+ *
+ * Translates SAI_OBJECT_TYPE_INSEG_ENTRY into VPP MPLS FIB programming via
+ * mpls_route_add_del(). Pop (IP next-hop, no out-labels) and swap/push
+ * (SAI_NEXT_HOP_TYPE_MPLS next-hop carrying an out-label stack) are supported.
+ */
+#include "SwitchVpp.h"
+#include "SwitchVppUtils.h"
+
+#include "meta/sai_serialize.h"
+
+#include "swss/logger.h"
+
+#include "vppxlate/SaiVppXlate.h"
+
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <cstdlib>
+#include <cstring>
+
+using namespace saivs;
+
+/* VPP MPLS FIB table used for all SONiC in-segment entries. */
+#define MPLS_TABLE_ID 0
+/* Reserved MPLS implicit-null label (RFC 3032): pop and forward on the payload. */
+#define MPLS_IMPLICIT_NULL_LABEL 3
+/* Default TTL for an imposed label; SAI carries no per-outsegment TTL here. */
+#define MPLS_DEFAULT_OUT_TTL 64
+
+sai_status_t SwitchVpp::ensureMplsTable()
+{
+    SWSS_LOG_ENTER();
+
+    if (m_mpls_table_created) {
+        return SAI_STATUS_SUCCESS;
+    }
+
+    int ret = mpls_table_add_del(MPLS_TABLE_ID, true);
+    if (ret != 0) {
+        SWSS_LOG_ERROR("Failed to create VPP MPLS table %u: %d", MPLS_TABLE_ID, ret);
+        return SAI_STATUS_FAILURE;
+    }
+
+    m_mpls_table_created = true;
+    SWSS_LOG_NOTICE("Created VPP MPLS table %u", MPLS_TABLE_ID);
+
+    return SAI_STATUS_SUCCESS;
+}
+
+sai_status_t SwitchVpp::fillMplsNexthop(
+        _In_ const SaiObject *nh_obj,
+        _Out_ vpp_mpls_nexthop_t *vnh)
+{
+    SWSS_LOG_ENTER();
+
+    sai_attribute_t attr;
+
+    memset(vnh, 0, sizeof(*vnh));
+    vnh->sw_if_index = (uint32_t)~0;   /* let VPP resolve egress recursively by IP */
+    vnh->hwif_name = NULL;
+    vnh->weight = 1;
+    vnh->preference = 0;
+    vnh->type = VPP_NEXTHOP_NORMAL;
+    vnh->n_labels = 0;
+
+    attr.id = SAI_NEXT_HOP_ATTR_TYPE;
+    CHECK_STATUS_W_MSG(nh_obj->get_attr(attr), "MPLS path nexthop missing TYPE");
+    int32_t nh_type = attr.value.s32;
+
+    if (nh_type != SAI_NEXT_HOP_TYPE_IP && nh_type != SAI_NEXT_HOP_TYPE_MPLS) {
+        SWSS_LOG_ERROR("Unsupported MPLS path nexthop type %d", nh_type);
+        return SAI_STATUS_NOT_IMPLEMENTED;
+    }
+
+    attr.id = SAI_NEXT_HOP_ATTR_IP;
+    CHECK_STATUS_W_MSG(nh_obj->get_attr(attr), "MPLS path nexthop missing IP");
+    sai_ip_address_t_to_vpp_ip_addr_t(attr.value.ipaddr, vnh->addr);
+
+    if (nh_type == SAI_NEXT_HOP_TYPE_MPLS) {
+        /*
+         * LABELSTACK is a list attribute: the caller must supply the output
+         * buffer (list pointer + count) before calling get_attr, otherwise the
+         * read fails and the out labels are silently dropped (turning a
+         * swap/push into a bare pop).
+         */
+        uint32_t label_buf[VPP_MPLS_MAX_LABELS];
+        attr.id = SAI_NEXT_HOP_ATTR_LABELSTACK;
+        attr.value.u32list.count = VPP_MPLS_MAX_LABELS;
+        attr.value.u32list.list = label_buf;
+        if (nh_obj->get_attr(attr) == SAI_STATUS_SUCCESS && attr.value.u32list.count > 0) {
+            uint32_t cnt = attr.value.u32list.count;
+            if (cnt > VPP_MPLS_MAX_LABELS) {
+                cnt = VPP_MPLS_MAX_LABELS;
+            }
+            vnh->n_labels = (uint8_t)cnt;
+            for (uint32_t i = 0; i < cnt; i++) {
+                vnh->label_stack[i].label = attr.value.u32list.list[i];
+                vnh->label_stack[i].ttl = MPLS_DEFAULT_OUT_TTL;
+                vnh->label_stack[i].exp = 0;
+                vnh->label_stack[i].is_uniform = 1;
+            }
+        }
+    }
+
+    return SAI_STATUS_SUCCESS;
+}
+
+sai_status_t SwitchVpp::MplsRouteAddRemove(
+        _In_ const SaiObject *inseg_obj,
+        _In_ const std::string &serializedObjectId,
+        _In_ bool is_add)
+{
+    SWSS_LOG_ENTER();
+
+    sai_inseg_entry_t inseg_entry;
+    sai_deserialize_inseg_entry(serializedObjectId, inseg_entry);
+
+    sai_attribute_t attr;
+    int32_t action = SAI_PACKET_ACTION_FORWARD;
+    attr.id = SAI_INSEG_ENTRY_ATTR_PACKET_ACTION;
+    if (inseg_obj->get_attr(attr) == SAI_STATUS_SUCCESS) {
+        action = attr.value.s32;
+    }
+    if (action != SAI_PACKET_ACTION_FORWARD) {
+        SWSS_LOG_NOTICE("Ignoring inseg label %u: packet action %d is not forward",
+                        inseg_entry.label, action);
+        return SAI_STATUS_SUCCESS;
+    }
+
+    auto nh_obj = inseg_obj->get_linked_object(SAI_OBJECT_TYPE_NEXT_HOP,
+                                               SAI_INSEG_ENTRY_ATTR_NEXT_HOP_ID);
+    if (!nh_obj) {
+        SWSS_LOG_NOTICE("Ignoring inseg label %u: no resolvable nexthop", inseg_entry.label);
+        return SAI_STATUS_SUCCESS;
+    }
+
+    CHECK_STATUS(ensureMplsTable());
+
+    vpp_mpls_route_t *route = (vpp_mpls_route_t *)
+        calloc(1, sizeof(vpp_mpls_route_t) + sizeof(vpp_mpls_nexthop_t));
+    if (!route) {
+        return SAI_STATUS_FAILURE;
+    }
+    route->table_id = MPLS_TABLE_ID;
+    route->label = inseg_entry.label;
+    route->is_multipath = false;
+    route->nexthop_cnt = 1;
+    route->eos_proto_af = AF_INET;
+
+    sai_status_t status = fillMplsNexthop(nh_obj.get(), &route->nexthop[0]);
+    if (status != SAI_STATUS_SUCCESS) {
+        free(route);
+        return status;
+    }
+
+    bool has_outlabels = (route->nexthop[0].n_labels > 0);
+
+    /*
+     * Resolve the next hop's router interface to its VPP egress hwif so the
+     * path is programmed as *attached* rather than recursive. This is required
+     * for both forwarding cases:
+     *   - pop  (disposition): VPP only inserts the MPLS disposition (and thus
+     *     honours the uniform LSP mode) for an attached next hop; a recursive
+     *     next hop collapses to a plain IP forward and ignores the mode.
+     *   - swap/push (imposition): a recursive labelled path fails to resolve
+     *     and the packet is dropped at the MPLS DROP DPO.
+     * If resolution fails we fall back to the recursive (~0) path.
+     */
+    std::string egress_hwif;
+    {
+        sai_attribute_t rif_attr;
+        rif_attr.id = SAI_NEXT_HOP_ATTR_ROUTER_INTERFACE_ID;
+        if (nh_obj->get_attr(rif_attr) == SAI_STATUS_SUCCESS) {
+            sai_object_id_t rif_id = rif_attr.value.oid;
+            sai_attribute_t port_attr;
+            port_attr.id = SAI_ROUTER_INTERFACE_ATTR_PORT_ID;
+            if (get(SAI_OBJECT_TYPE_ROUTER_INTERFACE, rif_id, 1, &port_attr) == SAI_STATUS_SUCCESS &&
+                vpp_get_hwif_name(port_attr.value.oid, 0, egress_hwif)) {
+                route->nexthop[0].hwif_name = egress_hwif.c_str();
+            }
+        }
+    }
+
+    if (!has_outlabels) {
+        /*
+         * Pop/disposition case (no out labels). Inject an implicit-null out
+         * label with uniform LSP mode so VPP builds a UNIFORM-mode MPLS
+         * disposition: the popped inner IP TTL is derived from the popped MPLS
+         * TTL (mpls_ttl - 1). This matches the SAI default POP TTL mode
+         * (SAI_INSEG_ENTRY_POP_TTL_MODE_UNIFORM) and typical hardware. Without
+         * it VPP defaults the disposition to PIPE (inner TTL left untouched and
+         * only decremented once by the IP forwarding stage).
+         */
+        route->nexthop[0].n_labels = 1;
+        route->nexthop[0].label_stack[0].label = MPLS_IMPLICIT_NULL_LABEL;
+        route->nexthop[0].label_stack[0].ttl = 0;
+        route->nexthop[0].label_stack[0].exp = 0;
+        route->nexthop[0].label_stack[0].is_uniform = 1;
+    }
+
+    /*
+     * The SAI INSEG entry carries a single local label with no End-of-Stack
+     * qualifier, but the VPP MPLS FIB is keyed by {label, eos}. Always program
+     * the eos=1 entry (single-label / disposition case). For swap/push (out
+     * labels present) also program eos=0 so a non-bottom label in a stack is
+     * handled (e.g. test_swap_labelstack).
+     */
+    uint8_t eos_list[2];
+    int n_eos = 0;
+    if (has_outlabels) {
+        eos_list[n_eos++] = 0;
+    }
+    eos_list[n_eos++] = 1;
+
+    int ret = 0;
+    for (int e = 0; e < n_eos; e++) {
+        route->eos = eos_list[e];
+        ret = mpls_route_add_del(route, is_add);
+        if (ret != 0) {
+            break;
+        }
+    }
+
+    SWSS_LOG_NOTICE("%s inseg label %u out_labels %u status %d",
+                    (is_add ? "Add" : "Remove"), inseg_entry.label,
+                    route->nexthop[0].n_labels, ret);
+
+    free(route);
+
+    return (ret == 0) ? SAI_STATUS_SUCCESS : SAI_STATUS_FAILURE;
+}
+
+sai_status_t SwitchVpp::addMplsRoute(
+        _In_ const std::string &serializedObjectId,
+        _In_ sai_object_id_t switch_id,
+        _In_ uint32_t attr_count,
+        _In_ const sai_attribute_t *attr_list)
+{
+    SWSS_LOG_ENTER();
+
+    SaiCachedObject inseg_obj(this, SAI_OBJECT_TYPE_INSEG_ENTRY, serializedObjectId, attr_count, attr_list);
+
+    bool route_programmed = false;
+
+    if (is_ip_nbr_active() == true) {
+        CHECK_STATUS(MplsRouteAddRemove(&inseg_obj, serializedObjectId, true));
+        route_programmed = true;
+    }
+
+    sai_status_t status = create_internal(SAI_OBJECT_TYPE_INSEG_ENTRY, serializedObjectId, switch_id, attr_count, attr_list);
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        // The entry was not committed, so undo the VPP programming to keep the
+        // MPLS FIB in sync with the SAI object database.
+        if (route_programmed) {
+            MplsRouteAddRemove(&inseg_obj, serializedObjectId, false);
+        }
+        return status;
+    }
+
+    return SAI_STATUS_SUCCESS;
+}
+
+sai_status_t SwitchVpp::removeMplsRoute(
+        _In_ const std::string &serializedObjectId)
+{
+    SWSS_LOG_ENTER();
+
+    auto inseg_obj = get_sai_object(SAI_OBJECT_TYPE_INSEG_ENTRY, serializedObjectId);
+    if (inseg_obj && is_ip_nbr_active() == true) {
+        CHECK_STATUS(MplsRouteAddRemove(inseg_obj.get(), serializedObjectId, false));
+    }
+
+    CHECK_STATUS(remove_internal(SAI_OBJECT_TYPE_INSEG_ENTRY, serializedObjectId));
+
+    return SAI_STATUS_SUCCESS;
+}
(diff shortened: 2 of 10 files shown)


augusdn:
Thanks — both review comments were valid, and I've pushed fixes in 6c02b43.

### 1. `SAI_STATUS_BUFFER_OVERFLOW` on the label stack

Confirmed by tracing the propagation path rather than assuming:

* `meta/SaiSerialize.cpp:151` — `transfer_list` returns `SAI_STATUS_BUFFER_OVERFLOW` when the caller's buffer is too small, overwrites `count` with the **required** size, and copies **nothing**.
* `vslib/SwitchStateBase.cpp:768` — propagates it as `final_status`.
* `vslib/vpp/SaiObjectDB.cpp:386` / `:398` / `:420` — all three `get_attr` implementations pass it through.

So with `VPP_MPLS_MAX_LABELS = 16`, a deeper stack failed the `== SAI_STATUS_SUCCESS` test, left `n_labels` at 0, and programmed a bare pop instead of swap/push — silent misforwarding, exactly as described. Both sites now detect `BUFFER_OVERFLOW` explicitly, log the required depth, and return `SAI_STATUS_NOT_SUPPORTED`. Any other failure still falls through to "no label stack", which is the legitimate pop/disposition path for a `SAI_NEXT_HOP_TYPE_IP` next hop.

One thing the review didn't mention: the `cnt > VPP_MPLS_MAX_LABELS` clamps in both functions were **dead code**. On `SAI_STATUS_SUCCESS` the returned count can never exceed the supplied 16-entry buffer, so those branches were unreachable and implied a bound check that never actually ran. Removed them rather than leave a misleading safety net.

### 2. Partial two-EOS rollback (found in a separate review pass)

`MplsRouteAddRemove` programs two VPP FIB entries for a swap/push — `eos=0` then `eos=1`. If `eos=0` succeeded and `eos=1` failed, the orphaned `eos=0` entry stayed in the MPLS FIB with no SAI object referring to it, because `addMplsRoute` returns through `CHECK_STATUS(...)` *before* setting `route_programmed = true`, so the existing create-time rollback never ran.

The loop now unwinds whatever it managed to program. Removing an entry that was never added is safe — `vpp_normalize_ret` (`vslib/vpp/vppxlate/SaiVppXlate.c:454`) maps `NO_SUCH_ENTRY` to success on delete.

### Verification

Rebuilt against VPP 2606 to match the sonic-vpp image (the tree's older helper linked 2510, which the DUT can't load) — clean compile, zero warnings under `-Werror -Wall -Wextra -Wconversion -Wshadow -Wcast-align=strict -Wformat=2`. The repo's own gates also pass locally: `aspellcheck.pl`, `checkwhitespace.sh`, `swsslogentercheck.sh`.

Functionally re-verified on a `vms-kvm-vpp-t1-lag` KVM testbed with this exact binary installed in `syncd`:

```
mpls/test_mpls.py::TestBasicMpls::test_pop_label        PASSED
mpls/test_mpls.py::TestBasicMpls::test_swap_label       PASSED
mpls/test_mpls.py::TestBasicMpls::test_push_label       SKIPPED
mpls/test_mpls.py::TestBasicMpls::test_swap_labelstack  PASSED
```

vslib/vpp/SwitchVppMpls.cpp
```diff
@@ -0,0 +1,279 @@
...
+                cnt = VPP_MPLS_MAX_LABELS;
+            }
+            vnh->n_labels = (uint8_t)cnt;
+            for (uint32_t i = 0; i < cnt; i++) {
+                vnh->label_stack[i].label = attr.value.u32list.list[i];
+                vnh->label_stack[i].ttl = MPLS_DEFAULT_OUT_TTL;
+                vnh->label_stack[i].exp = 0;
+                vnh->label_stack[i].is_uniform = 1;
+            }
+        }
```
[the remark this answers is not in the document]
augusdn: Addressed in 6c02b43.

Confirmed the concern by tracing the propagation path rather than assuming: `transfer_list` returns `SAI_STATUS_BUFFER_OVERFLOW` when the caller buffer is too small, overwrites `count` with the **required** size and copies nothing (`meta/SaiSerialize.cpp:151`); `SwitchStateBase::get()` propagates it (`vslib/SwitchStateBase.cpp:768`); and all three `get_attr` implementations pass it through (`vslib/vpp/SaiObjectDB.cpp:386`, `:398`, `:420`).

So with `VPP_MPLS_MAX_LABELS = 16` a deeper stack failed the `== SAI_STATUS_SUCCESS` test, left `n_labels` at 0, and programmed a bare pop instead of swap/push - silent misforwarding, exactly as described. It now detects `BUFFER_OVERFLOW` explicitly, logs the required depth and returns `SAI_STATUS_NOT_SUPPORTED`. Other failures still fall through to "no label stack", which is the legitimate pop/disposition path for a `SAI_NEXT_HOP_TYPE_IP` next hop.

One thing not in the original comment: the `cnt > VPP_MPLS_MAX_LABELS` clamp here was **dead code**. On `SAI_STATUS_SUCCESS` the returned count can never exceed the supplied 16-entry buffer, so that branch was unreachable and implied a bound check that never ran. Removed rather than left as a misleading safety net.

vslib/vpp/SwitchVppNexthop.cpp
```diff
@@ -187,13 +188,53 @@ SwitchVpp::fillNHGrpMember(nexthop_grp_member_t *nxt_grp_member, sai_object_id_t
...
+            attr.value.u32list.list = lbuf;
+            if (get(SAI_OBJECT_TYPE_NEXT_HOP, next_hop_oid, 1, &attr) == SAI_STATUS_SUCCESS &&
+                attr.value.u32list.count > 0) {
+                uint32_t cnt = attr.value.u32list.count > VPP_MPLS_MAX_LABELS ?
+                               VPP_MPLS_MAX_LABELS : attr.value.u32list.count;
+                nxt_grp_member->n_labels = (uint8_t)cnt;
+                for (uint32_t li = 0; li < cnt; li++) {
+                    nxt_grp_member->label_stack[li] = attr.value.u32list.list[li];
+                }
+            }
```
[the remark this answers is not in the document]
augusdn: Addressed in 6c02b43, same fix as the `SwitchVppMpls.cpp` site.

`fillNHGrpMember` now captures the status of the `SAI_NEXT_HOP_ATTR_LABELSTACK` read, returns `SAI_STATUS_NOT_SUPPORTED` on `SAI_STATUS_BUFFER_OVERFLOW` (logging the required depth), and only treats a genuine success with `count > 0` as "has labels". Previously a too-deep stack silently programmed a bare IP next hop, dropping the imposed labels.

Checked that returning an error here is safe for every caller: all three (`SwitchVppNexthop.cpp:90`, `:137`, `:494`) check the return value, and `IpRouteNexthopEntry` `free()`s the allocated `nexthop_grp_config_t` before propagating, so there is no leak on the new error path.

The redundant `cnt > VPP_MPLS_MAX_LABELS` clamp was removed here too, for the same reason as the other site.

augusdn:
Pushed `14401c7`, which addresses the remaining automated-review feedback. Going through all five suppressed comments from the last two review passes explicitly, since two of them were worth acting on and three I think should be declined.

## Fixed

**1. `ip_route_add_del_get_stats` did not clamp `n_labels` (`SaiVppXlate.c`)**

Valid, and the more important of the two. The code assigned `nexthop->n_labels` straight into `fib_path->n_labels` while the loop only ever populated up to `VPP_MPLS_MAX_LABELS` entries, so a deeper stack would have told VPP the message carried more labels than it actually did.

It is currently unreachable — `fillNHGrpMember` now fails on `BUFFER_OVERFLOW`, so `n_labels` can never exceed 16 — but the asymmetry was real: `mpls_route_add_del` validates this and returns `-EINVAL`, and the IP path had no equivalent. Now clamped before the assignment, so the count and the populated entries always agree.

**2. Misleading `out_labels` in the `MplsRouteAddRemove` log (`SwitchVppMpls.cpp`)**

Also valid. The pop case injects an implicit-null label and sets `n_labels` to 1, so a pop was logged as having one out-label. Now reports the SAI-visible count, which is what a reader of that log is actually after.

## Declined, with reasoning

**3. Roll back partial *deletes* the way partial adds are rolled back**

I do not think this one should be actioned. Rolling back a failed delete means re-installing MPLS FIB entries during a teardown that is already failing — re-adding forwarding state for a next hop that is on its way out is worse than the transient inconsistency it would fix.

It is also self-healing: `vpp_normalize_ret` maps `NO_SUCH_ENTRY` to success on delete, so when orchagent retries the remove, the already-deleted EOS entry is a no-op and the remaining one is retried. The add case is genuinely different — there the caller aborts before recording the route, so nothing in SAI refers to the orphaned entry, which is why that one was worth unwinding.

**4. `mr_n_paths` could truncate if `nexthop_cnt > 255`**

True in the abstract, but this matches the existing convention exactly: the neighbouring `ip_route_add_del_get_stats` does the identical unvalidated `ip_route->n_paths = (u8)path_count`. Adding a check to only the MPLS path would make the two inconsistent without fixing the general case. In practice INSEG always passes `nexthop_cnt = 1` (hardcoded in `MplsRouteAddRemove`). If this is worth tightening it should be done for both paths, which is beyond the scope of this PR.

**5. Treat an unresolvable `hwif_name` as an error rather than falling back to `~0`**

The fallback is deliberate and documented in `MplsRouteAddRemove`. It is also effectively unreachable: `hwif_name` is only set after `vpp_get_hwif_name()` has already succeeded for that port, so `get_swif_idx` failing afterwards would mean the interface disappeared mid-call. Turning that into a hard failure trades a rare degraded-forwarding case for a hard error on a path that today works, which I do not think is a good trade without evidence it happens.

Happy to revisit any of these three if a reviewer disagrees.

## Verification

Rebuilt against VPP 2606 to match the sonic-vpp image — clean, zero warnings under `-Werror -Wall -Wextra -Wconversion -Wshadow -Wcast-align=strict -Wformat=2`. The repo's own gates pass locally (`aspellcheck.pl`, `checkwhitespace.sh`, `swsslogentercheck.sh`), as does `make check` for `meta`, `lib`, `vslib` and `unittest/vslib`, and the full `tests` suite (all 11).

Both changes here are on paths the functional tests already cover: the label-count clamp sits in the IP-route imposition path exercised by `test_swap_labelstack`, and the log change is observational only.

augusdn:
Rebased onto current master (`11984240`) and pushed `de99c7a`. Flagging why, since the CI failure that prompted it was **not** a flake and the cause is worth knowing for anyone with an older VPP branch in flight.

## What broke

Two consecutive builds failed at *Compile sonic sairedis with coverage enabled*:

```
vpp/vppxlate/SaiVppXlate.c:5494:5: error: redefinition of 'get_sw_if_idx'
vpp/vppxlate/SaiVppXlate.c:2673:5: note: previous definition of 'get_sw_if_idx' with type 'int(const char *)'
make[4]: *** [Makefile:1029: vpp/vppxlate/libSaiVS_a-SaiVppXlate.o] Error 1
```

This is a **semantic merge conflict**, which is why it wasn't visible earlier:

- The merge base (`24c799da`) has **zero** definitions of `get_sw_if_idx`.
- This PR adds one.
- 51fe70f *"[vslib] Add support for SPAN mirror sessions"* (#1920) landed after this branch was cut and adds another.

Git merges both cleanly — the PR reported `MERGEABLE` throughout — but the merged result doesn't compile. Local builds passed because the branch base predated #1920.

## Fix

Dropped ours and kept the upstream definition, which is **strictly better**: it takes `VPP_LOCK()` around `get_swif_idx()`, which ours did not. Behaviour is otherwise identical — ours mapped `(u32)-1` to `-1` explicitly, and upstream's `return (int)idx;` yields the same value for that input.

No header change was needed: the `extern` declaration is identical on both sides, so this PR no longer touches `SaiVppXlate.h` for that symbol. Net effect of the new commit is `-11` lines.

## Verification

Full CI replication in `sonic-slave-trixie` against the rebased tree:

```
dpkg_buildpackage_rc=0        <- the step that was failing
9 debs produced
vslib/vpp/vppxlate/libSaiVS_a-SaiVppXlate.o   built clean
0 compile errors, 0 redefinition errors
```

`make check` reports one failure, `unittest/lib` → `ServerSai.stats_st_capability_query`, aborting on `Select::add_fd:epoll_ctl: error=(1):Operation not permitted`. That one is environmental and not related to this change:

- `ServerSai` mocks `getFd()` to return `0` (stdin), and `epoll_ctl` returns `EPERM` for a stdin that isn't pollable — which is what you get running the suite under `docker run` with stdin redirected.
- This PR touches **zero** files under `lib/` or `unittest/`.
- `unittest/lib/tests` doesn't even link `vslib`: `tests_LDADD = ... lib/libSaiRedis.a -lhiredis -lswsscommon -lpthread -L meta/.libs -lsaimetadata -lsaimeta -lzmq`. Nothing from this PR is in that binary.

The four functional commits are unchanged by the rebase; only the new fifth commit differs.

augusdn:
CI is green on the rebased head `de99c7a`: **19 checks pass, 0 fail** (ADO build 1184588).

That includes the two jobs that were failing before the rebase - `Build amd64` and `BuildAsan amd64` - plus `Test vstest` (1h26m) and `TestAsan vstest` (1h38m), and the ARM/Trixie/Docker/Swss builds.

For anyone reviewing the timeline: the earlier red builds were the `get_sw_if_idx` redefinition described in the previous comment, caused by #1920 landing after this branch was cut. That is fixed; no functional commit changed.

vslib/vpp/SwitchVppMpls.cpp
```diff
@@ -0,0 +1,312 @@
...
+    vpp_mpls_route_t *route = (vpp_mpls_route_t *)
+        calloc(1, sizeof(vpp_mpls_route_t) + sizeof(vpp_mpls_nexthop_t));
+    if (!route) {
+        return SAI_STATUS_FAILURE;
+    }
+    route->table_id = MPLS_TABLE_ID;
+    route->label = inseg_entry.label;
+    route->is_multipath = false;
+    route->nexthop_cnt = 1;
+    route->eos_proto_af = AF_INET;
```
yue-fred-gao, line 164: can you derive eos_proto_af from nexthop family, instead of hardcode?

vslib/vpp/SwitchVppNexthop.cpp
```diff
@@ -187,13 +188,66 @@ SwitchVpp::fillNHGrpMember(nexthop_grp_member_t *nxt_grp_member, sai_object_id_t
...
+                SWSS_LOG_ERROR("MPLS out-label stack of %u labels exceeds maximum %u",
+                        attr.value.u32list.count, VPP_MPLS_MAX_LABELS);
+                return SAI_STATUS_NOT_SUPPORTED;
+            }
+
+            if (label_status == SAI_STATUS_SUCCESS && attr.value.u32list.count > 0) {
+                uint32_t cnt = attr.value.u32list.count;
+                nxt_grp_member->n_labels = (uint8_t)cnt;
+                for (uint32_t li = 0; li < cnt; li++) {
+                    nxt_grp_member->label_stack[li] = attr.value.u32list.list[li];
```
yue-fred-gao, line 231: can we set label stack is_uniform and ttl value based on sai attribute SAI_NEXT_HOP_ATTR_OUTSEG_TTL_MODE and SAI_NEXT_HOP_ATTR_OUTSEG_TTL_VALUE?

vslib/vpp/SwitchVppNexthop.cpp
```diff
@@ -187,13 +188,66 @@ SwitchVpp::fillNHGrpMember(nexthop_grp_member_t *nxt_grp_member, sai_object_id_t
...
+            /*
+             * Program an attached path (resolve the router interface to its VPP
+             * egress hwif) so VPP actually imposes the label; a recursive
+             * labelled path is dropped at the MPLS DROP DPO.
+             */
+            std::string mpls_hwif;
+            sai_attribute_t port_attr;
+            port_attr.id = SAI_ROUTER_INTERFACE_ATTR_PORT_ID;
+            if (have_rif &&
+                get(SAI_OBJECT_TYPE_ROUTER_INTERFACE, nxt_grp_member->rif_oid, 1, &port_attr) == SAI_STATUS_SUCCESS &&
```
yue-fred-gao, line 243: minor: you can use get_linked_object to get rif object from nexthop

vslib/vpp/vppxlate/SaiVppXlate.c
```diff
@@ -2974,7 +3191,20 @@ int ip_route_add_del_get_stats (vpp_ip_route_t *prefix, bool is_add, uint32_t *s
...
+         * carries more labels than it actually does.
+         */
+        uint8_t n_labels = nexthop->n_labels > VPP_MPLS_MAX_LABELS ?
+                           VPP_MPLS_MAX_LABELS : nexthop->n_labels;
+        fib_path->n_labels = n_labels;
+        for (uint8_t l = 0; l < n_labels; l++) {
+            fib_path->label_stack[l].label = htonl(nexthop->label_stack[l]);
+            fib_path->label_stack[l].ttl = MPLS_DEFAULT_OUT_TTL;
+            fib_path->label_stack[l].exp = 0;
+            fib_path->label_stack[l].is_uniform = 1;
```
yue-fred-gao, line 3206: Can we support ttl/qos mode and ttl/exp value from SAI attribute instead of hardcode here? It seems not requiring a lot of code

vslib/vpp/SwitchVpp.h
```diff
@@ -754,6 +754,22 @@ namespace saivs
...
                     _In_ const std::string &serializedObjectId);
 
+            sai_status_t addMplsRoute(
+                    _In_ const std::string &serializedObjectId,
+                    _In_ sai_object_id_t switch_id,
+                    _In_ uint32_t attr_count,
+                    _In_ const sai_attribute_t *attr_list);
+            sai_status_t removeMplsRoute(
+                    _In_ const std::string &serializedObjectId);
+            sai_status_t MplsRouteAddRemove(
```
yue-fred-gao, line 764: camel case

vslib/vpp/SwitchVppMpls.cpp
```diff
@@ -0,0 +1,312 @@
...
+         * TTL (mpls_ttl - 1). This matches the SAI default POP TTL mode
+         * (SAI_INSEG_ENTRY_POP_TTL_MODE_UNIFORM) and typical hardware. Without
+         * it VPP defaults the disposition to PIPE (inner TTL left untouched and
+         * only decremented once by the IP forwarding stage).
+         */
+        route->nexthop[0].n_labels = 1;
+        route->nexthop[0].label_stack[0].label = MPLS_IMPLICIT_NULL_LABEL;
+        route->nexthop[0].label_stack[0].ttl = 0;
+        route->nexthop[0].label_stack[0].exp = 0;
+        route->nexthop[0].label_stack[0].is_uniform = 1;
```
yue-fred-gao, line 214: should we set is_uniform based on SAI_INSEG_ENTRY_ATTR_POP_TTL_MODE?

vslib/vpp/SwitchVppMpls.cpp
augusdn: Good catch, and this turned out to be a real bug rather than just a hardcode.

The surrounding plumbing already handled IPv6: `mpls_route_add_del()` explicitly accepts an `AF_INET6` nexthop (family validation in `SaiVppXlate.c`), and `mr_eos_proto` already maps `AF_INET6` onto `FIB_API_PATH_NH_PROTO_IP6`. So an IPv6 nexthop was being accepted end to end and then programmed with an IPv4 end-of-stack protocol.

Fixed in b26362e: derived from `route->nexthop[0].addr.sa_family`. The assignment had to move below `fillMplsNexthop()`, which is what populates the address.

For IPv4 next hops the derived value is `AF_INET`, i.e. exactly what was hardcoded, so there is no behaviour change on the tested path. To be clear about scope: this corrects the value, but I have not validated an IPv6 MPLS data path, so IPv6 stays listed as a known limitation rather than being claimed as supported.

vslib/vpp/SwitchVppNexthop.cpp
augusdn: Agreed - fixed in 41595f9, together with your related comment on `SaiVppXlate.c` (they turned out to be the two ends of the same pipe: this is where the value would be stored, that is where it was hardcoded on the way to VPP).

Added a small `getOutsegTtl()` helper that reads `SAI_NEXT_HOP_ATTR_OUTSEG_TTL_MODE` / `_TTL_VALUE` and `_EXP_MODE` / `_EXP_VALUE`, and called it from both imposition paths - `fillMplsNexthop()` for MPLS-to-MPLS and `IpRouteNexthopEntry()` for the IP-to-MPLS ingress LER case.

One design note worth flagging: SAI models this per next hop rather than per label, so I stored it as three `uint8_t` fields on the nexthop and applied them to every label when building the fib path, rather than widening `label_stack[16]` to a per-label struct. That keeps `vpp_ip_nexthop_t` small and matches the SAI object model.

The explicit TTL/EXP values are `@validonly` in PIPE mode, so the helper only takes a value when the mode calls for it. Unset attributes read back as `ITEM_NOT_FOUND`, leaving the SAI defaults - which are the values that were previously hardcoded - so nothing changes for callers that do not set them.

vslib/vpp/SwitchVppNexthop.cpp
augusdn: You are right, and it was inconsistent with our own code - `SwitchVppSRv6.cpp` resolves the RIF exactly this way, and `SwitchVppMpls.cpp` already uses `get_linked_object()` to resolve its nexthop.

Fixed in 9cf787d. It also let the `have_rif` bookkeeping flag go, since the returned object pointer now carries that state, and it removed a second manual OID lookup a few lines up.

vslib/vpp/vppxlate/SaiVppXlate.c
augusdn: Agreed - fixed in 41595f9, same commit as your `SwitchVppNexthop.cpp` comment since this is the consumer end of that path.

You were right that it is not much code. My initial hesitation was that `vpp_ip_nexthop_t.label_stack` was a flat `uint32_t[]` with nowhere to put the TTL, so it needed a struct change - but that struct is ours and was added in this PR, so the asymmetry with `vpp_mpls_nexthop_t` (which already carried ttl/exp/is_uniform end to end) was self-inflicted. Fixed rather than argued.

One thing I should disclose rather than leave implicit: this particular path is not covered by the test suite. `test_push_label` is skipped because orchagent does not install `ROUTE_TABLE`-injected push routes into ASIC_DB, so the IP-to-MPLS imposition path is not exercised end to end today. The code is reasoned and compiles clean, but I cannot claim it is tested. Happy to hear if you would rather it were gated differently until it can be.

vslib/vpp/SwitchVpp.h
augusdn: Fixed in 3b2a738. For what it is worth I checked before changing it: `SwitchVpp.h` has 159 camelCase member declarations against 12 PascalCase ones, so camelCase is clearly the convention - the original name was mirroring the adjacent `IpRouteAddRemove`. Renamed across all 5 sites.

vslib/vpp/SwitchVppMpls.cpp
augusdn: Agreed - fixed in 72bf00d. Reads `SAI_INSEG_ENTRY_ATTR_POP_TTL_MODE` off the INSEG object (same pattern already used a few lines above for `PACKET_ACTION`) and maps UNIFORM to `is_uniform=1`, PIPE to `0`.

The attribute defaults to `SAI_INSEG_ENTRY_POP_TTL_MODE_UNIFORM`, which is the value that was hardcoded, so behaviour is unchanged unless a caller explicitly asks for PIPE.

augusdn:
Thanks for the review @yue-fred-gao - all six comments are addressed in `3b2a738`. No push-backs; each one was valid, and two of them turned out to be more than cosmetic.

## Changes

| Commit | Comment |
|---|---|
| 9cf787d | `get_linked_object()` to resolve the nexthop RIF |
| 41595f9 | Honour `SAI_NEXT_HOP_ATTR_OUTSEG_{TTL,EXP}_{MODE,VALUE}` |
| b26362e | Derive `eos_proto_af` from the nexthop family |
| 72bf00d | Honour `SAI_INSEG_ENTRY_ATTR_POP_TTL_MODE` |
| 3b2a738 | `MplsRouteAddRemove` -> `mplsRouteAddRemove` |

Your two TTL comments (`SwitchVppNexthop.cpp` and `SaiVppXlate.c`) are fixed in one commit - they were the two ends of the same pipe, so they could not sensibly be done separately.

**`eos_proto_af` was a real bug, not just a hardcode.** `mpls_route_add_del()` already accepts an `AF_INET6` nexthop and `mr_eos_proto` already maps it to `FIB_API_PATH_NH_PROTO_IP6`, so an IPv6 nexthop was being accepted end to end and then programmed with an IPv4 end-of-stack protocol. Thanks for spotting it.

## Validation

Every one of these is behaviour-preserving on the tested path - the SAI defaults (`UNIFORM`, and `AF_INET` for an IPv4 nexthop) are exactly the values that were hardcoded. So I used "identical result" as the pass criterion rather than just "tests pass":

```
before (golden baseline):  3 passed, 1 skipped, 151 warnings, PYTEST_RC=0
after  (this push):        3 passed, 1 skipped, 151 warnings, PYTEST_RC=0
```

`mpls/test_mpls.py` on `vms-kvm-vpp-t1-lag`, freshly rebuilt `libsaivs.so` deployed to the DUT with a full config reload. The suite asserts exact TTL arithmetic (`MPLS.ttl -= 1`, `MPLS.ttl = IP.ttl - 1`, inner `IP.ttl` 64 -> 62 on pop), so an identical result is meaningful evidence that the TTL handling did not shift.

Also: clean `dpkg-buildpackage` (rc=0, 9 debs) and a clean `vslib` build with zero warnings under `-Werror -Wconversion -Wshadow`.

## Two things I want to be upfront about

1. **IPv6 is corrected, not claimed.** `eos_proto_af` now carries the right value, but I have not validated an IPv6 MPLS data path, so IPv6 stays in the known-limitations list.
2. **The non-default attribute paths are unexercised.** Nothing in sonic-swss sets `OUTSEG_TTL_MODE`, `OUTSEG_TTL_VALUE`, `POP_TTL_MODE`, `OUTSEG_EXP_MODE` or `POP_QOS_MODE` today, so they sit at their defaults and the PIPE branches never execute in the current suite. The IP-to-MPLS imposition path is additionally untested because `test_push_label` is skipped - orchagent does not install `ROUTE_TABLE`-injected push routes into ASIC_DB. The code is reasoned and compiles clean, but I would rather say plainly that it is not covered than imply it is.

I noticed #1950 landed a saithrift harness that drives saivpp directly - that looks like it could exercise these attribute paths properly. Happy to follow up there if you think it is worth it.

yue-fred-gao (approved):
looks good to me. Please trigger t1-lag-vpp sanity to make sure no breakage. Here is an example: https://github.com/sonic-net/sonic-buildimage/pull/27039

yue-fred-gao:
> I noticed #1950 landed a saithrift harness that drives saivpp directly - that looks like it could exercise these attribute paths properly. Happy to follow up there if you think it is worth it.
Yes, we are enabling saitest as UT for saivpp. #1950 brings in the UT infra. We will turn it on in azure-pipelines as PR checker. I am not sure about the coverage on mpls in saitest. Please follow up on that to enable mpls UT for saivpp. Thanks!

augusdn:
Thanks @yue-fred-gao. Ran the t1-lag-vpp sanity as requested, using the same mechanism as your example: sonic-net/sonic-buildimage#28806 pins `src/sonic-sairedis` at this PR's approved head `3b2a738`.

**`BuildVS vpp` succeeded** and the t1-lag-vpp Elastictest plan came back green:

```
Test plan 6a7c81c21c6297f01e854596   status FINISHED   result SUCCESS
tests 1838 | passes 1298 | failures 1 | errors 1 | skipped 538 | xfails 13
errors_ignore_retries = 0
```

Two modules were not clean, neither related to MPLS:

| Module | |
|---|---|
| `arp/test_arpall.py` | 1 failure |
| `vxlan/test_vxlan_multi_tunnel.py` | 1 error (passed on retry) |

`arp/test_arpall.py` also failed on the previous run of this same throwaway PR back on Aug 3 (plan `6a70bef6f481df03c4e5a5f3`, on an older sairedis commit), so it looks pre-existing on this topology rather than anything this change introduced.

**One caveat I want to state rather than let the green check imply too much:** this plan does not include the `mpls` module. It runs stock sonic-mgmt, which still carries the blanket `mpls` skip — lifting that is the companion PR sonic-net/sonic-mgmt#26619. So this run demonstrates *no breakage* from the saivpp change, but it is not MPLS coverage.

The MPLS evidence is the earlier combined run with both PRs in place — plan `6a69da68f481df03c4e59c5e` on `vms-kvm-vpp-t1-lag`, **17 passed / 5 skipped / 0 failed / 0 errors**, covering `test_pretest.py` + `mpls/test_mpls.py` + `test_posttest.py`.

On saitest: agreed, and thanks for the pointer. I'll follow up on MPLS coverage there once #1950's infra is switched on as a PR checker — that is the right place to exercise the attribute paths that cannot be reached today (`OUTSEG_TTL_MODE`/`POP_TTL_MODE` PIPE, and IPv6 end-of-stack), since nothing in sonic-swss sets them.

yue-fred-gao:
> Thanks @yue-fred-gao. Ran the t1-lag-vpp sanity as requested, using the same mechanism as your example: [sonic-net/sonic-buildimage#28806](https://github.com/sonic-net/sonic-buildimage/pull/28806) pins `src/sonic-sairedis` at this PR's approved head `3b2a738`.
> 
> **`BuildVS vpp` succeeded** and the t1-lag-vpp Elastictest plan came back green:
> 
> ```
> Test plan 6a7c81c21c6297f01e854596   status FINISHED   result SUCCESS
> tests 1838 | passes 1298 | failures 1 | errors 1 | skipped 538 | xfails 13
> errors_ignore_retries = 0
> ```
> 
> Two modules were not clean, neither related to MPLS:
> 
> Module	
> `arp/test_arpall.py`	1 failure
> `vxlan/test_vxlan_multi_tunnel.py`	1 error (passed on retry)
> `arp/test_arpall.py` also failed on the previous run of this same throwaway PR back on Aug 3 (plan `6a70bef6f481df03c4e5a5f3`, on an older sairedis commit), so it looks pre-existing on this topology rather than anything this change introduced.
> 
> **One caveat I want to state rather than let the green check imply too much:** this plan does not include the `mpls` module. It runs stock sonic-mgmt, which still carries the blanket `mpls` skip — lifting that is the companion PR [sonic-net/sonic-mgmt#26619](https://github.com/sonic-net/sonic-mgmt/pull/26619). So this run demonstrates _no breakage_ from the saivpp change, but it is not MPLS coverage.
> 
> The MPLS evidence is the earlier combined run with both PRs in place — plan `6a69da68f481df03c4e59c5e` on `vms-kvm-vpp-t1-lag`, **17 passed / 5 skipped / 0 failed / 0 errors**, covering `test_pretest.py` + `mpls/test_mpls.py` + `test_posttest.py`.
> 
> On saitest: agreed, and thanks for the pointer. I'll follow up on MPLS coverage there once #1950's infra is switched on as a PR checker — that is the right place to exercise the attribute paths that cannot be reached today (`OUTSEG_TTL_MODE`/`POP_TTL_MODE` PIPE, and IPv6 end-of-stack), since nothing in sonic-swss sets them.

test_arpall also passed on retry. It is probably not very stable.

PR #2019, opened 2026-08-04
[vpp] Support SAI_SWITCH_ATTR_ECMP_DEFAULT_HASH_SEED in saivpp

yejianquan:
### Description of PR

Summary: Implement `SAI_SWITCH_ATTR_ECMP_DEFAULT_HASH_SEED` in saivpp by mapping it onto VPP's global ECMP flow-hash router ID (`ip_flow_hash_router_id`), which VPP already mixes into the IPv4/IPv6 ECMP hash. Setting it re-distributes ECMP/LAG path selection — exactly the SAI hash-seed semantic, so no VPP change is needed.

Part of sonic-net/sonic-buildimage#25760.

> [!NOTE]
> Companion PR sonic-net/sonic-mgmt#26756 enables `ecmp/test_ecmp_balance.py` on sonic-vpp and depends on this being in the image, so it merges after this one.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

`test_udp_packets_ecmp` needs a settable ECMP hash seed. VPP has no vppctl CLI for the router ID — only the binary API behind SAI — so the seed must travel `swssconfig` → APPL_DB → orchagent → `SAI_SWITCH_ATTR_ECMP_DEFAULT_HASH_SEED` → saivpp → VPP. saivpp was the missing link.

##### Work item tracking

#### How did you do it?

3 files, +57:

| File | Change |
|---|---|
| `vppxlate/SaiVppXlate.h` | Declare `vpp_ip_flow_hash_router_id_set(uint32_t)`. |
| `vppxlate/SaiVppXlate.c` | Wrapper around `set_ip_flow_hash_router_id` (modelled on the adjacent `vpp_ip_flow_hash_set()`) + reply handler + message registration. |
| `SwitchVpp.cpp` | Handle the attribute, then fall through to `set_internal()` so it is cached for GET. |

Builds clean under the sairedis strict flags against current master; `libsaivs.so` exports `vpp_ip_flow_hash_router_id_set` and links `libvlib.so.2606`.

Deployed to `vlab-vpp-01` on the converged `vms-kvm-vpp-t1-lag` testbed (`SONiC.master.1151527`). syslog confirms the handler fires (`VPP set ECMP default hash seed=10`), and `ecmp/test_ecmp_balance.py` passes **4/4**.

On a stock image the same run gives `2 passed, 2 failed` — both `test_udp_packets_ecmp` cases failing on *"results are identical with and without ECMP hash offset change"*, i.e. orchagent sets the attribute but the seed never reaches VPP. That is the gap this PR closes.

Confined to `vslib/vpp/`; no other SAI backend or platform is affected.

### Documentation

No HLD required — implements an existing standard SAI attribute in an existing backend.

### DIFF
--- a/vslib/vpp/SwitchVpp.cpp
+++ b/vslib/vpp/SwitchVpp.cpp
@@ -2080,6 +2080,23 @@ sai_status_t SwitchVpp::set(
                     m_tunnel_mgr.set_vxlan_port(attr);
                     break;
                 }
+            case SAI_SWITCH_ATTR_ECMP_DEFAULT_HASH_SEED:
+                {
+                    // VPP mixes a global "router id" into the IPv4/IPv6 ECMP
+                    // flow hash (ip4_inlines.h / ip6_inlines.h). Map the SAI
+                    // ECMP hash seed onto it so that changing the seed
+                    // re-distributes ECMP/LAG path selection. Fall through to
+                    // set_internal() below so the attribute is also cached.
+                    uint32_t seed = attr->value.u32;
+                    int rc = vpp_ip_flow_hash_router_id_set(seed);
+                    if (rc != 0)
+                    {
+                        SWSS_LOG_ERROR("VPP set ECMP default hash seed=%u failed rc=%d", seed, rc);
+                        return SAI_STATUS_FAILURE;
+                    }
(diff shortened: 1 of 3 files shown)


lolyu (approved):
LGTM

Issue #1662, opened 2025-10-07
test_neighbor_mac.py failed on sonic-vpp


yue-fred-gao:
test_neighbor_mac.py failed due to a race condition in setting IP on Ethernet0. This is because in response to removing Ethernet0 from PortChannel, sonic-vpp needed to update all the affected routes which took a long time. In the end, setting IP on Ethernet0 happened before the interface was removed from the PortChannel. At the end, the IP on Ethernet0 was not set successfully so ping failed.

Below is sairedis.rec to manifest the out of order sequence.
Here is Ethernet0 added as LAG member.
```
2025-10-01.23:38:45.325507|c|SAI_OBJECT_TYPE_LAG_MEMBER:oid:0x1b0000000005ba|SAI_LAG_MEMBER_ATTR_LAG_ID=oid:0x20000000005b2|SAI_LAG_MEMBER_ATTR_PORT_ID=oid:0x100000
0000030|SAI_LAG_MEMBER_ATTR_EGRESS_DISABLE=true|SAI_LAG_MEMBER_ATTR_INGRESS_DISABLE=true
2025-10-01.23:38:45.456866|s|SAI_OBJECT_TYPE_LAG_MEMBER:oid:0x1b0000000005ba|SAI_LAG_MEMBER_ATTR_EGRESS_DISABLE=true
2025-10-01.23:38:45.457593|s|SAI_OBJECT_TYPE_LAG_MEMBER:oid:0x1b0000000005ba|SAI_LAG_MEMBER_ATTR_INGRESS_DISABLE=true
```
Here is creating router interface for Ethernet0 and set IP address
```
2025-10-01.23:43:47.059575|c|SAI_OBJECT_TYPE_ROUTER_INTERFACE:oid:0x60000000006d9|SAI_ROUTER_INTERFACE_ATTR_VIRTUAL_ROUTER_ID=oid:0x3000000000002|SAI_ROUTER_INTERFA
CE_ATTR_SRC_MAC_ADDRESS=22:5D:A7:7E:B7:8E|SAI_ROUTER_INTERFACE_ATTR_TYPE=SAI_ROUTER_INTERFACE_TYPE_PORT|SAI_ROUTER_INTERFACE_ATTR_PORT_ID=oid:0x1000000000030|SAI_RO
UTER_INTERFACE_ATTR_MTU=9100|SAI_ROUTER_INTERFACE_ATTR_NAT_ZONE_ID=0
[...]

PR #2020, opened 2026-08-04
fix(vpp): restore port tap MAC after LAG removal

auspham:
### Description of PR

Summary:
Restore the Linux hostif tap MAC to the switch source MAC after a physical port leaves a VPP LAG.

Related to #1662

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

`arp/test_neighbor_mac.py::TestNeighborMac::testNeighborMac[0]` is failing on `t1-lag-vpp` after removing `Ethernet0` from `PortChannel102` and using it as a routed interface.

Over the last 14 days, the signature affected 405 of 943 VPP PR plans (42.95%), with 167 plans exhausting all retries. A local run reproduced the exact 120-second PTF-to-DUT ping failure using image `SONiC.master-28819.1183023-66e2de6ac`.

Sample failing Elastictest plan: https://elastictest.org/scheduler/testplan/6a7168d4b63289adaeeee6a4?searchTestCase=arp%2F&testcase=arp%2Ftest_neighbor_mac.py%7C%7C%7C2&type=console

The failing state had a valid RIF, connected/local FIB entries, a learned PTF neighbor, and successful ARP replies. However, Linux teamd had restored the detached tap's permanent MAC (`52:54:00:44:86:c4`) while VPP retained the switch MAC (`22:5d:a7:7e:b7:8e`). PTF therefore learned the tap MAC, but unicast ICMP traffic to that MAC was not accepted by the VPP routed interface.

##### Work item tracking
- Microsoft ADO **(number only)**: 39081324

#### How did you do it?

After either the normal LAG detach path or the already-egress-disabled path, `removeLagMember()` now restores the port tap MAC to `SAI_SWITCH_ATTR_SRC_MAC_ADDRESS` before deleting the internal LAG member object.

The repair is best effort because the VPP detach has already completed. Failures are logged with the affected port or tap, while LAG object removal continues to avoid leaving the SAI object model inconsistent with the dataplane.

- Reproduced the exact failure locally after replaying the fleet predecessor sequence.
- Confirmed that enabling VPP promiscuous mode alone did not help: all 52 probes still failed.
- Changed only the Linux `Ethernet0` MAC to the switch/VPP MAC during a failing loop. Attempts 28-33 failed, attempt 34 passed immediately, `testNeighborMac[1]` passed on its first probe, and the module completed with 2 passed.
- Ran `git diff --check`.
- Ran a C++17 syntax and format harness against the changed helper and call site with `-Wall -Wextra -Wformat=2 -Werror`.
- CI amd64 and ASAN builds and unit tests passed.

A full `t1-lag-vpp` image run with a syncd binary built from this branch remains pending.

The change only affects the VPP virtual switch implementation.

### Documentation

N/A

### DIFF
--- a/vslib/vpp/SwitchVppFdb.cpp
+++ b/vslib/vpp/SwitchVppFdb.cpp
@@ -1505,11 +1505,64 @@ sai_status_t SwitchVpp::removeLagMember(
         CHECK_STATUS_QUIET(vpp_remove_lag_member(lag_member_oid));
     }
 
+    // teamd restores the member tap's permanent MAC when the port leaves the PortChannel while
+    // VPP keeps the switch MAC, so the routed port would drop L3 traffic sent to the tap MAC.
+    if (port_oid != SAI_NULL_OBJECT_ID)
+    {
+        restorePortTapMac(port_oid);
+    }
+
     auto sid = sai_serialize_object_id(lag_member_oid);
 
     CHECK_STATUS(remove_internal(SAI_OBJECT_TYPE_LAG_MEMBER, sid));
 
     return SAI_STATUS_SUCCESS;
+}
+
+void SwitchVpp::restorePortTapMac(
+        _In_ sai_object_id_t port_oid)
+{
+    SWSS_LOG_ENTER();
+
+    // Best effort repair, the caller must still drop the SAI LAG member because the VPP
+    // detach already happened and failing here would leave the object model inconsistent.
+    std::string if_name;
+
+    if (getTapNameFromPortId(port_oid, if_name) == false)
+    {
+        SWSS_LOG_ERROR("no tap found for port %s, switch MAC not restored",
+                sai_serialize_object_id(port_oid).c_str());
+        return;
+    }
+
+    sai_attribute_t attr;
+
+    attr.id = SAI_SWITCH_ATTR_SRC_MAC_ADDRESS;
+
+    sai_status_t status = get(SAI_OBJECT_TYPE_SWITCH, m_switch_id, 1, &attr);
+
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("failed to get SAI_SWITCH_ATTR_SRC_MAC_ADDRESS on switch %s: %s, "
+                "switch MAC not restored on %s",
+                sai_serialize_object_id(m_switch_id).c_str(),
+                sai_serialize_status(status).c_str(),
+                if_name.c_str());
+        return;
+    }
+
+    if (vs_set_dev_mac_address(if_name.c_str(), attr.value.mac) < 0)
+    {
+        SWSS_LOG_ERROR("failed to set MAC address %s on tap %s, the routed port will drop "
+                "traffic addressed to its tap MAC",
+                sai_serialize_mac(attr.value.mac).c_str(),
+                if_name.c_str());
+        return;
+    }
+
+    SWSS_LOG_NOTICE("restored switch MAC %s on tap %s after LAG member removal",
+            sai_serialize_mac(attr.value.mac).c_str(),
+            if_name.c_str());
 }
 
 sai_status_t SwitchVpp::vpp_remove_lag_member(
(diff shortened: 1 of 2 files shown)


lolyu (commented):
Nicely diagnosed fix for a real high-frequency flake (arp/test_neighbor_mac on t1-lag-vpp). I verified from source at `1ba3dfb`: `port_oid` is fetched once before the detach branch so `restorePortTapMac` correctly covers both the egress-disabled and normal `vpp_remove_lag_member` paths; `getTapNameFromPortId` and `vs_set_dev_mac_address` are used consistently with the rest of the file; and the best-effort void error handling is the right call (VPP detach already happened, so removal must still complete). Two non-blocking notes inline. Also worth gating merge on the pending full `t1-lag-vpp` image run, since that's the real proof for this fix.

vslib/vpp/SwitchVppFdb.cpp
```diff
@@ -1505,13 +1505,66 @@ sai_status_t SwitchVpp::removeLagMember(
         CHECK_STATUS_QUIET(vpp_remove_lag_member(lag_member_oid));
     }
 
+    // teamd restores the member tap's permanent MAC when the port leaves the PortChannel while
+    // VPP keeps the switch MAC, so the routed port would drop L3 traffic sent to the tap MAC.
+    if (port_oid != SAI_NULL_OBJECT_ID)
+    {
+        restorePortTapMac(port_oid);
```
lolyu, line 1512: 🟡 Non-blocking: silent skip when the port can't be resolved. Earlier in `removeLagMember`, `port_oid` stays `SAI_NULL_OBJECT_ID` if `get_lag_member_port(lag_member_oid, port_oid)` returns non-SUCCESS, in which case this guard skips the MAC restore entirely — with no log. That's a reasonable choice (you can't restore a tap you can't resolve), but given how thoroughly the new `restorePortTapMac` helper logs each failure mode, a one-line `SWSS_LOG_WARN` here ("port unresolved for LAG member %s, tap MAC not restored") would make the skipped-repair case visible in logs and consistent with the rest of the fix. Optional.

vslib/vpp/SwitchVppFdb.cpp
```diff
@@ -1505,13 +1505,66 @@ sai_status_t SwitchVpp::removeLagMember(
...
+    }
+
     auto sid = sai_serialize_object_id(lag_member_oid);
 
     CHECK_STATUS(remove_internal(SAI_OBJECT_TYPE_LAG_MEMBER, sid));
 
     return SAI_STATUS_SUCCESS;
 }
 
+void SwitchVpp::restorePortTapMac(
```
lolyu, line 1527: 🟢 Non-blocking symmetry check (not a change request): this restores the tap MAC to the switch MAC on LAG member *removal*. Please confirm the inverse is already handled — i.e. when a port *joins* a LAG (`createLagMember` / `vpp_create_lag_member`), the tap MAC is set to the switch MAC so add/remove stay symmetric and a port doesn't briefly run with a mismatched tap MAC while bonded. From a grep of this file the create path does manipulate tap MACs, so this looks fine; just calling it out so the round-trip is explicitly verified rather than assumed.

auspham: Confirmed, the add direction is already symmetric. Linux teamd sets the member netdev to the PortChannel MAC when it enslaves the port. On the live DUT after re-add, both `Ethernet0` and `PortChannel102` reported `22:5d:a7:7e:b7:8e`. Hostif creation also initially applies `SAI_SWITCH_ATTR_SRC_MAC_ADDRESS` to both the tap and VPP interface.

The mismatch occurs only on leave, when teamd restores the permanent virtio MAC but VPP retains the switch MAC. Changing only the tap MAC during the failing loop made the next probe pass and completed the module 2/2, which is the transition this PR repairs.

vslib/vpp/SwitchVppFdb.cpp
auspham: Good catch. Added this in 5a184f0.

If `port_oid` remains unresolved, the removal still proceeds but now emits `SWSS_LOG_WARN` with the LAG member OID and states that the tap MAC was not restored. This keeps the best-effort repair visible without changing object-removal behavior.

lolyu (approved):
LGTM

PR #2027, opened 2026-08-07
[vpp] Permit local traffic ahead of ACL deny rules

### LINKED ISSUE #28884: Bug: [vpp][dualtor] Mux drop ACL drops control-plane/for-us traffic — IN_PORTS ignored and L2 ACL runs before ip4-local punt (mux stuck unhealthy)
### Is it platform specific
generic (VPP platform / `docker-syncd-vpp`, `libsaivs` vslib + VPP graph)

### Importance or Severity
High

### Description of the bug
On the VPP platform, the dual-ToR **mux drop ACL** (`INGRESS_TABLE_DROP` / `mux_acl_rule`, MuxOrch) drops **control-plane / for-us (IP2ME)** traffic that should be trapped to the CPU. On a hardware ASIC these packets (ARP, ICMP-to-me, BGP, mux ICMP heartbeat) are trapped by the my-station / CoPP path **before / independently of** the ingress data ACL, so a standby mux port still delivers them to the control plane; the data-drop ACL only removes the *forwarding* copy. VPP does not preserve this ordering.

Two distinct VPP fidelity gaps combine here:

**(1) VPP ignores `SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS`.**
MuxOrch models the drop as one ACL table bound to *all* mux ports, with a per-rule `MATCH_IN_PORTS = {standby ports}` qualifier that scopes the DROP to standby ports only. In `vslib/vpp/SwitchVppAcl.cpp`, `SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS` is never read. The table→port binding is honored, but the `IN_PORTS` qualifier is silently dropped, so the rule degrades to a wildcard `ipv4 deny src 0/0 dst 0/0` applied to every bound mux port — **including active ports**. Editing `IN_PORTS` on switchover has no effect in VPP.

**(2) The mux drop ACL runs at `l2-input`, before the for-us / L3 punt at `ip4-local`.**
VPP realizes the mux ACL as an L2-stage ACL (`acl-plugin-in-ip4-l2`) at `l2-input` — the first stage after the frame enters the bridge domain. The for-us recognition and punt happen much later, at `ip4-input on BVI → ip4-lookup → ip4-local / ip4-punt` (VLAN-BVI HLD §5.6). So a for-us packet is denied at L2 before it can ever be recognized as IP2ME and punted. Even a correctly-scoped **standby** port would still drop its control-plane/heartbeat traffic, because there is no "trap for-us/CoPP before the data ACL" semantics in the VPP pipeline.

Concrete impact: the linkmgrd mux ICMP heartbeat reply (server → ToR loopback, a for-us packet) is dropped on the mux port, so linkmgrd never confirms the port and it is stuck `HEALTH=unhealthy`. Because the drop also eats the heartbeat that would clear the condition, an active port that still has the (mis-scoped) deny bound never recovers.

### Steps to Reproduce
1. Dual-ToR VPP KVM testbed (`subtype=DualToR`), mux ports on `Vlan1000`.
2. Observe `show mux status` — a port reports `active` but `HEALTH=unhealthy`.
3. On VPP:
   - `vppctl show acl-plugin acl` → a `ipv4 deny src 0.0.0.0/0 dst 0.0.0.0/0` rule (tag `sonic_acl_oid:...`) applied inbound on the mux bridge-member sw_if_indexes.
   - `vppctl show errors` → `acl-plugin-in-ip4-l2 ACL deny packets` incrementing in step with the port's ingress.
   - `vppctl show ip fib <ToR loopback>` → `dpo-receive on loopN` (for-us), i.e. it *would* be punted at `ip4-local` — after the L2 ACL that already dropped it.
4. `ASIC_DB` shows the active port's OID **removed** from `SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS`, yet VPP still has the deny ACL bound to that port.

**Actual:** Mux drop ACL is applied as an unscoped `deny any/any` at `l2-input` on every bound mux port; `IN_PORTS` is ignored, and control-plane/for-us traffic (ARP/ICMP-to-me/BGP/heartbeat) is dropped before the L3 for-us punt. Active ports drop too. Mux health cannot recover.

**Expected (match hardware semantics):**
- Honor `SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS` so the DROP is scoped to the standby ports actually in the set (bind/unbind the VPP ACL per member, re-derived on `set`), so active ports never drop.
- Control-plane / for-us (IP2ME) traffic must be trapped/punted **before** the mux data-drop ACL consumes it (as on hardware, where CoPP/my-station trapping precedes the ingress data ACL). The mux drop should only remove the forwarding copy, never the CPU-punt copy.

### Relevant log output
```
# mux says active+unhealthy:
show mux status | grep Ethernet4
  Ethernet4  active  active  unhealthy  consistent  ...

# ASIC_DB removed the active port from IN_PORTS (correct intent):
redis-cli -n 1 hget ASIC_STATE:SAI_OBJECT_TYPE_ACL_ENTRY:oid:0x8000000000692 SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS
  -> Ethernet4 oid 0x1000000000033 ABSENT

# ... but VPP still has the deny ACL bound to that port (sw_if 2 = bobm1 = Ethernet4):
vppctl show acl-plugin interface sw_if_index 2   -> input acl(s): 2, 3, 1   (3 = deny any/any)

# ... dropped at L2, before the for-us punt:
vppctl show errors | grep 'acl-plugin-in-ip4-l2'
  478550  acl-plugin-in-ip4-l2   ACL deny packets
vppctl show ip fib 10.1.0.36
  [@0]: dpo-receive: 10.1.0.36 on loop2      # for-us; would punt at ip4-local (after the L2 ACL)
```

```
SONiC Software Version: SONiC.master.1182394-fc9ea17bb
Distribution: Debian 13.5
Platform: x86_64-kvm_x86_64-r0
HwSKU: Force10-S6000
ASIC: vpp
ASIC Count: 1
```

### Fix direction
- `vslib/vpp/SwitchVppAcl.cpp`: implement `SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS` by binding/unbinding the VPP ACL to exactly the member interfaces in the set, re-derived on entry `set` (so removing a port from `IN_PORTS` unbinds the deny from that port).
- Ensure control-plane / for-us (IP2ME) punt precedes the mux data-drop ACL — e.g. a higher-priority `l2-input-classify` punt arc for for-us / control ethertypes+L4 (analogous to the existing LLDP/DHCP punt arcs in the `sonic_ext` plugin), so the CPU-punt copy survives the mux drop. See VPP VLAN-BVI HLD §5.6 (`doc/vpp/vlan-bvi-hld.md`); the HLD does not currently cover the ACL/mux/CoPP ordering.

Related: sonic-net/sonic-buildimage#28882 (VLAN SVI neighbor host route not installed) — separate VPP dual-ToR data-path gap.

lolyu:
Summary:
Fixes https://github.com/sonic-net/sonic-buildimage/issues/28884

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

the mux drop ACL blackholing ICMP replies at `l2-input`.
check issue: https://github.com/sonic-net/sonic-buildimage/issues/28884 for more details.

##### Work item tracking

#### How did you do it?

The fix emulates hardware "CoPP before ACL" purely through ACL rule ordering. VPP evaluates the rules of a single ACL in order, first-match-wins, so if permit rules for the switch's own addresses are placed ahead of a deny in the same ACL, traffic to local RIF is allowed through to the BVI and punted at ip4-local, while all other traffic still falls through to the deny unchanged. No extra datapath node and no per-packet lookup are introduced.

* **Learn the switch's own addresses.** The switch's local host addresses (loopback, SVI/BVI gateway, L3 router-interface) are exactly the interface addresses it programs into VPP, which VPP turns into local (ip2me) receive routes. The layer now records these addresses as they are added and removed, giving it an up-to-date set of "for-us" destinations for both IPv4 and IPv6.

* **Prepend permits when a table drops traffic.** When an ACL table is (re)built and it contains any deny/drop rule, host-scoped permit rules (one per known local address) are inserted ahead of that table's rules. Because the permits live in the same ACL that carries the deny, they are automatically applied to exactly the interfaces the table is bound to, and only traffic destined to the switch's own addresses matches them — everything else is unaffected.

* **Keep the two in sync.** The local-address set and the drop ACL are programmed by independent, unordered SAI events, so the solution reconciles them either way: tables that contain a deny are remembered, and a later change to the local-address set re-applies the permits to those tables. Equivalently, if the addresses are already known when the drop table is first built, the permits are emitted immediately. During bulk boot, before any drop ACL exists, addresses simply accumulate with no ACL churn.

Build and verify on dualtor testbed, the icmp replies are received on mux port:

```
root@vlab-vpp-04:~# tcpdump -i Ethernet8 -nev
tcpdump: listening on Ethernet8, link-type EN10MB (Ethernet), snapshot length 262144 bytes
12:59:27.644064 00:aa:bb:cc:dd:ee > 00:00:00:00:00:00, ethertype IPv4 (0x0800), length 69: (tos 0xb8, ttl 64, id 59341, offset 0, flags [none], proto ICMP (1), length 55)
    10.1.0.36 > 192.168.0.3: ICMP echo request, id 8, seq 3285, length 35
12:59:27.657001 00:00:00:00:00:00 > 00:aa:bb:cc:dd:ee, ethertype IPv4 (0x0800), length 69: (tos 0xb8, ttl 64, id 59341, offset 0, flags [none], proto ICMP (1), length 55)
    192.168.0.3 > 10.1.0.36: ICMP echo reply, id 8, seq 3285, length 35
```

### Documentation

### DIFF
--- a/vslib/vpp/SwitchVpp.h
+++ b/vslib/vpp/SwitchVpp.h
@@ -13,8 +13,11 @@
 #include "vppxlate/SaiVppXlate.h"
 #include "vppxlate/SaiRouteStats.h"
 
+#include "swss/ipaddress.h"
+
 #include <list>
 #include <map>
+#include <set>
 #include <unordered_map>
 #include <mutex>
 #include <atomic>
@@ -846,6 +849,9 @@ namespace saivs
             bool m_acl_default_created = false;
             uint32_t m_sflow_sample_rate = 0;
 
+            std::set<swss::IpAddress> m_local_ips;         // Local RIF addresses
+            std::set<sai_object_id_t> m_local_deny_tables; // ACL tables that contain a deny rule
+
         protected: // VPP
 
             sai_status_t createAclEntry(
@@ -867,6 +873,34 @@ namespace saivs
 
             sai_status_t AclTblConfig(
                     _In_ sai_object_id_t tbl_oid);
+
+            /**
+             * @brief Records or drops a local host address as router
+             * interface IPs are added/removed, and refreshes any ACL table that
+             * has a deny rule so its prepended local permits stay in sync.
+             *
+             * @param[in] prefix VPP interface prefix just programmed; only the
+             * host address is used (matched later as /32 or /128).
+             * @param[in] is_add True when the address was added, false on remove.
+             */
+            void trackLocalIp(
+                    _In_ const vpp_ip_route_t *prefix,
+                    _In_ bool is_add);
+
+            /**
+             * @brief If the assembled rule list contains any deny (drop) rule
+             * -- e.g. the dual-ToR mux drop -- prepend permit rules for the
+             * switch's own local addresses so local (for-us) traffic bypasses the
+             * deny and is punted at ip4-local.
+             *
+             * @param[in] tbl_oid ACL table being (re)configured.
+             * @param[in,out] acl_rules Ordered rule list; local permits are
+             * spliced onto the front when a deny rule is present.
+             * @return Number of permit rules prepended (0 if none).
+             */
+            size_t injectLocalPermits(
+                    _In_ sai_object_id_t tbl_oid,
+                    _Inout_ std::list<vpp_acl_rule_t> &acl_rules);
 
             sai_status_t AclTblRemove(
                     _In_ sai_object_id_t tbl_oid);
--- a/vslib/vpp/SwitchVppAcl.cpp
+++ b/vslib/vpp/SwitchVppAcl.cpp
@@ -21,10 +21,47 @@
 #include <list>
 #include <vector>
 #include <algorithm>
+#include <cstring>
 
 using namespace saivs;
 
 #define DEFAULT_PERMIT_RULES 2
+
+static bool acl_rule_is_deny(const vpp_acl_rule_t &r)
+{
+    return r.action == VPP_ACL_ACTION_API_DENY;
+}
+
+static vpp_acl_rule_t make_local_permit(const swss::IpAddress &ip)
+{
+    vpp_acl_rule_t rule = {};
+
+    rule.action = VPP_ACL_ACTION_API_PERMIT;
+
+    // VPP requires src and dst prefixes of an ACL rule to share the same address
+    // family, so set a same-family wildcard source (0.0.0.0/0 or ::/0). Leaving
+    // src unset defaults it to IPv4 in the xlate, which for an IPv6 dst yields a
+    // family mismatch that VPP rejects with VNET_API_ERROR_INVALID_SRC_ADDRESS.
+    if (ip.isV4()) {
+        rule.src_prefix.sa_family = AF_INET;        // src 0.0.0.0/0
+        rule.src_prefix_mask.sa_family = AF_INET;
+
+        rule.dst_prefix.sa_family = AF_INET;
+        rule.dst_prefix.addr.ip4.sin_addr.s_addr = ip.getV4Addr();
+        rule.dst_prefix_mask.sa_family = AF_INET;
+        rule.dst_prefix_mask.addr.ip4.sin_addr.s_addr = 0xFFFFFFFF; // /32
+    } else {
+        rule.src_prefix.sa_family = AF_INET6;       // src ::/0
+        rule.src_prefix_mask.sa_family = AF_INET6;
+
+        rule.dst_prefix.sa_family = AF_INET6;
+        memcpy(rule.dst_prefix.addr.ip6.sin6_addr.s6_addr, ip.getV6Addr(), 16);
+        rule.dst_prefix_mask.sa_family = AF_INET6;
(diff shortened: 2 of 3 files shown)


yue-fred-gao (commented):
vpp only supports about 80 rules. This can easily take half of that. Is there anyway to limit this feature only to where it is actually needed?
Have you run t1-lag sanity?

lolyu:
> vpp only supports about 80 rules. This can easily take half of that. Is there anyway to limit this feature only to where it is actually needed?
> Have you run t1-lag sanity?

Hi @yue-fred-gao, could you please elaborate on the 80 ACL rule limit? can we increase it?

I think this is a gap that any ACL drop rule can impact the traffic to local RIF addresses.

PR #2029, opened 2026-08-11
[vpp] Drive sonic_ext ip2me ACL-bypass from drop-ACL port binds

### LINKED ISSUE #28884: Bug: [vpp][dualtor] Mux drop ACL drops control-plane/for-us traffic — IN_PORTS ignored and L2 ACL runs before ip4-local punt (mux stuck unhealthy)

lolyu:
### Description of PR

Summary:
Fixes https://github.com/sonic-net/sonic-buildimage/issues/28884

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

SONiC VPP now drops ICMP replies to local RIF addresses (VLAN, Loopback, PortChannel, etc) if a drop ACL rule is installed. The correct behavior should be CoPP before ACL (any for-us traffic should be trapped to kernel before dataplane ACL rules).

This PR is to fix this issue.

##### Work item tracking
- Microsoft ADO **(number only)**: 39223945

#### How did you do it?

sonic-platforms-vpp PR: https://github.com/sonic-net/sonic-platform-vpp/pull/272 add sonic-ext-ip2me-ip4/ip6 nodes on the l2-input-ip4/ip6 feature arcs, running before acl-plugin-in-ip*-l2. Each frame gets a FIB receive-lookup on the inner dst IP; a DPO_RECEIVE (ip2me) result steers it to l2-input-feat-arc-end, bypassing the ACL so for-us traffic still reaches the BVI/ip*-local for punt while transit traffic hits the ACL as usual. This restores hardware-style CoPP-before-ACL ordering 

This PR enables this on exactly the interfaces where such a drop ACL is bound.

Build and setup an active-standby dualtor testbed, verify ICMP heartbeats can be received on standby mux ports.

### DIFF
--- a/vslib/vpp/SwitchVppAcl.cpp
+++ b/vslib/vpp/SwitchVppAcl.cpp
@@ -1198,6 +1198,21 @@ sai_status_t SwitchVpp::AclTblConfig(
     SWSS_LOG_INFO("Generated %ld regular ACL rules and %ld tunterm ACL rules",
                     acl_rules.size(), tunterm_acl_rules.size());
 
+    // ip2me: record whether this table now carries a DROP/deny rule so the
+    // feature can be (dis)enabled on the interfaces it is bound to. A table
+    // with a deny rule can discard ip2me (for-us) traffic on the l2-input
+    // arc before it is punted, which is exactly what the ip2me node guards.
+    {
+        bool has_deny = false;
+        for (const auto &rule : acl_rules) {
+            if (rule.action == VPP_ACL_ACTION_API_DENY) {
+                has_deny = true;
+                break;
+            }
+        }
+        ip2meUpdateDropTable(tbl_oid, has_deny);
+    }
+
     // Create and populate regular ACL if we have rules
     if (!acl_rules.empty()) {
         auto tbl_sid = sai_serialize_object_id(tbl_oid);
@@ -1441,6 +1456,11 @@ sai_status_t SwitchVpp::AclTblRemove(
         }
         m_acl_tbl_rules_map.erase(it);
     }
+
+    // ip2me: the table is going away -- clear its drop-table state and
+    // re-evaluate any interfaces that still reference it (defensive; the
+    // per-port unbind normally runs before the table is deleted).
+    ip2meUpdateDropTable(tbl_oid, false);
 
     status = tunterm_acl_delete(tbl_oid, true);
 
@@ -1824,6 +1844,118 @@ sai_status_t SwitchVpp::addRemovePortTblGrp(
     return SAI_STATUS_SUCCESS;
 }
 
+/*
+ * ip2me (receive-DPO check before ACL) -- sonic_ext plugin hook.
+ *
+ * Keep the sonic_ext ip2me feature enabled on exactly the VPP interfaces
+ * that have an ingress ACL containing a DROP/deny rule bound, so ip2me
+ * (destined-to-this-switch) traffic can bypass that drop ACL and still be
+ * punted (mirrors hardware CoPP-before-ACL). The feature is a no-op on
+ * every other interface and consumes zero ACL rules.
+ *
+ * State is split so the enable decision is robust to event ordering:
+ *   - m_ip2me_drop_tables   : tables that currently carry a deny rule
+ *                             (updated by AclTblConfig on every rule change).
+ *   - m_ip2me_port_tables   : per interface, the ingress tables bound to it
+ *                             (drop or not), so a table whose drop-ness flips
+ *                             after it was bound can be re-evaluated.
+ *   - m_ip2me_enabled_ports : interfaces where the feature is programmed on,
+ *                             so the vpp enable/disable stays idempotent.
+ */
+void SwitchVpp::ip2meRefreshPort(
+        _In_ const std::string &hwif_name)
+{
+    SWSS_LOG_ENTER();
+
+    // Desired state: enabled iff any ingress table bound to this interface
+    // is currently a drop table.
+    bool want_enable = false;
+    auto pit = m_ip2me_port_tables.find(hwif_name);
+    if (pit != m_ip2me_port_tables.end()) {
+        for (auto tbl_oid : pit->second) {
+            if (m_ip2me_drop_tables.count(tbl_oid)) {
+                want_enable = true;
+                break;
+            }
+        }
+    }
+
+    bool is_enabled = (m_ip2me_enabled_ports.count(hwif_name) != 0);
+    if (want_enable == is_enabled) {
+        return; // already in the desired state
+    }
+
+    int ret = vpp_sonic_ext_ip2me_enable_disable(hwif_name.c_str(), want_enable);
+    if (ret != 0) {
+        SWSS_LOG_ERROR("ip2me %s failed (%d) for interface %s",
+                       want_enable ? "enable" : "disable", ret, hwif_name.c_str());
+        return; // leave tracked state unchanged so a later refresh retries
+    }
+
+    if (want_enable) {
+        m_ip2me_enabled_ports.insert(hwif_name);
+    } else {
+        m_ip2me_enabled_ports.erase(hwif_name);
+    }
+    SWSS_LOG_NOTICE("ip2me skip ACL %s on interface %s",
+                    want_enable ? "enabled" : "disabled", hwif_name.c_str());
+}
+
+void SwitchVpp::ip2meUpdateDropTable(
+        _In_ sai_object_id_t tbl_oid,
+        _In_ bool has_deny)
+{
+    SWSS_LOG_ENTER();
+
+    bool was_drop = (m_ip2me_drop_tables.count(tbl_oid) != 0);
+    if (has_deny == was_drop) {
+        return; // drop-ness unchanged
+    }
+
+    if (has_deny) {
+        m_ip2me_drop_tables.insert(tbl_oid);
+    } else {
+        m_ip2me_drop_tables.erase(tbl_oid);
+    }
+
+    // Drop-ness flipped: re-evaluate every interface that currently has this
+    // table bound (the bind may have happened before the rules were added,
+    // or the last deny rule was just removed).
+    for (auto &kv : m_ip2me_port_tables) {
+        if (kv.second.count(tbl_oid)) {
+            ip2meRefreshPort(kv.first);
+        }
+    }
+}
+
+void SwitchVpp::ip2meUpdatePortBinding(
+        _In_ const std::string &hwif_name,
+        _In_ sai_object_id_t tbl_oid,
+        _In_ bool is_input,
+        _In_ bool is_bind)
+{
+    SWSS_LOG_ENTER();
+
+    // ip2me runs on the l2-input arc, so only ingress binds are relevant.
+    if (!is_input) {
+        return;
+    }
+
+    if (is_bind) {
+        m_ip2me_port_tables[hwif_name].insert(tbl_oid);
+    } else {
+        auto pit = m_ip2me_port_tables.find(hwif_name);
+        if (pit != m_ip2me_port_tables.end()) {
+            pit->second.erase(tbl_oid);
+            if (pit->second.empty()) {
+                m_ip2me_port_tables.erase(pit);
+            }
+        }
+    }
+
+    ip2meRefreshPort(hwif_name);
+}
+
 sai_status_t SwitchVpp::aclBindUnbindPort(
         _In_ sai_object_id_t port_oid,
         _In_ sai_object_id_t tbl_grp_oid,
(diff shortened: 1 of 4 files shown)


vslib/vpp/vppxlate/SaiVppXlate.c
```diff
@@ -3497,6 +3534,47 @@ int vpp_sflow_sampling_rate_set(uint32_t sampling_n)
...
+        idx = get_swif_idx(vam, hwif_name);
+        if (idx != (u32) -1) {
+            mp->sw_if_index = htonl(idx);
+        } else {
+            SAIVPP_ERROR("Unable to get the sw_index for %s\n", hwif_name);
+            VPP_UNLOCK();
+            return -EINVAL;
+        }
+    } else {
+        VPP_UNLOCK();
```
yue-fred-gao, line 3791: Please add an error message

vslib/vpp/SwitchVppAcl.cpp
```diff
@@ -1824,6 +1847,118 @@ sai_status_t SwitchVpp::addRemovePortTblGrp(
...
+{
+    SWSS_LOG_ENTER();
+
+    // ip2me runs on the l2-input arc, so only ingress binds are relevant.
+    if (!is_input) {
+        return;
+    }
+
+    if (is_bind) {
+        m_ip2me_port_tables[hwif_name].insert(tbl_oid);
```
yue-fred-gao: technically, we can navigate SaiObject graph to find the mapping from port to table or reverse. port (get_linked_object) -> table_group (get_child_objects) -> table_group_member (get_linked_object) -> table.  For the reverse lookup, table (get_child_objects) -> table_group_member (get_linked_object) -> table_group (get_child_objects) -> port. The amount of code is not less but the benefit is that we don't maintain custom mappings.

vslib/vpp/SwitchVppAcl.cpp
```diff
@@ -1917,7 +2055,12 @@ sai_status_t SwitchVpp::aclBindUnbindPort(
             return SAI_STATUS_FAILURE;
         }
         SWSS_LOG_NOTICE("ACL swindex %u %s to port %s (priority %u)", acl_swindex,
-                        is_bind ? "bound" : "unbound", hwif_name.c_str(), member.first);
+                        is_bind ? "bound" : "unbound", hwif_name.c_str(), priority);
+
+        // ip2me: only now that the (un)bind actually succeeded in VPP do we
+        // track this (interface, table) ingress binding and enable/disable
+        // the feature, so the tracked state matches what VPP is enforcing.
+        ip2meUpdatePortBinding(hwif_name, tbl_oid, is_input, is_bind);
```
yue-fred-gao: The mapping is not specific to ip2me. I think we should generalize the member variable name and the functions. It can be used in other places where we need to find port to table or table to port. If you do want to keep the custom mapping, let's make it generic and include egress ACL as well.

vslib/vpp/vppxlate/SaiVppXlate.c
lolyu: Added an `SAIVPP_ERROR` in the `hwif_name == NULL` branch of `vpp_sonic_ext_ip2me_enable_disable()`, before releasing the lock and returning `-EINVAL`:

```c
} else {
    SAIVPP_ERROR("%s: hwif_name is NULL, cannot %s ip2me", __func__,
                 enable ? "enable" : "disable");
    VPP_UNLOCK();
    return -EINVAL;
}
```

vslib/vpp/SwitchVppAcl.cpp
```diff
@@ -1198,6 +1198,21 @@ sai_status_t SwitchVpp::AclTblConfig(
...
+    {
+        bool has_deny = false;
+        for (const auto &rule : acl_rules) {
+            if (rule.action == VPP_ACL_ACTION_API_DENY) {
+                has_deny = true;
+                break;
+            }
+        }
+        ip2meUpdateDropTable(tbl_oid, has_deny);
+    }
```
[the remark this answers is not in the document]
lolyu: Good catch — both in-memory state updates now happen only *after* the VPP programming succeeds, so they can no longer diverge from what VPP is enforcing:

- `ip2meUpdateDropTable()` is now inside the `if (status == SAI_STATUS_SUCCESS)` guard in `AclTblConfig()` (after `acl_add_replace()` / `tunterm_acl_add_replace()`). A failed ACL program no longer flips the drop-table bit.
- The per-port binding update (`updatePortAclTableBinding()` + `ip2meRefreshPort()`) now runs only after `vpp_acl_interface_bind()` / `vpp_acl_interface_unbind()` return success (the `if (ret != 0) return SAI_STATUS_FAILURE;` guard is above it).

Resolving as addressed.

vslib/vpp/SwitchVppAcl.cpp
lolyu: We considered navigating the SaiObject graph (port → table_group → table_group_member → table, and the reverse). As you note it isn't less code, and the custom map buys us two things the traversal doesn't: (1) it lets us re-evaluate a table whose drop-ness *flips after* it was already bound (event-ordering robustness) without re-walking the graph on every rule change, and (2) the reverse lookup is a simple `getPortsWithAclTable()` over an in-memory map.

So we kept the mapping but generalized it per your other comment (now `m_port_acl_tables` with generic `updatePortAclTableBinding()` / `getPortsWithAclTable()` helpers, tracking both ingress and egress). Resolving as addressed.

vslib/vpp/SwitchVppAcl.cpp
lolyu: Generalized as suggested:

- Renamed `m_ip2me_port_tables` → `m_port_acl_tables`, now a generic per-interface `struct PortAclTables { std::set<sai_object_id_t> ingress; std::set<sai_object_id_t> egress; }` that tracks **both** directions.
- Replaced the ip2me-specific `ip2meUpdatePortBinding()` with feature-agnostic helpers:
  - `updatePortAclTableBinding(hwif_name, tbl_oid, is_input, is_bind)` — maintains the forward map for either direction.
  - `getPortsWithAclTable(tbl_oid, is_input)` — reverse (table → ports) lookup for a direction.
- Both ACL bind paths (`aclBindUnbindPort` / `aclBindUnbindPorts`) now record ingress *and* egress binds. ip2me is now just one consumer: `ip2meRefreshPort()` reads the `.ingress` set and `ip2meUpdateDropTable()` uses `getPortsWithAclTable(tbl_oid, /*is_input*/ true)`.

Resolving as addressed.

yue-fred-gao (dismissed):

yue-fred-gao:
please trigger t1-lag-vpp sanity to avoid breakage before merging it

yue-fred-gao:
the change looks good. Please trigger t1-lag-vpp sanity and share the result.

yue-fred-gao (approved):

PR #1992, opened 2026-07-15
[Draft Only - Do Not Merge/Review] SAI VPP UT Harness - Phase 3

nicholasching:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

##### Work item tracking
- Microsoft ADO **(number only)**:

PR #1952, opened 2026-06-23
[vpp] VPP SAI backend fixes: CPU queue, host-interface admin-up, L3-over-LAG/host-route forwarding, LAG egress-disable

nicholasching:
### Description of PR

Summary:
VPP SAI backend (`vslib/vpp/`) correctness fixes uncovered while bringing the OCP `sai_test` suite to green against `libsaivs` — neighbor/route/LAG/ECMP programming and forwarding. Pure backend; no harness or docs. Reviewers can start in `vslib/vpp/SwitchVppNbr.cpp` and `vslib/vpp/SwitchVppFdb.cpp`. Development context, before/after results, and the compatibility matrix are in the harness PR's `.azure-pipelines/docker-sai-test-vpp/devdocs/`.

Fixes # (N/A — no upstream issue)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

#### What is the motivation for this PR?
Several VPP SAI backend gaps caused neighbor/route/LAG/ECMP tests to fail at create time or in the dataplane (e.g. host routes dropped at `local0`, LAG-member egress-disable not enforced, missing CPU-queue parent). These fixes make the corresponding `sai_test` cases pass.

#### How did you do it?
All changes under `vslib/vpp/`:
- `SwitchVppNbr.cpp`: neighbor host route was pinned to `local0` because the host-route nexthop `sw_if_index` was left at 0; set `sw_if_index = ~0` so `ip_route_add_del()` resolves the egress via `hwif_name`. Also program the implicit host route on neighbor add/remove and check RIF type before reading `PORT_ID` (avoids `-7` on non-PORT/SUB_PORT RIFs).
- `SwitchVpp.cpp/.h`, `SwitchVppFdb.cpp`: LAG-RIF neighbor/route create and `SAI_LAG_MEMBER_ATTR_EGRESS_DISABLE` (detach/reattach the VPP bond member); BVI/bond FDB handling for L3-over-LAG forwarding (backend portion — the harness IP-assignment half is in the harness PR).
- `SwitchVppRif.cpp`, `SwitchVppRoute.cpp`, `SwitchVppNexthop.cpp`: ROUTER_INTERFACE next-hop + LAG `BondEthernet`/`be<N>` interface-IP support, NHG-member idempotency, `vppProcessEvents` sleep tuning.
[...]

PR #1951, opened 2026-06-23
[vpp] Dedicated binary-API socket and thread for asynchronous VPP events

nicholasching:
### Description of PR

Summary:
Part of the SAIVPP unit-test framework (`docker-sai-test-vpp`, #1950, see its `devdocs/`). Carries the asynchronous VPP event stream on a dedicated binary-API socket and background thread, separate from the synchronous request/reply command path. This resolves a mutex deadlock between the command path and the async event path that blocked bring-up of the OCP `sai_test` suite at `PORT_COUNT=32`. Reviewers can start in `vslib/vpp/vppxlate/SaiVppXlate.c`.

Fixes # (N/A — no upstream issue)

### Type of change

- [x] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

#### What is the motivation for this PR?
The synchronous command path and the asynchronous VPP event stream shared a single binary-API socket and the global `__plugin_msg_base`. At scale this deadlocked between the command path and the async event path, preventing the standalone SAIVPP UT (`SaiSanityTest`) from passing.

#### How did you do it?
In `vslib/vpp/vppxlate/SaiVppXlate.c`: introduced a dedicated event binary-API socket (`event_socket_client_main` / `vat_event_main`) and background thread; added event-connection message macros (`M_EV` / `PING_EV` / `S_EV` / `WR_EV`) that operate on the explicit per-connection socket via the `vl_socket_client_*2()` APIs; added a separate lock for the event connection; and made `__plugin_msg_base` thread-local so the command and event threads cannot race between setting the base and the following `M()`/`PING()`.

#### How did you verify/test it?
Built into the `docker-sai-test-vpp` image and run against the OCP `sai_test` suite; the command-vs-event deadlock at 32 ports no longer occurs and `SaiSanityTest` passes. The change is orthogonal to the recently-merged route-flow-counter context-table rework (#1894) and ipinip support (#1860); rebased on current `master` and verified both coexist.

#### Any platform specific information?
VPP platform only (`vslib/vpp/`). No change to the synchronous command path or SAI semantics; the event connection is additive.

### Documentation
Development notes for this work are in `.azure-pipelines/docker-sai-test-vpp/devdocs/` in #1950.

PR #2032, opened 2026-08-12
[vpp] SAIVPP harness compatibility: VPP 26.06 startup plugins, backend-restart reaping, harness log routing

nicholasching:
### Description of PR

Summary:
Follow-up to the merged SAIVPP unit-test framework ([#1950](https://github.com/sonic-net/sonic-sairedis/pull/1950), [#1951](https://github.com/sonic-net/sonic-sairedis/pull/1951), [#1952](https://github.com/sonic-net/sonic-sairedis/pull/1952)). Restores the compatibility that the harness lost as `master` and VPP moved forward: two VPP 26.06 startup plugins the harness needs, a backend-restart race in `run_test.sh`, harness log routing, and the corresponding documentation refresh. Small and mostly mechanical; reviewers can start at `.azure-pipelines/docker-sai-test-vpp/vpp_startup.conf.template`.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation update
- [x] Test improvement

The Phase 3 development branch was rebuilt on a newer `master` and a newer VPP than Phase 1/2 targeted. That surfaced two independent breakages: VPP 26.06 moved the TAP driver into its own plugin and `saiserver` began asserting on a missing sflow message-id base, so the harness could not create `linux_cp` host interfaces; and the per-test backend recycle could overlap a dying VPP instance. Neither is a VPP dataplane behavior change — they are the cost of catching up to current upstream.

#### How did you do it?
- `.azure-pipelines/docker-sai-test-vpp/vpp_startup.conf.template`: enable `tap_plugin.so` (VPP 26.06 relocated the TAP driver `linux_cp` needs; without it `lcp_itf_pair_add_del` returns `FEATURE_DISABLED` and host-interface bring-up fails) and `sflow_plugin.so` (saiserver startup asserts if the sflow message-id base is absent).
- `.azure-pipelines/docker-sai-test-vpp/run_test.sh`: `terminate_process()` now `wait`s on each VPP/saiserver/Redis PID after it dies, so the next `start_vpp()` cannot overlap a dying instance during the per-test recycle.
- `.azure-pipelines/docker-sai-test-vpp/swss_log_stdout_preload.cpp`: route `SWSS_LOG_*` to **stderr** instead of stdout. `run_test.sh` redirects both streams into `saiserver.log`, and keeping `SWSS_LOG_ENTER` lines off stdout stops them from polluting `swss::exec()` captured output (e.g. the `find_new_bond_id` shell pipeline used during LAG create).
- `.azure-pipelines/docker-sai-test-vpp/README.md`, `build_harness.sh`, `gen_compatibility_matrix.py`: refresh the harness documentation for the current state, retire the stale `phase1` image tag in favour of Docker's default `docker-sai-test-vpp:latest` (per review feedback), document the per-test isolation default and the newly required VPP plugins, and add a Monitor section for detached long matrix runs.

Four commits that were originally in this PR moved into the SAI submodule-bump PR. Three are metadata initializers (`syncd/SwitchNotifications.h`, `unittest/meta/TestSaiSerialize.cpp`, `unittest/syncd/TestAttrVersionChecker.cpp`), because the members they initialize do not exist in the SAI revision `master` currently pins. The fourth is the `pyext/py3/Makefile.am` `-Wno-error=disabled-optimization` flag: the newer SAI headers are what enlarge the SWIG-generated `pysairedis` wrapper past GCC's `max-gcse-memory` limit on armhf, so that flag belongs with the bump rather than here.

`bash -n` on the harness scripts, `tests/checkwhitespace.sh`, `tests/swsslogentercheck.sh`, and `git diff --check` all pass. The compatibility fixes were originally found and validated on the Phase 3 integration branch, where the harness could not build the image or complete a backend recycle without them. No new pass counts are claimed here: this PR restores the ability to build and run, and the gated matrix numbers live in the CI PR.

VPP unit-test harness only. No `vslib/vpp/` dataplane behavior changes, so production VPP forwarding is untouched. The two newly enabled VPP plugins apply to the harness-generated `startup.conf` only, not to any product configuration.

### Documentation
Updates the harness `README.md`: per-test isolation vs grouped mode, the required VPP startup plugins and why they cannot be trimmed, the `docker-sai-test-vpp:latest` tag convention, and detached-run/monitoring guidance.

Based on current upstream `master`; no required merge order. File-disjoint from the backend PR below. This PR and the CI PR both touch `docker-sai-test-vpp/README.md`, `run_test.sh`, and `build_harness.sh`, so whichever merges second needs a rebase — that is a textual overlap, not a functional dependency. Development context for this work is in the Phase 3 integration draft [#1992](https://github.com/sonic-net/sonic-sairedis/pull/1992), which stays open as history and should not be merged.

### DIFF
--- a/.azure-pipelines/docker-sai-test-vpp/build_harness.sh
+++ b/.azure-pipelines/docker-sai-test-vpp/build_harness.sh
@@ -8,7 +8,7 @@ BUILDIMAGE_ROOT="${SONIC_BUILDIMAGE:-$(cd "${SAIREDIS_ROOT}/../.." && pwd)}"
 DEB_STAGING="${SCRIPT_DIR}/debs"
 DOCKERFILE="${SAIREDIS_ROOT}/.azure-pipelines/docker-sai-test-vpp/Dockerfile"
 BLDENV="${BLDENV:-trixie}"
-IMAGE_TAG="${IMAGE_TAG:-docker-sai-test-vpp:phase1}"
+IMAGE_TAG="${IMAGE_TAG:-docker-sai-test-vpp:local}"
 STAGE_DEBS="${STAGE_DEBS:-1}"
 BUILD_SAIREDIS_DEBS="${BUILD_SAIREDIS_DEBS:-0}"
 AUTO_BUILD_SAIREDIS_DEBS="${AUTO_BUILD_SAIREDIS_DEBS:-1}"
@@ -35,7 +35,7 @@ Options:
   --no-auto-build       Do not invoke make when sonic-sairedis .debs are
                         missing (validate only; fail with hints)
   --bldenv <name>       Debian suite for target/debs (default: trixie)
-  --image-tag <tag>     Docker image tag (default: docker-sai-test-vpp:phase1)
+  --image-tag <tag>     Docker image tag (default: docker-sai-test-vpp:local)
   --vpp-deb-dir <path>  Directory of VPP .debs (default: search buildimage
                         target/debs/<bldenv> and VPP_DEB_DIR env)
   -h, --help            Show this help
--- a/.azure-pipelines/docker-sai-test-vpp/build_harness.sh
+++ b/.azure-pipelines/docker-sai-test-vpp/build_harness.sh
@@ -8,7 +8,7 @@ BUILDIMAGE_ROOT="${SONIC_BUILDIMAGE:-$(cd "${SAIREDIS_ROOT}/../.." && pwd)}"
 DEB_STAGING="${SCRIPT_DIR}/debs"
 DOCKERFILE="${SAIREDIS_ROOT}/.azure-pipelines/docker-sai-test-vpp/Dockerfile"
 BLDENV="${BLDENV:-trixie}"
-IMAGE_TAG="${IMAGE_TAG:-docker-sai-test-vpp:local}"
+IMAGE_TAG="${IMAGE_TAG:-docker-sai-test-vpp:latest}"
 STAGE_DEBS="${STAGE_DEBS:-1}"
 BUILD_SAIREDIS_DEBS="${BUILD_SAIREDIS_DEBS:-0}"
 AUTO_BUILD_SAIREDIS_DEBS="${AUTO_BUILD_SAIREDIS_DEBS:-1}"
@@ -35,7 +35,7 @@ Options:
   --no-auto-build       Do not invoke make when sonic-sairedis .debs are
                         missing (validate only; fail with hints)
   --bldenv <name>       Debian suite for target/debs (default: trixie)
-  --image-tag <tag>     Docker image tag (default: docker-sai-test-vpp:local)
+  --image-tag <tag>     Docker image tag (default: docker-sai-test-vpp:latest)
   --vpp-deb-dir <path>  Directory of VPP .debs (default: search buildimage
                         target/debs/<bldenv> and VPP_DEB_DIR env)
   -h, --help            Show this help
--- a/.azure-pipelines/docker-sai-test-vpp/vpp_startup.conf.template
+++ b/.azure-pipelines/docker-sai-test-vpp/vpp_startup.conf.template
@@ -45,6 +45,7 @@ plugins {
   plugin vxlan_plugin.so { enable }
   plugin tunterm_acl_plugin.so { enable }
   plugin ip_validate_plugin.so { enable }
+  plugin sflow_plugin.so { enable }
 }
 
 linux-cp {
--- a/.azure-pipelines/docker-sai-test-vpp/vpp_startup.conf.template
+++ b/.azure-pipelines/docker-sai-test-vpp/vpp_startup.conf.template
@@ -38,6 +38,7 @@ ip6 {
 plugins {
   plugin default { disable }
   plugin af_packet_plugin.so { enable }
+  plugin tap_plugin.so { enable }
   plugin linux_cp_plugin.so { enable }
   plugin linux_nl_plugin.so { enable }
   plugin acl_plugin.so { enable }
--- a/.azure-pipelines/docker-sai-test-vpp/README.md
+++ b/.azure-pipelines/docker-sai-test-vpp/README.md
@@ -57,7 +57,7 @@ The table below lists OCP `sai_test` classes that **pass** on the current VPP SA
 
 ## b) Building the framework
 
-The image bundles pre-built `.deb` packages from `debs/` (git-ignored locally). You only need to regenerate `.deb`s when the corresponding source changes. All examples below use the local image tag **`docker-sai-test-vpp:local`** (the `build_harness.sh` default).
+The image bundles pre-built `.deb` packages from `debs/` (git-ignored locally). You only need to regenerate `.deb`s when the corresponding source changes. All examples below use **`docker-sai-test-vpp:latest`**, the tag `build_harness.sh` applies by default. Because `latest` is Docker's implicit tag you can also drop it and write just `docker-sai-test-vpp`; CI images are tagged with the pipeline definition and build number instead, so they never collide with a locally built one.
 
 ### Use cases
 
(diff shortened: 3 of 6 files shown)


.azure-pipelines/docker-sai-test-vpp/build_harness.sh
```diff
@@ -8,7 +8,7 @@ BUILDIMAGE_ROOT="${SONIC_BUILDIMAGE:-$(cd "${SAIREDIS_ROOT}/../.." && pwd)}"
 DEB_STAGING="${SCRIPT_DIR}/debs"
 DOCKERFILE="${SAIREDIS_ROOT}/.azure-pipelines/docker-sai-test-vpp/Dockerfile"
 BLDENV="${BLDENV:-trixie}"
-IMAGE_TAG="${IMAGE_TAG:-docker-sai-test-vpp:phase1}"
+IMAGE_TAG="${IMAGE_TAG:-docker-sai-test-vpp:local}"
```
yue-fred-gao: should we use latest?

.azure-pipelines/docker-sai-test-vpp/vpp_startup.conf.template
```diff
@@ -38,12 +38,14 @@ ip6 {
 plugins {
   plugin default { disable }
   plugin af_packet_plugin.so { enable }
+  plugin tap_plugin.so { enable }
```
yue-fred-gao, line 41: This cross dependency is concerning. If vpp is upgraded and requires a new plugin, what's the correct procedure? we want to add saitest to sonic-platform-vpp pr check, which downloads docker-sai-test-vpp container from sairedis pipeline artifacts, which has old startup without the new plugin. pr check fails. If we add the plugin to this template first, will it crash due to the plugin is missing?

syncd/SwitchNotifications.h
```diff
@@ -169,6 +169,7 @@ namespace syncd
                             .on_ipsec_post_status = &Slot<context>::onIpsecPostStatus,
                             .on_switch_macsec_post_status = &Slot<context>::onSwitchMacsecPostStatus,
                             .on_switch_ipsec_post_status = &Slot<context>::onSwitchIpsecPostStatus,
+                            .on_next_hop_group_hw_protection_switchover = nullptr,
```
yue-fred-gao: is this due to upgrading to the new SAI submodule? does this function pointer exist in older SAI?

unittest/meta/TestSaiSerialize.cpp
```diff
@@ -1299,8 +1299,8 @@ TEST(SaiSerialize, serialize_qos_map)
     attr.id = SAI_QOS_MAP_ATTR_MAP_TO_VALUE_LIST;
 
     sai_qos_map_t qm = {
-        .key   = { .tc = 1, .dscp = 2, .dot1p = 3, .prio = 4, .pg = 5, .queue_index = 6, .color = SAI_PACKET_COLOR_RED, .mpls_exp = 0, .fc = 2 },
-        .value = { .tc = 11, .dscp = 22, .dot1p = 33, .prio = 44, .pg = 55, .queue_index = 66, .color = SAI_PACKET_COLOR_GREEN, .mpls_exp = 0, .fc = 2 } };
+        .key   = { .tc = 1, .dscp = 2, .dot1p = 3, .prio = 4, .pg = 5, .queue_index = 6, .color = SAI_PACKET_COLOR_RED, .mpls_exp = 0, .fc = 2, .dei = 0, .vc = 0 },
```
yue-fred-gao: is this also due to sai version change?

syncd/SwitchNotifications.h
nicholasching: Hi Fred, yes, this is because of the SAI submodule pointer update:

`sai_switch_notifications_t` is generated by SAI metadata from the `SAI_SWITCH_ATTR_*_NOTIFY` attributes; master's pinned SAI (`c67f1152`) declares 21 attributes; the newer SAI adds `SAI_SWITCH_ATTR_NEXT_HOP_GROUP_HW_PROTECTION_SWITCHOVER_NOTIFY` (`inc/sainexthopgroup.h`, `inc/saiswitch.h`). 

Looking into this more, the SAI pointer update will need to be bundled with this change as they are interdependent. There are a total of 3 commits with similar SAI update changes. I will create a separate PR for that to make review easier and clear CI failures.

https://github.com/sonic-net/sonic-sairedis/pull/2035

unittest/meta/TestSaiSerialize.cpp
nicholasching: Yes, `sai_uint8_t dei` and `vc` were added to `sai_qos_map_params_t` in `inc/saitypes.h` in the newer SAI commits. I will also move this change into the other PR.

https://github.com/sonic-net/sonic-sairedis/pull/2035

syncd/SwitchNotifications.h
yue-fred-gao: > Hi Fred, yes, this is because of the SAI submodule pointer update:
> 
> `sai_switch_notifications_t` is generated by SAI metadata from the `SAI_SWITCH_ATTR_*_NOTIFY` attributes; master's pinned SAI (`c67f1152`) declares 21 attributes; the newer SAI adds `SAI_SWITCH_ATTR_NEXT_HOP_GROUP_HW_PROTECTION_SWITCHOVER_NOTIFY` (`inc/sainexthopgroup.h`, `inc/saiswitch.h`).
> 
> Looking into this more, the SAI pointer update will need to be bundled with this change as they are interdependent. There are a total of 3 commits with similar SAI update changes. I will create a separate PR for that to make review easier and clear CI failures.

yes, SAI submodule update should be a separate PR. Can you check the past submodule update PRs? is it adhoc or on schedule?

nicholasching: Looks like previous updates:
- occurred in irregular timing patterns; there are updates as close as 2 weeks, but also as far as 3 months.
- were authored by many contributors from engineers across Nokia, Nexthop, Marvell, Nvidia, and Microsoft
- define a motivation that is linked to specific needs

So it seems submodule updates are adhoc and done when there's a need to do so.

yue-fred-gao: > Looks like previous updates:
> 
> * occurred in irregular timing patterns; there are updates as close as 2 weeks, but also as far as 3 months.
> * were authored by many contributors from engineers across Nokia, Nexthop, Marvell, Nvidia, and Microsoft
> * define a motivation that is linked to specific needs
> 
> So it seems submodule updates are adhoc and done when there's a need to do so.

Great. Let's create a PR for submodule update.

.azure-pipelines/docker-sai-test-vpp/vpp_startup.conf.template
nicholasching: Hi Fred, that is true, very good catch. 

I just finished testing w/VPP: enabling VPP plugin that is not installed will not crash VPP and does not result in a warning either, so pre-staging new VPP plugins on `sonic-sairedis` before `sonic-platform-vpp` release will work and clear `sonic-platform-vpp` checks. But this is an additional maintenance step that will need to be documented; where is the best place to include this documentation?

syncd/SwitchNotifications.h
nicholasching: Done! https://github.com/sonic-net/sonic-sairedis/pull/2035.

.azure-pipelines/docker-sai-test-vpp/build_harness.sh
nicholasching: Sure! All done.

nicholasching:

Github Run: https://github.com/sonic-net/sonic-buildimage/runs/94833752331
Azure Build: https://dev.azure.com/mssonic/build/_build/results?buildId=1193017&view=logs&j=314bc92a-d36a-5eac-5c45-43e8f1288876

yue-fred-gao (approved):

PR #2033, opened 2026-08-12
[vpp] VPP SAI backend fixes: detect bond-create failure, program route DROP actions

nicholasching:
### Description of PR

Summary:
Two VPP SAI backend (`vslib/vpp/`) correctness fixes found while bringing the OCP `sai_test` suite forward on the SAIVPP unit-test framework ([#1950](https://github.com/sonic-net/sonic-sairedis/pull/1950)), and not included in the earlier backend PR ([#1952](https://github.com/sonic-net/sonic-sairedis/pull/1952)). Pure backend — no harness, CI, or docs. Reviewers can start in `vslib/vpp/SwitchVppRoute.cpp`.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

Two gaps. First, `vpp_create_lag()` ignored the return code of `create_bond_interface()` and only rejected `sw_if_index == ~0`. When VPP rejects `bond_create` the reply carries a non-zero `retval` with `sw_if_index` 0 (`local0`), so the guard passed, the LAG was recorded against `local0`, and the real failure only surfaced later as a confusing lag-member create error. Second, a SAI route with `SAI_PACKET_ACTION_DROP` was silently ignored: the code took the "action is not forward" path and never programmed anything, so traffic to a supposedly dropped prefix kept following whatever less-specific route existed.

#### How did you do it?
- `vslib/vpp/SwitchVppFdb.cpp`: capture the `create_bond_interface()` return code and treat a non-zero return, `~0`, or `0` as failure, logging both `ret` and `swif_idx` so the VPP-side rejection is visible at the point of failure.
- `vslib/vpp/SwitchVppRoute.cpp`: handle `SAI_PACKET_ACTION_DROP` before the forward-only early return. Resolve the VRF, build a single-path `vpp_ip_route_t` with `VPP_NEXTHOP_DROP` and `sw_if_index = ~0`, and program it through the existing `ip_route_add_del_get_stats()` path so add and remove are symmetric. The `sai_deserialize_route_entry()` call moved above the packet-action branch because both paths now need the parsed entry.
- `vslib/vpp/vppxlate/SaiVppXlate.{c,h}`: add `VPP_NEXTHOP_DROP` to `vpp_nexthop_type_e` and map it to `FIB_API_PATH_TYPE_DROP` when building the FIB path.

`tests/checkwhitespace.sh`, `tests/swsslogentercheck.sh`, and `git diff --check` pass. Both fixes were exercised on the Phase 3 integration branch, where the bond-create guard converted a late, misleading lag-member error into an immediate and accurate LAG create failure, and the DROP path made `sai_route_test.DropRouteTest` / `DropRoutev6Test` program a real VPP drop route instead of being ignored. Those two selectors run in the full matrix but are deliberately **not** in the gated stable baseline yet (see the CI PR), so no gated pass-count change is claimed here.

VPP platform only (`vslib/vpp/`). The bond-create change is a strictly stronger error check on an existing failure path. The DROP change adds programming where the backend previously did nothing, so a VRF that relied on a DROP route being a no-op will now actually drop.

### Documentation
No new docs in this PR (kept backend-only). Development context and validation notes are in `sonic-buildimage/devdocs/merge-p3.md`.

Based on current upstream `master`; no required merge order. File-disjoint from both other Phase 3 PRs, verified by merging this branch and the compatibility branch onto `master` in both orders with no conflicts. Development context is in the Phase 3 integration draft [#1992](https://github.com/sonic-net/sonic-sairedis/pull/1992), which stays open as history.

### DIFF
--- a/vslib/vpp/SwitchVppFdb.cpp
+++ b/vslib/vpp/SwitchVppFdb.cpp
@@ -1202,10 +1202,11 @@ sai_status_t SwitchVpp::vpp_create_lag(
     mode = VPP_BOND_API_MODE_XOR;
     lb = VPP_BOND_API_LB_ALGO_L34_INNER;
 
-    create_bond_interface(bond_id, mode, lb, &swif_idx);
-    if (swif_idx == static_cast<uint32_t>(~0))
+    int ret = create_bond_interface(bond_id, mode, lb, &swif_idx);
+    if (ret != 0 || swif_idx == static_cast<uint32_t>(~0) || swif_idx == 0)
     {
-        SWSS_LOG_ERROR("failed to create bond interface in VPP for %s", sai_serialize_object_id(lag_id).c_str());
+        SWSS_LOG_ERROR("failed to create bond interface in VPP for %s (ret=%d, swif_idx=%u)",
+                sai_serialize_object_id(lag_id).c_str(), ret, swif_idx);
         return SAI_STATUS_FAILURE;
     }
 
--- a/vslib/vpp/SwitchVppRoute.cpp
+++ b/vslib/vpp/SwitchVppRoute.cpp
@@ -162,7 +162,35 @@ sai_status_t SwitchVpp::IpRouteAddRemove(
         packet_action = attr.value.s32;
     }
 
-    // We should program drop routes
+    sai_route_entry_t route_entry;
+    sai_deserialize_route_entry(serializedObjectId, route_entry);
+
+    if (packet_action == SAI_PACKET_ACTION_DROP) {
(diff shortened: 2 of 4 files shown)


yue-fred-gao (approved):

nicholasching:

Github Run: https://github.com/sonic-net/sonic-buildimage/runs/94612473195
Azure Build: https://dev.azure.com/mssonic/build/_build/results?buildId=1191981&view=logs&jobId=314bc92a-d36a-5eac-5c45-43e8f1288876

PR #2016, opened 2026-08-03
[ci]: Cut sairedis Build over to shared buildenv_setup (CI unification Phase 2A)

theasianpianist:
## Summary

Phase 2A of the dataplane CI-unification migration, following the merged
sonic-swss-common tool/canary PR #1222.

This PR makes sonic-sairedis's `build-env/` the single source of truth for its
Build-stage environment and replaces the duplicated apt/artifact/Redis/VPP setup
in `.azure-pipelines/build-template.yml` with the shared `buildenv_setup` tool
hosted by sonic-swss-common.

### What changes

- Add declarative `build-env/` configuration:
  - `packages/base.yaml` — sairedis build/link dependencies that cascade downstream.
  - `packages/tooling.yaml` — Build-only doc/rsyslog/sswsyncd setup.
  - `upstream-artifacts.yaml` — common-libs, sonic-swss-common, and VPP artifacts.
- Add canonical `build-env/build.sh`, used by both CI and local development.
- Add Docker Compose local-development support using the same setup config + build command as CI.
- Cut the complete sairedis Build matrix over (`Build`, `BuildAsan`, `BuildArm`, `BuildTrixie`).
- Publish `build-env/` additively beside the existing top-level DEBs so sonic-swss can cascade it later.
- Make the Build template reusable via `repo` + `staged_upstreams`, including required-staged fail-fast behavior for future same-run downstream builds.

### Behavior preserved / intentionally inherited

- **libnl:** preserve today's split — SONiC-built common-lib DEBs on amd64; stock apt libnl plus CLI headers on arm.
- **VPP:** remains amd64-only/master and installs with `VPP_INSTALL_SKIP_SYSCTL=1`; VPP stays in its own dpkg group.
- **Result filters:** common-libs requires success; swss-common and VPP continue allowing partially-successful builds.
- **Redis:** reuse sonic-swss-common's canonical `configure-redis-for-tests.sh` through the published artifact cascade (including `notify-keyspace-events AKE`, previously PoC-verified against sairedis).
- ASAN vs coverage build flags, post-build rsyslog rewrite, setcap/make-check tests, coverage and syslog publication remain unchanged.

### Scope boundaries

This PR changes the **Build-family stages only**. `BuildSwss`, `BuildDocker*`, and `Test*` remain on their existing local templates and setup paths; those are separate migration slices.

## Validation

- YAML parsing and `build-env/build.sh` shell syntax.
- `docker compose config` validation.
[...]

PR #2034, opened 2026-08-12
[vpp] Add VPP SAI compatibility CI: build and publish the test image, run the matrix, gate on a stable baseline

nicholasching:
### Description of PR

Summary:
Phase 3 of the SAIVPP unit-test framework: wires the harness added in [#1950](https://github.com/sonic-net/sonic-sairedis/pull/1950) into the pipeline. Adds a Trixie `saithriftv2` package build, a `BuildSaiTestVpp` stage that publishes a self-contained `docker-sai-test-vpp` image with provenance, and a `TestSaiVpp` stage that runs the full 85-selector OCP `sai_test` matrix and fails the check only on regressions against a reviewed 30-selector stable baseline. Reviewers can start at `azure-pipelines.yml` and `.azure-pipelines/build-docker-sai-test-vpp-template.yml`.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [x] Documentation update
- [x] Test improvement

[#1950](https://github.com/sonic-net/sonic-sairedis/pull/1950) landed the harness but left CI wiring as explicit follow-up work — its README recorded use case 2 (`sonic-sairedis` PR CI) as documented intent only. Without it, VPP SAI regressions are caught only when someone runs the harness by hand. This PR closes that gap and makes the result reviewable rather than advisory: the matrix always runs in full for visibility, and the gate is a small, explicitly reviewed set of selectors that must not regress.

#### How did you do it?
- `.azure-pipelines/build-template.yml`, `azure-pipelines.yml`: add an opt-in `saithrift_v2` step that builds `saiserverv2` / `python-saithriftv2` in the Trixie job, and enable it there. `.gitignore` gains the resulting root-level package artifacts.
- `.azure-pipelines/build-docker-sai-test-vpp-template.yml` (new) and the `BuildSaiTestVpp` stage: assemble an explicit Docker context, require exactly one of each runtime `.deb`, build the image, and publish it with `provenance.txt`, `SHA256SUMS`, and a `ci-contract/` directory so `sonic-platform-vpp` can consume the same image and evaluator.
- `.azure-pipelines/test-sai-vpp-template.yml` (new) and the `TestSaiVpp` stage: clean stale labelled containers/images, load the published image, run all four modules at `PORT_COUNT=32` with `ISOLATE_EACH_TEST=1`, publish JUnit results, and always archive logs and provenance.
- `.azure-pipelines/docker-sai-test-vpp/ci-matrix-tests.txt`, `ci-pass-tests.txt`, `evaluate_ci_baseline.py`, `test_evaluate_ci_baseline.py`: the selector contract. The evaluator first asserts the observed JUnit selector set matches the expected 85, then fails only on a baseline selector that is missing, failed, errored, or skipped. Non-baseline failures stay visible, new passes are reported as promotion candidates, and a harness exit code of 2 or more is treated as infrastructure failure rather than a test result. JUnit XML is parsed with `defusedxml`.
- `.azure-pipelines/docker-sai-test-vpp/derive_ptf_version.py`, `test_derive_ptf_version.py`, `Dockerfile`, `build_harness.sh`: derive a deterministic PEP 440 PTF version from the pinned `SAI/test/ptf` checkout and pass it as a build arg, so `pip install` no longer depends on `setuptools_scm` finding usable git metadata inside the image build. The CI checkout is `submodules: recursive` because `ptf` is nested under `SAI`.
- One VPP build for the whole run: a `ResolveVpp` job resolves the VPP artifact once, republishes it as `vpp-resolved-trixie`, and both the Trixie compile (via the shared `buildenv_setup` `staged_upstreams` hook) and the image build consume that same current-run artifact. This keeps compile-time and runtime VPP packages identical without reverting any part of the shared build-environment contract added in #2016. `build-template.yml` also accepts an optional `vpp_run_id`; this repository's stages do not use it, and it is retained as an override hook for consumers that must pin a specific VPP run.
- `azure-pipelines.yml`: `BuildArm` and `BuildTrixie` no longer serialize behind earlier stages. This is a repo-wide ordering change rather than a VPP-specific one, included here because the two new stages hang off `BuildTrixie` and would otherwise wait on the full ARM matrix first. Happy to split it out if maintainers prefer.
- `.azure-pipelines/docker-sai-test-vpp/README.md`: document the CI workflow, the selector contract, baseline promotion rules, and the flaky-but-not-gated ECMP classes.

Commit-history note: `fix: gate CI on 34-test baseline` first drops three dataplane-flaky ECMP classes, and `docs: set strict 30-test Phase 3 baseline` later lowers the gate to 30 after strict PTF validation exposed four more masked assertions. The intermediate counts are kept for reviewability; **30** is the final contract.

Locally: both Python helper suites pass (5 tests for `derive_ptf_version.py`, 6 for `evaluate_ci_baseline.py`, covering regressions, missing selectors, malformed JUnit, duplicate baseline entries, matrix-contract mismatch, and infrastructure exit codes); `bash -n` on the harness scripts; YAML parsing of all four pipeline files; `tests/checkwhitespace.sh`, `tests/swsslogentercheck.sh`, and `git diff --check`. The 30-selector baseline itself was established on the Phase 3 integration branch under strict PTF semantics — see [#1992](https://github.com/sonic-net/sonic-sairedis/pull/1992) for those runs. I have not re-run the full ~45–50 minute matrix from this split, so the gate needs one hosted burn-in run before it should be made mandatory.

VPP platform CI only. The test job needs a `--privileged` container for raw AF_PACKET on veths and TAPs, which is why the harness Dockerfile carries a justified `# nosemgrep` for its root ENTRYPOINT. No production runtime code is changed by this PR. Runtime green additionally requires the `SAI` submodule to include the merged `simulate_sonic` work from [opencomputeproject/SAI#2316](https://github.com/opencomputeproject/SAI/pull/2316) and the IPv6 control-traffic isolation from [opencomputeproject/SAI#2321](https://github.com/opencomputeproject/SAI/pull/2321); that pointer bump is maintainer-owned and intentionally not part of this PR.

### Documentation
Extends the harness `README.md` with the CI workflow, the producer/consumer artifact contract, the matrix and baseline files, how to run the evaluator locally, and the rule that baseline promotions are reviewed changes and never automatic. The Phase 3 split ledger, the VPP-artifact design decision, and every conflict resolution are recorded in `sonic-buildimage/devdocs/merge-p3.md`.

Based on current upstream `master`; no required Git merge order. This PR and the compatibility PR both touch `docker-sai-test-vpp/README.md`, `run_test.sh`, and `build_harness.sh`, so whichever merges second needs a rebase. Functionally, a green matrix depends on the compatibility PR (the harness cannot build or recycle without it) and on the maintainer-managed SAI pointer above — both are runtime prerequisites, not merge blockers. Recommended sequence for a green check: compatibility PR, then SAI pointer bump, then this PR.


.azure-pipelines/build-docker-sai-test-vpp-template.yml
```diff
@@ -0,0 +1,219 @@
...
+              echo "File=$(basename "$deb")"
+              dpkg-deb -f "$deb" Package Version Architecture
+              sha256sum "$deb"
+          done
+      } > "$provenance"
+
+      rm -rf "$context_dir"
+      context_harness="$context_dir/.azure-pipelines/docker-sai-test-vpp"
+      mkdir -p "$context_harness" "$context_dir/SAI/test"
+      for harness_file in \
```
yue-fred-gao, line 166: is port_config.ini missing?

azure-pipelines.yml
```diff
@@ -74,8 +77,7 @@ stages:
       debian_version: ${{ parameters.debian_version }}
 
 - stage: BuildArm
-  dependsOn: Build
```
yue-fred-gao, line 77: is there really no dependency on Build stage?

.azure-pipelines/build-docker-sai-test-vpp-template.yml
nicholasching: Very good catch, thank you Fred. I must have missed it when hunk-splitting my dev branch commits into the feature branches. Added back now.

azure-pipelines.yml
nicholasching: Yes, `BuildArm` does not consume any artifacts from `Build`. Only `BuildSwss` consumes artifacts published `Build` and they have been left to run sequentially. The CI checks continued to pass after the change; here is one run result from my development branch after the change: https://dev.azure.com/mssonic/build/_build/results?buildId=1185010&view=results

PR #2035, opened 2026-08-12
[submodule] Update SAI to latest upstream master and align generated-metadata initializers

nicholasching:
### Description of PR

Summary:
Moves the `SAI` submodule from `c67f1152` (tag `v1.18.1`) to upstream `opencomputeproject/SAI` master `6dd7381`, and makes the repository build and pass its tests against the newer headers. Three commits: the pointer move plus the four generated-metadata initializers it requires (these must be atomic, since neither half builds alone), serialization support for the new `SAI_ATTR_VALUE_TYPE_PORT_ILT_LANE_TRAINING_STATUS_LIST` value type, and one build flag for the SWIG-generated Python wrapper that the larger headers make necessary on armhf. Reviewers can start at `syncd/SwitchNotifications.h`, and please read the version-string note under "Any platform specific information?" before approving.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [x] Test improvement

The SAIVPP unit-test framework needs two `sai_test` changes that are already merged upstream — the opt-in `SIMULATE_SONIC` control-plane emulation ([opencomputeproject/SAI#2316](https://github.com/opencomputeproject/SAI/pull/2316)) and the simulated IPv6 control-traffic isolation ([opencomputeproject/SAI#2321](https://github.com/opencomputeproject/SAI/pull/2321)) — and neither is reachable at the currently pinned revision. Bumping the pointer also pulls in newer generated metadata, which requires three initializer updates in this repository; without them the build fails, so they belong in the same commit.

There is also a build-blocking reason beyond the test content. The VPP SAI CI PR adds a Trixie `saithriftv2` package build, and on the pinned `c67f1152` that build fails outright: `gensairpc.pl` throws in `assign_attr_types` while generating `sai.thrift`. The fix is upstream SAI `82a8d7e` ("Fix gensairpc.pl crash on Doxygen 1.9.8+", [opencomputeproject/SAI#2282](https://github.com/opencomputeproject/SAI/pull/2282)), reachable from master but not from the current pin. So this bump is a prerequisite for that PR's CI to pass at all, not only for its runtime matrix.

#### How did you do it?
- `SAI`: `c67f1152` (tag `v1.18.1`) → `6dd7381` (upstream master).
- `syncd/SwitchNotifications.h`: add the `on_next_hop_group_hw_protection_switchover` initializer. `sai_switch_notifications_t` is generated from the `SAI_SWITCH_ATTR_*_NOTIFY` attributes; the newer SAI adds `SAI_SWITCH_ATTR_NEXT_HOP_GROUP_HW_PROTECTION_SWITCHOVER_NOTIFY` (`inc/sainexthopgroup.h`, `inc/saiswitch.h`), taking the count from 21 to 22, so the slot-table aggregate initialization needs the new member. It is set to `nullptr` because syncd does not yet handle this notification.
- `unittest/meta/TestSaiSerialize.cpp`: initialize the `dei` and `vc` members added to `sai_qos_map_params_t` in `inc/saitypes.h`.
- `unittest/syncd/TestAttrVersionChecker.cpp`: initialize the `valueprecision` member added to `meta/saimetadatatypes.h`.
- `syncd/SwitchNotifications.h`, second slot: add `on_ocs_port_state_change`. The newer SAI also adds `SAI_SWITCH_ATTR_OCS_PORT_STATE_CHANGE_NOTIFY` in `experimental/saiswitchextensions.h` for the optical-circuit-switch API, and it lands last in the generated notification struct. Also `nullptr`, for the same reason.
- Second commit, `meta/SaiSerialize.cpp`, `meta/sai_serialize.h`, `meta/Meta.cpp`, and `unittest/meta/TestSaiSerialize.cpp`: implement `SAI_ATTR_VALUE_TYPE_PORT_ILT_LANE_TRAINING_STATUS_LIST`, added upstream by [opencomputeproject/SAI#2303](https://github.com/opencomputeproject/SAI/pull/2303). It mirrors the existing `PORT_LANE_LATCH_STATUS_LIST` and `PORT_SNR_LIST` handling: transfer, serialize, deserialize, the free path, the five validation sites in `Meta.cpp`, and round-trip unit tests. The list serializes as a JSON object keyed by lane, with each status rendered by enum name. The deserializer differs from its `PORT_LANE_LATCH_STATUS_LIST` sibling in one respect: it builds the entry array in a local pointer and publishes it to the caller only on success. Both a malformed lane key and an unrecognized status name throw (`sai_deserialize_enum` falls back to `sai_deserialize_number`), and freeing on that path keeps the malformed-input tests clean under ASAN.
- Third commit, `pyext/py3/Makefile.am`: add `-Wno-error=disabled-optimization` for the SWIG-generated `pysairedis` wrapper. The newer headers enlarge the generated `pysairedis_wrap.cpp` enough that GCC skips its const/copy-propagation and PRE passes on `max-gcse-memory`, which `-Werror` turns into a build failure on 32-bit armhf. The flag is scoped to the generated wrapper only.

This mirrors how previous bumps in this repository were structured: [#1742](https://github.com/sonic-net/sonic-sairedis/pull/1742) touched `SAI` plus `syncd/SwitchNotifications.h`, and [#1849](https://github.com/sonic-net/sonic-sairedis/pull/1849) paired `SAI` with `meta/*` and `unittest/meta/TestSaiSerialize.cpp`, including serialization support for the value types that bump introduced.

`git diff --check`, `tests/checkwhitespace.sh`, and `tests/swsslogentercheck.sh` pass locally; the compile and unit-test stages are verified by this PR's CI, which drove the last three additions. The first run failed to compile on the missing `on_ocs_port_state_change` initializer. The second run compiled and then failed `SaiSerialize.transfer_attributes` and `SaiSerialize.sai_serialize_attr_value`, both throwing "`SAI_ATTR_VALUE_TYPE_PORT_ILT_LANE_TRAINING_STATUS_LIST` is not implemented, FIXME" — those tests walk every value type, so the `default:` labels that kept the build green throw at run time. The third run passed `Build amd64`, `BuildAsan amd64`, `BuildArm arm64`, and `BuildSwss amd64`, and failed only `BuildArm armhf` on the generated wrapper.

Rather than keep chasing one error per run, I enumerated the whole breaking surface between the two pins directly: exactly one new `sai_attr_value_type_t` enumerator (the ILT list), exactly two new `*_NOTIFY` switch attributes (both initialized here), and no other exhaustive-switch site — the `syncd/FlexCounter.cpp` reference to the analogous latch-status type is feature-specific and has no ILT equivalent.

Note that no earlier pin avoids this work: the ILT value type arrives with SAI #2303 on 2026-07-25, which predates SAI #2321 on 2026-07-28, so every revision carrying the `sai_test` changes needed here also carries the new value type.

None specific to a platform, but the blast radius is repository-wide: every consumer of the SAI headers and generated metadata recompiles against the new revision. The two `nullptr` notification slots are inert — they change no runtime behavior for any vendor. The new serialization support is additive: it replaces a throwing `default:` path for a value type that nothing in this repository could previously handle.

**Version-string note, please read before approving.** The current pin `c67f1152` is tag `v1.18.1`, which lives on the `v1.18` release branch and is **not** an ancestor of SAI master. Moving to master therefore looks like it drops commits, so I checked all 8 that `v1.18.1` has and master does not: 7 are patch-identical to commits already on master (CEILR interfaces `#2256`, strict-flag enum meta `#2264`, PRBS pattern `#2262`, DASH ENI/port counters `#2251`, `SAI_HASH_ALGORITHM_END` `#2255`, RIF QoS map `#2208`, MACSec attribute enhancements `#2213`). The only one absent from master is `c67f115` itself, "Update SAI version to v1.18.1". No functional content is lost.

The consequence is that `SAI_API_VERSION` reports **1.18.0 instead of 1.18.1** after this change, because master has not taken that version-string bump, even though the tree is 32 commits newer in content. I could not find a better target: the `v1.18` release-branch head is the commit we are already pinned to, there is no tag newer than `v1.18.1`, and the metadata this work needs (`dei` / `vc`, `valueprecision`, `SAI_SWITCH_ATTR_NEXT_HOP_GROUP_HW_PROTECTION_SWITCHOVER_NOTIFY`) exists only on master. If maintainers would rather wait for upstream to cut a tag containing these changes, I am happy to park this PR — that would block the VPP SAI CI matrix until then.

### Documentation
No documentation changes. Context for why this bump is needed by the SAIVPP unit-test work is in `sonic-buildimage/devdocs/merge-p3.md`.

Self-contained and based on current upstream `master`. Both pointed-to SAI changes are already merged upstream, so this points at an upstream SHA rather than a fork commit. It unblocks the runtime matrix in the VPP SAI CI PR, but nothing else in the Phase 3 series has to merge before or after it.

### DIFF
--- a/meta/Meta.cpp
+++ b/meta/Meta.cpp
@@ -2036,6 +2036,7 @@ void Meta::meta_generic_validation_post_remove(
                 break;
 
             case SAI_ATTR_VALUE_TYPE_PORT_LANE_LATCH_STATUS_LIST:
+            case SAI_ATTR_VALUE_TYPE_PORT_ILT_LANE_TRAINING_STATUS_LIST:
             case SAI_ATTR_VALUE_TYPE_PORT_SNR_LIST:
             case SAI_ATTR_VALUE_TYPE_TAPS_LIST:
                 // no special action required
@@ -3827,6 +3828,10 @@ sai_status_t Meta::meta_generic_validation_create(
                 VALIDATION_LIST(md, value.portlanelatchstatuslist);
                 break;
 
+            case SAI_ATTR_VALUE_TYPE_PORT_ILT_LANE_TRAINING_STATUS_LIST:
+                VALIDATION_LIST(md, value.port_ilt_lane_training_status_list);
+                break;
+
             case SAI_ATTR_VALUE_TYPE_PORT_SNR_LIST:
                 VALIDATION_LIST(md, value.portsnrlist);
                 break;
@@ -4480,6 +4485,10 @@ sai_status_t Meta::meta_generic_validation_set(
             VALIDATION_LIST(md, value.portlanelatchstatuslist);
             break;
 
+        case SAI_ATTR_VALUE_TYPE_PORT_ILT_LANE_TRAINING_STATUS_LIST:
+            VALIDATION_LIST(md, value.port_ilt_lane_training_status_list);
+            break;
+
         case SAI_ATTR_VALUE_TYPE_PORT_SNR_LIST:
             VALIDATION_LIST(md, value.portsnrlist);
             break;
@@ -4890,6 +4899,10 @@ sai_status_t Meta::meta_generic_validation_get(
                 VALIDATION_LIST(md, value.portlanelatchstatuslist);
                 break;
 
+            case SAI_ATTR_VALUE_TYPE_PORT_ILT_LANE_TRAINING_STATUS_LIST:
+                VALIDATION_LIST(md, value.port_ilt_lane_training_status_list);
+                break;
+
             case SAI_ATTR_VALUE_TYPE_PORT_SNR_LIST:
                 VALIDATION_LIST(md, value.portsnrlist);
                 break;
@@ -5206,6 +5219,10 @@ void Meta::meta_generic_validation_post_get(
 
             case SAI_ATTR_VALUE_TYPE_PORT_LANE_LATCH_STATUS_LIST:
                 VALIDATION_LIST_GET(md, value.portlanelatchstatuslist);
+                break;
+
+            case SAI_ATTR_VALUE_TYPE_PORT_ILT_LANE_TRAINING_STATUS_LIST:
+                VALIDATION_LIST_GET(md, value.port_ilt_lane_training_status_list);
                 break;
 
             case SAI_ATTR_VALUE_TYPE_PORT_SNR_LIST:
--- a/meta/SaiSerialize.cpp
+++ b/meta/SaiSerialize.cpp
@@ -350,6 +350,10 @@ sai_status_t transfer_attribute(
             RETURN_ON_ERROR(transfer_list(src_attr.value.portlanelatchstatuslist, dst_attr.value.portlanelatchstatuslist, countOnly));
             break;
 
+        case SAI_ATTR_VALUE_TYPE_PORT_ILT_LANE_TRAINING_STATUS_LIST:
+            RETURN_ON_ERROR(transfer_list(src_attr.value.port_ilt_lane_training_status_list, dst_attr.value.port_ilt_lane_training_status_list, countOnly));
+            break;
+
         case SAI_ATTR_VALUE_TYPE_LATCH_STATUS:
             transfer_primitive(src_attr.value.latchstatus, dst_attr.value.latchstatus);
             break;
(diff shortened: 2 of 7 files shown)


yue-fred-gao (approved):
looks good to me

nicholasching:

Github Run: https://github.com/sonic-net/sonic-buildimage/runs/94978778223
Azure Build: https://dev.azure.com/mssonic/build/_build/results?buildId=1193281&view=logs&jobId=3ab7146d-ea64-5d52-0097-33a9aae59fb6&j=314bc92a-d36a-5eac-5c45-43e8f1288876

lolyu:
Hi @tjchadaga, could you please help review?

tjchadaga (approved):
LGTM. The SAI version falls back to v1.18.0 from v1.18.1, but this should not have a functional impact

PR #2036, opened 2026-08-13
[vpp] Clear stale VPP ACL when a table's last rule is removed

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

On a dual-ToR VPP setup, the mux ingress drop ACL is bound to the mux member ports on the VPP l2-input arc. When a port transitions standby → active, orchagent deletes that drop entry, so the ACL table's last regular rule is removed.

In `AclTblConfig()` an empty rule set produces a `NULL` converted `vpp_acl_t`, and the apply step was guarded by if (acl != NULL). As a result the reprogram was skipped entirely and the previously programmed VPP ACL (deny 0.0.0.0/0 -> 0.0.0.0/0) was left bound to the interface. That stale deny then keeps silently dropping for-us / ip2me punts (mux/BFD/ICMP heartbeat) even though SONiC believes the rule is gone — so the now-active ToR's mux/heartbeat traffic is black-holed and the mux state flaps/goes unhealthy.

##### Work item tracking
- Microsoft ADO **(number only)**: 39251238
- Tracking issue: sonic-net/sonic-buildimage#28996

#### How did you do it?

The fix widens the guard in AclTblConfig() that if the table is NULL but programmed before, still call `acl_add_replace` to reprogram the VPP dataplane.

* Reproduced on the VPP dual-ToR KVM testbed: flap a mux port standby → active, then confirm a stale deny 0/0 -> 0/0 ACL remains bound to the mux member interfaces (docker exec syncd vppctl show acl-plugin interface) and for-us/heartbeat punts are dropped, driving the mux unhealthy.
* After the fix: on the same transition the table's VPP swindex is reprogrammed to the placeholder ACL (permit dst 0.0.0.0/32), the stale deny is gone, punt traffic to the CPU is restored, and the mux returns/stays HEALTHY on both ToRs.

### DIFF
--- a/vslib/vpp/SwitchVppAcl.cpp
+++ b/vslib/vpp/SwitchVppAcl.cpp
@@ -1246,8 +1246,15 @@ sai_status_t SwitchVpp::AclTblConfig(
         SWSS_LOG_INFO("Allocated tunterm ACL with %u rules", tunterm_acl->count);
     }
 
-    // Apply the ACL configurations
-    if (acl != NULL) {
+    // Apply the ACL configurations. acl is NULL when the table has no regular
+    // rules left - e.g. the mux drop entry is deleted when a port transitions
+    // standby -> active. In that case acl_add_replace() replaces the previously
+    // programmed VPP ACL with an empty placeholder (see its acl == NULL branch),
+    // so a stale rule such as deny 0/0->0/0 is not left bound to the mux ports
+    // and silently dropping for-us/punt traffic. Only route a NULL acl through
+    // acl_add_replace() when the table was already programmed in VPP, otherwise
+    // there is nothing to clear and vpp_acl_add_replace() must not see a NULL.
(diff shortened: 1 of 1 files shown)


yue-fred-gao (approved):

yue-fred-gao:
Please trigger sonic-mgmt sanity to make sure no breakage

lolyu:
> Please trigger sonic-mgmt sanity to make sure no breakage

Here is the pipeline, the vpp t1-lag is passing:

https://github.com/sonic-net/sonic-buildimage/pull/29031

It has the following three PRs:

https://github.com/sonic-net/sonic-sairedis/pull/2029
https://github.com/sonic-net/sonic-sairedis/pull/2036
https://github.com/sonic-net/sonic-sairedis/pull/2037

PR #2040, opened 2026-08-17
[2039] FDB Notification becomes stale in notification enqueue buffer due to continuous mac movement between local and vtep syncd: fix lost condition-variable wakeup in NotificationProcessor 

skannan-CLS:
### Description of PR
[2039] FDB Notification becomes stale in notification enqueue buffer due to continuous mac movement between local and vtep 

NotificationProcessor::ntf_process_function() waited on a std::condition_variable with a predicate-less m_cv.wait() using a function-local mutex, while signal() called notify_all() without holding that mutex. A condition variable does not remember a notify sent when no thread is waiting, so a notification enqueued in the window after the consumer drains the queue but before it re-parks in wait() is lost. The tail of a notification burst (e.g. an FDB event storm) then stays stranded in NotificationQueue and is not forwarded to orchagent. When traffic stops there is no next enqueue, so it is never processed; it only self-heals when a later FDB event re-signals the thread (which then drains both the stale and the new notifications).

Impact: dropped/stuck SAI notifications (FDB and others) between syncd and orchagent -> stale state; under storm conditions notification processing can stall.

Involved APIs: producer NotificationHandler::enqueueNotification() ->
NotificationQueue::enqueue() -> NotificationProcessor::signal() -> notify_all(); consumer NotificationProcessor::ntf_process_function() -> m_cv.wait() -> NotificationQueue::tryDequeue() -> processNotification().

Summary:
Fixes # (issue)

Fix
---
- Wait on a predicate (queue non-empty || !m_runThread) so an already-enqueued item (or shutdown) is handled without sleeping.
- Add a member mutex m_mtx paired with m_cv, held only around the wait (never during processNotification(), to avoid the syncd mutex lock-ordering hazard).
- Take m_mtx in signal() and in stopNotificationsProcessingThread() around notify_all(), so a wakeup cannot slip in between the consumer evaluating its predicate and blocking in wait().

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

  During internal testing, a MAC-move loop caused an FDB entry to flap continuously between the VTEP (remote) and the local port. Under that burst, syncd stopped propagating FDB notifications to orchagent, and both devices ended up stuck
  showing the MAC as local — the final "move" notification was never delivered, so APPL_DB/STATE_DB kept the stale (local) state.

  Root cause: NotificationProcessor used a predicate-less std::condition_variable::wait() with an unsynchronized signal(). A notification enqueued in the window between the consumer draining the queue and re-parking in wait() is lost (a
  CV doesn't remember a notify sent when no thread is waiting), so the tail of the burst is stranded in NotificationQueue and never forwarded. With traffic settled, there's no further enqueue to re-signal the thread, so it never recovers
  on its own.

##### Work item tracking
2039

#### How did you do it?
This test was validated using celestica DS2000 device.

  Unit tests — added 4 gtests in unittest/syncd/TestNotificationProcessor.cpp targeting the wakeup path (a counting synchronizer is the observation point, so no SAI/redis needed):
  - LostWakeup_EnqueueBeforeStart_NoSignal — item enqueued before the thread parks, with no signal(), must still be processed (the core regression).
  - LostWakeup_ConcurrentEnqueueSignal_NoneLost — 500 concurrent enqueue()+signal(), none lost.
  - BurstThenSingleSignal_AllDrained — 1000 items + a single signal(), all drained.
  - ShutdownWakesWaitingThread — stop() wakes a parked thread (no hang).

  Built and run (./configure --with-sai=vs --enable-code-coverage && make):
  [==========] 4 tests from NotificationProcessor
  [  PASSED  ] 4 tests.

  Negative control — reverting to the old predicate-less m_cv.wait(ulock) makes LostWakeup_EnqueueBeforeStart_NoSignal hang and FAIL (5 s timeout), confirming the test actually catches the bug; restoring the fix makes it pass again.

  Coverage — changed functions are 100% line-covered (ntf_process_function + wait-predicate lambda, signal, start/stopNotificationsProcessingThread, processNotification).

  Integration image — built the full docker-sonic-vs image from master with the fix for DVS/FDB validation.

None generic

### DIFF
--- a/syncd/NotificationProcessor.cpp
+++ b/syncd/NotificationProcessor.cpp
@@ -1138,13 +1138,24 @@ void NotificationProcessor::ntf_process_function()
 {
     SWSS_LOG_ENTER();
 
-    std::mutex ntf_mutex;
-
-    std::unique_lock<std::mutex> ulock(ntf_mutex);
-
     while (m_runThread)
     {
-        m_cv.wait(ulock);
+        {
+            std::unique_lock<std::mutex> ulock(m_mtx);
+
+            // Wait on a predicate so a signal() that races ahead of this wait
+            // (queue already made non-empty by enqueue) is not lost. With the
+            // previous predicate-less m_cv.wait(), the tail of a notification
+            // burst could sit stuck in the queue until the next enqueue happened
+            // to re-signal the thread. m_mtx is held ONLY around the wait, never
+            // during processNotification(), to stay clear of the syncd mutex
+            // taken while processing.
+
+            m_cv.wait(ulock, [this]()
+            {
+                return !m_runThread || m_notificationQueue->getQueueSize() > 0;
+            });
+        }
 
         // this is notifications processing thread context, which is different
         // from SAI notifications context, we can safe use syncd mutex here,
@@ -1173,7 +1184,11 @@ void NotificationProcessor::stopNotificationsProcessingThread()
 {
     SWSS_LOG_ENTER();
 
-    m_runThread = false;
+    {
+        std::lock_guard<std::mutex> lock(m_mtx);
+
+        m_runThread = false;
+    }
 
     m_cv.notify_all();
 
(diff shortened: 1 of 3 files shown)


skannan-CLS:
Hi @kcudnik could you please review

PR #2047, opened 2026-08-20
fix(vpp): reconcile tap MAC on active IP path

auspham:
### Description of PR

Summary:
Reconcile the Linux LCP tap MAC with the switch source MAC on the active physical-port IP programming path.

Related: sonic-net/sonic-mgmt#27062

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

When the Linux team driver removes a port from a PortChannel, it publishes the port-removed event before restoring the port's saved MAC. That event can reach `teamsyncd`, orchagent, and the #2020 repair before the kernel performs its final MAC restore. If that happens, the kernel write wins the race and leaves the Linux tap using the per-port MAC while VPP still uses the switch MAC.

This change adds a final reconciliation immediately before the physical port IP is programmed, so the port cannot start carrying L3 traffic with a mismatched tap MAC.

##### Work item tracking
- Microsoft ADO **(number only)**: Not applicable

#### How did you do it?

- Changed `restorePortTapMac()` to return a SAI status.
- Kept LAG-member removal best effort because the VPP detach has already occurred.
- Moved the final reconciliation to the active `vpp_add_del_intf_ip_addr_norif()` path.
- Propagated its status through `IpRouteAddRemove()`.
- Updated the prefix map only after VPP IP programming succeeds.

- Built and loaded `libsaivs` against the same VPP ABI as the reproducing image.
- Kept the port routed with no LAG transition, forced its tap to a different MAC, then added the interface IP.
- Confirmed the active-path log fired, the tap returned to the switch MAC, and VPP received the interface IP.
- Ran `git diff --check`.
- Ran `tests/swsslogentercheck.sh`.

This change is limited to the VPP implementation and physical port router interfaces.

### Documentation

Not applicable. This is a VPP behavior fix with no user-facing interface change.

### DIFF
--- a/vslib/vpp/SwitchVppRif.cpp
+++ b/vslib/vpp/SwitchVppRif.cpp
@@ -1086,6 +1086,14 @@ sai_status_t SwitchVpp::vpp_add_del_intf_ip_addr (
         hw_ifname = hwifname;
     }
 
+    if (is_add &&
+        rif_type == SAI_ROUTER_INTERFACE_TYPE_PORT &&
+        ot == SAI_OBJECT_TYPE_PORT)
+    {
+        // Reassert the tap MAC immediately before the port starts carrying L3 traffic.
+        CHECK_STATUS(restorePortTapMac(obj_id));
+    }
+
     int ret = interface_ip_address_add_del(hw_ifname, &vpp_ip_prefix, is_add);
 
     if (ret == 0)
--- a/vslib/vpp/SwitchVppRif.cpp
+++ b/vslib/vpp/SwitchVppRif.cpp
@@ -1086,14 +1086,6 @@ sai_status_t SwitchVpp::vpp_add_del_intf_ip_addr (
         hw_ifname = hwifname;
     }
 
-    if (is_add &&
-        rif_type == SAI_ROUTER_INTERFACE_TYPE_PORT &&
-        ot == SAI_OBJECT_TYPE_PORT)
-    {
-        // Reassert the tap MAC immediately before the port starts carrying L3 traffic.
-        CHECK_STATUS(restorePortTapMac(obj_id));
-    }
-
     int ret = interface_ip_address_add_del(hw_ifname, &vpp_ip_prefix, is_add);
 
     if (ret == 0)
@@ -1195,6 +1187,7 @@ sai_status_t SwitchVpp::vpp_add_del_intf_ip_addr_norif (
 
     std::string full_if_name;
     std::string ip_prefix_str;
+    std::string intf_data;
 
     if (is_add)
     {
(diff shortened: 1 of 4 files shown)


yue-fred-gao:
"PR https://github.com/sonic-net/sonic-sairedis/pull/2020 repairs the tap during LAG-member removal, but a later teamd update can overwrite that one-time repair". Does teamd set the member interface removed from LAG back to the switch Mac? What is the sequence of events?

vslib/vpp/SwitchVppRif.cpp
```diff
@@ -1086,6 +1086,14 @@ sai_status_t SwitchVpp::vpp_add_del_intf_ip_addr (
         hw_ifname = hwifname;
     }
 
+    if (is_add &&
```
aaronber0614: The new restorePortTapMac()  call is added to  vpp_add_del_intf_ip_addr() , but that function has no active caller; its only route call at  SwitchVppRoute.cpp:224  is commented out. The active physical-port path calls  vpp_add_del_intf_ip_addr_norif()  at line 232, which programs the address without restoring the tap MAC. Its return value is also discarded, so failures cannot propagate to the route operation. Please place the reconciliation on the active physical-port path and propagate its status through  IpRouteAddRemove() .

Review co-authored with CoPilot

aaronber0614 (changes requested):

vslib/vpp/SwitchVppRif.cpp
auspham: Thanks, you were right. I moved the call to `vpp_add_del_intf_ip_addr_norif()`, which is the active physical-port path, and now return its status from `IpRouteAddRemove()`. I also update the prefix map only after VPP programming succeeds. I rebuilt the matching `libsaivs`, forced Ethernet0 to a wrong MAC while it was already routed, and added the IP again. The new path ran, restored the switch MAC, and programmed the address in VPP.

auspham:
@yue-fred-gao Correction to my earlier reply: the important ordering is inside the Linux team driver. It publishes the port-removed event first, then calls `team_port_set_orig_dev_addr()` to restore the saved member MAC. `teamsyncd` can process that event and trigger the #2020 repair before the kernel performs the final MAC restore, so the kernel write can overwrite #2020. teamd starts the removal, but the later writer is the kernel team driver. I updated the PR description with this sequence.

aaronber0614 (approved):

yue-fred-gao:
as always, please trigger vpp sanity through sonic-buildimage PR

yue-fred-gao (approved):

PR #2050, opened 2026-08-21
vpp: support proxy arp

yue-fred-gao:
### Description of PR

Summary:
Support proxy-arp in vpp

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

SONiC enables proxy ARP on a VLAN interface with "config interface proxy_arp <Vlan> enabled". IntfsOrch::setIntfVlanFloodType translates that into two SAI attributes on the VLAN object:
SAI_VLAN_ATTR_BROADCAST_FLOOD_CONTROL_TYPE
SAI_VLAN_ATTR_UNKNOWN_MULTICAST_FLOOD_CONTROL_TYPE
set to [SAI_VLAN_FLOOD_CONTROL_TYPE_NONE when enabled and back to ..._ALL when disabled. The VPP platform silently ignored both, so proxy ARP had no dataplane effect. When proxy-arp is enabled, the expectation is ARP request and IPv6 ND multicast packets are not flooded in the bridge so the host in the same bridge won't reply with its own MAC. SONiC will answer the request with its MAC. All that is required on dataplane is stopping flood such packets.
##### Work item tracking

#### How did you do it?
**Implemented with `l2-input-classify` punt tables**
classify hit is terminal —`l2_input_classify.c` clears the feature bit and dispatches straight to the session's next node to lcp host interface of the ingress interface.
So punting a traffic class *is* flood suppression for exactly that class.
Members of an affected VLAN are re-bound to punting variants of the classify
tables:

| Attribute set to `NONE` | New table | Match |
| --- | --- | --- |
| `BROADCAST_FLOOD_CONTROL_TYPE` | `*_arp_table` | DA `ff:ff:ff:ff:ff:ff`, ethertype `0x0806` |
| `UNKNOWN_MULTICAST_FLOOD_CONTROL_TYPE` | `*_ip6mc_table` | DA `33:33:*` |

The two are independent knobs. Untagged members get one classify slot each
(ARP in the OTHER slot, chaining to the existing LLDP table; IPv6 multicast in
the IP6 slot), so they never interact. Tagged members carry outer ethertype
`0x8100` and so have only the OTHER slot available; since a table's
`next_table_index` is fixed at creation, the tagged IPv6 head exists in two
variants to make all four on/off combinations reachable:

```
both      -> tag_ip6mc_table      -> tag_arp_table -> tag_dhcp_table
IPv6 only -> tag_ip6mc_only_table ->                  tag_dhcp_table
ARP only  -> tag_arp_table        ->                  tag_dhcp_table
neither   -> tag_dhcp_table
```

Plumbing:

- New `SwitchVpp::vpp_set_vlan_attribute()`, hooked into `SwitchVpp::set()`
  *before* `set_internal()`
- VLAN members are enumerated through `SaiObjectDB` — a `VLAN_MEMBER -> VLAN`
  entry was added to `sai_child_relation_defs` — instead of keeping a shadow
  member map in the switch object.
- Members created *after* the attribute is set pick the configuration up at
  create time via `vlan_flood_punt_enabled()`, which reads the stored VLAN
  object. No cached copy of either attribute is kept anywhere.

**Refactor**
added `SwitchVpp::bridge_port_to_port_id()` and used it to replace five hand-rolled copies of

```cpp
m_objectHash.at(SAI_OBJECT_TYPE_BRIDGE_PORT).at(oid)[meta->attridname]->getAttr()->value.oid
```
which is unsafe twice over — `.at()` throws `std::out_of_range` if the bridge
port is not in the store, and `operator[]` default-inserts a null attribute that
is then dereferenced. Both are now plain error returns.

**Lower log message when an attribute is not found in an object**
In SwitchStateBase::get(), if an attribute is not found in the object, lower the log message to DEBUG. It could be a normal case. Let the caller decide if this is normal.

arp/test_arp_extended.py::test_proxy_arp passed for both v4 and v6

vpp

### DIFF
--- a/vslib/vpp/SwitchVppFdb.cpp
+++ b/vslib/vpp/SwitchVppFdb.cpp
@@ -64,9 +64,24 @@ using namespace saivs;
  * the BD via the same linux-cp-punt-xc shortcut.  Only untagged LLDP
  * (which arrives on a parent that IS in the BD) needs the classifier.
  *
- * ARP is handled by the sonic_ext VPP plugin via the arp arc ->
- * sonic-ext-aggr-tap-redirect on the BVI tap.  See
- * platform/vpp/vppbld/plugins/sonic_ext/.
+ * ARP and IPv6 ND are normally handled by the sonic_ext VPP plugin via
+ * the arp arc -> sonic-ext-aggr-tap-redirect on the BVI tap.  See
+ * platform/vpp/vppbld/plugins/sonic_ext/.  That path depends on the
+ * BD flood copy reaching the BVI, so the BD flood bit is never cleared,
+ * not even when orchagent asks for FLOOD_CONTROL_TYPE_NONE: VPP shares
+ * one L2INPUT_FEAT_FLOOD bit between broadcast and multicast, so it
+ * cannot express either SAI attribute without also applying the other.
+ * Instead the members of such a VLAN are switched over to the punting
+ * table variants below.  A classify hit is terminal, so each punt
+ * suppresses the flood for exactly its own traffic class:
+ *
+ *   SAI_VLAN_ATTR_BROADCAST_FLOOD_CONTROL_TYPE        == NONE
+ *       -> *_arp_table   (dst ff:ff:ff:ff:ff:ff, ethertype 0x0806)
+ *   SAI_VLAN_ATTR_UNKNOWN_MULTICAST_FLOOD_CONTROL_TYPE == NONE
+ *       -> *_ip6mc_table (dst 33:33:*)
+ *
+ * The two are independent knobs.  See
+ * SwitchVpp::vpp_set_vlan_attribute().
  *
  * Slot selection in l2-input-classify (vnet/l2/l2_input_classify.c):
  * VPP picks the per-interface table by *outer* ethertype at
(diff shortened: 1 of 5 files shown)


vslib/vpp/SwitchVppFdb.cpp
```diff
@@ -113,12 +144,24 @@ static uint32_t s_trap_fixup_next_index = ~0; /* sonic-ext-l2-trap-fixup (tagged
...
- * attached directly to the tagged member's OTHER slot. */
-static uint32_t s_tag_dhcp_table = ~0;        /* DHCPv4 broadcast over .1Q */
+ * attached directly to the tagged member's OTHER slot.
+ *
+ * The ARP and IPv6-multicast punts are independent knobs, but a tagged
+ * member has only the one OTHER slot to bind, so they have to share a
+ * miss-chain.  A classify table's next_table_index is fixed at creation,
+ * so the IPv6-multicast head exists in two variants -- one that chains
+ * through the ARP table and one that skips it -- to cover all four
+ * on/off combinations. */
```
dypet, line 160: Are there tests that cover all four combinations?

dypet (approved):

vslib/vpp/SwitchVppFdb.cpp
yue-fred-gao: yes. They are used in various tests, including test_dhcp_relay, test_lldp, test_tagged_arp, test_arp_extended

yue-fred-gao:
vpp sanity passed in PR https://github.com/sonic-net/sonic-buildimage/pull/27039

PR #2052, opened 2026-08-24
[syncd]: Use bulk redis operations in warm boot APPLY_VIEW

### LINKED ISSUE #29101: Bug: [warm-reboot] APPLY_VIEW takes ~94s at ~418K routes due to Redis full-table rewrite
### Is it platform specific

generic

### Importance or Severity

Critical

### Description of the bug

During warm-reboot with full IPv6 route scale (~418K short /64 prefixes,
~421.5K ASIC objects), syncd `APPLY_VIEW` takes ~94 s end-to-end.

Warmboot still completes with the orchagent Mellanox response-timeout
workaround (60 s → 480 s around APPLY_VIEW). ASIC programming and route
comparison are not the bottleneck:

```text
getAsicView ASIC_STATE ............... ~8.6 s   (421,551 objects)
getAsicView TEMP_ASIC_STATE .......... ~8.9 s   (421,547 objects)
ComparisonLogic ...................... ~5.7 s
executeOperationsOnAsic .............. ~0.1 s   (84 ops)
updateRedisDatabase .................. ~59.6 s  <--- dominant
------------------------------------------------
applyView total ...................... ~94.3 s
```

Almost all of the cost is Redis I/O: full ASIC_STATE wipe + rewrite, plus
loading both views. Route matching is already scale-safe (dedicated O(N)
key path). This is unlike the ACL / NH / NHGM warmboot failures, which are
O(n²) comparison bugs.

At this runtime:

* syncd's 30 s TimerWatchdog emits ERR for `notify:APPLY_VIEW` (non-fatal)
* without the 480 s orchagent workaround, APPLY_VIEW would fail at the
  default 60 s response timeout (parent functional failures)
* warmboot finalizer wait is 5 min; this run used ~3 min from finalizer
  start to orchagent `reconciled`, leaving limited margin if APPLY grows

The issue is triggered by route / ASIC-object scale; small route counts do
not reproduce the multi-minute APPLY_VIEW path.

### Investigation and suspected root cause

During warmboot, syncd reconciles:

* **current view** (`ASIC_STATE`) — objects present after warm restart
* **temporary view** (`TEMP_ASIC_STATE`) — objects expected by orchagent

Measured scale in the failing/slow dump:

* 418,002 unique IPv6 /64 routes (from warmboot Redis snapshot)
* ASIC_STATE / TEMP_ASIC_STATE object counts: 421,551 / 421,547
* Comparison produced only **84** ASIC operations; hardware apply ~0.1 s

#### Why `updateRedisDatabase` dominates

`Syncd::updateRedisDatabase()` (`sonic-sairedis/syncd/Syncd.cpp`) currently:

1. `removeAsicStateTable()` — `KEYS` + one `DEL` per key (~421K RTTs)
2. `removeTempAsicStateTable()` — same for TEMP (~421K RTTs)
3. per-object `createAsicObject()` — one `HSET` per attribute (no pipeline)
4. per-entry VID↔RID map `HSET`s

The same tree already has pipelined bulk helpers used by the normal bulk
create/remove path, but APPLY_VIEW rewrite does not use them:

* `RedisClient::createAsicObjects()` → pipelined `HMSET`
* `RedisClient::removeAsicObjects()` → pipelined chunked `DEL`

There is also an in-tree TODO on `updateRedisDatabase`:

```text
// TODO: We can make LUA script for this which will be much faster.
```

`getAsicView` is already Lua-backed (`table_dump.lua`) and still costs
~17.5 s combined at this scale — secondary target after the rewrite.

#### Why this is not the ACL/NH O(n²) class

Route / neighbor / FDB already use dedicated O(N) matchers
(`m_soRoutes` / `m_soNeighbors` / `m_soFdbs`). In this run ComparisonLogic
finished in ~5.7 s. The residual scale problem is Redis dump + full-table
rewrite of ~421.5K objects, not candidate search.

#### Optimization breakdown (analytical; no PoC yet)

| Approach | Target | Est. impact on rewrite / APPLY |
|---|---|---|
| A. Use existing bulk APIs in `updateRedisDatabase` (`createAsicObjects` / `removeAsicObjects`) | rewrite ~59.6 s | ~5–10× on rewrite → ~6–12 s; APPLY still ~35–45 s |
| B. Lua server-side wipe+rewrite (existing TODO) | rewrite | ~10–20× → rewrite ~3–6 s; APPLY still ~35–40 s |
| C. TEMP→ASIC rename/swap or incremental apply of the tiny delta (only 84 ASIC ops here) | rewrite | ~15–30× → rewrite ~2–4 s |
| D. Faster / avoided double `getAsicView` (~17.5 s already Lua) | view load | required together with A/B/C to get APPLY comfortably under ~30 s |

Floor without touching view load + comparison: ~23–25 s. So rewrite-only
fixes remove most of the 94 s, but clearing the 30 s watchdog / leaving
healthy finalizer margin needs rewrite **and** view-load improvement.

#### Proposed fix

1. **Primary:** stop the chatty APPLY_VIEW Redis rewrite
   * minimum: switch `updateRedisDatabase` to existing pipelined bulk APIs
   * preferred: Lua bulk rewrite and/or TEMP→ASIC swap / incremental update
     using ComparisonLogic’s already-computed delta
2. **Secondary:** reduce `getAsicView` cost at ~400K objects (streaming /
   avoid redundant full dumps where safe)
3. **Goal:** APPLY_VIEW at ~418K routes in seconds to low tens of seconds —
   well under orchagent default 60 s, syncd 30 s warn span, and warmboot
   finalizer 5 min

Do not treat watchdog severity changes as the fix; they only hide the
symptom.

### Steps to Reproduce

1. DUT with warm-reboot enabled
2. Bring up BGP toward a traffic generator; advertise ~418K IPv6 /64 routes
3. Forward traffic through the routes; wait until ASIC install is complete
   and there is no loss
4. Warm-reboot
5. Observe syncd APPLY_VIEW timers (`getAsicView`, `comparison logic`,
   `redis update`, `applyView: apply took`) and orchagent reconcile time

### Actual Behavior:
At ~418K IPv6 routes / ~421.5K ASIC objects, APPLY_VIEW takes ~94 s. Redis
full-table rewrite alone is ~60 s because APPLY_VIEW uses per-key / per-attr
Redis calls instead of the existing bulk/pipeline path. Warmboot succeeds
only because orchagent’s response timeout is temporarily raised to 480 s;
syncd also logs 30 s TimerWatchdog ERR for the long notify. Finalizer still
passes, but with limited margin.

### Expected Behavior:
APPLY_VIEW at this route scale finishes in seconds to low tens of seconds.
Redis update uses bulk/Lua/incremental mechanisms; view load is not a
multi-ten-second tax; orchagent reconciles within default timeouts without
relying on a 480 s APPLY_VIEW workaround.

### Relevant log output

```shell
# orchagent starts APPLY_VIEW (timeout raised for Mellanox)
NOTICE swss#orchagent: :- setWarmStartState: orchagent warm start state changed to restored
NOTICE swss#orchagent: :- syncd_apply_view: Notify syncd APPLY_VIEW
NOTICE swss#orchagent: :- syncd_apply_view: Set SAI REDIS response timeout to 480000 for Mellanox platform, to extend the timeout for apply view
NOTICE swss#orchagent: :- notifySyncd: sending syncd: APPLY_VIEW

# ~421.5K-object views
NOTICE syncd#SDK: :- getAsicView: oid:0x21000000000000: objects count: 421551
NOTICE syncd#SDK: :- getAsicView: get asic view from ASIC_STATE took 8.611079 sec
NOTICE syncd#SDK: :- getAsicView: oid:0x21000000000000: objects count: 421547
NOTICE syncd#SDK: :- getAsicView: get asic view from TEMP_ASIC_STATE took 8.903004 sec

# comparison / ASIC apply are cheap; Redis rewrite dominates
NOTICE syncd#SDK: :- applyViewTransition: comparison logic took 5.726386 sec
NOTICE syncd#SDK: :- compareViews: ASIC operations to execute: 84
NOTICE syncd#SDK: :- executeOperationsOnAsic: asic apply took 0.097552 sec
NOTICE syncd#SDK: :- updateRedisDatabase: redis update took 59.643072 sec
NOTICE syncd#SDK: :- applyView: apply took 94.252157 sec

# secondary symptom: 30s TimerWatchdog (non-fatal; APPLY still succeeds)
ERR    syncd#SDK: :- threadFunction: time span WD exceeded 30255 ms for notify:APPLY_VIEW
ERR    syncd#SDK: :- setEndTime: event 'notify:APPLY_VIEW' took 94379 ms to execute

# warmboot still succeeds with the 480s workaround
NOTICE swss#orchagent: :- sai_redis_notify_syncd: switched ASIC to APPLY VIEW
NOTICE swss#orchagent: :- setWarmStartState: orchagent warm start state changed to reconciled
NOTICE root: WARMBOOT_FINALIZER : Finalizing warmboot...
```

* N/A

### Attach files (if any)

* N/A

PRIEYAN:
### Description of PR

Summary:
Fixes sonic-net/sonic-buildimage#29101

During warm boot at full IPv6 route scale (~418K routes, ~421.5K ASIC objects), syncd `APPLY_VIEW` takes ~94s end to end, and `updateRedisDatabase()` alone accounts for ~59.6s of that. ASIC programming is not the bottleneck — the comparison logic produced only 84 ASIC operations and applying them took ~0.1s. The cost is redis I/O: the current view rewrite talks to redis one key, and even one attribute, at a time.

This PR switches that rewrite onto the pipelined bulk helpers that already exist in this tree.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [x] Test improvement

`Syncd::updateRedisDatabase()` is on the warm boot critical path, and at route scale it is dominated entirely by redis round trips:

| call | current behaviour | round trips at ~421.5K objects |
| --- | --- | --- |
| `removeAsicStateTable()` | `KEYS` + one `DEL` per key | ~421K |
| `removeTempAsicStateTable()` | `KEYS` + one `DEL` per key | ~421K |
| `createAsicObject()` | one `HSET` per **attribute**, unpipelined | ~600K+ |
| `setVidAndRidMap()` | two `HSET` per entry, unpipelined | 2 per entry |

At this runtime syncd's 30s `TimerWatchdog` logs an ERR for `notify:APPLY_VIEW`, and warm boot only completes because orchagent temporarily raises its response timeout to 480s for the Mellanox platform. Without that workaround `APPLY_VIEW` would fail at the default 60s response timeout.

The tree already has pipelined bulk helpers, used by the normal bulk create/remove path — they were simply never wired into the `APPLY_VIEW` rewrite. There is also a long-standing in-tree `TODO` on this function.

#### How did you do it?

- **Bulk table wipe.** `removeAsicStateTable()` / `removeTempAsicStateTable()` now remove all keys with a single `del()` on the vector of keys, which issues chunked pipelined `DEL` commands (`KEY_DEL_CHUNK_SIZE` = 128 keys per command in swss-common) instead of one round trip per key.

  Note the keys returned by `keys()` already carry the table prefix, so they are handed to `m_dbAsic->del()` directly rather than through `RedisClient::removeAsicObjects()`, which prepends the prefix itself and would double it.

- **Bulk view rewrite.** `updateRedisDatabase()` now accumulates objects and writes them with `createAsicObjects()` (pipelined `HSET` per object, all fields in one command) instead of a `createAsicObject()` per object.

  Because the bulk helper materialises an intermediate hash of everything it is given, objects are written in chunks of `ASIC_STATE_BULK_WRITE_CHUNK_SIZE` (10,000). Writing all ~421K objects in a single call would materialise the map twice — once by the caller and once inside the helper, which prepends the table prefix into a fresh map — for a peak of several hundred MB during warm boot. Chunking keeps that bounded at a few MB while still keeping the pipeline saturated.

- **Pipelined VID/RID map.** `setVidAndRidMap()` now uses `swss::RedisPipeline`, exactly as the neighbouring `insertVidsAndRids()` already does. The default pipeline size is used deliberately so replies are flushed periodically and memory stays bounded. This also benefits the cold boot path, since `HardReiniter` calls the same method.

**Behaviour is unchanged.** The same keys are written with the same table prefix, and objects with no attributes are still stored as `NULL:NULL` — `createAsicObjects()` handles that case identically to `createAsicObject()`. `ZmqRedisClient` overrides both the single and bulk variants and its two implementations are equivalent, so the ZMQ path is unaffected; `DisabledRedisClient` stubs both as no-ops.

Estimated effect on the rewrite: round trips drop from roughly 1.4M to a few thousand pipeline flushes, which should take `updateRedisDatabase()` from ~59.6s to low single-digit seconds, and `APPLY_VIEW` from ~94s to roughly ~35s.

That clears the orchagent 60s default response timeout and removes the need for the 480s workaround, but it does **not** by itself get under syncd's 30s watchdog warn span — the remaining time is the ~17.5s combined `getAsicView` of both views plus ~5.7s of comparison logic. That view-load cost is the secondary target called out in the issue and is intentionally left out of this PR to keep the change reviewable.

Added `syncd/tests/TestRedisClient.cpp`, covering the paths this PR changes:

- `removeAsicStateTable()` removes every ASIC_STATE key, using more objects than a single `DEL` chunk holds so the chunked bulk removal is actually exercised
- `removeAsicStateTable()` on an already empty table is a no-op and does not throw
- `removeTempAsicStateTable()` removes only TEMP keys and leaves the current view intact
- `createAsicObjects()` writes the same keys as `createAsicObject()`, including the empty-attribute `NULL:NULL` case
- `setVidAndRidMap()` writes both directions and an empty map clears a previously stored one

I do not have a SONiC build environment available, so I have not been able to compile the tree or run the suite locally — I would appreciate a CI run. The changed logic was checked in isolation against this project's exact warning set from `configure.ac` (`-Wall -Wextra -Wconversion -Werror`), and the tests follow the live-redis pattern already used by `TestZmqRedisClient.cpp`.

The ~94s / ~59.6s figures are from the syncd logs in the linked issue, not from a run of my own; I do not have access to a 418K-route setup to measure the improvement directly.

None. The change is platform generic, though the issue was reported against Mellanox, where orchagent's 480s `APPLY_VIEW` timeout workaround currently masks the problem.

### Documentation

No documentation change — this is an internal performance fix with no user-facing behaviour or config change.

### DIFF
--- a/syncd/RedisClient.cpp
+++ b/syncd/RedisClient.cpp
@@ -669,14 +669,36 @@ void RedisClient::setVidAndRidMap(
     m_dbAsic->del(VIDTORID);
     m_dbAsic->del(RIDTOVID);
 
+    if (map.empty())
+    {
+        return;
+    }
+
+    // Use a pipeline to avoid a round trip per map entry, same as done in
+    // insertVidsAndRids(). Default pipeline size is used on purpose here, so
+    // that replies are flushed periodically and memory stays bounded.
+
+    swss::RedisPipeline pipe(m_dbAsic.get());
+
     for (auto &kv: map)
     {
         std::string strVid = sai_serialize_object_id(kv.first);
         std::string strRid = sai_serialize_object_id(kv.second);
 
-        m_dbAsic->hset(VIDTORID, strVid, strRid);
-        m_dbAsic->hset(RIDTOVID, strRid, strVid);
+        {
+            swss::RedisCommand hset;
+            hset.format("HSET %s %s %s", VIDTORID, strVid.c_str(), strRid.c_str());
+            pipe.push(hset, REDIS_REPLY_INTEGER);
+        }
+
+        {
+            swss::RedisCommand hset;
+            hset.format("HSET %s %s %s", RIDTOVID, strRid.c_str(), strVid.c_str());
+            pipe.push(hset, REDIS_REPLY_INTEGER);
+        }
     }
+
+    pipe.flush();
 }
 
 std::vector<std::string> RedisClient::getAsicStateKeys() const
@@ -889,10 +911,17 @@ void RedisClient::removeAsicStateTable()
 
     const auto &asicStateKeys = m_dbAsic->keys(ASIC_STATE_TABLE ":*");
 
-    for (const auto &key: asicStateKeys)
+    if (asicStateKeys.empty())
     {
-        m_dbAsic->del(key);
+        return;
     }
+
+    // Keys returned by keys() already contain the table prefix. Removing them
+    // with a single del() call issues chunked pipelined DEL commands, instead
+    // of one round trip per key, which at route scale dominates the time spent
+    // in warm boot APPLY_VIEW.
+
+    m_dbAsic->del(asicStateKeys);
 }
 
 void RedisClient::removeTempAsicStateTable()
@@ -901,10 +930,14 @@ void RedisClient::removeTempAsicStateTable()
 
     const auto &tempAsicStateKeys = m_dbAsic->keys(TEMP_PREFIX ASIC_STATE_TABLE ":*");
 
-    for (const auto &key: tempAsicStateKeys)
+    if (tempAsicStateKeys.empty())
     {
-        m_dbAsic->del(key);
+        return;
     }
+
+    // Same as in removeAsicStateTable(), remove all keys in bulk.
+
+    m_dbAsic->del(tempAsicStateKeys);
 }
 
 std::map<sai_object_id_t, swss::TableDump> RedisClient::getAsicView()
--- a/syncd/Syncd.cpp
+++ b/syncd/Syncd.cpp
@@ -47,6 +47,12 @@
 #define SAI_FAILURE_DUMP_SCRIPT "/usr/bin/sai_failure_dump.sh"
 #define SYNCD_ZMQ_RESPONSE_BUFFER_SIZE (128*1024*1024)
 
+// Number of ASIC objects written to redis database in a single bulk
+// call when saving temporary view as current view. Bulk write builds an
+// intermediate hash of the objects, so writing all of them at once at
+// route scale would cause a significant memory spike during warm boot.
+#define ASIC_STATE_BULK_WRITE_CHUNK_SIZE (10000u)
+
 using namespace syncd;
 using namespace saimeta;
 using namespace sairediscommon;
@@ -5936,6 +5942,13 @@ void Syncd::updateRedisDatabase(
     m_client->removeTempAsicStateTable();
 
     // Save temporary views as current view in redis database.
+    //
+    // Objects are written in bulk, since createAsicObject() issues a separate
+    // redis call per object attribute, which at route scale is the dominant
+    // cost of the whole warm boot APPLY_VIEW. Writing is done in chunks to keep
+    // the memory used by the intermediate hash bounded.
+
+    std::unordered_map<std::string, std::vector<swss::FieldValueTuple>> multiHash;
 
     for (auto& tv: temporaryViews)
     {
@@ -5954,8 +5967,23 @@ void Syncd::updateRedisDatabase(
                 entry.emplace_back(saiAttr->getStrAttrId(), saiAttr->getStrAttrValue());
             }
 
-            m_client->createAsicObject(obj->m_meta_key, entry);
+            // NOTE: key must be without table prefix, since createAsicObjects()
+            // is adding it internally
+
+            multiHash[sai_serialize_object_meta_key(obj->m_meta_key)] = std::move(entry);
+
+            if (multiHash.size() >= ASIC_STATE_BULK_WRITE_CHUNK_SIZE)
+            {
+                m_client->createAsicObjects(multiHash);
+
+                multiHash.clear();
+            }
         }
+    }
+
+    if (multiHash.size() > 0)
+    {
+        m_client->createAsicObjects(multiHash);
     }
 
     /*
--- a/syncd/tests/Makefile.am
+++ b/syncd/tests/Makefile.am
@@ -5,7 +5,7 @@ LDADD_GTEST = -L/usr/src/gtest -lgtest -lgtest_main
 bin_PROGRAMS = tests
 
 tests_SOURCES = \
-    main.cpp TestSyncdBrcm.cpp TestSyncdMlnx.cpp TestSyncdNvdaBf.cpp TestSyncdLib.cpp TestSyncdLinkEventDamping.cpp TestDisabledRedisClient.cpp TestZmqRedisClient.cpp
+    main.cpp TestSyncdBrcm.cpp TestSyncdMlnx.cpp TestSyncdNvdaBf.cpp TestSyncdLib.cpp TestSyncdLinkEventDamping.cpp TestDisabledRedisClient.cpp TestZmqRedisClient.cpp TestRedisClient.cpp
 tests_CXXFLAGS = \
     $(DBGFLAGS) $(AM_CXXFLAGS) $(CXXFLAGS_COMMON)
 tests_LDADD = \
--- a/syncd/tests/TestRedisClient.cpp
+++ b/syncd/tests/TestRedisClient.cpp
@@ -1,0 +1,195 @@
+#include <gtest/gtest.h>
+#include <memory>
+#include <string>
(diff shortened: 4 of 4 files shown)


PRIEYAN:
> Hi, there are workflow run(s) waiting for approval, you may be first-time contributor. I will notify maintainers to help approve once PR is approved. Thanks!
> ---Powered by SONiC BuildBot

Yeah, Thanks! I'll wait until someone approve that flow

PR #2056, opened 2026-08-25
[flexcounter] Remove stale ACL counters after ACL removal

### LINKED ISSUE #1982: Stale ACL counter entries permanently leaked in COUNTERS_DB after ACL rule deletion
## Description

`COUNTERS_DB` (redis DB 2) accumulates permanent orphaned hashes for deleted ACL rule counters, e.g.:

```
COUNTERS:oid:0x90000000a720c
  SAI_ACL_COUNTER_ATTR_BYTES: 0
  SAI_ACL_COUNTER_ATTR_PACKETS: 0
```

These entries are never cleaned up even though the corresponding ACL rule, SAI object, and flex-counter registration have all been correctly removed. On a production device with ACL rule churn, this reached **582,805 orphaned `COUNTERS:oid:0x9*` entries vs. only 111 live ACL rules** (>99.98% leak rate), causing unbounded COUNTER_DB growth and misleading monitoring/telemetry data.

## Root Cause

In `syncd/FlexCounter.cpp`, `FlexCounter::removeCounter()`, the `SAI_OBJECT_TYPE_ACL_COUNTER` branch only calls `removeObject()` on the internal counter-context (which just erases syncd's in-memory polling map). It never calls `removeDataFromCountersDB()`, the helper that actually issues `COUNTERS_DB.del("COUNTERS:oid:<oid>")`:

```cpp
else if (objectType == SAI_OBJECT_TYPE_ACL_COUNTER)
{
    if (hasCounterContext(ATTR_TYPE_ACL_COUNTER))
    {
        getCounterContext(ATTR_TYPE_ACL_COUNTER)->removeObject(vid);
    }
}
```

`removeDataFromCountersDB()` is already correctly wired up for other object types — `ROUTER_INTERFACE` (RIF), `SAI_OBJECT_TYPE_COUNTER` (flow/trap), and `ENI` (DASH, fixed recently in #1836, same root cause). It was simply never added for `ACL_COUNTER`.

## Steps to Reproduce

1. Create an ACL rule with a counter (e.g. via `ACL_RULE|<table>|<rule>` in CONFIG_DB, or `config acl` / loaded ACL json).
2. Confirm the counter stack is created: SAI object in ASIC_DB, polling entry in FLEX_COUNTER_DB, stats hash in `COUNTERS:oid:<oid>` in COUNTERS_DB, and name-map entry in `ACL_COUNTER_RULE_MAP`.
3. Delete the ACL rule.
4. Observe:
   - ASIC_DB object: removed ✓
   - FLEX_COUNTER_DB entry: removed ✓
   - `ACL_COUNTER_RULE_MAP` entry: removed ✓
   - `COUNTERS:oid:<oid>` hash in COUNTERS_DB: **still present**, orphaned permanently ✗

This was verified live on a physical testbed switch and is 100% reproducible.

## Expected Behavior

When an ACL counter object is destroyed, its corresponding `COUNTERS:oid:<oid>` hash should also be deleted from COUNTERS_DB, consistent with how RIF, flow/trap counters, and ENI counters are already handled.

## Suggested Fix

Add a `removeDataFromCountersDB()` call to the `ACL_COUNTER` branch in `FlexCounter::removeCounter()`, mirroring the pattern used for `ENI` in #1836:

```cpp
else if (objectType == SAI_OBJECT_TYPE_ACL_COUNTER)
{
    if (hasCounterContext(ATTR_TYPE_ACL_COUNTER))
    {
        removeDataFromCountersDB(vid, "");
        getCounterContext(ATTR_TYPE_ACL_COUNTER)->removeObject(vid);
    }
}
```

## Related / Prior Art

#1836 fixed the identical root cause for `SAI_OBJECT_TYPE_ENI` (DASH) counters — same missing `removeDataFromCountersDB()` call, same fix pattern, and included a unit test in `unittest/syncd/TestFlexCounter.cpp` that can serve as a template for testing this fix.

## Additional Notes

While auditing `FlexCounter::removeCounter()`, several other object types appear to have the same gap and may warrant follow-up investigation (each would need independent reproduction/verification before fixing):
- `MACSEC_SA` (SAs rotate periodically via rekeying)
- `HA_SET` (DASH, same class as ENI)
- `TUNNEL` (dynamic VXLAN/EVPN/dual-ToR tunnels)
- `POLICER` (bound/unbound with ACL/storm-control config changes)
- `MACSEC_FLOW` (created/destroyed per MACsec session)

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

Fixes https://github.com/sonic-net/sonic-sairedis/issues/1982

When an ACL rule is deleted, orchagent removes the SAI object, the
`ACL_COUNTER_RULE_MAP` entry, and the `FLEX_COUNTER_DB` polling entry, but the
`COUNTERS:oid:<oid>` hash written by syncd's flex counter poll is never removed.
`FlexCounter::removeCounter()` only erased its in-memory polling maps for
`SAI_OBJECT_TYPE_ACL_COUNTER`; no component issues the `DEL`, so the entries
survive until `COUNTERS_DB` is flushed on a cold restart.

On a production device with ACL rule churn this reached 582,805 orphaned
`COUNTERS:oid:0x9*` entries against 111 live ACL rules (>99.98% leak rate),
causing unbounded `COUNTERS_DB` growth and misleading monitoring/telemetry data.

##### Work item tracking
- Microsoft ADO **(number only)**: 39402687

#### How did you do it?

Added the missing `removeDataFromCountersDB()` call to the
`SAI_OBJECT_TYPE_ACL_COUNTER` branch of `FlexCounter::removeCounter()`, so the
`COUNTERS:oid:<oid>` hash is deleted when the ACL counter is deregistered.

- `unittest/syncd/TestFlexCounter.cpp`: the two `SAI_OBJECT_TYPE_ACL_COUNTER`
  cases in `TEST(FlexCounter, addRemoveCounter)` (single-add and bulk-add) now pass
  `autoRemoveDbEntry = true`. The helper therefore no longer deletes the key on
  the test's behalf, and its closing `ASSERT_TRUE(keys.empty())` asserts that
  syncd itself removed the `COUNTERS:oid:` entry. Without this fix those cases fail.
- Built with `--with-sai=vs` and ran `unittest/syncd` against a local redis.

- Verified on hardware: Broadcom 7050CX3, `internal-202605` image built with this fix
  (`202605` + this commit), syncd container replaced on a switch otherwise running the
  unpatched build. The test creates ACL rules with counters, waits for the flex counter to
  publish the `COUNTERS:oid:` hashes, deletes the rules and table, waits until ASIC_DB /
  FLEX_COUNTER_DB / `ACL_COUNTER_RULE_MAP` are cleaned up, then inspects COUNTERS_DB:

  | syncd | 5 rules x 1 cycle | 20 rules x 3 cycles |
  | --- | --- | --- |
  | before (unpatched) | **5/5 leaked** | - |
  | after (this fix) | **0/5 leaked** | **0/60 leaked** |

  Before the fix every OID reported `ASIC_DB=removed FLEX_COUNTER_DB=removed COUNTERS_DB=STALE`,
  leaving behind the all-zero `SAI_ACL_COUNTER_ATTR_BYTES` / `SAI_ACL_COUNTER_ATTR_PACKETS`
  hash described in the issue. After the fix every OID reports `COUNTERS_DB=removed` and the
  device ends with 0 ACL counter keys in COUNTERS_DB.

  The reproducer script used here is posted on
  https://github.com/sonic-net/sonic-sairedis/issues/1982 (exit 1 = leak reproduced,
  exit 0 = fix present, so it works as a before/after gate).

### DIFF
--- a/syncd/FlexCounter.cpp
+++ b/syncd/FlexCounter.cpp
@@ -4374,6 +4374,7 @@ void FlexCounter::removeCounter(
     {
         if (hasCounterContext(ATTR_TYPE_ACL_COUNTER))
         {
+            removeDataFromCountersDB(vid, "");
             getCounterContext(ATTR_TYPE_ACL_COUNTER)->removeObject(vid);
         }
     }
--- a/unittest/syncd/TestFlexCounter.cpp
+++ b/unittest/syncd/TestFlexCounter.cpp
@@ -788,7 +788,7 @@ TEST(FlexCounter, addRemoveCounter)
         {"SAI_ACL_COUNTER_ATTR_PACKETS"},
         {"1000"},
         counterVerifyFunc,
-        false);
+        true);
 
     // Bulk create mode to satisfy the coverage requirement
     testAddRemoveCounter(
@@ -798,7 +798,7 @@ TEST(FlexCounter, addRemoveCounter)
         {"SAI_ACL_COUNTER_ATTR_PACKETS"},
         {"1000"},
         counterVerifyFunc,
-        false,
+        true,
         STATS_MODE_READ,
         true);
 


bingwang-ms (approved):

bingwang-ms:
The fix to ACL part LGTM. Are we going to fix the leak for other resources as well? Such as
- MACSEC_SA (SAs rotate periodically via rekeying)
- HA_SET (DASH, same class as ENI)
- TUNNEL (dynamic VXLAN/EVPN/dual-ToR tunnels)
- POLICER (bound/unbound with ACL/storm-control config changes)
- MACSEC_FLOW (created/destroyed per MACsec session)

PR #1115, opened 2022-08-25
[202205][SAI] advance SAI version to 1.10 to pick up saiserver and syncd-rpc support

richardyu-ms:
* 566d4a8ef2 2022-08-11 | [SAI-PTF] Enable saiserverv2 with syncd-rpc and fix saithriftv2 build (#1552) (#1533) (#1514) (#1492) (#1558) (#1557) [Richard.Yu]

PR #2057, opened 2026-08-25
[syncd] Do not abort syncd when a FlexCounter poll cycle fails

### LINKED ISSUE #2055: [syncd] A transient Redis protocol error aborts the entire syncd process
### Description

On production devices running SONiC, `syncd` exits with SIGABRT (exit 134):

```
terminate called after throwing an instance of 'std::system_error'
  what(): Expected to get redis type 3 got type 6,
          err: ERR Protocol error: expected '$', got 'N': Input/output error
```

Redis rejected a frame from the FlexCounter counter-polling path with a protocol
error, `RedisReply::checkReplyType()` threw, and nothing caught it — so the whole
`syncd` process died, losing every counter group and the entire SAI object state,
rather than losing a single poll cycle.

Seen on Arista 7060CX-32S devices.

### The poll loop has no exception handling

`FlexCounter::flexCounterThreadRunFunction()` in `syncd/FlexCounter.cpp` runs the
counter poll loop with no exception handling:

```cpp
swss::DBConnector db(m_dbCounters, 0, m_isTcpConn);
swss::RedisPipeline pipeline(&db);
swss::Table countersTable(&pipeline, COUNTERS_TABLE, true);

{
    // ...
    collectCounters(countersTable);      // can throw
    runPlugins(db);                      // can throw
    // ...
}
```

Any exception raised while draining the COUNTERS_DB pipeline therefore escapes the
thread's top-level function, and `std::terminate()` runs.

### Why a bare `try`/`catch` is not sufficient

Redis calls `setProtocolError()` on a framing error, which sets
`CLIENT_CLOSE_AFTER_REPLY` (`redis/src/networking.c`). The server therefore closes
the connection after sending that error reply. A handler that swallows the
exception but keeps the existing `DBConnector` / `RedisPipeline` would retry
forever on a dead socket: the poll loop never produces counters again and
`RedisPipeline::m_remaining` grows without bound.

We measured exactly this. With a catch-only handler: zero successful poll cycles
and pending replies climbing 2 → 5 → 8 → 11. A silent permanent counter outage
plus a leak — arguably worse than the crash, because nothing alerts.

The connection has to be **rebuilt**, not just the exception caught.

### Affected branches

Verified present — unchanged and unguarded — on `202411`, `202505`, `202511`,
`202605` and `master`.

### How we plan to handle this

1. **Stop the process abort.** Hold the connector, pipeline and table in
   `unique_ptr` so they can be rebuilt, catch exceptions around the poll body, and
   reconnect after a failed cycle — connecting and releasing **outside** the
   counter mutex so a slow or refused connect cannot block `addCounter()` /
   `removeCounter()`. The existing poll-interval sleep doubles as retry backoff;
   failure logging is rate limited. PR to follow.
2. **Find out why the protocol error happens at all** — still under investigation
   on our side; we will report back separately.

Step 1 does not depend on step 2: whatever the trigger turns out to be, one failed
poll cycle should not kill `syncd`.

lipxu:
### Description of PR

Summary:
Fixes #2055

`FlexCounter::flexCounterThreadRunFunction()` calls `collectCounters()` and
`runPlugins()` with no exception handling, so a transient Redis error while
draining the COUNTERS_DB pipeline escapes the thread top-function and
`std::terminate()` kills the whole `syncd` process. This PR makes a failed poll
cycle a logged, self-healing reconnect instead.

**Dependencies / merge order:** none for `master` and `202605`. On `202511`,
`202505` and `202411`, `~RedisPipeline()` is still unguarded, so the reconnect
introduced here would terminate the process; `sonic-swss-common`
[#1115](https://github.com/sonic-net/sonic-swss-common/pull/1115) — already merged
on `master` — must be backported to those branches first. Details under
*How did you do it?*.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

A failed counter poll is transient and single-cycle. Aborting `syncd` over it is
disproportionate, because SONiC does **not** restart `syncd` in place — its death
tears down the container and reprograms the ASIC:

```
poll throws -> std::terminate() -> SIGABRT
  -> supervisord does NOT restart it        supervisord.conf: [program:syncd] autorestart=false
  -> proc-exit-listener kills the container critical_processes: program:syncd
  -> auto_restart brings the service back   init_cfg.json.j2: FEATURE|syncd|auto_restart=enabled
  -> SAI re-init, ASIC reprogrammed, swss/orchagent cycles with it -> dataplane impact
```

(paths under `sonic-buildimage`: `platform/broadcom/docker-syncd-brcm/`,
`files/build_templates/`; `syncd.service.j2` itself has no `Restart=`.)

So one bad Redis reply costs a container restart and a dataplane hit. This PR keeps
the recovery but moves it from container granularity down to rebuilding three Redis
handles.

Seen in production on an Arista 7060CX-32S:

```
terminate called after throwing an instance of 'std::system_error'
  what(): Expected to get redis type 3 got type 6,
          err: ERR Protocol error: expected '$', got 'N': Input/output error
```

This is not a deliberate crash-only design — the original code states no such
contract, it is simply an unhandled exception. The same pattern exists elsewhere in
SONiC: `orchagent`, `eventd` and `dhcpmon` all abort on the same fault.

##### Work item tracking
- Microsoft ADO **(number only)**: 38233337

#### How did you do it?

Catching the exception alone is **not** sufficient, for two reasons.

**1. The connection is dead, not merely errored.** Redis calls `setProtocolError()`,
which sets `CLIENT_CLOSE_AFTER_REPLY`, so the server closes the connection after
that error reply.

**2. `RedisPipeline` cannot be resynchronised.** It exposes no `reset` / `clear` /
`resync`, and its state cannot self-correct, because `pop()` updates the bookkeeping
*after* the failure point:

```cpp
int rc = redisGetReply(m_db->getContext(), (void**)&reply);
if (rc != REDIS_OK) throw RedisError(...);   // throws here
m_remaining--;                               // never reached on failure
m_expectedTypes.pop();
```

`push()` meanwhile keeps succeeding, because `redisAppendCommand` only appends to
hiredis's `obuf` and never touches the socket — so `m_remaining`, `m_expectedTypes`
and `obuf` grow monotonically. Destroy-and-rebuild is therefore the only available
reset, which is what forces the `unique_ptr` change.

So:

- `DBConnector`, `RedisPipeline` and `Table` move from stack objects to
  `unique_ptr`, so they can be rebuilt.
- The poll body is wrapped in `try`/`catch`; on failure the objects are torn down
  and reconnected on the next iteration.
- **Connect and release happen outside the counter mutex**, so a slow or refused
  connect cannot block `addCounter()` / `removeCounter()`.
- The existing poll-interval sleep doubles as retry backoff.
- Failure logging **from this loop** is rate limited to 1 in
  `FLEX_COUNTER_POLL_FAILURE_LOG_INTERVAL` (60) cycles — this does not cover
  `swss-common`'s own destructor log, see *Known limitations*.

Two files change: `syncd/FlexCounter.cpp` **+96 / −5**, and
`unittest/syncd/TestFlexCounter.cpp` **+123** (the regression test).

<details>
<summary><b>Why 202511 / 202505 / 202411 additionally need sonic-swss-common #1115</b></summary>

The reconnect calls `pipeline.reset()` → `~RedisPipeline()`, which differs by branch:

```cpp
// master / 202605 - guarded, cannot throw
~RedisPipeline() noexcept {
    try {
        if (m_ownerTid == gettid()) {
            try { flush(); } catch (const std::exception& e) { SWSS_LOG_ERROR(...); }
        }
        delete m_db;                    // still reached
    } catch (...) { fprintf(stderr, ...); }
}

// 202511 / 202505 / 202411 - no guard
~RedisPipeline() {
    if (m_ownerTid == gettid()) { flush(); }   // throws on a dead socket
    delete m_db;
}
```

Destructors are implicitly `noexcept(true)`, so on those branches the doomed
`flush()` terminates the process — the same SIGABRT, merely relocated.
`m_ownerTid == gettid()` is always true here, as the reset runs on the thread that
constructed the object. A `try`/`catch` at the call site does not help either: an
exception escaping a `noexcept` function terminates at the throw point and never
reaches a handler. Backporting `swss-common` #1115 is therefore a hard prerequisite
for 202511, 202505 and 202411.

</details>

**1. A new unit test in this repo's own suite.**

`FlexCounter.pollLoopSurvivesThrowAndRecovers` (added here) makes the mocked
`getStats` throw the same exception the swss Redis layer raises on a
desynchronised reply stream:

```cpp
throw std::runtime_error("Protocol error: expected '$', got 'N'");
```

It then asserts three things: the process is still alive after repeated throws;
the counters go **stale** rather than blank; and after the throw stops, the loop
publishes **new, different** values — proving it actually reconnected rather than
quietly going idle.

Note the shape of this test: **without the fix it does not fail, it aborts the
test binary.** That is the bug.

| Run | Tree | Result |
|---|---|---|
| `make -C syncd` | patched | builds |
| `./tests --gtest_filter='FlexCounter*'` | patched | **19/19 pass** (18 pre-existing + new) |
| `./tests --gtest_filter='FlexCounter.pollLoopSurvivesThrowAndRecovers'` | **unpatched** | **exit 134 — SIGABRT** |
| full suite (175 tests), new test excluded | unpatched | 103 pass · 72 fail |
| full suite (176 tests), new test included | patched | 104 pass · 72 fail |

The 72 failures are pre-existing and unrelated (`VendorSai`, `Syncd`,
`SyncdLinkEventDampingTest`, …) — they need a real vendor SAI library. The two
failure lists were diffed line by line and are **identical**, so this change
introduces **no regression**.

The unpatched run reproduces the production signature exactly:

```
[ RUN      ] FlexCounter.pollLoopSurvivesThrowAndRecovers
  what():  Protocol error: expected '$', got 'N'
Aborted (core dumped)                      <- exit 134
```

**2. A userspace harness across all five branches.**

It compiles **each branch's own `flexCounterThreadRunFunction()`, extracted
verbatim from git**, links it against the **real upstream `swss::RedisPipeline`**
header, and injects the exact field fault — one framing byte flipped in hiredis's
`obuf` immediately before `send()`, via an `LD_PRELOAD` hook. Same code, same
fault, before and after the patch.

Run on `master`, `202605`, `202511`, `202505` and `202411`. Each branch is compiled
against **its own** `swss-common` `common/redispipeline.h`, asserted by git blob
hash before anything is built:

| | master | 202605 | 202511 | 202505 | 202411 |
|---|---|---|---|---|---|
| `~RedisPipeline()` guarded (#1115) | yes | yes | **no** | **no** | **no** |
| BEFORE / fault → **SIGABRT 134** | ✔ | ✔ | ✔ | ✔ | ✔ |
| AFTER / clean → no regression | ✔ | ✔ | ✔ | ✔ | ✔ |
| **AFTER / fault → recovered** | ✔ | ✔ | ✘ | ✘ | ✘ |
| AFTER / 3 faults → repeatable | ✔ | ✔ | ✘ | ✘ | ✘ |

The crash reproduces on all five; the fix is sufficient on `master` and `202605`
only. On the three unguarded branches the syncd-side guard fires correctly, but the
process still dies while tearing the pipeline down:

```
[ERROR ] FC ...: poll cycle failed (1 consecutive), reconnecting COUNTERS_DB: ...
terminate called after throwing an instance of 'swss::RedisError'
  what():  Failed to redisGetReply in RedisPipeline::pop, err=3: Server closed the connection
```

The first line shows the new guard working; the process still aborts because
`~RedisPipeline()` on those branches has no handler. That is direct empirical
confirmation that backporting #1115 is a hard prerequisite, not defence in depth.

`flexCounterThreadRunFunction()` itself has only two distinct forms across the five
branches: `master` / `202605` / `202511` are byte-identical, and `202505` / `202411`
differ by exactly one line — `DBConnector` has no `m_isTcpConn` argument yet — which
the harness adapts before building.

Also measured, as a control: a **catch-only** handler produces zero successful
cycles and pending replies climbing 2 → 5 → 8 → 11 — a silent permanent counter
outage plus a leak. That is why this PR rebuilds the connection.

#### On device

A private SONiC image built from `202605` carrying this change was installed on an
Arista 7060CX-32S (`Arista-7060CX-32S-C32`, topology `t1-lag`) and put through a
dataplane run. `flexCounterThreadRunFunction()` in that image is byte-identical to the
version in this PR, so the code exercised on hardware is the code under review.

**42 tests: 30 passed, 0 failed, 12 skipped.** Thirteen of the passes are real
forwarding cases (`fib/test_fib` 12, `decap/test_decap` 1); the other 17 are
pretest/posttest housekeeping. All 12 skips are collection-time topology
deselections — 6 of them in `ipfwd/test_dip_sip`, whose `topology` marker lists `t1`
but not `t1-lag` — so they are a property of the testbed and the test plan, not of this
change. No dataplane regression.

**Not covered:** the hardware run is a regression check, not a fault-injection test. No
Redis outage was injected on device, so the crash path itself is exercised only in the
unit test and the harness, where the SAI layer is mocked or stubbed and the fault is
injected deliberately.

None. The change is in generic syncd code. Originally observed on Broadcom
(Arista 7060CX-32S), but nothing in the fix is vendor specific.

### Known limitations

Disclosed deliberately; all are bounded and none is fatal.

1. **The log rate limiting does not cover `swss-common`.** Each reconnect runs
   `~RedisPipeline()`, whose own `SWSS_LOG_ERROR` is not governed by
   `FLEX_COUNTER_POLL_FAILURE_LOG_INTERVAL`. During a sustained outage with the 8
   counter groups seen on hardware at a 1 s poll interval, this loop logs
   ~0.13 msg/s while `swss-common` logs ~8 msg/s.

2. **Reconnects have no jitter or exponential backoff.** Each instance holds two
   Redis connections (`RedisPipeline` opens its own via `newConnector()` rather than
   reusing the passed `DBConnector`), so one blip makes all 8 groups reconnect on the
   same cycle — 16 connections per poll interval, repeated indefinitely. PFC_WD
   polls at 200 ms.

3. **Connect has no timeout.** `DBConnector(m_dbCounters, 0, ...)` resolves to
   `redisConnect()` / `redisConnectUnix()` (`dbconnector.cpp:712`, `:585`, `:601`),
   and `RedisPipeline`'s own connection is fixed at `NEWCONNECTOR_TIMEOUT = 0`
   (`redispipeline.h:23`), which sairedis cannot override. The value is inherited
   unchanged from the pre-patch code, but now runs on every reconnect instead of
   once per thread. A refused connection still fails fast; a *wedged* Redis (blocked
   on fork/AOF rewrite, full accept backlog) can block the thread and stall
   `endFlexCounterThread()`'s `join()`.

4. **New steady state: `syncd` healthy, counters stale.** This is the intended trade
   and is strictly better than a container restart, but it is a state that could not
   previously occur, and SONiC publishes no freshness signal for it
   (`FLEX_COUNTER_STATUS_FIELD` is a CONFIG_DB input; `RedisPipeline::getIdleTime()`
   has no callers). Staleness is detectable with
   `redis-cli -n 2 OBJECT IDLETIME COUNTERS:<oid>` (0-1 while healthy). Monitoring
   that relied on `syncd` crash/restart signals should watch for
   `FC <id>: poll cycle failed` in syslog and for counter staleness instead.

**Follow-ups:** jittered exponential backoff on the reconnect path; a bounded connect
timeout (needs a `RedisPipeline` constructor overload in `swss-common`); a FlexCounter
heartbeat field so staleness becomes a first-class signal.

### Documentation

No documentation change needed — internal error handling only.

---


lolyu:
Hi @justin-wong-ce, could you please help review this one as well?

lolyu (commented):
Excellent, rigorously-validated fix — verified the whole chain against source. The root cause is correct (unhandled exception on the `std::thread` top-function → `std::terminate()` → SIGABRT, and because syncd is `autorestart=false` the death tears down the container and reprograms the ASIC — disproportionate to a transient Redis protocol error). The "catch alone is insufficient" reasoning is right and I confirmed it: Redis sets `CLIENT_CLOSE_AFTER_REPLY` so the connection is dead, and `RedisPipeline` can't resync because `pop()` updates `m_remaining`/`m_expectedTypes` after the throw while `push()` keeps appending to `obuf` — there's no `reset`/`resync`, so destroy-and-rebuild is genuinely the only option and the `unique_ptr` change is forced, not stylistic. The five-branch harness proving the syncd-side guard fires but the process still aborts in `~RedisPipeline()` on the unguarded branches is exactly the evidence that makes swss-common #1115 a hard backport prerequisite rather than defence-in-depth — very well done.

All three Copilot comments are already resolved at head `1436306`: `catch(...)` on both paths, the test-comment reworded, and `m_pollInterval` snapshotted under `MUTEX` before the backoff. The regression test is strong — it asserts survival, staleness (not blanking), and recovery with distinguishable values, and aborts the binary without the fix.

Two non-blocking notes inline. Net: approve-quality; flagging as Comment only to give you a chance to reply on the wedged-Redis-under-mutex point and to track the jitter follow-up before this propagates on 202605.

syncd/FlexCounter.cpp
```diff
@@ -4188,21 +4195,139 @@ void FlexCounter::flexCounterThreadRunFunction()
...
-            collectCounters(countersTable);
+            bool pollFailed = false;
+            std::string failureReason;
+
+            // A counter poll must never take down syncd. Any failure here is
+            // transient by nature (Redis I/O, protocol desync, malformed reply), so
+            // log it, drop the connection and retry on the next cycle.
+            try
+            {
+                collectCounters(*countersTable);
```
lolyu, line 4325: Non-blocking (pre-existing): the poll drain here runs while holding `m_mtx` (the `MUTEX` above this `if`). The PR's stated design goal is that connect/release happen *outside* the counter mutex so a slow/refused connect can't block `addCounter()`/`removeCounter()` — and the release path does correctly reset after `MUTEX_UNLOCK`. But the poll itself (`collectCounters` → `countersTable.flush()` + pipeline drain / `redisGetReply`) still holds `m_mtx`, so a *wedged* (not refused) Redis — Known-limitation #3's blocked-on-fork/AOF case — can still stall `addCounter()`/`removeCounter()` for the hang duration, now on every poll. This matches the old behavior (the pre-patch code also polled under the mutex), so it's not a regression — worth a comment noting the "outside the mutex" guarantee covers connect/release but not the poll drain itself.

syncd/FlexCounter.cpp
```diff
@@ -4213,6 +4338,15 @@ void FlexCounter::flexCounterThreadRunFunction()
             correction = m_pollInterval - correction;
             MUTEX_UNLOCK; // explicit unlock
 
+            if (pollFailed)
+            {
+                // Release the handles outside the counter mutex; the next iteration
+                // reconnects. The sleep below doubles as the retry backoff.
```
lolyu: Non-blocking follow-up (your Known-limitation #2): on a sustained outage all 8 counter groups reset and reconnect on the same cycle with no jitter — 16 connections/poll-interval (PFC_WD at 200ms), repeated indefinitely. Agreed this is still far better than a container restart, so fine for the fix. But since this is labeled for 202605, I'd like the jittered/exponential-backoff follow-up tracked (an issue reference here is enough) before it propagates widely on a release branch. The `m_cvSleep.wait_for` here is the natural place to add the jitter later.

syncd/FlexCounter.cpp
lipxu: Good catch -- fixed in e130470. You're right that the poll runs under `MUTEX`; only the connect and release are outside it. The commit narrows the comment and adds a note on the wedged-Redis case, including that the retry loop means the stall can now recur every cycle. Comments only, no behaviour change.

syncd/FlexCounter.cpp
lipxu: Agreed -- filed as #2065 and referenced it from both wait sites in 9003c6f. The post-poll wait you pointed at doubles as the backoff on a failed cycle, so it's covered too.

lipxu:
> Hi @justin-wong-ce, could you please help review this one as well?

Hi, @justin-wong-ce , could you please help to review this PR, thanks a lot

justin-wong-ce:
> > Hi @justin-wong-ce, could you please help review this one as well?
> 
> Hi, @justin-wong-ce , could you please help to review this PR, thanks a lot

I will be building an image and taking a look, it seems to be a pretty important change so I will spend some time understanding it. I will probably put comments next week.

syncd/FlexCounter.cpp
```diff
@@ -4188,21 +4195,168 @@ void FlexCounter::flexCounterThreadRunFunction()
...
+            //
+            // Note that, unlike the connect and the release, this runs with the
+            // counter mutex held: collecting the counters flushes the pipeline, and
+            // that blocks while reading the replies. A Redis that refuses the
+            // connection fails fast and is handled outside the lock, but one that
+            // accepts and then stops answering, blocked on a background save for
+            // instance, holds up addCounter()/removeCounter() until the socket gives
+            // up. The guard does not change where the poll is serialized, so this is
+            // no worse than before; it is worth knowing that retrying forever also
+            // means such a stall can now repeat every cycle rather than ending in an
```
justin-wong-ce, line 4321: How long could this stall potentially be? 

Would poll interval < stall time cause multiple FC run threads to accumulate over time?

lipxu: Good questions — I measured both rather than reasoning about them. Numbers are from an instrumented run of the syncd unit tests against a real Redis.

**Q2 — no, threads can't accumulate.** The loop is fixed-*delay*, not fixed-*rate*: one thread, created once (`FlexCounter.cpp:4402`), running a sequential `while (m_runFlexCounterThread)` (`:4222`) with no timer and no work queue — so a slow cycle only delays the next one, it has nothing to overlap with. Measured at a 100 ms interval with every poll stalled 1000 ms (10x overrun):

```
threads: steady = 2 | max while stalling = 2
polls completed in a 5 s window: 5   (fixed-rate would have wanted ~50)
```

Related: `:4366-4367` sleeps `m_pollInterval - (delay % m_pollInterval)`, so after an overrun the loop re-aligns to the interval grid rather than spinning. Pre-existing, untouched here.

**Q1 — unbounded in the worst case, and equally unbounded before this PR.** The COUNTERS_DB connection carries no read timeout and can't be given one at this layer: `FlexCounter.cpp:4214` passes 0, `RedisPipeline` hardcodes `NEWCONNECTOR_TIMEOUT = 0` (`redispipeline.h:23`), and `dbconnector.cpp:711` turns 0 into `ptv = NULL`, which selects the plain blocking `redisConnect()` / `redisConnectUnix()` (`:577-607`). Read straight off the socket:

```
DBConnector("COUNTERS_DB", 0)    -> SO_RCVTIMEO = 0 s
DBConnector("COUNTERS_DB", 5000) -> SO_RCVTIMEO = 0 s
after redisSetTimeout(7 s)       -> SO_RCVTIMEO = 7 s   (probe sanity check)
```

The middle line is the interesting one: a non-zero `DBConnector` timeout still leaves `SO_RCVTIMEO` at zero — it's a *connect* timeout and never bounds a read. The third confirms the probe isn't broken. End to end, against a server that accepts the connection and then stops answering, the read was still blocked after 10 s, and only an explicit `shutdown()` released it.

So the honest answer is: as long as the peer stays silent, with no bound short of the socket closing. In practice the stalls we see are short (BGSAVE / AOF-rewrite latency spikes), but nothing in the code enforces that.

None of this is new — pre-patch the poll ran under the same mutex on an identically constructed connection. What changed is only the outcome once the RESP error surfaces: it used to escape and abort syncd (the crash this fixes); now it's caught, the connection is rebuilt, and polling resumes.

Bounding the stall itself is worth doing, but it needs an swss-common change — `DBConnector`'s timeout argument demonstrably won't do it, so it would take `redisSetTimeout()` on the context plus a decision on what a read timeout means mid-pipeline-flush. Happy to file that as a follow-up issue.

syncd/FlexCounter.cpp
```diff
@@ -4188,21 +4195,96 @@ void FlexCounter::flexCounterThreadRunFunction()
...
+                pollFailed = true;
 
-            runPlugins(db);
+                // Rate limit: a persistent failure would otherwise log every poll interval.
+                if ((consecutivePollFailures++ % FLEX_COUNTER_POLL_FAILURE_LOG_INTERVAL) == 0)
+                {
+                    SWSS_LOG_ERROR("FC %s: poll cycle failed (%" PRIu64 " consecutive), reconnecting COUNTERS_DB: %s",
+                            m_instanceId.c_str(), consecutivePollFailures, e.what());
+                }
+            }
```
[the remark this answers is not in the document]
lipxu: Fixed in `1c36661` — the poll loop now has a `catch (...)` fallback alongside the `catch (const std::exception&)`, setting the same `pollFailed` flag so it takes the identical rate-limited log + reconnect path. The connect path got the same treatment. Test coverage for it is in `c899fbd`.

unittest/syncd/TestFlexCounter.cpp
```diff
@@ -2995,3 +2996,125 @@ TEST(FlexCounter, failedPollsCountAndCleanUp)
...
+    sai->mock_getStats = [&](sai_object_type_t, sai_object_id_t,
+                             uint32_t number_of_counters, const sai_stat_id_t *,
+                             uint64_t *counters) -> sai_status_t
+    {
+        if (throwFromGetStats.load())
+        {
+            throwCount++;
+
+            // Same exception type and message the swss Redis layer raises on a
+            // desynchronised reply stream.
```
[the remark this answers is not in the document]
lipxu: Fixed in `8fb88fc` — the comment now says the test injects the Redis error signature through the SAI `mock_getStats` hook, rather than implying swss raises it.

syncd/FlexCounter.cpp
```diff
@@ -4188,21 +4195,96 @@ void FlexCounter::flexCounterThreadRunFunction()
...
+            catch (const std::exception& e)
+            {
+                SWSS_LOG_ERROR("FC %s: failed to connect to COUNTERS_DB: %s",
+                        m_instanceId.c_str(), e.what());
+
+                uint32_t backoff = (m_pollInterval > 0) ? m_pollInterval : FLEX_COUNTER_RECONNECT_BACKOFF_MS;
+
+                std::unique_lock<std::mutex> lk(m_mtxSleep);
+
+                m_cvSleep.wait_for(lk, std::chrono::milliseconds(backoff));
```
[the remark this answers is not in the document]
lipxu: Fixed in `c00b09e` — the reconnect path now takes a snapshot of `m_pollInterval` under `m_mtx` and releases the lock before computing the backoff, so the connect itself still runs lock-free.

lolyu (approved):

lolyu:
Hi @justin-wong-ce, needs your review for this PR, thanks.

justin-wong-ce:
> Hi @justin-wong-ce, needs your review for this PR, thanks.

I currently have this change patched on my testbed and are running more tested. The tests should finish when I wake up tomorrow.

lolyu:
> > Hi @justin-wong-ce, needs your review for this PR, thanks.
> 
> I currently have this change patched on my testbed and are running more tested. The tests should finish when I wake up tomorrow.

Thanks :D

justin-wong-ce:
This change runs well on my testbed, I don't see any regressions.

syncd/FlexCounter.cpp
justin-wong-ce: I think it is important to address the potential stall as a follow-up, even if it is not a common occurence. This is because there won't be any log nor visible issue if the stall happens. 

Any potential issues will now go from a `syncd` crash which forces a recovery, to a masked status where counters will remain stale and sent out on telemetry, not triggering any alarms. On a large scale production environment, this will be pretty scary.

justin-wong-ce:
I can approve after this issue is filed and assigned to be addressed
https://github.com/sonic-net/sonic-sairedis/pull/2057#discussion_r3994728250

lolyu:
Hi @lipxu, please help address the comment.

syncd/FlexCounter.cpp
lipxu: Agreed it's worth tracking, filed as #2075.

One correction though: a failed poll does log (`FlexCounter.cpp:4349`). The silent case is a wedged redis — we block in `redisGetReply()`, nothing throws, nothing logs. Still stuck after 10s when I tested it.

And that isn't new. Pre-patch (`371977b^`) had the same connection and no try/catch, so a wedged redis hung identically and logged just as little. It didn't crash either; the SIGABRT needs a malformed reply to arrive, and a stall is the absence of one.

lipxu:
> I can approve after this issue is filed and assigned to be addressed

Filed as #2075 and assigned to myself.

lolyu:
> I can approve after this issue is filed and assigned to be addressed [#2057 (comment)](https://github.com/sonic-net/sonic-sairedis/pull/2057#discussion_r3994728250)

Hi @justin-wong-ce, could you please help check the update? thanks

justin-wong-ce (approved):

lipxu:
https://github.com/sonic-net/sonic-sairedis/pull/2078 create a PR for 202605 backport manually

PR #2059, opened 2026-08-25
syncd: Avoid false nonexisting-object log during FlexCounter removal

### LINKED ISSUE #2053: syncd: Avoid false “Trying to remove nonexisting” notice during FlexCounter removal
## Description

`syncd` can incorrectly log that an object is nonexisting during FlexCounter removal when the object was successfully removed from the normal object-ID map but is not present in the bulk-stats context.

  ## Problem

  `FlexCounter::removeObject()` removes an object from two independent tracking structures:

  - `m_objectIdsMap`
  - Bulk-stats context, via `removeBulkStatsContext()`

An object may legitimately exist only in `m_objectIdsMap`. In that case, removal from the normal map succeeds, while removal from the bulk context returns `false`.

The current logic uses only the bulk-context removal result to decide whether to emit:
`Trying to remove nonexisting <object-type> <object-id>`

  This produces a misleading notice even though the object was successfully removed.

  ## Steps to Reproduce

  1. Enable PHY counter polling: `counterpoll phy enable`

  2. Disable PHY counter polling: `counterpoll phy disable`

  3. Check the Syncd logs for a message similar to: `Trying to remove nonexisting SAI_OBJECT_TYPE_PORT <object-id>`

  ## Expected Behavior

  The notice should be emitted only when the object is absent from both m_objectIdsMap and the bulk-stats context.

  ## Actual Behavior

A false Trying to remove nonexisting notice can be emitted when the object was successfully removed from m_objectIdsMap but was not present in the bulk-stats context.

  ## Proposed Fix

  Track the removal result from both tracking structures and emit the notice only if removal fails in both locations.

  ## Validation

  Validated on hardware by disabling and re-enabling PHY counter polling:

  counterpoll phy disable
  counterpoll phy enable

  No false Trying to remove nonexisting SAI_OBJECT_TYPE_PORT messages were observed.

mrinal-arista:
## Description

 Fix a false `Trying to remove nonexisting` notice emitted by `FlexCounter::removeObject()`.

An object can legitimately be tracked in `m_objectIdsMap` without being present in the bulk-stats context. Previously,   Syncd used only the bulk-context removal result to decide whether the object was nonexisting, which could produce a misleading notice after successful removal from `m_objectIdsMap`.

The fix tracks removal results from both locations and emits the notice only when the object is absent from both.

Fixes #2053

## How I Tested
Validated on hardware by disabling and re-enabling PHY counter polling:

  ```bash
  counterpoll phy disable
  counterpoll phy enable
 ```

No false Trying to remove nonexisting SAI_OBJECT_TYPE_PORT messages were observed in Syncd logs.

### DIFF
--- a/syncd/FlexCounter.cpp
+++ b/syncd/FlexCounter.cpp
@@ -1509,7 +1509,8 @@ public:
         SWSS_LOG_ENTER();
 
         auto iter = m_objectIdsMap.find(vid);
-        if (iter != m_objectIdsMap.end())
+        const bool removedFromObjectMap = iter != m_objectIdsMap.end();
+        if (removedFromObjectMap)
         {
             auto rid = iter->second->rid;
             m_failedPolls.erase({rid, vid});
@@ -1518,7 +1519,9 @@ public:
 
         // An object can be in both m_objectIdsMap and the bulk context
         // when bulk polling is supported by some counter prefixes but unsupported by some others
-        if (!removeBulkStatsContext(vid) && log)
+        const bool removedFromBulkContext = removeBulkStatsContext(vid);
+
+        if (!removedFromObjectMap && !removedFromBulkContext && log)
         {
             SWSS_LOG_NOTICE("Trying to remove nonexisting %s %s",
                             sai_serialize_object_type(m_objectType).c_str(),


dhanasekar-arista:
@StormLiangMS @lolyu pls assign reviewers for this PR.

lolyu (approved):
LGTM

lolyu:
Hi @mrinal-arista, do we need this to 202605?

syncd/FlexCounter.cpp
```diff
@@ -1518,7 +1519,9 @@ class CounterContext : public BaseCounterContext
 
         // An object can be in both m_objectIdsMap and the bulk context
         // when bulk polling is supported by some counter prefixes but unsupported by some others
```
lolyu, line 1521: Good — capturing `removeBulkStatsContext(vid)` into its own bool here (rather than folding it into the `if` as `... || removeBulkStatsContext(vid)`) is what preserves the side effect: the bulk context is still unconditionally cleaned up regardless of `m_objectIdsMap` membership. That's the one trap this fix had to avoid, and it does. Suggest a follow-up unit test in `TestFlexCounter.cpp` covering the object-in-`m_objectIdsMap`-only removal path so the no-false-notice behavior is guarded against future refactors.

lolyu (commented):
Verified against the full `removeObject` function (not just the hunk) — the fix is correct. "Nonexisting" should mean absent from **both** `m_objectIdsMap` and the bulk-stats context, which is exactly what the new `!removedFromObjectMap && !removedFromBulkContext` condition checks; the pre-fix code gated the notice only on the bulk-context result, so an object tracked solely in `m_objectIdsMap` (legitimate per the in-code comment about mixed bulk/non-bulk prefix support) logged a false "Trying to remove nonexisting" after a successful removal — the #2053 PHY-counterpoll symptom.

Importantly, no regression: `removeBulkStatsContext(vid)` is still called unconditionally (its result captured into a bool before the branch rather than short-circuited into a `||`), so the bulk-context cleanup side effect is preserved, and `cleanupCounterGroupMapping(vid)` still runs on every path. Clean, minimal, and avoids the obvious short-circuit trap.

One non-blocking follow-up: a `TestFlexCounter.cpp` case asserting "remove an object present only in `m_objectIdsMap` → no notice emitted" would be the durable guard, since the whole bug is that the two tracking structures can legitimately diverge. Not a blocker for a targeted log-gating fix. LGTM.

mrinal-arista:
> Hi @mrinal-arista, do we need this to 202605?

Hi @lolyu, yes, we need this fix on 202605 as well.

syncd/FlexCounter.cpp
mrinal-arista: Agreed, I will add a unit test for this.

PR #2067, opened 2026-09-01
vpp: interface registry to consolidate scattered interface maps

yue-fred-gao:
### Description of PR

Summary:
Consolidate VPP interface identity into VppInterfaceRegistry

### Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

The VPP backend kept interface identity in multiple side maps and helper paths (tap map, ifname map, bond/bd/swif maps, name derivation helpers), which made lifecycle handling fragile and hard to reason about.
The goal is to make one registry the source of truth for interface identity and lookups, while keeping compatibility writes for inherited base paths that still consult legacy maps.

##### Work item tracking

#### How did you do it?
1. Introducing interface identity model used by VppInterfaceRegistry.
###
Base attributes on every record:
- hwif name   (always present, primary key)
- osif name   (SONiC side name like Ethernet0, PortChannel0, Vlan100, Ethernet0.10)
- tap name    (optional, present only when vpp lcp creates the host tap)
- oid         (optional; PORT/LAG/VLAN identity)
- rif oid     (optional router interface oid)
- sw_if_index (optional VPP runtime interface index)
- bd_id       (optional bridge domain id)
###
Subclasses:
- VppPhysicalPort: front panel port identity seeded from ifmap.
- VppBondInterface: LAG identity with bond id and lcp-created state.
- VppSubInterface: parent-linked <parent>.<subId> interface identity.
- VppVlanInterface: VLAN BVI identity (bvi<VLAN>, Vlan<VLAN>).
###
Lookup coverage by subclass (through registry indexes):
- Physical/LAG/BVI: hwif + osif + optional tap + optional oid + optional sw_if_index.
- Sub-interface:    hwif + osif + optional tap + optional sw_if_index, and findSubIf(parentHwif, subId).

2. Introduced a dedicated interface identity model and registry for VPP.
3. Migrated VPP call sites to registry-based lookups for interface resolution.
4. Renamed resolvePortOid to resolveIfOid and made it generic so callers can apply explicit type checks where needed.
5. Moved OS interface name to VPP hardware interface resolution into the registry (including logical interface derivation and sub-interface suffix handling).
6. Removed legacy/dead helper wrappers that were replaced by registry lookups.
7. Kept legacy map writes on hostif create/remove only for compatibility with inherited base logic:

- setIfNameToPortId and setPortIdToTapName are still written.
- removeIfNameToPortId and removePortIdToTapName are still removed on teardown.
- VPP read paths now use the interface registry instead of those maps.
- Hardened ifmap parsing with bounded scanf widths.

8. Fixed a BFD string lifetime issue by keeping the hardware interface name in owning string scope.
Run full regression through PR https://github.com/sonic-net/sonic-buildimage/pull/27039
vpp

### DIFF
--- a/vslib/vpp/SwitchVppRif.cpp
+++ b/vslib/vpp/SwitchVppRif.cpp
@@ -434,8 +434,8 @@ bool SwitchVpp::getPortHwifNameFromLane(
         return false;
     }
 
-    const char *mapped_hwifname = tap_to_hwif_name(port_name.c_str());
-    if (mapped_hwifname == nullptr || strcmp(mapped_hwifname, "Unknown") == 0)
+    const char *mapped_hwifname = osif_name_to_hwif_name(port_name.c_str());
+    if (mapped_hwifname == nullptr)
     {
         SWSS_LOG_ERROR("port %s has no VPP mapping", port_name.c_str());
         return false;
@@ -452,6 +452,22 @@ bool SwitchVpp::getPortHwifNameFromLane(
     SWSS_LOG_INFO("resolved port %s lane set to %s/%s",
             sai_serialize_object_id(port_id).c_str(), port_name.c_str(),
             if_name.c_str());
+
+    /*
+     * The one place that holds all three parts of a physical port's identity at
+     * once -- the oid, the SONiC name from the lane set, and the hwif name from
+     * sonic_vpp_ifmap.ini -- and that has just confirmed the VPP interface
+     * really exists. Registering here rather than in createPort() is deliberate:
+     * create_ports() creates ports with no attributes at all and only later sets
+     * SAI_PORT_ATTR_HW_LANE_LIST, so a port has no derivable name inside
+     * createPort() and attempting it there would fail for every front panel
+     * port.
+     */
+    if (!m_ifaceRegistry.findByOid(port_id))
+    {
+        m_ifaceRegistry.addPhysicalPort(mapped_hwifname, port_name, port_id);
+    }
+
     return true;
 }
 
@@ -464,9 +480,32 @@ bool SwitchVpp::vpp_get_hwif_name (
 {
     SWSS_LOG_ENTER();
 
-    std::string tap_name;
+    std::string osif_name;
     std::string hwifname;
-    if (objectTypeQuery(object_id) == SAI_OBJECT_TYPE_PORT &&
+
+    sai_object_type_t ot = objectTypeQuery(object_id);
+
+    if (ot == SAI_OBJECT_TYPE_VLAN)
+    {
+        /*
+         * A VLAN is represented in VPP by its BVI, which is named after the
+         * VLAN id and has no host interface behind it, so it never goes
+         * through the osif name path below.
+         */
+        sai_attribute_t vlan_attr;
+
+        vlan_attr.id = SAI_VLAN_ATTR_VLAN_ID;
+
+        if (get(SAI_OBJECT_TYPE_VLAN, object_id, 1, &vlan_attr) != SAI_STATUS_SUCCESS)
+        {
+            SWSS_LOG_ERROR("SAI_VLAN_ATTR_VLAN_ID not found for %s",
+                    sai_serialize_object_id(object_id).c_str());
+            return false;
+        }
+
+        hwifname = std::string(BVI_PREFIX) + std::to_string(vlan_attr.value.u16);
+    }
+    else if (ot == SAI_OBJECT_TYPE_PORT &&
                 getPortHwifNameFromLane(object_id, attr_count, attr_list, hwifname))
     {
         SWSS_LOG_DEBUG("using lane-based VPP interface %s for port %s",
@@ -474,21 +513,53 @@ bool SwitchVpp::vpp_get_hwif_name (
     }
     else
     {
-        if (getTapNameFromPortOrLagId(object_id, tap_name) == false)
+        /*
+         * Registry first. Both kinds that reach here record their oid at the
+         * moment they acquire an identity -- addPhysicalPort() from the lane
+         * path above, addLag() from vpp_create_lag() -- so the registry answers
+         * straight from the oid without going through a name at all, and
+         * without needing get_lag_bond_info() to rebuild "PortChannel<id>"
+         * only for osif_name_to_hwif_name() to take it apart again.
+         */
+        auto rec = m_ifaceRegistry.findByOid(object_id);
+
+        if (rec)
         {
-            SWSS_LOG_ERROR("host interface for port/lag id %s not found",
-                    sai_serialize_object_id(object_id).c_str());
-            return false;
+            hwifname = rec->getHwifName();
         }
-
-        const char *mapped_hwifname = tap_to_hwif_name(tap_name.c_str());
-
-        if (mapped_hwifname == NULL || strcmp(mapped_hwifname, "Unknown") == 0)
+        else
         {
-            return false;
-        }
+            /*
+             * Legacy fallback, deliberately still here: the registry is
+             * in-memory and only populated by the create paths, so anything it
+             * has not seen must still resolve exactly as before. The WARN is
+             * the parity harness -- a VS run that never logs it is the signal
+             * that Phase 5 may delete this arm.
+             *
+             * osif_name_to_hwif_name() is keyed on the SONiC netdev name, so
+             * the LAG must be resolved to PortChannel<id> and not to its
+             * linux-cp tap be<id>.
+             */
+            if (getOsIfFromPortOrLagId(object_id, osif_name) == false)
+            {
+                SWSS_LOG_ERROR("host interface for port/lag id %s not found",
+                        sai_serialize_object_id(object_id).c_str());
+                return false;
+            }
 
-        hwifname = mapped_hwifname;
+            const char *mapped_hwifname = osif_name_to_hwif_name(osif_name.c_str());
+
+            if (mapped_hwifname == nullptr)
+            {
+                return false;
+            }
+
+            hwifname = mapped_hwifname;
+
+            SWSS_LOG_WARN("registry miss for %s, legacy path resolved %s to %s",
+                    sai_serialize_object_id(object_id).c_str(),
+                    osif_name.c_str(), hwifname.c_str());
+        }
     }
     if (vlan_id) {
         ifname = hwifname + "." + std::to_string(vlan_id);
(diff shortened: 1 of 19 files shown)


vslib/vpp/SwitchVppRif.cpp
aaronber0614, line 714: SwitchVpp::asyncIntfStateUpdate (SwitchVppRif.cpp ~713-731) now calls send_port_oper_status_notification(rec->getOid(), ...) for any registry record. A subinterface has no oid, so SAI_NULL_OBJECT_ID; a LAG/BVI returns a non-PORT oid. The old ifmap held only physical ports, so these events were ignored. Tagged-VLAN-member / LAG link up-down is a common path.
Possible Fix: gate to physical-port records with a real PORT oid (type==PORT and hasOid()); ignore subif/LAG/BVI.

_Co-authored with Copilot_

vslib/vpp/VppInterfaceRegistry.cpp
```diff
@@ -0,0 +1,964 @@
...
+
+    if (rec && rec->hasOsIf())
+    {
+        return rec->getOsIf();
+    }
+
+    return std::string();
+}
+
+std::string VppInterfaceRegistry::resolveHwIfName(
```
aaronber0614, line 759: New resolveHwIfName() (VppInterfaceRegistry.cpp:735) is a pure index lookup returning "" on miss; the old vpp_get_hwif_name() actively resolved via lane/tap. In hostif create (SwitchVppHostif.cpp:545-557), if findByHwif misses and getPortHwifNameFromLane fails (dump-cache miss), setTapName logs an error and returns false (return is ignored), so no registry record, no tap, but the legacy sonic_vpp_ifmap/tap maps were written = drift plus later resolve failures. setTapName requiring a pre-existing record is confirmed in source.

Possible Fix: on miss, addPhysicalPort(hwif, name, obj_id) from the ifname already in hand, then setTapName, and honor the bool. Confirm/doc that ACL/nbr/mirror/sflow/route/mpls never program a port before it registers, or add a lane fallback for PORT oids.

_Co-authored with CoPilot_

vslib/vpp/SwitchVppRif.cpp
```diff
@@ -1907,51 +1652,56 @@ sai_status_t SwitchVpp::vpp_create_router_interface(
...
+        /*
+         * For a port-channel sub-port the LCP host tap must be be<id>.<vlan>
+         * (a VLAN netdev on the be<id> bond tap), NOT PortChannel<id>.<vlan>:
+         * that name collides with the kernel 8021q netdev owned by the Linux
+         * bond/team stack. linux-cp-punt-xc lands on be<id>.<vlan> and the
+         * sonic_ext aggr-tap-redirect steers the punted copy to the
+         * originating member tap (SONiC PR #2440 §5.3). For a plain port the
+         * tap name is the SONiC name, so this is just Ethernet<n>.<vlan>.
+         */
+        std::string tap_name = m_ifaceRegistry.resolveTapName(obj_id);
```
aaronber0614, line 1703: PORT removal does not deregister from the registry, unlike LAG and sub-port.

SwitchVpp::remove dispatches SAI_OBJECT_TYPE_PORT to removePort here (SwitchVpp.cpp:1737), which is not overridden, so it runs SwitchStateBase::removePort (vslib/SwitchStateBase.cpp:368). That base removal never touches m_ifaceRegistry, so the physical port record stays behind. LAG removal calls removeByOid (SwitchVppFdb.cpp:1928) and sub-port removal calls remove (SwitchVppRif.cpp:2092), so this is the one interface class that is not cleaned up.

The stale record is not benign on re-add: addPhysicalPort now refuses a hwif that is already registered and returns nullptr (VppInterfaceRegistry.cpp:137-143), so a port remove followed by a re-add of the same hwif leaves the new port with no registry record, which then breaks resolveHwIfName and resolveTapName for it. The old map was an operator[] overwrite, so a re-seed was idempotent.

I do not think physical ports are removed at runtime on this platform today, so this is latent rather than live, but the asymmetry is cheap to close: override removePort in SwitchVpp to call m_ifaceRegistry.removeByOid(objectId) after the base removal succeeds, matching the LAG and sub-port paths. If you prefer to leave it, a short comment on why PORT removal is not a real path would help

_Co-authored with CoPilot_

vslib/vpp/SwitchVppRif.cpp
```diff
@@ -687,17 +715,16 @@ sai_status_t SwitchVpp::asyncIntfStateUpdate(const char *hwif_name, bool link_up
 {
     SWSS_LOG_ENTER();
 
-    std::string tap_str;
-    const char *tap;
-
-    tap = hwif_to_tap_name(hwif_name);
-    auto port_oid = getPortIdFromIfName(std::string(tap));
+    auto rec = m_ifaceRegistry.findByHwif(hwif_name);
```
lolyu: Net-new (relates to @aaronber0614's #1, same function): this touches `m_ifaceRegistry.findByHwif()` from the event thread. `asyncIntfStateUpdate` runs out of `vppProcessEvents` (SwitchVppRif.cpp:574), which executes on its own dedicated `m_vpp_thread` (started at :705), and I don't see `m_apimutex` acquired anywhere in that loop or here. But the registry header explicitly states (VppInterfaceRegistry.h:32-35) that it takes no lock of its own and may only be touched by "the main/command thread and the FDB aging thread" — the event thread is a third thread not on that list. The registry mutates `unordered_map`s of `shared_ptr` (rehash on `indexRecord`, erase on `remove`) from the main thread, so a concurrent link event here is a real data race, not just a stale read. To be fair the old code had the same lock-free access from this thread (`getPortIdFromIfName` on the legacy map), so it's not strictly worse than before — but the refactor introduces a written invariant that its own code violates. Either take `m_apimutex` on this path, or correct the header contract to include the event thread and explain why it's safe (e.g. these maps are quiescent by the time link events flow). Right now code and contract disagree.

vslib/vpp/SwitchVppRif.cpp
```diff
@@ -687,17 +715,16 @@ sai_status_t SwitchVpp::asyncIntfStateUpdate(const char *hwif_name, bool link_up
...
+    {
+        SWSS_LOG_NOTICE("No interface record for hwif %s. Ignore the update.", hwif_name);
         return SAI_STATUS_SUCCESS;
     }
 
+    const sai_object_id_t port_oid = rec->getOid();
+
     auto state = link_up ? SAI_PORT_OPER_STATUS_UP : SAI_PORT_OPER_STATUS_DOWN;
 
     send_port_oper_status_notification(port_oid, state, false);
```
lolyu, line 740: Confirming @aaronber0614's #1 with the before/after: on `origin/master` this path was `getPortIdFromIfName(tap)` which returned `SAI_NULL_OBJECT_ID` for non-physical-ports and the caller ignored the event. Now `rec->getOid()` returns null for a subif and a non-PORT oid for a LAG/BVI, and the notification fires regardless. Suggest gating: only emit when `rec->getType() == PHYSICAL_PORT && rec->hasOid()`; ignore subif/LAG/BVI so VLAN-member/LAG link edges don't generate spurious or mis-typed SAI_PORT_OPER_STATUS notifications.

lolyu (commented):
Verified this against source at head `87dca23` (cloned the branch, not just the diff). Strong refactor overall — collapsing the scattered tap/ifname/bond/bd/swif side-maps into one typed registry is the right direction and exactly the fragility source behind the tap-MAC bug family. Design is careful: `m_parent` is a `weak_ptr` (avoids cycles; `remove()` collects children before unindexing the parent), both `bindSwIfIndex` and `setTapName` proactively detach stale holders for VPP's recycled `sw_if_index` / recycled Linux netdev names, `by_oid` partiality is documented and enforced in `setOid`, and the threading contract is written down. The BFD string-lifetime fix (owning-string scope instead of a dangling `c_str()`) is a real correctness catch on its own.

I independently confirmed all three of @aaronber0614's findings against source, and they're all accurate:

1. **Oper-status for non-PORT records (SwitchVppRif.cpp:726-730) — confirmed, and it's a genuine regression.** I checked `origin/master`: the old `asyncIntfStateUpdate` called `getPortIdFromIfName(tap)` which returned `SAI_NULL_OBJECT_ID` for anything that wasn't a physical port → the event was *ignored*. The new code emits `send_port_oper_status_notification(rec->getOid(), ...)` for any record, so a subif (null oid) or LAG/BVI (non-PORT oid) now produces spurious/mis-typed notifications on a common path (tagged-VLAN-member / LAG link up-down). Gating to `type==PHYSICAL_PORT && hasOid()` is the right fix. This is the one I'd block on.
2. **Hostif-create drift (SwitchVppHostif.cpp:545-557) — confirmed.** If `findByHwif` misses and `getPortHwifNameFromLane` fails, `setTapName` runs on an unregistered hwif, returns false, and the return is ignored (line 557) — while the legacy maps were already written (537-538). Drift + later resolve failures. Fallback-to-`addPhysicalPort`-then-honor-the-bool is sound.
3. **PORT removal doesn't deregister (SwitchVpp.cpp:1737) — confirmed.** `removePort` isn't overridden in `SwitchVpp`; base `SwitchStateBase::removePort` never touches `m_ifaceRegistry`, whereas LAG (`removeByOid`) and sub-port (`remove`) both do. And it's not fully benign: `addPhysicalPort` now refuses an already-registered hwif, so remove→re-add of the same hwif leaves the new port with no record (old map was an idempotent `operator[]`). Latent today but cheap to close (override → `removeByOid` after base success) or worth a comment.

One net-new note inline that compounds #1 at the same call site (threading-contract violation). Flagging as Comment rather than a second Request Changes since @aaronber0614's review already stands — but I do think #1 should land before merge, and my threading note should be reconciled alongside it.

nhegde-microsoft:
Looks very good to me. Keeping map of SONiC<->VPP interface names/IDs has the inherent problem of information becoming stale in the event of unexpected VPP restart. I think a big advantage of this PR is that it makes the task of converging/syncing configuration between SONiC and VPP simpler and more efficient.

nhegde-microsoft (dismissed):

dypet (dismissed):

vslib/vpp/SwitchVppRif.cpp
yue-fred-gao: will do

vslib/vpp/VppInterfaceRegistry.cpp
yue-fred-gao: create hostIf is after port creation (it has port OID). When port is created, it registers the hwif to oid mapping so findByHwif should work. If that indeep misses the entry, there must be a bug that needs to be fixed.

vslib/vpp/SwitchVppRif.cpp
yue-fred-gao: added the suggested check on physical port

aaronber0614 (dismissed):
LGTM

dypet (dismissed):

dypet (approved):

aaronber0614 (approved):

PR #1766, opened 2026-02-11
[test] Fix flaky FlexCounter.bulkChunksize by replacing usleep with poll-wait

rustiqly:
## Description

Fix flaky `FlexCounter.bulkChunksize` unit test that intermittently fails in CI due to timing-dependent `usleep(1000*1050)`.

**Issue:** #1765

## Root Cause

The test uses `usleep(1000*1050)` (1.05s) to wait for the FlexCounter polling thread (1s poll interval) to complete counter collection. Under CI load, the polling thread may not finish within this window, causing assertion failures like:

```
TestFlexCounter.cpp:1390: Failure
Expected equality of these values:
  object_count
    Which is: 6
  unifiedBulkChunkSize
    Which is: 3
```

This failure reproduces on completely unrelated PRs ([#1763](https://github.com/sonic-net/sonic-sairedis/pull/1763), [#1764](https://github.com/sonic-net/sonic-sairedis/pull/1764)), confirming it is timing-dependent and not caused by code changes.

## Fix

Replace all `usleep(1000*1050)` calls with deterministic poll-wait helpers:

- **`waitForCounterKeys(table, expectedCount)`** — polls until the expected number of counter keys appear in COUNTERS_DB
- **`waitForCounterValue(table, key, field)`** — polls until a specific counter field has a non-empty value

Both helpers poll every 100ms with a 5-second timeout. This eliminates timing sensitivity while keeping tests fast on unloaded machines (typically completes in 1-2 polls).

Also replaces the similar `usleep(1000*1000)` and `usleep(1000*2000)` in `queryCounterCapability` / `addRemoveCounterForPort` which have the same timing vulnerability.

## Changes

- `unittest/syncd/TestFlexCounter.cpp`:
  - Add `#include <chrono>` for `steady_clock`
  - Add `waitForCounterKeys()` helper
  - Add `waitForCounterValue()` helper  
  - Replace 8x `usleep(1000*1050)`, 1x `usleep(1000*1000)`, 1x `usleep(1000*2000)` across `testAddRemoveCounter`, `FlexCounter.counterIdChange`, `FlexCounter.addRemoveCounterForPort`, and `testDashMeterAddRemoveCounter`
  - Retain `usleep(60*1000)` in `updateSwitchDebugCounterIdList` (50ms poll interval, not flaky)

PR #2069, opened 2026-09-02
[syncd]: Reject empty warm boot state and fall back to cold start

### LINKED ISSUE #29310: Bug: [warm-reboot] syncd restart during pre-shutdown still kexecs warm and leaves the dataplane dead
### Is it platform specific

generic

### Importance or Severity

Critical

### Description of the bug

If `syncd` is killed while `warm-reboot` is waiting for pre-shutdown to
complete, the switch still reboots in warm mode and comes back with a dead
dataplane. No traffic is forwarded, and no Ethernet host interfaces are
created. The CLI can still report ports as oper-up, because that state is
read from the Redis snapshot restored from `dump.rdb`, while the neighbors
on the other side of the links see them as down.

The system never falls back to a cold boot on its own. Only an operator
cold reboot brings the switch back. Killing `syncd` again during the warm
startup of the next life leads to the same stuck state.

This is not a missed branch in the existing swss/syncd failover logic.
SONiC currently has no mechanism that detects incomplete warm state and
converts the reboot into a cold one.

### Investigation and suspected root cause

The sequence below comes from a techsupport dump of a failing test and
from a live reproduction on the same image.

1. `warm-reboot` enables warm restart very early. It sets
   `WARM_RESTART_ENABLE_TABLE|system` to `true` during its setup phase,
   roughly 35 seconds before pre-shutdown is even requested, so the flag
   is already committed before any ASIC state has been saved.
2. The test kills the container (`docker kill syncd`) once `warm-reboot`
   logs that it is waiting for pre-shutdown.
3. systemd restarts `syncd.service` in the *old* life, as a scheduled
   restart. Nothing inhibits that restart while `warm-reboot` is still
   polling for the pre-shutdown result.
4. Because the kill happens one or two seconds after pre-shutdown was
   requested, SAI never finishes writing the warm state. The warmboot
   directory itself is still present, either empty or holding leftover
   files from an earlier successful warm boot.
5. The restarted `syncd` sees the warm restart flag, so
   `performStartupLogic` overrides the start type to
   `SAI_START_TYPE_WARM_BOOT`. The only condition that forces cold is a
   failing `access(SAI_WARM_BOOT_READ_FILE, F_OK)`. Since that path is a
   directory, a directory that merely exists — even an empty one — is
   enough to keep the start type warm. This was confirmed live.
6. The warm restore then fails, sairedis throws, and the process enters
   shutdown-wait mode: it stays alive and only answers restart queries.
   It does not retry as cold, does not clear the warm restart flag, and
   does not exit.
7. On the `warm-reboot` side, `wait_for_pre_shutdown_complete_or_fail`
   times out with the state still at `requesting`. That failure is only
   logged; the script continues. The parked `syncd` answers the following
   warm shutdown request in microseconds without ever having created a
   switch object, so STATE_DB ends up recording
   `warm-shutdown-succeeded`, and the box kexecs with
   `SONIC_BOOT_TYPE=warm`.
8. In the new life the same thing happens again: warm restore fails and
   `syncd` parks. This time orchagent's `INIT_VIEW` is answered
   immediately with a failure rather than timing out, and orchagent
   aborts.

The failover between swss and syncd only works in one direction, which is
why nothing recovers from step 8. The watchdog started for swss is
`docker-wait-any-rs -s swss -d syncd teamd`, so when the `syncd` container
exits, swss is torn down with it. The reverse is not true: when swss or
orchagent dies, `syncd` is not restarted. `syncd.service` only declares
`After=swss`, and during a warm shutdown `swss.sh stop` deliberately just
kills the swss container and leaves `syncd` running. A `syncd` that is
alive but parked is the worst case for this design, because it looks
healthy to systemd and to the watchdog while it can no longer program the
ASIC. The fact that orchagent then hangs in `do_coredump` is a secondary
problem: even a clean abort would not push `syncd` onto a cold path.

The live reproduction confirmed the file-check part of this. With an empty
warmboot directory and the kill issued at pre-shutdown, the start type was
still warm, `syncd` parked, the box kexec'd warm, and no Ethernet host
interfaces came up.

Handling this inside SAI alone does not fix it. Even if the SAI
implementation were to fall back to its own cold initialization and return
success, sairedis still expects the switch RID from the previous life and
still constructs `SaiSwitch` in warm mode. Deleting the warmboot directory
from SAI would at most let the *next* `syncd` process take the cold branch
of the `access()` check, while Redis, orchagent, the restored `dump.rdb`
and the kexec boot type would all still be warm. That mixed state is not a
mode the product is tested in.

There is also no reliable automatic recovery if `syncd` is killed during
the warm startup of the next life, since it parks in exactly the same way.
A cold reboot by the operator does recover the switch. Manually restarting
`syncd` or swss only helps after the warm restart flag has been cleared by
the finalizer, and the finalizer does not restart `syncd` itself.

#### Proposed direction

1. **sairedis:** when a warm `create_switch` or state restore fails, do
   not just park in shutdown-wait mode. Either fall back to cold by
   dropping the stale VID/RID mappings and reinitializing, or exit so that
   a restart can re-read the flags. A warmboot directory that exists but
   is empty should not be accepted as valid warm state.
2. **warm-reboot:** if pre-shutdown did not succeed, do not kexec in warm
   mode. Either abort the reboot, or continue as a cold boot by clearing
   the warm restart flag and skipping the warm Redis restore.
3. **Flag ordering:** set `WARM_RESTART_ENABLE_TABLE` only after the state
   has actually been saved, and clear it again if pre-shutdown fails.

### Steps to Reproduce

1. Start from a DUT with a working dataplane, forwarding traffic.
2. Run `warm-reboot -v`.
3. From a second session, as soon as the log shows that it is waiting for
   pre-shutdown, run `docker kill syncd`.
4. Let the box kexec into the next life.
5. Check the result: whether any Ethernet netdevs exist (`ip link`), how
   the neighbor or traffic generator sees the link state compared to
   `show interface status`, whether `syncd` is running normally or parked
   in restart-query-only mode, and whether orchagent aborted.

The same kill issued after a successful kexec, during the warm startup of
the next life, leaves the switch stuck in the same way, with no Ethernet
interfaces.

### Actual Behavior:
`syncd` is restarted in the middle of pre-shutdown, so no warm state is
ever saved, while the warm restart flag is already set. The restore fails
and `syncd` parks instead of falling back to cold. `warm-reboot` only logs
the pre-shutdown failure and still kexecs in warm mode, and the next life
repeats the same failure. Orchagent aborts, and because a parked `syncd`
is not seen as a crashed `syncd`, the swss failover does not recover the
ASIC. The dataplane stays down, while the CLI still shows stale oper-up
state from Redis. No automatic cold fallback runs at any point.

### Expected Behavior:
The system should detect that the warm ASIC state is missing or incomplete
and react by falling back to a cold boot or by aborting the warm reboot.
This applies whether `syncd` was killed during pre-shutdown, the warm
state files were never written, or the warm restore failed at startup. In
all these cases the dataplane should not be left permanently down.

### Relevant log output

```shell
# failing run - syncd is killed while warm-reboot waits for pre-shutdown
NOTICE root: Waiting for pre-shutdown ...
INFO   systemd: syncd.service: Deactivated successfully.
INFO   systemd: syncd.service: Scheduled restart job, restart counter is at 1.
NOTICE root: Warm boot flag: syncd true.

# the start type stays warm because the warmboot directory still exists
WARNING syncd#syncd: override command line startType=cold via SAI_START_TYPE_WARM_BOOT
NOTICE  syncd#syncd: using warmBootReadFile: '/var/warmboot/...'

# the warm restore fails and syncd parks instead of retrying as cold
ERR syncd#SDK: Failed to initialize switch
ERR syncd#SDK: Runtime error during syncd init: failed to create switch RID
NOTICE syncd#SDK: starting main loop, ONLY restart query

# warm-reboot logs the pre-shutdown failure but still reboots warm
NOTICE root: Syncd pre-shutdown failed, state: requesting ...
NOTICE syncd#SDK: handleRestartQuery: received WARM switch shutdown event
NOTICE root: Rebooting with /sbin/kexec -e ...    # SONIC_BOOT_TYPE=warm

# next life - syncd parks again, so INIT_VIEW is answered with an immediate failure
WARNING syncd: processEventInShutdownWaitMode: op=notify, key=INIT_VIEW
ERR swss#orchagent: Aborting orchagent due to critical SAI API failure...
```

* N/A

### Attach files (if any)

* N/A

PRIEYAN:
### Description of PR

Summary:
Fixes sonic-net/sonic-buildimage#29310

`performStartupLogic` guards the warm start path with `access(warmBootReadFile, F_OK)`, which only tests that the path exists. Several platforms point `SAI_WARM_BOOT_READ_FILE` at a directory such as `/var/warmboot/`, and a directory that exists but holds no warm state passes that check, so the start type stays `SAI_START_TYPE_WARM_BOOT`. An empty regular file passes it too.

That is what the linked issue hits. If syncd is restarted while `warm-reboot` is still waiting for pre-shutdown to complete, `WARM_RESTART_ENABLE_TABLE|system` is already set but SAI never finished writing the warm state, so the warmboot directory is empty or holds leftovers from an earlier warm boot. The restarted syncd still picks warm boot, the restore fails, and syncd parks in shutdown-wait mode rather than coming back cold.

This is a partial fix for the issue. It addresses the start type selection only. It does not change what happens when a warm restore fails for some other reason, where syncd still enters `inShutdownWaitMode` without a cold retry, and it does not touch the `warm-reboot` side, which lives in sonic-utilities and still kexecs warm after a failed pre-shutdown.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

A syncd that selects warm boot without any saved warm state cannot recover on its own, and the box ends up with a dead dataplane that only an operator cold reboot clears. Refusing the warm start when the state is not actually there keeps that failure inside a cold start, which is recoverable.

##### Work item tracking

#### How did you do it?

Replaced the `access(..., F_OK)` call with a `hasWarmBootState()` helper. It stats the path, requires a directory to contain at least one entry other than `.` and `..`, and requires a regular file to be non-zero length. Anything else logs a warning and returns false, which takes the existing cold start branch. The surrounding control flow is unchanged.

Added five cases to `unittest/syncd/TestSyncd.cpp` covering a missing path, an empty directory, an empty file, a populated file and a directory holding a file.

I could not build the full sairedis suite locally, `libswsscommon` and the SAI headers are not available on this machine, so CI is the first real run of the gtest suite here. What I did run was the extracted helper and the new test block compiled standalone under `-Wall -Wextra -Werror`, all five assertions passing, and I confirmed separately that `access()` returns 0 on an empty directory, which is the behaviour this change relies on being wrong. `tests/swsslogentercheck.pl` is clean on the three edited files and the diff adds no trailing whitespace or tabs.

I do not have a DUT, so the end to end warm-reboot reproduction from the issue is not verified by me.

Only affects platforms where the configured warm boot read path is a directory, or where the warm state file is present but empty. Platforms that write a normal non-empty file behave exactly as before.

### Documentation

No, this is a bug fix with no user facing interface change.

### DIFF
--- a/syncd/Syncd.cpp
+++ b/syncd/Syncd.cpp
@@ -38,6 +38,10 @@
 
 #include <unistd.h>
 #include <inttypes.h>
+#include <dirent.h>
+#include <errno.h>
+#include <string.h>
+#include <sys/stat.h>
 
 #include <iterator>
 #include <algorithm>
@@ -322,6 +326,64 @@ Syncd::~Syncd()
     stopDampingTimerThread();
 }
 
+bool Syncd::hasWarmBootState(
+        _In_ const char* warmBootReadFile)
+{
+    SWSS_LOG_ENTER();
+
+    struct stat st;
+
+    if (stat(warmBootReadFile, &st) == -1)
+    {
+        SWSS_LOG_WARN("failed to stat warmBootReadFile '%s': %s", warmBootReadFile, strerror(errno));
+
+        return false;
+    }
+
+    if (S_ISDIR(st.st_mode))
+    {
+        DIR* dir = opendir(warmBootReadFile);
+
+        if (dir == NULL)
+        {
+            SWSS_LOG_WARN("failed to open warmBootReadFile directory '%s': %s", warmBootReadFile, strerror(errno));
+
+            return false;
+        }
+
+        bool hasEntry = false;
+
+        while (const struct dirent* entry = readdir(dir))
+        {
+            if (strcmp(entry->d_name, ".") == 0 || strcmp(entry->d_name, "..") == 0)
+            {
+                continue;
+            }
+
+            hasEntry = true;
+            break;
+        }
+
+        closedir(dir);
+
+        if (!hasEntry)
+        {
+            SWSS_LOG_WARN("warmBootReadFile directory '%s' is empty", warmBootReadFile);
+        }
+
+        return hasEntry;
+    }
+
+    if (st.st_size == 0)
+    {
+        SWSS_LOG_WARN("warmBootReadFile '%s' is empty", warmBootReadFile);
+
+        return false;
+    }
+
+    return true;
+}
+
 void Syncd::performStartupLogic()
 {
     SWSS_LOG_ENTER();
@@ -343,9 +405,9 @@ void Syncd::performStartupLogic()
 
         SWSS_LOG_NOTICE("using warmBootReadFile: '%s'", warmBootReadFile);
 
-        if (warmBootReadFile == NULL || access(warmBootReadFile, F_OK) == -1)
+        if (warmBootReadFile == NULL || !hasWarmBootState(warmBootReadFile))
         {
-            SWSS_LOG_WARN("user requested warmStart but warmBootReadFile is not specified or not accessible, forcing cold start");
+            SWSS_LOG_WARN("user requested warmStart but warmBootReadFile is not specified or holds no warm boot state, forcing cold start");
 
             m_commandLineOptions->m_startType = SAI_START_TYPE_COLD_BOOT;
         }
--- a/syncd/Syncd.h
+++ b/syncd/Syncd.h
@@ -34,6 +34,7 @@
 
 class SyncdTest;
 class SyncdLinkEventDampingTest;
+class SyncdWarmBootStateTest;
 
 namespace syncd
 {
@@ -41,6 +42,7 @@ namespace syncd
     {
         friend class ::SyncdTest;
         friend class ::SyncdLinkEventDampingTest;
+        friend class ::SyncdWarmBootStateTest;
 
         private:
 
@@ -103,6 +105,9 @@ namespace syncd
                     _Out_ const char** value);
 
             void performStartupLogic();
+
+            static bool hasWarmBootState(
+                    _In_ const char* warmBootReadFile);
 
             void sendShutdownRequest(
                     _In_ sai_object_id_t switchVid);
--- a/unittest/syncd/TestSyncd.cpp
+++ b/unittest/syncd/TestSyncd.cpp
@@ -24,6 +24,8 @@
 
 #include <chrono>
 #include <thread>
+#include <fstream>
+#include <cstdlib>
 
 #include "swss/table.h"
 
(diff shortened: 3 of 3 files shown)


PRIEYAN:
the red check here is `FlexCounter.bulkChunksize`, which isn't in anything this pr touches. 179 of the 180 tests passed and the five `SyncdWarmBootStateTest` cases i added all came back OK, so the change compiles and the cold fallback behaves the way it's meant to.

that test has form for this — #1766 was literally "fix flaky FlexCounter.bulkChunksize", and #1887 added a re-probe skip after it kept firing on unrelated prs on both master and 202511. this run died at `TestFlexCounter.cpp:1622`, 30 times:

```
TestFlexCounter.cpp:1622: Failure
Expected equality of these values:
  object_count
    Which is: 6
  unifiedBulkChunkSize
    Which is: 3
```

reading it, i think it's the `bulkChunkSizeAfterPort` path rather than the re-probe case #1887 covered. the ports get added and the poll thread starts before `addCounterPlugin` applies the chunk size, so a poll landing in that gap sees all six objects in one unbounded chunk with all five counters, while the test set `unifiedBulkChunkSize = 3` before the call ever started. `initialCheckCount` is a fixed budget of 6 and won't absorb an extra real poll slipping in there on a slow agent.

could someone re-run the failed job? i'd rather not paper over that assertion myself — skipping `object_count == 6` would also swallow the case where the chunk size genuinely never gets applied, and that seems like the maintainers' call not mine.

PR #2071, opened 2026-09-05
[syncd] read attribute counter groups with one bulk get

pinky-nexthop:
#### Summary

Two changes to the flex counter path, both aimed at the poll cadence of attribute
counter groups.

**1. Read attribute groups with one bulk get.** `VendorSai::get()` takes the
process-wide mutex on every call, so an attribute group polling N objects takes
the lock N times per sweep and interleaves with any other group holding it.
Collecting the group in a single `bulkGet` takes the lock once. Falls back to the
per-object path when the vendor SAI does not implement the bulk entry point, so
it cannot regress an implementation that lacks it.

**2. Hand the poll cycle start and collection time to plugins.** A plugin can
only read a clock when it runs, which is after collection, so an interval it
measures itself carries the change in collection time and can read shorter than
the configured one. `ARGV[5]` is the monotonic cycle start, taken before
collection, so differencing it gives the true period. `ARGV[6]` is the collection
time, the term that scales with polled object count. Both are additive; the
secondary poll factor moves to a fixed `ARGV[4]` so every position is indexable.

The first bounds how long a sweep holds the lock. The second makes the resulting
cadence measurable, which is what shows whether a given poll interval is
sustainable at a given scale.

The plugin side that consumes `ARGV[5]`/`ARGV[6]` is sonic-net/sonic-swss#4881.
Both arguments are additive and a plugin falls back when they are absent, so the
two can merge in either order.

#### Why I did it

`VendorSai::get()` takes the process-wide `m_apimutex` on every call, while `getStats()`/`getStatsExt()` deliberately do not. Attribute counter groups are therefore the only flex counter groups that contend for that lock.

On a 256 port box the PFC watchdog reads `SAI_QUEUE_ATTR_PAUSE_STATUS` on every lossless queue at its poll interval — roughly 512 acquisitions per sweep — while the two port PHY attribute groups hold the same lock across per-lane SerDes firmware reads. The mutex is not fair, so the fast group advances about one object per turn and its sweep stretches toward the length of the slow one. We observed a 2.6 s effective poll interval against a configured 100 ms, recurring once per PHY sweep, which makes PFC storm detection unbounded against a 200 ms detection time.

#### How I did it

Collect the whole group in a single `bulkGet` instead of one `get()` per object. The lock is then taken once, which bounds the wait at one in-flight foreign get rather than a whole foreign sweep. The hardware work is unchanged; only the contention goes away.

`bulkGet` gains a `SAI_OBJECT_TYPE_QUEUE` case alongside the existing `SAI_OBJECT_TYPE_PORT` one. A vendor that does not implement the bulk entry point returns `SAI_STATUS_NOT_SUPPORTED` through the existing null check; `AttrContext` latches that once and keeps using the per-object path, so this is **inert until the vendor SAI provides it** and cannot regress an implementation that lacks it.

#### How to verify it

Measured on a 256 port Broadcom TH6 box at a 100 ms watchdog interval, against a SAI providing `get_queues_attribute`:

| | PHY attr groups enabled | disabled |
|---|---|---|
| before | 2181–2797 ms | 271–304 ms |
| after | **415 ms** | 301 ms |

Worst watchdog poll drops ~6x, and the penalty the PHY groups impose on it goes from ~9.2x to 1.38x. The PHY-disabled column is essentially unchanged, which is what shows the gain is removed lock contention rather than faster collection.

Existing unit tests pass; the fallback path is exercised by any vendor without the entry point.

### DIFF
--- a/syncd/FlexCounter.cpp
+++ b/syncd/FlexCounter.cpp
@@ -2329,6 +2329,11 @@ public:
     {
         SWSS_LOG_ENTER();
 
+        if (m_bulkGetSupported && collectDataBulk(countersTable))
+        {
+            return;
+        }
+
         for (const auto &kv : Base::m_objectIdsMap)
         {
             const auto &vid = kv.first;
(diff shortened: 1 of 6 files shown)


pinky-nexthop:
@vmittal-msft @bingwang-ms  Kumaresh kindly review.

syncd/FlexCounter.cpp
```diff
@@ -2355,18 +2360,134 @@ class AttrContext : public CounterContext<AttrType>
...
+        sai_status_t status = Base::m_vendorSai->bulkGet(
+                Base::m_objectType,
+                static_cast<uint32_t>(count),
+                rids.data(),
+                attrCounts.data(),
+                attrList.data(),
+                SAI_BULK_OP_ERROR_MODE_IGNORE_ERROR,
+                statuses.data());
+
+        if (status == SAI_STATUS_NOT_IMPLEMENTED || status == SAI_STATUS_NOT_SUPPORTED)
```
kperumalbfn, line 2461: How about other error status like SAI_STATUS_FAILURE for the above bulkGet query? We should still return failure before processing the bulk statuses

pinky-nexthop: done.

kperumalbfn (dismissed):

kperumalbfn:
@pinky-nexthop Please resolve the conflicts

kperumalbfn (approved):

pinky-nexthop:
@lolyu @yejianquan  kindly merge. thanks.

kperumalbfn:
@tjchadaga Could you merge

PR #2076, opened 2026-09-14
[vslib] Everflow mirroring

taran9999:
### Description of PR

- Add ERSPAN mirroring to create/removeMirrorSession backed by gre tunnel
- Extend MirrorSessionInfo struct with gre tunnel and monitor pin fields
- Support SAI_ACL_ENTRY_ATTR_ACTION_MIRROR_INGRESS/EGRESS action
- Add acl_table_get_ip_version to avoid duplicating address-less mirror rules (supports MIRROR/MIRRORv6 tables as separate tables)
- Add gre tunnel and sonic-ext plugin bindings and types to SaiVppXlate
- Maintain oid/neighbour IP/MAC map in SwitchVppNbr to resolve ERSPAN monitor port nexthop from DST_MAC
- Move getLagFromPort to protected scope in SwitchStateBase to use in mirror monitor resolution path

Summary:
Addresses sonic-net/sonic-buildimage#25780 and sonic-net/sonic-buildimage#26047. Implements the feature described in Phase II in this HLD: https://github.com/yue-fred-gao/SONiC/blob/13c854acd86063cd9e11beacd2945515f0b28ac6/doc/vpp/vpp-mirror.md

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

Enable Everflow feature for dataplane testing.

Passes 8/8 everflow/test_everflow_per_interface.py tests on both t0 and t1-lag topologies.
Passes 48/48 applicable everflow/test_everflow_testbed.py tests (EgressAclIngressMirror class skipped, dscp_with_policer tests skipped due to out of scope, frwd_with_bkg_trf skipped due to an open issue) on t0 and t1-lag topologies.

Skips are pending merge from these PRs:
- sonic-net/sonic-mgmt#27795
- sonic-net/sonic-mgmt#27797

IngressAclEgressMirror tests require dest route configuration skip to be removed, which is pending in this PR:
- sonic-net/sonic-mgmt#27796

VPP patches, sonic-ext additions, and plugin additions are required, added in this PR:
- sonic-net/sonic-platform-vpp#283


nhegde-microsoft (dismissed):

aaronber0614:
Review note (blocking):

1. The ERSPAN monitor pin currently installs the collector destination /32 or /128 in VRF 0 with is_multipath=false. This uses the same VPP API FIB source as the ordinary SAI route: adding the pin replaces the existing exact route, and removing the pin deletes that route without restoring it. Two mirror sessions using the same collector also share this prefix, so removing either session can break the other. Please give the pin independent ownership, such as a dedicated outer FIB, or preserve and restore the existing route with shared lifetime management.

2. The pin update is also reported as successful when route programming fails. applyErspanMonitor() sets monitor_pinned before calling pinErspanMonitor(), ignores both removal and addition failures, and setMirrorSession() persists the attribute regardless. Please propagate these failures and update the stored state only after the dataplane transition succeeds.

3. Finally, m_port_neighbor_mac is maintained only during neighbor create/remove. SAI_NEIGHBOR_ENTRY_ATTR_DST_MAC_ADDRESS updates currently bypass this index. When MirrorOrch subsequently updates the mirror session DST MAC, resolution uses the stale MAC, removes the working pin, and falls back to FIB/ECMP forwarding. Please handle neighbor MAC SET by updating both the VPP adjacency and this index.

The current Build and BuildAsan failures are caused by the expected dependency on the new gre_tunnel_add_del_v3 API from sonic-net/sonic-platform-vpp#283. This PR should be rerun after that dependency lands and its VPP packages are available.

Reviewed by AI agent on behalf of @aaronber0614.

aaronber0614 (changes requested):
Review note (blocking):

The ERSPAN monitor pin currently installs the collector destination /32 or /128 in VRF 0 with is_multipath=false. This uses the same VPP API FIB source as the ordinary SAI route: adding the pin replaces the existing exact route, and removing the pin deletes that route without restoring it. Two mirror sessions using the same collector also share this prefix, so removing either session can break the other. Please give the pin independent ownership, such as a dedicated outer FIB, or preserve and restore the existing route with shared lifetime management.

The pin update is also reported as successful when route programming fails. applyErspanMonitor() sets monitor_pinned before calling pinErspanMonitor(), ignores both removal and addition failures, and setMirrorSession() persists the attribute regardless. Please propagate these failures and update the stored state only after the dataplane transition succeeds.

Finally, m_port_neighbor_mac is maintained only during neighbor create/remove. SAI_NEIGHBOR_ENTRY_ATTR_DST_MAC_ADDRESS updates currently bypass this index. When MirrorOrch subsequently updates the mirror session DST MAC, resolution uses the stale MAC, removes the working pin, and falls back to FIB/ECMP forwarding. Please handle neighbor MAC SET by updating both the VPP adjacency and this index.

The current Build and BuildAsan failures are caused by the expected dependency on the new gre_tunnel_add_del_v3 API from https://github.com/sonic-net/sonic-platform-vpp/pull/283. This PR should be rerun after that dependency lands and its VPP packages are available.

Reviewed by AI agent on behalf of @aaronber0614.

vslib/vpp/SwitchVppMirror.cpp
```diff
@@ -81,6 +213,34 @@ sai_status_t SwitchVpp::removeMirrorSession(
...
+        tunnel.outer_table_id = 0;
+
+        uint32_t sw_if_index = 0;
+        int ret = vpp_gre_tunnel_add_del(&tunnel, false, &sw_if_index);
+        if(ret != 0) {
+            // Not usefully retryable, so drop the entry anyway rather than leak.
+            SWSS_LOG_ERROR("Failed to remove GRE tunnel for ERSPAN session, ret=%d; freeing session id and dropping entry anyway", ret);
+        }
+
+        // After the delete attempt, so a re-alloc cannot collide with a live tunnel.
```
dypet: Is this guaranteed? This is still exectued after a failed delete from vpp_gre_tunnel_add_del

vslib/vpp/SwitchVppMirror.cpp
```diff
@@ -123,6 +283,178 @@ sai_status_t SwitchVpp::removeMirrorSession(
...
+        }
+        else if (attr->id == SAI_MIRROR_SESSION_ATTR_DST_MAC_ADDRESS)
+        {
+            // Neighbor MAC changed; re-resolve against the current monitor port.
+            applyErspanMonitor(info, info.monitor_port, attr->value.mac);
+            SWSS_LOG_NOTICE("ERSPAN session %s dst mac updated", sid.c_str());
+        }
+    }
+
+    return set_internal(SAI_OBJECT_TYPE_MIRROR_SESSION, sid, attr);
```
dypet: Does set_internal() update or recreate the VPP dataplane tunnel, or does it only update the cached SAI state?

taran9999:
@aaronber0614 , I've addressed your items in 83aec4c237eb80087620daf1b2e01e74581fdb61. Please review

aaronber0614 (dismissed):
LGTM

PR #2068, opened 2026-09-02
[syncd]: Drop host_tx_ready ntf when port RID is not mapped yet

nazariig:
Signed-off-by: Nazarii Hnydyn <nazariig@nvidia.com>

## The bug

`syncd` can `SIGABRT` during warm-reboot when a `PORT_HOST_TX_READY`
notification arrives after switch create and before PORT objects exist.

`NotificationProcessor::process_on_port_host_tx_ready_change()` translated
SAI RIDs with `translateRidToVid(rid, SAI_NULL_OBJECT_ID)`. That API is a
mutating discovery helper: on a miss it calls
`allocateNewObjectId(object_type, switchVid)`. The second argument is switch
**context** for allocating a VID, not an in-place destination.
`SAI_NULL_OBJECT_ID` is an invalid context, so a missing port RID throws:

```text
allocateNewObjectId: object type of switch oid:0x0 is SAI_OBJECT_TYPE_NULL,
should be SWITCH
```

uncaught in the notification thread → `std::terminate` → `SIGABRT`.

Port RID→VID is only inserted in `SaiSwitch::onPostPortsCreate`. Hardware
links stay up across warm-reboot, so a flap in that window is a real SAI
RID that sairedis has not mapped yet. The sibling
`process_on_port_state_change()` already uses read-only
`tryTranslateRidToVid()` and only WARNs on a miss.

## Why dropping the notification is safe

orchagent does not depend on that one-shot edge for the starting picture.
After PORT create it GETs `SAI_PORT_ATTR_HOST_TX_READY_STATUS`
(`initializePortHostTxReadyBulk`). At warm restore end it GETs
`SAI_PORT_ATTR_OPER_STATUS` (`refreshPortStatus`). A later flap still
delivers a mapped-RID notification.

## The fix

Use `tryTranslateRidToVid()` for both port and switch. On a miss, WARN and
return without allocating a VID or calling `sendNotification`.

```cpp
if (!m_translator->tryTranslateRidToVid(port_id, port_vid))
{
    SWSS_LOG_WARN("Port RID %s translated to null VID!!!",
            sai_serialize_object_id(port_id).c_str());
    return;
}

if (!m_translator->tryTranslateRidToVid(switch_id, switch_vid))
{
    SWSS_LOG_WARN("Switch RID %s translated to null VID!!!",
            sai_serialize_object_id(switch_id).c_str());
    return;
}
```

Unit test: unknown port RID and unknown switch RID must not throw
[...]

PR #2079, opened 2026-09-15
syncd: Skip VID-to-RID translation for FlexCounter removal

### LINKED ISSUE #29260: Bug:[Broadcom TH5 CPO][processFlexCounterEvent] Syncd error log when config reload
### Is it platform specific

broadcom

### Importance or Severity

Medium

### Description of the bug

flex counter error log when config reload
```
2026 Aug 31 13:36:04.846891 M2-W6940-64X1-FR4 ERR syncd#syncd: :- processFlexCounterEvent: port VID oid:0x57000000000465, was not found (probably port was removed/splitted) and will remove from counters now
2026 Aug 31 13:36:05.184476 M2-W6940-64X1-FR4 ERR syncd#syncd: :- processFlexCounterEvent: port VID oid:0x570000000007ad, was not found (probably port was removed/splitted) and will remove from counters now
2026 Aug 31 13:36:05.488881 M2-W6940-64X1-FR4 ERR syncd#syncd: :- processFlexCounterEvent: port VID oid:0x570000000006b8, was not found (probably port was removed/splitted) and will remove from counters now
2026 Aug 31 13:36:05.807262 M2-W6940-64X1-FR4 ERR syncd#syncd: :- processFlexCounterEvent: port VID oid:0x570000000005a0, was not found (probably port was removed/splitted) and will remove from counters now
2026 Aug 31 13:36:06.119552 M2-W6940-64X1-FR4 ERR syncd#syncd: :- processFlexCounterEvent: port VID oid:0x5700000000090b, was not found (probably port was removed/splitted) and will remove from counters now
2026 Aug 31 13:36:06.446704 M2-W6940-64X1-FR4 ERR syncd#syncd: :- processFlexCounterEvent: port VID oid:0x570000000001ef, was not found (probably port was removed/splitted) and will remove from counters now
2026 Aug 31 13:36:06.782352 M2-W6940-64X1-FR4 ERR syncd#syncd: :- processFlexCounterEvent: port VID oid:0x570000000002c1, was not found (probably port was removed/splitted) and will remove from counters now
2026 Aug 31 13:36:07.390039 M2-W6940-64X1-FR4 ERR syncd#syncd: :- processFlexCounterEvent: port VID oid:0x570000000001a9, was not found (probably port was removed/splitted) and will remove from counters now
2026 Aug 31 13:36:08.947806 M2-W6940-64X1-FR4 ERR syncd#syncd: :- processFlexCounterEvent: port VID oid:0x57000000000393, was not found (probably port was removed/splitted) and will remove from counters now
2026 Aug 31 13:36:08.972376 M2-W6940-64X1-FR4 ERR syncd#syncd: :- processFlexCounterEvent: port VID oid:0x57000000000140, was not found (probably port was removed/splitted) and will remove from counters now
```

### Steps to Reproduce

config reload -y

Expect no error log

### Relevant log output

```shell

```

```shell

```

### Attach files (if any)

_No response_

Griffin-micas:
Fixes sonic-net/sonic-buildimage#29260
FlexCounter removal only requires the VID, but processFlexCounterEvent currently performs VID-to-RID translation for both SET and DEL events.

When a dynamically recreated object such as PORT_SERDES is removed, its mapping may be deleted before the corresponding stop_poll event is processed. This causes a misleading missing-VID error even though the counter cleanup succeeds.

Only translate VID to RID for SET events and add a regression test ensuring DEL does not query the mapping.

### Description of PR

Summary:
Avoid false `processFlexCounterEvent` errors when removing FlexCounter entries whose SAI objects and VID-to-RID mappings have already been removed.

For `stop_poll`/`DEL`, FlexCounter removal only requires the VID. However, `Syncd::processFlexCounterEvent()` previously attempted VID-to-RID translation for both `SET` and `DEL`.

For dynamically recreated objects such as `SAI_OBJECT_TYPE_PORT_SERDES`, the old VID-to-RID mapping may legitimately be removed before the corresponding FlexCounter `DEL` event is processed. This produced misleading errors such as:
```
processFlexCounterEvent: port VID oid:0x57..., was not found
(probably port was removed/splitted) and will remove from counters now
```

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

CPO platforms may register PORT_SERDES objects with the PORT_PHY_SERDES_ATTR FlexCounter group. These objects can be removed and recreated when updated SerDes parameters are applied.
The stop_poll event for the old object may therefore be processed after its VID-to-RID mapping has been removed. Since the removal path does not use the RID, attempting the translation is unnecessary and generates a misleading error.

##### Work item tracking
- Microsoft ADO **(number only)**:N/A

#### How did you do it?
Changed Syncd::processFlexCounterEvent() so that VID-to-RID translation is performed only for SET/start_poll.
The DEL/stop_poll path now directly removes the counter using the VID.
The existing SET behavior is preserved: if its VID-to-RID translation fails, Syncd logs the error and falls back to removing the stale counter entry.
Added a unit test using a mock Redis client to verify that FlexCounter removal does not call getRidForVid().

Validated on a 64-port CPO platform running docker-syncd-brcm-rpc.
Test procedure:
1. Installed the updated syncd-rpc package.
2. Restarted the syncd container.
3. Triggered the PORT_SERDES remove/recreate workflow.
4. Observed two PORT_PHY_SERDES_ATTR start_poll events.
5. Confirmed that no new processFlexCounterEvent: port VID ... was not found errors were generated.
Result:
```
post_upgrade_flexcounter_vid_errors=0
post_upgrade_serdes_start_poll_events=2
```
The last occurrence of the target error was from before the updated syncd binary was installed.
A regression unit test was also added:
```
SyncdTest.FlexCounterRemoveDoesNotRequireVidToRidTranslation
```

The issue was reproduced on a Broadcom-based CPO platform where PORT_SERDES objects participate in PORT_PHY_SERDES_ATTR FlexCounter polling.
The fix is platform-independent and applies to all dynamically removed SAI object types processed by this FlexCounter path.


syncd/Syncd.cpp
```diff
@@ -4177,9 +4177,12 @@ sai_status_t Syncd::processFlexCounterEvent(
...
+        sai_object_id_t rid = SAI_NULL_OBJECT_ID;
 
-        if (!m_translator->tryTranslateVidToRid(vid, rid))
+        // RID is only needed when adding a counter. During removal the SAI
+        // object and its VID-to-RID mapping may already have been removed.
+        if (effective_op == SET_COMMAND &&
+                !m_translator->tryTranslateVidToRid(vid, rid))
         {
             if (fromAsicChannel)
             {
```
lolyu, line 4188: Good gate — verified the DEL path is VID-only: `m_manager->removeCounter(vid, groupName)` doesn't take the rid, so skipping translation here is correct and the short-circuit means `tryTranslateVidToRid` isn't even called on a native DEL. Initializing `rid = SAI_NULL_OBJECT_ID` also closes the previously-uninitialized declaration. SET-failure→DEL fallback is preserved.

unittest/syncd/TestSyncd.cpp
```diff
@@ -428,6 +435,24 @@ class SyncdTest : public ::testing::Test
     std::shared_ptr<Syncd> m_syncd;
 };
 
+TEST_F(SyncdTest, FlexCounterRemoveDoesNotRequireVidToRidTranslation)
```
lolyu, line 438: This covers `fromAsicChannel = false` only. Please add a `fromAsicChannel = true` variant: that branch additionally runs `m_flexCounterTable->del(singleKey)` + `sendApiResponse(...)`, and it's the path that was emitting the misleading `SWSS_LOG_ERROR` in the first place — so it's the more faithful reproduction of the reported bug. Same assertions (`getRidForVid` Times(0), returns `SAI_STATUS_SUCCESS`) would guard both DEL branches.

lolyu (commented):
Correct, minimal, well-diagnosed fix — verified against the full `processFlexCounterEvent`. The root-cause analysis holds: the DEL path calls `m_manager->removeCounter(vid, groupName)`, which takes **only the VID**, so translating VID→RID on removal was doing work whose result is never consumed — and its miss produced the misleading "port VID … was not found" error for a removal that actually succeeds. For dynamically recreated objects (CPO `PORT_SERDES`) the old mapping can legitimately be gone before the `stop_poll`/DEL arrives, so gating the translation on `SET_COMMAND` is the right fix.

Details check out:
- Initializing `rid = SAI_NULL_OBJECT_ID` is necessary and also closes a latent uninitialized-read (previously `rid` was only written by `tryTranslateVidToRid`).
- The SET-failure→DEL fallback is preserved: a native DEL now skips translation via short-circuit, while a SET that can't translate still logs and demotes to DEL exactly as before.
- Null rid on the DEL path is never forwarded (`removeCounter` is VID-only), so it's harmless.
- The regression test pins the right contract — `getRidForVid` called `Times(0)` on DEL — not just an incidental output.

**One request before merge: close the test-coverage gap.** The new test only exercises `fromAsicChannel = false`. The `fromAsicChannel = true` DEL branch takes an additional path (`m_flexCounterTable->del(singleKey)` + `sendApiResponse(...)`), and since the ASIC channel is exactly where the misleading ERROR was being logged (the `SWSS_LOG_ERROR` vs `SWSS_LOG_WARN` split is on `fromAsicChannel`), it's the more representative reproduction of the reported bug. Please add a second variant with `fromAsicChannel = true` asserting the same `getRidForVid` `Times(0)` + `SAI_STATUS_SUCCESS`, so both DEL branches are guarded against a future re-introduction of the translation. Everything else is approve-quality; this is the one thing I'd like tightened.

(Context: this is the third fix in this syncd FlexCounter/translator area recently — #2057 poll-guard, #2068 notification `tryTranslateRidToVid` — all sharing the same lesson: use the translator's read-only/try variants where a miss is legitimate, and don't translate where the result is unused. This is the cleanest instance.)

Griffin-micas:
> Correct, minimal, well-diagnosed fix — verified against the full `processFlexCounterEvent`. The root-cause analysis holds: the DEL path calls `m_manager->removeCounter(vid, groupName)`, which takes **only the VID**, so translating VID→RID on removal was doing work whose result is never consumed — and its miss produced the misleading "port VID … was not found" error for a removal that actually succeeds. For dynamically recreated objects (CPO `PORT_SERDES`) the old mapping can legitimately be gone before the `stop_poll`/DEL arrives, so gating the translation on `SET_COMMAND` is the right fix.
> 
> Details check out:
> 
> * Initializing `rid = SAI_NULL_OBJECT_ID` is necessary and also closes a latent uninitialized-read (previously `rid` was only written by `tryTranslateVidToRid`).
> * The SET-failure→DEL fallback is preserved: a native DEL now skips translation via short-circuit, while a SET that can't translate still logs and demotes to DEL exactly as before.
> * Null rid on the DEL path is never forwarded (`removeCounter` is VID-only), so it's harmless.
> * The regression test pins the right contract — `getRidForVid` called `Times(0)` on DEL — not just an incidental output.
> 
> **One request before merge: close the test-coverage gap.** The new test only exercises `fromAsicChannel = false`. The `fromAsicChannel = true` DEL branch takes an additional path (`m_flexCounterTable->del(singleKey)` + `sendApiResponse(...)`), and since the ASIC channel is exactly where the misleading ERROR was being logged (the `SWSS_LOG_ERROR` vs `SWSS_LOG_WARN` split is on `fromAsicChannel`), it's the more representative reproduction of the reported bug. Please add a second variant with `fromAsicChannel = true` asserting the same `getRidForVid` `Times(0)` + `SAI_STATUS_SUCCESS`, so both DEL branches are guarded against a future re-introduction of the translation. Everything else is approve-quality; this is the one thing I'd like tightened.
> 
> (Context: this is the third fix in this syncd FlexCounter/translator area recently — #2057 poll-guard, #2068 notification `tryTranslateRidToVid` — all sharing the same lesson: use the translator's read-only/try variants where a miss is legitimate, and don't translate where the result is unused. This is the cleanest instance.)

Thank you for the thorough review! @lolyu 
You're right — although the VID-to-RID translation gate is shared, the existing test only covers the non-ASIC-channel path. The fromAsicChannel=true variant is a more faithful reproduction of the reported issue and should also verify the FLEX_COUNTER_TABLE deletion and API response path.
I’ll add coverage for fromAsicChannel=true with the same getRidForVid Times(0) and SAI_STATUS_SUCCESS assertions, then rerun the unit tests.

Griffin-micas:
@lolyu Thanks again for the careful review. I’ve added a fromAsicChannel=true regression test. It verifies that DEL does not call getRidForVid(), removes the FLEX_COUNTER_TABLE entry, sends a successful API response, and returns SAI_STATUS_SUCCESS.

Issue #2074, opened 2026-09-11
Enhancement:sonic-vpp needs to support crm tracking on acl and fdb


yue-fred-gao:
crm/test_crm.py exposed 3 unsupported crm tracking in sonic-vpp platform.
FAILED crm/test_crm.py::test_acl_entry[vlab-vpp-02-None] 
FAILED crm/test_crm.py::test_acl_counter[vlab-vpp-02-None]
FAILED crm/test_crm.py::test_crm_fdb_entry[vlab-vpp-02-None]

These are needed to enable vpp crm test in t0 topo.

yue-fred-gao:
Update: 2 acl failures are due to mismatched version between sairedis and sonic-platform-vpp.

PR #2081, opened 2026-09-15
vpp: fix CRM accounting for FDB entries

yue-fred-gao:
### Description of PR

Summary: fix CRM accounting for FDB entries

Fixes # (issue) https://github.com/sonic-net/sonic-sairedis/issues/2074

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

CRMTracker::m_fdbCount was maintained by onFdbCreated()/onFdbRemoved() hooked into SwitchVpp::create/remove. Three paths bypass those hooks:

- fdbclear → flush_fdb_entries(), where VirtualSwitchSaiInterface::flushFdbEntries() erases entries directly from m_objectHash

- MACs learned by VPP, added via create_internal() in generateFdbLearnedOrMoveEvent()

- MACs aged by VPP, dropped via remove_internal()

so the counter never recovered after a flush, and under-counted learned MACs that FdbOrch does count in crm_stats_fdb_entry_used.

##### Work item tracking

#### How did you do it?
CRMTracker::onFdbCreated()/onFdbRemoved() replaced by syncFdbCount(uint32_t), and SwitchVpp::refresh_read_only() feeds it m_objectHash[SAI_OBJECT_TYPE_FDB_ENTRY].size() before computing max - used for SAI_SWITCH_ATTR_AVAILABLE_FDB_ENTRY. Deriving the count on the read path is correct for every mutation path, including flush, learn and age, so the hooks in create/remove were dropped.

Two additional changes in SwitchVppAcl.cpp for post failure recovery. 
Run crm/test_crm.py and all passed in t0 vpp.

vpp specific


vslib/vpp/SwitchVppAcl.cpp
```diff
@@ -1829,7 +1830,13 @@ sai_status_t SwitchVpp::removeAclEntry(
...
         status);
 
-    return status;
+    // The object is gone from the state either way, and
+    // SAI_ACL_TABLE_ATTR_AVAILABLE_ACL_ENTRY is recomputed from what is left, so
+    // the entry has already been given back to the table. Reporting a failure
+    // here would stop the caller from dropping its own accounting for the entry,
+    // leaving crm_stats_acl_entry_used stuck above what the table actually holds.
+    // The VPP reprogramming error is logged above.
+    return SAI_STATUS_SUCCESS;
```
nhegde-microsoft, line 1869: I'm not sure I followed the reason for this change. `status` here is the result of the VPP call, so if it is not successful should we assume that the entries still exist in VPP? If yes, then will this not result in under-counting? Maybe internal state should stay unchanged if VPP failed.

yue-fred-gao: You are right. I have changed it. Thanks!

nhegde-microsoft (approved):
LGTM

aaronber0614 (approved):
LGTM

